PR #26, opened 2020-01-05
use recv character event instead of keyboard input

unrealhoang:
I was checking [alacritty](https://github.com/jwilm/alacritty/blob/1cfb0740bc530da3162292bbf682a234240f26a3/alacritty/src/event.rs#L511) and see they use ReceivedCharacter to send to tty and only use KeyboardInput for hotkey.
This is because for my keyboard, the Colon key is comprise from:
```
Construct keybinding: KeyboardInput { scancode: 39, state: Released, virtual_keycode: Some(Colon), modifiers: ModifiersState { shift: true, ctrl: false, alt: false, logo: false } }
```
and not `Semicolon` with shift modifier, same for most of other shift-based character like `*`, `(`, `)`, ..

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -79,16 +79,11 @@ pub fn ui_loop(editor: Arc<Mutex<Editor>>, nvim: Neovim, initial_size: (u64, u64
             },
 
             Event::WindowEvent {
-                event: WindowEvent::KeyboardInput {
-                    input,
-                    ..
-                },
+                event: WindowEvent::ReceivedCharacter(c),
                 ..
             } => {
-                if let Some(string) = construct_keybinding_string(input) {
-                    nvim.input(&string).expect("Input call failed...");
-                }
-            },
+                nvim.input(&c.to_string()).expect("Input call failed...");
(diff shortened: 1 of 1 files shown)


Kethku:
Interesting. How does this interact with the manually listed shift characters found here: https://github.com/Kethku/neovide/blob/master/src/keybindings.rs#L95

Let me do some manual testing and see how things work.

Kethku:
I am going to reject this PR since I'm pretty sure it will break other key combinations, however I will push a change that I think might fix it. Can you create an issue to track it and give feedback explaining if my solution fixed your problem?

Kethku:
https://github.com/Kethku/neovide/issues/27 comment here if you have further issues after trying the new change.

unrealhoang:
Yes, I can help with reporting the error. I don't think it would break vim's key combinations though, as the characters coming through alacritty to vim are from recv character events.
Regarding the design though, shouldn't you make the keyboard input for special cases/combinations and let all the characters pass through to vim?

PR #54, opened 2020-01-27
Fix a few typos and inconsistencies in README.md

meribold:
This looks like a very interesting project!

I noticed a couple typos while skimming the readme file, hence this PR. I also changed the formatting in a few places, so feel free to reject this PR if you disagree with those changes.


Kethku:
Thanks! Apparently I need to do a better job proof reading

PR #81, opened 2020-01-30
Remove unnecessary clones and allocations from bridge module

luisholanda:
Thanks for writing this awesome project! 

Exploring the code I saw that the `bridge` module make a lot of `.clone`s and owned values when references are enough. This PR fixes what I found.

The commits are very isolated and can be reviewed in isolation.

As a tip, dunno if you knew the tool but [clippy](https://github.com/rust-lang/rust-clippy) pointed many of the issues that this PR fixes.

### DIFF
--- a/src/bridge/events.rs
+++ b/src/bridge/events.rs
@@ -159,95 +159,97 @@ fn unpack_color(packed_color: u64) -> Color4f {
     }
 }
 
-fn parse_array(array_value: &Value) -> Result<Vec<Value>> {
-    if let Value::Array(content) = array_value.clone() {
-        Ok(content.to_vec())
+fn parse_array(array_value: &Value) -> Result<&[Value]> {
+    if let Value::Array(content) = array_value {
+        Ok(&content[..])
     } else {
         Err(EventParseError::InvalidArray(array_value.clone()))
     }
 }
 
-fn parse_map(map_value: &Value) -> Result<Vec<(Value, Value)>> {
-    if let Value::Map(content) = map_value.clone() {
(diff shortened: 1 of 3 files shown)


Kethku:
Thanks for the PR! This looks great. My only issue is that I am currently in the middle of rewriting the keyboard management, so I can't merge this unless that file's changes are removed. Otherwise I suspect this will improve performance non trivially. I appreciate the help!

Kethku:
Welp I've given up on the keybinding rewrite again. So I might as well merge this.

KillTheMule:
Huh, you're fast. I'll make my comment anyways, maybe you're interested in revisiting:

My recommendation is to take `Value`s by value, and return the owned, decoded value, rather than taking and returning a reference. I don't see a usecase for keeping the old `Value` around, and that way you can elide even more clones (e.g. [this one](https://github.com/Kethku/neovide/pull/81/files#diff-adf795efa9299f343af5f94871dc2b46R213)). It also feels much cleaner to me.

You can use the [`TryFrom`](https://docs.rs/rmpv/0.4.3/rmpv/enum.Value.html#impl-TryFrom%3CValue%3E) implementations for nice error reporting, although for `nvim-rs` I had to resort to [my own variant of the trait](https://github.com/KillTheMule/nvim-rs/blob/master/src/rpc/unpack.rs) because you can't very well implement `TryFrom` recursively because of the ambiguity introduced by msgpack's `Binary` Value.

luisholanda:
Yeah, we could defitely do that. I will see what I can do.

PR #99, opened 2020-02-01
Different implementation of scroll_region

jonvaldes:
Avoids allocations and temporary copies of the scrolled data, although it's probably harder to read, unfortunately. I could add comments if desired.

Note: one difference with the previous implementation is that this one clears the grid after itself. The area that the scrolled text will no longer cover will be left with a value of None, and it will be marked as dirty. I did this as the neovim-qt implementation was doing that too, and seemed safer. If this is not desired, I can easily remove it.

### DIFF
--- a/src/editor/mod.rs
+++ b/src/editor/mod.rs
@@ -204,48 +204,35 @@ impl Editor {
     }
 
     fn scroll_region(&mut self, top: u64, bot: u64, left: u64, right: u64, rows: i64, cols: i64) {
-        let (top, bot) =  if rows > 0 {
-            (top as i64 + rows, bot as i64)
-        } else if rows < 0 {
-            (top as i64, bot as i64 + rows)
+        let (mut y, end_y, step_y) = if rows > 0 {
+            (top as i64 + rows, bot as i64, 1)
         } else {
-            (top as i64, bot as i64)
+            (bot as i64 + rows - 1, top as i64 - 1, -1)
         };
 
-        let (left, right) = if cols > 0 {
-            (left as i64 + cols, right as i64)
-        } else if rows < 0 {
-            (left as i64, right as i64 + cols)
+        let (start_x, end_x, step_x) = if cols > 0 {
+            (left as i64 + cols, right as i64, 1)
         } else {
-            (left as i64, right as i64)
+            (right as i64 + cols - 1, left as i64 - 1, -1)
         };
 
-        let mut region = Vec::new();
-        for y in top..bot {
-            let row = &self.grid[y as usize];
-            let mut copied_section = Vec::new();
-            for x in left..right {
-                copied_section.push(row[x as usize].clone());
-            }
-            region.push(copied_section);
-        }
-
-        let new_top = top as i64 - rows;
-        let new_left = left as i64 - cols;
-
-        for (y, row_section) in region.into_iter().enumerate() {
-            for (x, cell) in row_section.into_iter().enumerate() {
-                let y = new_top + y as i64;
-                if y >= 0 && y < self.grid.len() as i64 {
-                    let row = &mut self.grid[y as usize];
-                    let dirty_row = &mut self.dirty[y as usize];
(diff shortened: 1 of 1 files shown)


jonvaldes:
By the way, I feel like the editor structure is using a lot of allocations because of the use of nested Vecs. If you're OK with that, I might refactor it so Editor::grid is a Vec<GridCell> of width*height length, instead of a Vec<Vec<GridCell>>.

Kethku:
Taking a look now

src/editor/mod.rs
```diff
@@ -204,48 +204,35 @@ impl Editor {
...
-                if y >= 0 && y < self.grid.len() as i64 {
-                    let row = &mut self.grid[y as usize];
-                    let dirty_row = &mut self.dirty[y as usize];
-                    let x = new_left + x as i64;
-                    if x >= 0 && x < row.len() as i64 {
-                        row[x as usize] = cell;
-                        dirty_row[x as usize] = true;
+        while y != end_y {
+            let dest_y = y - rows;
+            if dest_y >= 0 && dest_y < self.size.1 as i64 {
```
Kethku: nit: might as well destructure size into width and height variables so that readers don't have to puzzle through what self.size.1 means

src/editor/mod.rs
```diff
@@ -204,48 +204,35 @@ impl Editor {
...
+                let mut x = start_x;
+                while x != end_x {
+                    let dest_x = x - cols;
+                    if dest_x >= 0 && dest_x < self.size.0 as i64 {
+                        let cell = std::mem::replace(&mut self.grid[y as usize][x as usize], None);
```
Kethku: I know you called this out in the request, but the docs for the scroll event explicitly say that clearing the scrolled in cells is unnecessary: https://github.com/neovim/neovim/blob/master/runtime/doc/ui.txt#L406

src/editor/mod.rs
```diff
@@ -204,48 +204,35 @@ impl Editor {
...
-            for (x, cell) in row_section.into_iter().enumerate() {
-                let y = new_top + y as i64;
-                if y >= 0 && y < self.grid.len() as i64 {
-                    let row = &mut self.grid[y as usize];
-                    let dirty_row = &mut self.dirty[y as usize];
-                    let x = new_left + x as i64;
-                    if x >= 0 && x < row.len() as i64 {
-                        row[x as usize] = cell;
-                        dirty_row[x as usize] = true;
+        while y != end_y {
```
Kethku: My only real hesitation here is that these while loops feel like they are itching to be for loops. Other than that this looks great.

Kethku:
I only have very minor feedback. Thanks for the PR! As for the extra allocations in grid, I'm sorta inclined to hold off because semantically separating the rows helps keep the code clean. I have done some extensive profiling on the render loops, and the allocations don't reach levels to where they are important in terms of runtime. I could be wrong cause profiling is hard to get right, but I feel like the tradeoff of fewer allocations in return for more confusing data structure feels not worth it.

What do you think?

src/editor/mod.rs
jonvaldes: Reimplemented as a ranged for loop. Not terribly happy about the allocations, but `rev()` returning an entirely different type forced me to use a `Box`, and I haven't been able to find a cleaner way

src/editor/mod.rs
jonvaldes: Removed the clearing. Seems to work fine indeed

src/editor/mod.rs
jonvaldes: Destructured it into variables. Good idea.

Kethku:
Thanks again!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

src/editor/mod.rs
Kethku: I'll spend some time thinking if there is something cleaner. Merged for noq

PR #102, opened 2020-02-03
Idea for linearization of the Editor grid storage

jonvaldes:
Converts `Editor::grid` from a nested `Vec<Vec<GridCell>>` to a `Vec<GridCell>` structure.
Does the same thing with `Editor::dirty`.

I'm opening this pull request just to discuss this as a possible approach. I believe this specific implementation can be improved (maybe by encapsulating a bit more the `grid` itself? Maybe extract `grid` and `dirty` into a separate struct?), but I also think it shows promise.

### DIFF
--- a/src/editor/grid.rs
+++ b/src/editor/grid.rs
@@ -78,11 +78,10 @@ impl CharacterGrid {
         self.dirty.resize_with((self.width * self.height) as usize, || value);
     }
 
-    pub fn rows<'a>(&'a self) -> Vec<&'a [GridCell]> {
+    pub fn rows<'a>(&'a self) -> impl Iterator<Item=&'a[GridCell]> {
         (0..self.height)
-            .map(|row| {
+            .map(move |row| {
                 &self.characters[(row * self.width) as usize..((row + 1) * self.width) as usize]
             })
-            .collect()
     }
 }
--- a/src/editor/grid.rs
+++ b/src/editor/grid.rs
@@ -1,0 +1,75 @@
+use std::sync::Arc;
+
+use super::style::Style;
+
+type GridCell = Option<(String, Option<Arc<Style>>)>;
+
+pub struct CharacterGrid {
+    pub characters: Vec<GridCell>,
+    pub width: u64,
+    pub height: u64,
+    pub should_clear: bool,
+
+    dirty: Vec<bool>,
+}
+
+impl CharacterGrid {
+    pub fn new() -> CharacterGrid {
+        CharacterGrid {
+            characters: vec![],
+            dirty: vec![],
+            width: 0,
+            height: 0,
+            should_clear: true,
+        }
+    }
+
+    pub fn resize(&mut self, new_size: (u64, u64)) {
+        trace!("Editor resized");
+        self.width = new_size.0;
+        self.height = new_size.1;
+        self.clear();
+    }
+
+    pub fn clear(&mut self) {
+        trace!("Editor cleared");
+        let cell_count = (self.width * self.height) as usize;
+        self.characters = vec![None; cell_count];
+        self.dirty = vec![true; cell_count];
+        self.should_clear = true;
+    }
+
+    pub fn cell_index(&self, x: u64, y: u64) -> Option<usize> {
+        if x >= self.width || y >= self.height {
+            None
+        } else {
+            Some((x + y * self.width) as usize)
+        }
+    }
+
+    pub fn is_dirty_cell(&self, x: u64, y: u64) -> bool {
+        if let Some(idx) = self.cell_index(x, y) {
+            self.dirty[idx]
+        } else {
+            false
+        }
+    }
+
+    pub fn set_dirty_cell(&mut self, x: u64, y: u64) {
+        if let Some(idx) = self.cell_index(x, y) {
+            self.dirty[idx] = true;
+        }
+    }
+
+    pub fn set_dirty_all(&mut self, value: bool) {
+        self.dirty.resize(self.dirty.len(), value);
+    }
+
+    pub fn rows<'a>(&'a self) -> Vec<&'a [GridCell]> {
+        (0..self.height)
+            .map(|row| {
+                &self.characters[(row * self.width) as usize..((row + 1) * self.width) as usize]
+            })
+            .collect()
+    }
+}
--- a/src/editor/grid.rs
+++ b/src/editor/grid.rs
@@ -1,4 +1,5 @@
 use std::sync::Arc;
+use log::trace;
 
 use super::style::Style;
 
--- a/src/editor/grid.rs
+++ b/src/editor/grid.rs
@@ -3,15 +3,15 @@ use log::trace;
 
 use super::style::Style;
 
-type GridCell = Option<(String, Option<Arc<Style>>)>;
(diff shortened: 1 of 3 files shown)


src/editor/mod.rs
```diff
@@ -65,6 +65,36 @@ impl Editor {
...
+        }
+    }
+
+    pub fn set_dirty_cell(&mut self, x: u64, y: u64) {
+        if let Some(idx) = self.cell_index(x, y) {
+            self.dirty[idx] = true;
+        }
+    }
+
+    fn rows<'a> (&'a self) -> Vec<&'a [GridCell]> {
```
jonvaldes: This is a way to preserve the row-by-row access pattern, without cloning the actual data. It just creates one slice per row, so the overhead should be pretty minimal.

Kethku:
I really like the idea of encapsulating the grid structure in its own type with access functions to make it easy to understand while still reducing the need for extra allocations.

I've been thinking about the ways in which the Vec of Vec approach is inefficient:
1. Each Vec if not initialized to a given size, likely has allocated extra space for additions to the array
2. Size management overhead at the beginning of the Vec (negligible)
3. Extra bounds checks for both the row and column

Those were the ones I thought of off the top of my head. Can you help me understand further problems? So part of me feels the local change isn't worth the extra minor complexity since the above issues feel small to me. Not to mention, none of them show up on perf measurements as an issue.

Regardless, abstracting the storage to some general purpose 2d datastructure seems completely reasonable and represents the best of both worlds to me. The internals can be whatever is most performant while the externals can expose an interface that makes sense for the domain. This would also have the added benefit that both the cells themselves and the dirty tracking can use the same data structure and have the same api. This way any perf improvements or what have you can be shared etc.

In any case, thanks so much for your continued help. I appreciate the rigor you're lending. Hopefully I can figure out the visual issues youve been running into.

Kethku:
I would be happy to take a stab at abstracting the data structure soonish. You've already done a lot

src/editor/mod.rs
Kethku: Looks awesome to me. See my top level comment

jonvaldes:
It's no trouble at all. Neovide is the first vim client that fits my specific needs:
- Good hi-dpi support (I use an extremely high resolution screen at work)
- Fast rendering at high resolution (same reason as above)
- Good font rendering
- Doesn't use OpenGL (OpenGL doesn't work at all through VPN on Windows)
- Its source code is small enough for me to wrap my head around it

No other client fits those, so even though I have the line ordering issue, I'm already using it as my main driver on Windows. Once the issue with the keyboard modifiers is solved on Linux, I'll start using it there too :)

I don't have much free time (yay kids!), but I'll happily hack on neovide when I can to improve my own life :smile:

jonvaldes:
As for the reasons for the linearization, there's a couple more.

First, going through the nested `Vec`s will require extra pointer chasing (with associated cache invalidations), which will hurt performance.

Second, a sizeable chunk of religious zealotry on my part. Neovide's performance does plummet quite quickly under heavy cpu load, so I'm also looking into ways of improving that. And because proper performance analysis is a lot more work than seeing a usual suspect and start shouting "Wiiitch! Bring the torches!"... well, that's how we end up with this pull request.

Still, I agree that splitting the grid into its own structure will be beneficial for both performance and readability. I think I'll give it a try today if I have some time

jonvaldes:
Still unfinished, but starting to get there

jonvaldes:
I still have an issue with scrolling, I think I introduced a bug in the extraction. I'll try to look into it today

src/editor/grid.rs
```diff
@@ -0,0 +1,76 @@
...
+        self.width = new_size.0;
+        self.height = new_size.1;
+        self.clear();
+    }
+
+    pub fn clear(&mut self) {
+        trace!("Editor cleared");
+        let cell_count = (self.width * self.height) as usize;
+        self.characters = vec![None; cell_count];
+        self.dirty = vec![true; cell_count];
```
luisholanda: Couldn't we recycle the previous allocations? We could use [`Vec::resize_with`] to do that.

```suggestion
	    self.characters.clear();
        self.characters.resize_default(cell_count);
        self.dirty.clear();
        self.dirty.resize_with(cell_count, || true); // default of bool is false.
```

[`Vec::resize_with`]: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize_with

src/editor/grid.rs
```diff
@@ -0,0 +1,76 @@
...
+        if let Some(idx) = self.cell_index(x, y) {
+            self.dirty[idx] = true;
+        }
+    }
+
+    pub fn set_dirty_all(&mut self, value: bool) {
+        self.dirty.resize(self.dirty.len(), value);
+    }
+
+    pub fn rows<'a>(&'a self) -> Vec<&'a [GridCell]> {
```
luisholanda: Do we need to collect the iterator to a `Vec`? It seems from the code that we only collect to call `iter` again. Couldn't we return `impl Iterator<Item = &'a [GridCell]>`?

src/editor/grid.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    pub fn clear(&mut self) {
+        trace!("Editor cleared");
+        let cell_count = (self.width * self.height) as usize;
+        self.characters = vec![None; cell_count];
+        self.dirty = vec![true; cell_count];
+        self.should_clear = true;
+    }
+
+    pub fn cell_index(&self, x: u64, y: u64) -> Option<usize> {
```
luisholanda: Instead of this, couldn't we have a method to return a reference for the cell? From the code, I see the following pattern:

1. Get the index.
2. Use the index to get the cell.

This doesn't abstract the underlying storage of the cells and is very repetitive.

src/editor/grid.rs
jonvaldes: Yep, that sounds like an even better idea :+1:

src/editor/grid.rs
jonvaldes: Yep, returning an `Option<&'a mut GridCell>` could be a better interface. I'll look into that. Thanks for the suggestion :+1:

src/editor/grid.rs
jonvaldes: According to the docs, `resize_with` only applies the new values to the newly allocated elements. The old ones retain their old value. Same is true for the other resize functions, actually, which makes my other uses of them wrong

luisholanda: > The old ones retain their old value.

Exactly, just call `Vec::clear` before.

src/editor/grid.rs
luisholanda: I suggest making two methods, one for `&GridCell` and another for `&mut GridCell`, this way we can write functions that don't need mutable access to `CharacterGrid`.

jonvaldes: Yep, that was my plan

src/editor/grid.rs
jonvaldes: Huh. Ok, sounds good :+1:

jonvaldes:
Ok, scrolling issue solved. PR is looking much better now after the feedback :-)

Kethku:
Looks awesome. This should be a big improvement.

PR #103, opened 2020-02-03
Github Workflow to generate binaries

rahuliyer95:
Simple GitHub workflow to generate Windows / macOS binaries. It will run with each push and create a binary against the master branch. Hope this helps to create releases :)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,33 @@
+name: Rust
+
+on: [push]
+
+jobs:
+  build-windows:
+
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build
+      run: cargo build --release
+    - uses: actions/upload-artifact@v1
+      with:
+        name: neovide-windows.exe
+        path: ./target/release/neovide.exe
+
+  build-mac:
+
+    runs-on: macos-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install Vulkan SDK
+      run: |
+        brew cask install apenngrace/vulkan/vulkan-sdk
+    - name: Build
+      run: cargo build --release
+    - uses: actions/upload-artifact@v1
+      with:
+        name: neovide-macos
+        path: ./target/release/neovide


.github/workflows/rust.yml
```diff
@@ -0,0 +1,33 @@
+name: Rust
+
+on: [push]
```
luisholanda: Wouldn't it be better to only run in pushes to the `master` branch?

luisholanda:
Is there a reason to not build Linux binaries?

.github/workflows/rust.yml
Kethku: Actually it's useful for any branch I think. For situations when building might fail on a platform you don't have access to

luisholanda: Yeah, makes sense 👍

justinmk:
> Is there a reason to not build Linux binaries

Would need to be an appimage, else it is unlikely to work on most systems except Travis version of Ubuntu.

rahuliyer95:
I'm not very familiar with building on Linux and the instructions in the README were not very clear to me. If you can provide me steps for building **appimage**  I can add it to the workflow. 

I was reading online and found out that Cargo can build native app bundles (like **neovide.app** on macOs) using [cargo-bundle](https://github.com/burtonageo/cargo-bundle) library. If you can set that up (or help get that working) we can have the workflow directly upload the app bundles as artifacts. 

Let me know what you think about it.

justinmk:
> If you can provide me steps for building the **appimages** I can add it to the workflow

I can't provide steps, nor would I want you to spend your time on it :)

It requires reading appimage docs and takes some effort. Neovim's is [here](https://github.com/neovim/neovim/blob/master/scripts/genappimage.sh) for reference.

Linux has better dev tooling that mac/win should building neovide shouldn't be a problem. Though an appimage is great to have once someone puts the effort into it.

Kethku:
This is a beautiful first start, so I'm going to merge it and split out the remaining tasks: build full packages using cargo-bundle, and add support for linux builds.

This is fantastic help. Thanks so much @rahuliyer95

PR #104, opened 2020-02-03
Switch to parking_lot::Mutex. It is faster and has a better API than the std one

jonvaldes:
Also, no possibility of poisoning, so it removes a bunch of unwraps.
parking_lot is already a transitive dependency for neovide, so this is not adding more dependencies, really.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1234,6 +1234,7 @@ dependencies = [
  "log",
  "lru",
  "nvim-rs",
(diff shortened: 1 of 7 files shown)


Kethku:
I've been meaning to do this. Thanks!

Kethku:
I'll merge this once it is rebased

jonvaldes:
Rebased!

PR #108, opened 2020-02-04
Update README with build instructions for Ubuntu

rahuliyer95:
I tried this on Docker and it successfully generates a binary. I don't have a way to test if the binary actually works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,11 +5,11 @@ functionally like the terminal UI.
 
 ![Basic Screen Cap](./assets/BasicScreenCap.png)
 
-I've been using this as my daily driver since November 2019. It should be relatively stable, but I'm still working out some kinks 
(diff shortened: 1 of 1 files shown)


Kethku:
Only thing is that neovide has been built on distros other than Ubuntu, so my preference would be to have general instructions for Linux and then use specifics for generating the appimage. What do you think?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rahuliyer95:
Based on my limited understanding of Appimages, it first requires a binary to be built and then can be packaged into an Appimage to be shared agnostic of the distribution. Since it needs to be built first, I updated the README with instructions to manually build it on Ubuntu.

Kethku:
Okeydokey

PR #112, opened 2020-02-05
Add missing keycode translation for the Add (plus) key

jonvaldes:
As reported in https://github.com/Kethku/neovide/issues/27


jonvaldes:
I'm finding a bunch more keycodes that are missing. I'll do a PR that adds all of them

PR #115, opened 2020-02-05
Cargo bundle for macOS

### LINKED ISSUE #105: Add cargo-bundle to actions
As mentioned here: https://github.com/Kethku/neovide/pull/103#issuecomment-581736977 cargo bundle seems like a good option for producing macos and linux release assets. On windows the single exe with everything linked in should be more than sufficient, but the other platforms need some help.

In particular it is important that we add the icon to the resulting bundle, but that appears to be well supported.

rahuliyer95:
- Should fix #105 for macOS
- `Neovide.app` available as an artifact to be downloaded
- `nvim` binary should available in path while executing the App
- Need to run `chmod +x Neovide.app/Contents/MacOS/neovide` after downloading the artifact. Open issue in Github's [upload-artifact](https://github.com/actions/upload-artifact/issues/38) action

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -23,11 +23,23 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Install Vulkan SDK
-      run: |
-        brew cask install apenngrace/vulkan/vulkan-sdk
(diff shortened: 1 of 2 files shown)


Kethku:
Thank you so much for setting this up! This is awesome :)

PR #118, opened 2020-02-05
Add all missing VirtualKeycodes to the translation function

jonvaldes:
Left a bunch unsupported, as I have no idea how to translate the "Calculator" key, for example, to something Vim understands


Kethku:
Perfect. I am planning on rewriting the modifiers code to handle global events (current modifier handling is deprecated as per warnings) but this should make our current support more complete

PR #127, opened 2020-02-11
fixed < key

### LINKED ISSUE #27: Missing Key Support
Issue to track missing keybinding support

awsomearvinder:
Other then for numerics, keycodes seemed to arrive with shift applied. Issue seems to have been caused by the match{} statement not actually activating on a press of "," as it was sent as a "<". Hence "," couldn't be swapped to "lt".
This may be the final step towards being able to close #27?

### DIFF
--- a/src/bridge/keybindings.rs
+++ b/src/bridge/keybindings.rs
@@ -195,23 +195,10 @@ pub fn append_modifiers(modifiers: ModifiersState, keycode_text: &str, special: 
             "8" => "*".to_string(),
             "9" => "(".to_string(),
             "0" => ")".to_string(),
-            "'" => "\"".to_string(),
-            "Bslash" => {
-                special = false;
-                "|".to_string()
-            },
-            "," => {
+            "<" => {
                 special = true;
                 "lt".to_string()
-            },
-            "=" => "+".to_string(),
-            "`" => "~".to_string(),
-            "[" => "{".to_string(),
-            "-" => "_".to_string(),
-            "." => ">".to_string(),
-            "]" => "}".to_string(),
-            ";" => ":".to_string(),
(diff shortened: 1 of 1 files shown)


Kethku:
@jonvaldes I think you touched this most recently. Do you have an idea of the effects of a change like this?

Kethku:
As a side note, I am in the process of investigating moving to sdl instead of winit. Initial tests are very promising as the input management is way more complete and consistent.

awsomearvinder:
This may be dumb but can I ask the reason that window.rs tries to remove the shift modifier? From my quick testing it dosen't change behavior much, and it just takes up a single line of code. You still need to apply the shift on the numerical values yourself, and the other chars mostly are unaffected by it.

Kethku:
I'm not sure I'm following what line of code you are talking about...

jonvaldes:
Well, the idea of manually applying the effect of the shift modifier to the captured character is extremely brittle and breaks down as soon as you switch keyboard layout to anything other than English, so I'm definitely not against finding a better way to do this.

I haven't had time yet to check whether this change improves things, unfortunately.

Kethku:
I just tried this, but it breaks shifted keys on windows. Adding the S- prefix to a key does not convert it to the shift modified version of the key. For instance sending `<S-;>` to neovim via an input does not result in `:` but instead `;`. The code you removed is meant to handle this.

As I mentioned above, this is all changing to be more complete and standard with the change to sdl.

jonvaldes:
@awsomearvinder Just proposed a different implementation above, based on your work. This should still fix the < key on linux, without (hopefully) causing trouble in other systems.

Kethku:
Closing in favor of https://github.com/Kethku/neovide/pull/128

PR #128, opened 2020-02-12
Fix less than key on linux, simpler version

jonvaldes:
Based on @awsomearvinder's PR: https://github.com/Kethku/neovide/pull/127
But doing a simpler change that's unlikely to cause any issues on Windows


Kethku:
I see nothing wrong with this. Especially since its changing shortly. Thanks

PR #132, opened 2020-02-15
Write nvim version requirement messages directly to stderr

vivlim:
The error messages for having a version of neovim which is too old are not printed to the console. They are logged to file, and only if the `--log` argument is passed.

This PR makes those messages more visible by writing them directly to stderr.


Kethku:
Totally my bad. That came in when I swapped to the logging system. Thanks!

PR #143, opened 2020-02-18
Adding a setting to change the refresh rate

jonvaldes:


Kethku:
My only concern here is that the cursor animation algorithm is currently framerate dependent. But thats probably a separate issue.

Kethku:
Merging and creating a new issue

PR #150, opened 2020-02-24
Structured settings

Kethku:
Making PR to facilitate discussion

### DIFF
--- a/src/redraw_scheduler.rs
+++ b/src/redraw_scheduler.rs
@@ -4,11 +4,40 @@ use std::time::Instant;
 
 use log::trace;
 
-use crate::settings::SETTINGS;
+use crate::settings::{SETTINGS, Value};
 
 lazy_static! {
     pub static ref REDRAW_SCHEDULER: RedrawScheduler = RedrawScheduler::new();
 }
+
+#[derive(Clone)]
+struct RedrawSettings {
+    extra_buffer_frames: u64,
+}
+
+fn parse_changed_setting(name: &str, value: Option<Value>) -> Value {
+    match name {
+        "extra_buffer_frames" => {
+            let mut settings = SETTINGS.get::<RedrawSettings>();
+            if let Some(value) = value {
+                settings.extra_buffer_frames= value.as_u64().unwrap(); // TODO -- handle wrong data type
+                SETTINGS.set(&settings);
+            }
+            Value::from(settings.extra_buffer_frames)
+        }
+        _ => {
+            panic!(format!("Unknown setting: {}", name));
+        }
+    }
+}
+
+pub fn initialize_settings() {
+    SETTINGS.set(&RedrawSettings { extra_buffer_frames: 60 });
+
+    SETTINGS.add_listener("extra_buffer_frames", parse_changed_setting);
+}
+
+
 
 pub struct RedrawScheduler {
     frames_queued: AtomicU16,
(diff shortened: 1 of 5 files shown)


src/redraw_scheduler.rs
```diff
@@ -4,12 +4,41 @@ use std::time::Instant;
...
+struct RedrawSettings {
+    extra_buffer_frames: u64,
+}
+
+fn parse_changed_setting(name: &str, value: Option<Value>) -> Value {
+    match name {
+        "extra_buffer_frames" => {
+            let mut settings = SETTINGS.get::<RedrawSettings>();
+            if let Some(value) = value {
+                settings.extra_buffer_frames = value.as_u64().unwrap(); // TODO -- handle wrong data type
```
Kethku: I thought about this some more. I suspect for most cases, there are a finite number of basic types that we store as top level variables in vim: u64, string, int representing u64, etc

Given that thats the case, I feel like creating a generic trait which we can implement for the base types would be a better solution than creating a macro for this code. What do you think? It seems to make sense to me that macros would be a last resort when no other solution exists.

src/redraw_scheduler.rs
```diff
@@ -37,8 +66,8 @@ impl RedrawScheduler {
 
     pub fn queue_next_frame(&self) {
         trace!("Next frame queued");
-        let buffer_frames = SETTINGS.get("extra_buffer_frames").read_u16();
-        self.frames_queued.store(buffer_frames, Ordering::Relaxed);
+        let buffer_frames = SETTINGS.get::<RedrawSettings>().extra_buffer_frames;
```
Kethku: This is WAY better. No strings, type safe. Good stuff

src/window.rs
```diff
@@ -251,6 +251,45 @@ impl WindowWrapper {
...
 
+#[derive(Clone)]
+struct WindowSettings {
+    refresh_rate: u64,
+    no_idle: bool,
+}
+
+fn parse_changed_setting(name: &str, value: Option<Value>) -> Value {
+    let mut settings = SETTINGS.get::<WindowSettings>();
+    match name {
```
Kethku: Minor, but you could probably simplify this by passing a lambda directly to the add_listener function which does the parsing for each setting directly. That way you don't have to do the string comparison, and don't have to have the match statement.

Further, if you do that, I think you could probably remove the name argument to the listeners type completely. add_listener specifies the name of the variable in the hookup, so its sorta redundant to have the name in the callback as well.

src/window.rs
```diff
@@ -251,6 +251,45 @@ impl WindowWrapper {
...
+        "refresh_rate" => {
+            if let Some(value) = value {
+                settings.refresh_rate = value.as_u64().unwrap(); // TODO -- handle wrong data type
+                SETTINGS.set(&settings);
```
Kethku: You mentioned this above, that the listeners are in charge of setting and parsing the default values. This feels extra weird to me because the purpose of the listener doesn't flow from the type of the listener. Just reading the settings code, it isn't immediately obvious that the listener should also set the current value.

Unfortunately, I don't have a suggestion for how to change this to make it better... I'll think some more.

src/settings.rs
```diff
@@ -15,120 +16,86 @@ lazy_static! {
     pub static ref SETTINGS: Settings = Settings::new();
 }
 
-#[derive(Debug)]
-pub enum Setting {
-    Bool(bool),
-    U16(u16),
-    String(String)
+struct SettingsObject {
```
Kethku: A high level comment at the top of this file somewhere describing the plan and reasoning would help make this understandable for anyone trying to puzzle through the settings architecture in the future.

jonvaldes:
Ok, I think this should be pretty much done now.

jonvaldes:
Ok, I think this should be pretty much done now.

PR #155, opened 2020-02-27
Shaper includes system symbol fonts into cache

### LINKED ISSUE #153: Neovide fails to render classic unicode glyphs when using nerd font
I noticed that neovide seems to not render some characters correctly when I use the "[Meslo Nerd Font patched for Powerlevel10k](https://github.com/romkatv/powerlevel10k#meslo-nerd-font-patched-for-powerlevel10k)". 

image

Some glyphs seem to work as seen at bottom tux but the chars that I use for `listchars` are not.

When I set the same font in the "new Windows Terminal" everything renders just fine.

image

My `listchars` variable is set to

```viml
set listchars=tab:➞\ ,space:⬝,trail:~,nbsp:+
```

I have also no idea where that "NORMAL" thing in the top right corner came from. I don't see this in other neovim clients.

I'm running Windows 10 (Build 1909) with NVIM v0.4.3 and the latest version of Neovide that I just pulled from Github and built.

EDIT: Could be related to HarfBuzz because fvim also can't render my listchars, but nvim-qt does fine.

SirJson:
This should fix #153 on Windows but needs testing on Linux and macOS.

This is how it looks on Windows:

b

I can test this branch on on macOS later if you want. On Linux I don't even know if I can run something using Vulkan thanks to Nvidia 👍 

I also did a test with Quiva as default font for symbols but it didn't mesh well with monospaced fonts.

### DIFF
--- a/src/renderer/caching_shaper.rs
+++ b/src/renderer/caching_shaper.rs
@@ -17,6 +17,15 @@ const WIDE_FONT: &str = "NotoSansMonoCJKjp-Regular.otf";
 const WIDE_BOLD_FONT: &str = "NotoSansMonoCJKjp-Bold.otf";
 
 #[cfg(target_os = "windows")]
+const SYSTEM_SYMBOL_FONT: &str = "Segoe UI Symbol";
+
+#[cfg(target_os = "linux")]
+const SYSTEM_SYMBOL_FONT: &str = "Unifont";
+
+#[cfg(target_os = "macos")]
+const SYSTEM_SYMBOL_FONT: &str = "Apple Symbols";
+
+#[cfg(target_os = "windows")]
 const SYSTEM_EMOJI_FONT: &str = "Segoe UI Emoji";
 
 #[cfg(target_os = "macos")]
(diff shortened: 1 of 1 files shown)


Kethku:
Thanks for the change! Testing on macOS would be great. On linux I don't know either way. For now I'm going to merge. Could you ping me here when you've gotten a chance to test on mac?

Kethku:
Thanks for the PR btw! It helps me out a ton

SirJson:
Ok so I managed to test it on both macOS and Linux / Ubuntu 19.10.

On macOS this patch had no effect. As a quick way of testing if Apple Symbols even includes the missing glyph I copy pasted it into this "Text Editor" application that is pre-installed and it switched to STIXGeneral. So I modified the renderer to use that font for Symbols and tried again: Same result.

As of now I have no idea why the renderer isn't picking those glphys up but they both loaded just fine.

On Linux on the other hand...

Screenshot from 2020-02-28 09-02-09

it looks like it works. But only if you install Unifont if it is not pre-installed. I think mentioning this in the README would be fine since I'm pretty sure there is no "standard font" on Linux, and Unifont is a GNU Project so I'm certain that most distros package that Font one way or another.

One thing to note though is that the installation guides for both Linux and macOS are incomplete, on my Mac I had to install cmake: `brew install cmake`

And on my Linux system I had to install `cmake` , `libsndio-dev` (?) and as mentioned above `unifont`

PR #171, opened 2020-03-02
nicer install instructions

DhruvDh:
theoretically, it should do the exact same thing in one command. haven't tried it though


Kethku:
Just tried on windows. This works!

Kethku:
I'm torn. Eventually I'd like to be in a place where users either build from a git clone which means they should use the normal flow so they know where the source is, or they download from a release. So unfortunately I think I'm gonna close this to lead people down better paths.

Kethku:
Feel free to reopen if you want to chat more about it tho!

PR #183, opened 2020-03-07
Add workaround for Nordic keyboard

mewhhaha:
This is just the result from my testing. I would call it dangerous, because I have no idea what anything does. But it does allow me to do brackets. 💯


src/window.rs
```diff
@@ -126,8 +126,11 @@ impl WindowWrapper {
 
     pub fn handle_key_down(&mut self, keycode: Keycode, modifiers: Mod) {
         trace!("KeyDown Received: {}", keycode);
+        const ALTGR: Mod = Mod::from_bits_truncate(0x0240);
```
Kethku: Could you try https://docs.rs/sdl2/0.33.0/sdl2/keyboard/struct.Mod.html#associatedconstant.MODEMOD instead of using the from bits truncate path? I think it represents the AltGR key like you are doing

Kethku: https://github.com/Rust-SDL2/rust-sdl2/blob/73afe22a08443d6e494e24ed4b8133ca868b2f2e/src/sdl2/keyboard/mod.rs#L28 welp maybe not. Docs say it should, but it looks different...

mewhhaha: I thought it'd be that too, but MODEMOD seem to correspond to 0x4000 and not 0x0240

mewhhaha: 0x0240 would be RALTMOD and LCTRLMOD together

Kethku: Does AltGr only trigger if it is the Left Ctrl and the Right Alt? If you hold Left Alt and Right Ctrl does it do the same thing?

mewhhaha: Yes, I can accomplish the same thing with Left Alt and Right Ctrl as with Alt Gr when writing in VS Code for example.

mewhhaha: It seems any combination of ALT and CTRL will work as Alt Gr.

Kethku: ok. Heres my proposal then:
```
let ctrl = modifiers.contains(Mod::LCTRLMOD) || modifiers.contains(Mod::RCTRLMOD);
let alt = modifiers.contains(Mod::LALTMOD) || modifiers.contains(Mod::RALTMOD);
let gui = modifiers.contains(Mod::LGUIMOD) || modifiers.contains(Mod::RGUIMOD);
let will_text_input = 
    modifiers.contains(Mod::MODEMOD) ||
    (ctrl && alt) ||
    !ctrl && !alt && !gui;
```
Thoughts?

mewhhaha: I will try it out! 😄

Kethku: For extra explicit-ness we could define altGr to equal ctrl && alt, but I could take that change or leave it

mewhhaha: It seems to be working for all the combinations that I could think of. Should I update this draft with that?

Kethku: Yes! And I'll get it merged

src/bridge/keybindings.rs
```diff
@@ -46,9 +46,9 @@ pub fn parse_keycode(keycode: Keycode) -> Option<(&'static str, bool)> {
         Keycode::Num9 => Some(("9", false)),
         Keycode::Colon => Some((":", false)),
         Keycode::Semicolon => Some((";", false)),
-        Keycode::Less => Some(("lt", true)),
```
Kethku: Special in this case means that the resulting keybinding should be wrapped in brackets. This is necessary because < and > are the escaping characters themselves. So we turn Keycode::Less into <lt> instead of the < character directly. I suspect that this change would mess up being able to type those characters at all. What made you make this change, and what happens if it is not present?

mewhhaha: I see! The issue that I'm having is that < and > are on the same button. So to get > I press shift + <. When `lt` returns false I can get > by pressing that key combination. `gt` being false I don't think has an effect but is there from when I was testing.

Kethku: Hrmmm... Your change would cause problems if someone binds < or > with a modifier. In your branch does holding ctrl and pressing < do something weird?

mewhhaha: If `lt` is true then I will get < even if I press shift + <. I'm guessing anything that will make this key combination return would work.

mewhhaha: I can't see any weird behaviour. But I'm not used to Vim. I'm only trying it out plainly in normal and insert mode.

mewhhaha: I get the same behaviour as when I run nvim in my Powershell terminal, that it tabs the row left and right one step if I do << and >> respectively.

Kethku: Interesting. In that case I'll pull and test it out locally

mewhhaha: Oh, sorry. I guess the weird behaviour is that I'm _able_ to output these characters. When I use nvim in my terminal I get the characters ^\ as output.

Kethku: What I'm confused then. Which is the desired behavior?

mewhhaha: I don't know for Ctrl + <. The only issue is that on a Nordic(?) keyboard you need to press Shift + < to output >.

mewhhaha: I guess you can do an extra case for (shift || !special) but that will affect a lot of characters and I don't know if that'd cause more trouble? The least intrusive but oddly specific I guess would be (key_text == "lt" && shift || !special). Apparently the behaviour I observed in terminal nvim for Ctrl + < and Ctrl + > doesn't exist on master for NeoVide either. So not exclusive for this request.

Kethku: its actually not too odd to special case < and > because they are the only keys which produce a character and also need to be escaped

So I actually think thats the best option.

mewhhaha: Thank you!

Kethku: No thank you! This has been bugging me for a while

mewhhaha: I will do some extra testing on my Linux machine on Monday, but at least I can use it on my Windows machine now! 👍 Been great using it so far!

PR #188, opened 2020-03-14
feat: add desktop entry for linux systems

acheronfail:
This PR just adds a `.desktop` file that Linux users can use to start Neovide from their Desktop manager.

We can add this to the AUR for Linux so it's installed there.

### DIFF
--- a/assets/neovide.desktop
+++ b/assets/neovide.desktop
@@ -1,8 +1,9 @@
 [Desktop Entry]
 Type=Application
-Exec=neovide
+Exec=neovide %F
 Icon=neovide
 Name=Neovide (nvim)
 Keywords=Text;Editor;
 Categories=Utility;TextEditor;
 Comment=No Nonsense Neovim Client in Rust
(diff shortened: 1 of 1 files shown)


ericonr:
You should probably add a `%F` argument to the `Exec` line, so it can be used to open files as well. Ideally, a mime-type as well.

For an example, the .desktop file for nvim also contains this:

```
TryExec=nvim
Exec=nvim %F
Terminal=true
Type=Application
Keywords=Text;editor;
Icon=nvim
Categories=Utility;TextEditor;
StartupNotify=false
MimeType=text/english;text/plain;text/x-makefile;text/x-c++hdr;text/x-c++src;text/x-chdr;text/x-csrc;text/x-java;text/x-moc;text/x-pascal;text/x-tcl;text/x-tex;application/x-shellscript;text/x-c;text/x-c++;
```

acheronfail:
Good suggestions, thanks @ericonr. :)

Kethku:
I have no idea what this does, but I see no problem adding it ;)

PR #192, opened 2020-03-15
Make embeddable

j4qfrost:
* Added check for bridge connection still running
* Changed inner window to Option for dropping
* Take sdl context as a parameter for new function to tie event loops together
* Add lib file


src/bridge/mod.rs
```diff
@@ -139,11 +150,13 @@ impl Bridge {
         runtime.spawn(async move {
             start_process(receiver).await;
         });
-
-        Bridge { _runtime: runtime, sender }
+        Bridge { _runtime: runtime, sender, running: Arc::new(AtomicBool::new(true)) }
```
Kethku: being an atomic value, you can share references. No need to use an arc or clone the reference.

src/bridge/mod.rs
```diff
@@ -72,7 +75,8 @@ async fn start_process(mut receiver: UnboundedReceiver<UiCommand>) {
             },
             Ok(Ok(())) => {}
         };
-        std::process::exit(0);
+        // std::process::exit(0);
```
Kethku: If this is the play, we should probably just remove the comment.

src/editor/mod.rs
```diff
@@ -37,52 +37,75 @@ pub struct Editor {
...
             RedrawEvent::ModeInfoSet { cursor_modes } => self.cursor.mode_list = cursor_modes,
             RedrawEvent::OptionSet { gui_option } => self.set_option(gui_option),
-            RedrawEvent::ModeChange { mode_index } => self.cursor.change_mode(mode_index, &self.defined_styles),
+            RedrawEvent::ModeChange { mode_index } => {
+                self.cursor.change_mode(mode_index, &self.defined_styles)
+            }
             RedrawEvent::BusyStart => {
                 trace!("Cursor off");
                 self.cursor.enabled = false;
-            },
```
Kethku: I assume most of the changes on this file are from rustfmt?

Kethku: I can't tell if there are any functional changes in this diff. Maybe revert any formatting changes? I'm interested in adding rustfmt globally at some point, but doing it on only one file seems distracting for this PR

Kethku:
Overall I'm interested in these changes, but the formatting changes make it too difficult to tell what actually changed. So either we should make a separate PR which implements rustfmt changes across the whole codebase, and then rebase this PR, or you should remove the formatting specific changes so that I can actually tell whats going on.

All that said, I'm very interested in implementing changes to enable this use case! Thanks for the effort!

src/editor/mod.rs
```diff
@@ -37,52 +37,75 @@ pub struct Editor {
...
+            RedrawEvent::Scroll {
+                top,
+                bottom,
+                left,
+                right,
+                rows,
+                columns,
+                ..
+            } => self.scroll_region(top, bottom, left, right, rows, columns),
+            RedrawEvent::WindowClose { .. } => self.grid.clear(),
```
j4qfrost: This line is a bit preemptive. I was still learning a bit about how neovide works and figured it might be useful someday to add debugging info here.

src/editor/mod.rs
j4qfrost: Sorry about that. I started working on this before it crossed my mind about whether or not I would even be sharing this. I'm reverting the formatting for most of the file. The only major change is that I switched out the global variable for a getter.

src/window.rs
```diff
@@ -105,50 +121,66 @@ impl WindowWrapper {
...
+            let fullscreen = { SETTINGS.get::<WindowSettings>().fullscreen };
+            if self.fullscreen != fullscreen {
+                let state = match fullscreen {
+                    true => FullscreenType::Desktop,
+                    false => FullscreenType::Off,
+                };
+                window.set_fullscreen(state).ok();
+                self.fullscreen = fullscreen;
+            }
+            self.window = Some(window);
```
j4qfrost: I tried to look for a more elegant solution to retain the window, but it looks like this is the only way to do it.

Kethku:
Can you rebase and force push to fix the diffs?

j4qfrost:
On it.

PR #196, opened 2020-03-16
Tweak to mac instructions to include how to build the .app file

svermeulen:


Kethku:
Correct me if I'm wrong, but this probably requires installing the cargo bundle tool right?

svermeulen:
I am not sure since I still haven't learned rust  :|  But it sounds like it

I just know that as a mac I find it a lot more useful to run it from the generated `.app` folder rather than the executable, and took me some time to figure this out, so thought I'd add this

j4qfrost:
https://github.com/burtonageo/cargo-bundle

Looking at the usage for the cargo-bundle tool, it appears this isn't as simple as just calling cargo bundle. It's not hard, but it's probably not a pressing issue to get to. Where exactly did you find the instruction to just call cargo bundle?

svermeulen:
This was taken from the `.github/workflows/build.yml` file which does this to create a build:

```
        cargo test
        cargo build --release
        cargo install cargo-bundle
        cargo bundle --release
```

This does already produce an icon for the app as well

j4qfrost:
Huh, thanks didn't see that.

Kethku:
My only concern with this is that for most, downloading either the release or the github action version should be the way to build an app folder if they want it. Building from source should probably default to just using `cargo run` so that you don't have to go through the bundling process.

What do you think? Maybe a better change would be to add a comment to the end of the existing instructions describing how to build an app folder if desired? That way the documentation is there but the normal strategy is to run `cargo run` or download a release or ci build artifact?

I suspect part of the issue is that its not obvious the github actions produce actual artifacts. So maybe I could add comments to the readme describing this strategy if someone wants a build without having to build themselves.

svermeulen:
Yeah that makes a lot of sense to me.  Currently it suggests either grabbing it from releases or building yourself but does not mention getting it through github actions.  So for people like me that just want an up to date .app build that would be enough.  I'll close this

PR #199, opened 2020-03-17
Embedding changes

j4qfrost:
Rebasing went belly up, so I just branched off the fmt branch.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -1,21 +1,23 @@
+use std::sync::atomic::Ordering;
 use std::thread::sleep;
 use std::time::{Duration, Instant};
 
 use log::{debug, error, info, trace};
 use skulpin::sdl2;
+use skulpin::sdl2::Sdl;
+use skulpin::{LogicalSize,PhysicalSize};
+
 use skulpin::sdl2::event::{Event, WindowEvent};
 use skulpin::sdl2::keyboard::Keycode;
 use skulpin::sdl2::video::{FullscreenType, Window};
-use skulpin::sdl2::Sdl;
 use skulpin::{dpis, CoordinateSystem, PresentMode, Renderer as SkulpinRenderer, RendererBuilder};
-use skulpin::{LogicalSize, PhysicalSize};
 
 use crate::bridge::{produce_neovim_keybinding_string, UiCommand, BRIDGE};
 use crate::editor::EDITOR;
+use crate::get_initial_dimensions;
 use crate::redraw_scheduler::REDRAW_SCHEDULER;
 use crate::renderer::Renderer;
 use crate::settings::*;
-use crate::INITIAL_DIMENSIONS;
 
 #[derive(RustEmbed)]
 #[folder = "assets/"]
@@ -42,29 +44,36 @@ fn handle_new_grid_size(new_size: LogicalSize, renderer: &Renderer) {
     }
 }
 
-struct WindowWrapper {
-    context: Sdl,
-    window: Window,
-    skulpin_renderer: SkulpinRenderer,
-    renderer: Renderer,
-    mouse_down: bool,
-    mouse_position: LogicalSize,
-    title: String,
-    previous_size: LogicalSize,
-    previous_dpis: (f32, f32),
-    transparency: f32,
-    fullscreen: bool,
+pub struct WindowWrapper {
+    pub context: Sdl,
+    pub window: Option<Window>,
+    pub skulpin_renderer: SkulpinRenderer,
+    pub renderer: Renderer,
+    pub mouse_down: bool,
+    pub mouse_position: LogicalSize,
+    pub title: String,
+    pub previous_size: LogicalSize,
+    pub previous_dpis: (f32, f32),
+    pub transparency: f32,
+    pub fullscreen: bool,
 }
 
 impl WindowWrapper {
-    pub fn new() -> WindowWrapper {
-        let context = sdl2::init().expect("Failed to initialize sdl2");
+    pub fn new(context: Option<Sdl>) -> WindowWrapper {
+        let ctx = match context {
+            Some(ctx) => ctx,
+            None => sdl2::init().expect("Failed to initialize sdl2"),
+        };
+
+        Self::with_context(ctx)
+    }
+
+    pub fn with_context(context: Sdl) -> WindowWrapper {
         let video_subsystem = context
             .video()
             .expect("Failed to create sdl video subsystem");
-        video_subsystem.text_input().start();
 
-        let (width, height) = INITIAL_DIMENSIONS;
+        let (width, height) = get_initial_dimensions();
 
         let renderer = Renderer::new();
         let logical_size = LogicalSize {
@@ -112,7 +121,7 @@ impl WindowWrapper {
 
         WindowWrapper {
             context,
-            window,
+            window: Some(window),
             skulpin_renderer,
             renderer,
             mouse_down: false,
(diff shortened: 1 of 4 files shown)


src/window.rs
```diff
@@ -42,29 +44,36 @@ fn handle_new_grid_size(new_size: LogicalSize, renderer: &Renderer) {
...
 impl WindowWrapper {
-    pub fn new() -> WindowWrapper {
-        let context = sdl2::init().expect("Failed to initialize sdl2");
+    pub fn new(context: Option<Sdl>) -> WindowWrapper {
+        let ctx = match context {
+            Some(ctx) => ctx,
+            None => sdl2::init().expect("Failed to initialize sdl2"),
+        };
+
+        Self::with_context(ctx)
```
Kethku: Minor: I'm not sure why this was extracted into a separate method. Is with_context used outside of the constructor?

src/window.rs
```diff
@@ -42,29 +44,36 @@ fn handle_new_grid_size(new_size: LogicalSize, renderer: &Renderer) {
...
-    mouse_down: bool,
-    mouse_position: LogicalSize,
-    title: String,
-    previous_size: LogicalSize,
-    previous_dpis: (f32, f32),
-    transparency: f32,
-    fullscreen: bool,
+pub struct WindowWrapper {
+    pub context: Sdl,
+    pub window: Option<Window>,
```
Kethku: The introduction of the option here is rough because it forces any usage of the window to be wrapped in an if statement. Can you help me understand what the state with no available window looks like? Pretty sure neovide wouldn't work very well without a rendering surface or input?

src/window.rs
```diff
@@ -129,30 +138,30 @@ impl WindowWrapper {
...
-                .set_title(&self.title)
-                .expect("Could not set title");
-        }
-
-        let transparency = { SETTINGS.get::<WindowSettings>().transparency };
-        if let Ok(opacity) = self.window.opacity() {
-            if opacity != transparency {
-                self.window.set_opacity(transparency).ok();
-                self.transparency = transparency;
+        if let Some(mut window) = self.window.take() {
```
Kethku: This pattern seems weird to me. This SO answer suggestions there may be better ways: https://stackoverflow.com/a/27363626

src/window.rs
```diff
@@ -251,52 +262,62 @@ impl WindowWrapper {
...
+                {
+                    error!("Render failed. Closing");
+                    return false;
+                }
             }
+            self.window = Some(window);
+            return true;
         }
-        return true;
+        return false;
```
Kethku: Are you sure false is the correct value here? The render didn't fail, it just didn't draw. Same thing as when no redraw is scheduled. As is this will stop the draw loop. Is that the correct response?

src/window.rs
```diff
@@ -251,52 +262,62 @@ impl WindowWrapper {
...
     }
 }
 
 #[derive(Clone)]
-struct WindowSettings {
-    refresh_rate: u64,
+pub struct WindowSettings {
+    pub refresh_rate: u64,
```
Kethku: Why was this made public? It is the user configured refresh_rate, not the current fps in case thats the reason.

src/window.rs
```diff
@@ -42,29 +44,36 @@ fn handle_new_grid_size(new_size: LogicalSize, renderer: &Renderer) {
...
-    renderer: Renderer,
-    mouse_down: bool,
-    mouse_position: LogicalSize,
-    title: String,
-    previous_size: LogicalSize,
-    previous_dpis: (f32, f32),
-    transparency: f32,
-    fullscreen: bool,
+pub struct WindowWrapper {
+    pub context: Sdl,
```
Kethku: Do all of these need to be public?

Kethku:
This was way easier to review :) My only concern is about the overall added complexity. I'm curious if there would be a way to abstract the individual components of neovide to fit your purposes without making any of the individual components more complex.

One way we could do that would be to abstract the renderer, editor, and bridge components and then have your system implement the windowing logic yourself. Your app would be somewhat more complicated, but it would effectively be pumping the other components itself instead of forcing Neovide's window wrapper into a form that works for it.

I'm not suuuper against adding these changes, just trying to be intentional about where the burden of complexity lies. Any thoughts?

src/window.rs
j4qfrost: I wanted to keep neovide open so that I could create new windows without having to run the other initialization steps again, so I wanted to keep neovide alive and close the window. The cleanest way to do that was to allow rust to just drop the window. I could have also invoked SDL through sys to drop the window.

src/window.rs
j4qfrost: I'm not too clear on the suggestion. Would match be better?

Kethku: I may be missing something, but to my eyes this:

```
impl Foo {
    fn send(&mut self) {
        if let Some(x) = &mut self.stream {
            *x += 1;
        }
    }
}
```

Feels better to me.

src/window.rs
j4qfrost: You're right. Just checked the logic with ui loop.

src/window.rs
Kethku: This is related to my comment below. I wonder if a better way of achieving this would be to implement your own window, and send UICommands to the bridge, and do equivalent draw logic in your app. Then whenever the editor component resizes, you get imput events etc you filter and pass the relevant ones to neovide's systems. This way neovide doesn't have to have more complexity, and you get full control over the renderer etc.

src/window.rs
j4qfrost: Gotcha.

src/window.rs
j4qfrost: Sorry, this is where my lack of familiarity with rust sort of fails me. It looks better, but it looks like something I've tried and got dinged for, because SDL window doesn't derive Copy. I'll give the pattern a shot.

Kethku: pretty sure this shouldn't require copy... But what do I know

src/window.rs
j4qfrost: Totally down for that. Though would this entail a full reengineering? Are you thinking of using sender and receiver logic to decouple the event handling from the renderer? I also had some ideas about using a texture instead of a full window, but the windowing is nice to have.

src/window.rs
j4qfrost: Probably not. I'll take a deeper look.

src/window.rs
j4qfrost: I'll give this a try!

src/window.rs
Kethku: Today I believe that the editor, bridge, and renderer components don't depend on the window. They are consumed by the window system. You'd have to pull in skribo (or produce your own skia-safe canvas) so that you can pass  a canvas to the renderer draw function. Similarly youd have to push ui events to the bridge and pull draw commands from the editor, but neither of those require anything specific to the window

Kethku: One bit thats somewhat weird is that the bridge and editor components are hard coded singletons. So maybe an action item on my side that I wouldn't mind doing would be to remove the singleton relationship. That way you could initialize a new neovide process and spin it down multiple times over the course of your app's lifetime

j4qfrost: That sounds great. I'm stepping back and looking at the WindowWrapper, and I think it can be converted into a trait and impl.

Kethku: Feel free to ping me with questions or w/e. I can respond here or on the gitter chat if you prefer.

j4qfrost: This also eliminates the need for making public struct members.

Kethku: Totally! I think its a more principled change overall.

Kethku: Note: I'm not proposing that we change the windowmanager into a trait because it isn't used by anything. In neovide today, the windowmanager uses and manages everything else. So your app would implement something similar, but doesn't need to use a unified trait because it wouldn't be consumed by anything on the neovide side

Kethku: The only interface is the skia-safe renderer which uses the sdl2 window. Other than that, everything goes from the window manager out to the other components. Not the other way around.

j4qfrost: That sounds like a good entrypoint for anyone who wants to sort of plug into the other components. Not sure if we're in agreement on whether to make a WindowWrapper component or not. As a person who is more of a "customer", I think I would find a window trait to be useful in helping cover the full functionality for text editing. If the trait is too restrictive, then I could just use the individual components.

Kethku: Its not really about being restrictive so much as that trait wouldn't be used by anything. Its very possible I'm misunderstanding your idea though so don't let that stop you :)

j4qfrost:
I'll check to see if the send pattern works. For now, I think I'll branch off master and checkout the changes for the bridge. That sound good?

Kethku:
Sounds great to me! Let me know if theres anything I can do to help!

j4qfrost:
Closing this since it's stale and changes should be made more granular

PR #201, opened 2020-03-18
Allow user provided geometry

aelkazdadi:
Allow the user to pass the window geometry as a command line flag `--geometry=<width>x<height>`

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -56,6 +56,48 @@ struct WindowWrapper {
     fullscreen: bool,
 }
 
+pub fn window_geometry() -> Result<(u64, u64), String> {
+    let prefix = "--geometry=";
+
+    std::env::args()
+        .filter(|arg| arg.starts_with(prefix))
+        .next()
+        .map_or(Ok(INITIAL_DIMENSIONS), |arg| {
+            let input = &arg[prefix.len()..];
+            let invalid_parse_err = format!(
+                "Invalid geometry: {}\nValid format: <width>x<height>",
+                input
+            );
+
+            input
+                .split('x')
+                .map(|dimension| {
+                    dimension
+                        .parse::<u64>()
+                        .or(Err(invalid_parse_err.as_str()))
+                        .and_then(|dimension| {
+                            if dimension > 0 {
+                                Ok(dimension)
+                            } else {
+                                Err("Invalid geometry: Window dimensions should be greater than 0.")
+                            }
+                        })
+                })
+                .collect::<Result<Vec<_>, &str>>()
+                .and_then(|dimensions| {
+                    if let [width, height] = dimensions[..] {
+                        Ok((width, height))
+                    } else {
+                        Err(invalid_parse_err.as_str())
+                    }
+                })
+                .map_err(|msg| msg.to_owned())
+        })
+}
+pub fn window_geometry_or_default() -> (u64, u64) {
+    window_geometry().unwrap_or(INITIAL_DIMENSIONS)
+}
+
 impl WindowWrapper {
     pub fn new() -> WindowWrapper {
         let context = sdl2::init().expect("Failed to initialize sdl2");
@@ -64,7 +106,7 @@ impl WindowWrapper {
             .expect("Failed to create sdl video subsystem");
         video_subsystem.text_input().start();
 
-        let (width, height) = INITIAL_DIMENSIONS;
+        let (width, height) = window_geometry_or_default();
 
         let renderer = Renderer::new();
         let logical_size = LogicalSize {
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -17,7 +17,7 @@ use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender};
 
 use crate::error_handling::ResultPanicExplanation;
 use crate::settings::*;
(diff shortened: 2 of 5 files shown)


src/window.rs
```diff
@@ -56,6 +56,48 @@ struct WindowWrapper {
...
 }
 
+pub fn window_geometry() -> Result<(u64, u64), String> {
+    let prefix = "--geometry=";
+    let arg = std::env::args()
+        .filter(|arg| arg.starts_with(prefix))
+        .next();
+    match arg {
+        Some(sizes_str) => {
+            let input = sizes_str.split_at(prefix.len()).1;
```
Kethku: This could probably be simplified with something like https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=827c58281dd17cf14eb720675b255f3c

src/window.rs
```diff
@@ -56,6 +56,48 @@ struct WindowWrapper {
...
+                    .and_then(|s| s.parse::<u64>().ok())
+                    .ok_or_else(err_msg)
+            };
+            width = parse_next()?;
+            height = parse_next()?;
+
+            if sizes.next().is_some() {
+                Err(err_msg())
+            } else if width == 0 || height == 0 {
+                Err("Invalid geometry: Window dimensions should be greater than 0.".into())
```
Kethku: This error information doesn't appear to be surfaced to the user anywhere

Kethku: My preference is to surface it by an error! log which I believe is configured to output to stderr by default

Kethku:
I made some minor comments. Overall looks like a good change. Minor thing I was wondering: maybe we should call it window dimensions? Geometry feels like too broad a term.

aelkazdadi:
I "borrowed" the term from neovim-qt, for the sake of consistency. But if you want to change it, then I don't mind.

src/window.rs
aelkazdadi: The error is returned to `main` in case of failure, then printed with `eprintln!`. Would you rather we exit the process from inside the function?

Kethku:
I don't have a strong opinion either way. If theres existing patterns then thats fine.

src/window.rs
Kethku: Ah gotcha. No thats fine and I think your handling is reasonable.

aelkazdadi:
I've cleaned up the code a bit. But I'm wondering if we should apply a scale factor to the input dimensions. `--geometry=50x50` results in a tall window, while `--geometry=100x50` gives me a square one.
Should we multiply the width by two so that equal dimensions result in a square window?

Kethku:
This isn't about the scale factor. Its about how fonts work. Fonts tend to be twice as tall as they are wide. I think it makes more sense to give window sizes either in terms of pixels or grid sizes. Trying to fix the grid size to match a pixel ratio seems like it would be confusing

src/window.rs
```diff
@@ -56,6 +56,39 @@ struct WindowWrapper {
...
+    match arg {
+        Some(arg) => {
+            let input = &arg[prefix.len()..];
+
+            if let [Some(width), Some(height)] = input
+                .split('x')
+                .map(|dimension| dimension.parse::<u64>().ok())
+                .collect::<Vec<_>>()[..]
+            {
+                if width > 0 && height > 0 {
```
Kethku: maybe pull the width and height check into the map function? No reason to encode the operation twice. Further, since they both have the same error, you could use `or_else` to turn the parse error into the invalid format string and the greater than zero error into the invalid geometry string. That way you don't have to have control flow here. Even further, the outer match can be turned into a single map with an `unwrap_or` at the end for INITIAL_DIMENSIONS

Kethku: My goal here isn't to nitpick, but to make the code more concise. I think its important for maintainability and at a glance understandability.

aelkazdadi: No worries. I'm not experienced enough with Rust yet, so I really appreciate your feedback.

aelkazdadi: I removed the outer match, and rewrote the inner logic, but I'm not sure if the `match` block is better this way. I need to make sure there's no `ParseError` when checking for zeros because I want to return a parsing error message on an input like `0xabc`.
What do you think?

Kethku: Maybe something like this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=75303220a1d75e15ae903e9a51abec9b

aelkazdadi: Yeah, looks good to me. Anything else?

src/window.rs
```diff
@@ -56,6 +56,48 @@ struct WindowWrapper {
...
+                        .and_then(|dimension| {
+                            if dimension > 0 {
+                                Ok(dimension)
+                            } else {
+                                Err("Invalid geometry: Window dimensions should be greater than 0.")
+                            }
+                        })
+                })
+                .collect::<Result<Vec<_>, &str>>()
+                .and_then(|dimensions| {
```
Kethku: welp :l I missed that we needed a tuple instead of a vec was a thing, so my proposal isn't much better really. Whatever. I'll just merge it and think about it some more to see if theres something to be done.

Kethku:
Thank you so much for putting up with my nitpicks :)

aelkazdadi:
Not at all. It's been very educational. Thanks!

PR #215, opened 2020-03-27
Add WSL option for neovide on Windows

NicholasJarr:
Added an command line argument to start the neovim process on the Windows WSL. I have little experience in Rust, so feel free to tell me if I did something wrong. This is my first pull request to an open source project, so any tips would be appreciated 😃.


Kethku:
Does this work? I thought somebody tried this with mixed success...

NicholasJarr:
My first try resulted in an error too. However, I figured out that my neovim version on WSL was lower than 0.4. After compiling from source it worked.

PR #222, opened 2020-03-31
Skulpin master

jotacosta:
* Updated the window.rs to Skulpin master codebase (v0.9.0)
* Updated the toml to fetch skulpin v0.9.0 with just the sdl2
* Changed mouse / resize based on LogicalSize
* Remove dpis from the codebase

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,8 +7,7 @@ build = "build.rs"
 description = "A simple GUI for Neovim."
 
 [features]
-default = ["sdl2-static-link", "embed-fonts"]
-sdl2-static-link = ["skulpin/sdl2-static-link"]
+default = ["embed-fonts"]
 embed-fonts = []
 
 [dependencies]
@@ -16,7 +15,7 @@ euclid = "0.20.7"
 font-kit = "0.5.0"
 skribo = { git = "https://github.com/linebender/skribo" }
 lru = "0.4.3"
-skulpin = { git = "https://github.com/Kethku/skulpin", branch = "sdl2" }
+skulpin = {version = "0.9.0", features=["skulpin_sdl2"]}
 # skulpin = { path = "../skulpin" }
 derive-new = "0.5"
 rmpv = "0.4.4"
--- a/src/window.rs
+++ b/src/window.rs
@@ -38,7 +38,6 @@ fn handle_new_grid_size(new_size: LogicalSize, renderer: &Renderer) {
     if new_size.width > 0 && new_size.height > 0 {
         let new_width = ((new_size.width + 1) as f32 / renderer.font_width) as u32;
         let new_height = ((new_size.height + 1) as f32 / renderer.font_height) as u32;
-        // Add 1 here to make sure resizing doesn't change the grid size on startup
         BRIDGE.queue_command(UiCommand::Resize {
             width: new_width,
             height: new_height,
@@ -46,7 +45,6 @@ fn handle_new_grid_size(new_size: LogicalSize, renderer: &Renderer) {
     }
 }
 
-// removed window and dpi
 struct WindowWrapper {
     context: Sdl,
     window: sdl2::video::Window,
@@ -223,18 +221,12 @@ impl WindowWrapper {
 
     pub fn handle_pointer_motion(&mut self, x: i32, y: i32) {
         let previous_position = self.mouse_position;
-        // TODO : This is just so that it compiles
         let physical_size = PhysicalSize::new(
             (x as f32 / self.renderer.font_width) as u32,
             (y as f32 / self.renderer.font_height) as u32,
         );
-        // if let Ok(new_mouse_position) = physical_size.to_logical(0.5) {
-        // LogicalSize::from_physical_size_tuple( (
-        //     (x as f32 / self.renderer.font_width) as u32,
-        //     (y as f32 / self.renderer.font_height) as u32,
-        // ),
-        // &self.window,
-        // {
+
+        // TODO : not sure this is the best way to get mouse position
         self.mouse_position = physical_size.to_logical(1.0);
         if self.mouse_down && previous_position != self.mouse_position {
             BRIDGE.queue_command(UiCommand::Drag(
@@ -242,7 +234,6 @@ impl WindowWrapper {
                 self.mouse_position.height,
             ));
         }
-        // }
     }
 
     pub fn handle_pointer_down(&mut self) {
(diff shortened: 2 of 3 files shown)


src/window.rs
```diff
@@ -218,20 +221,18 @@ impl WindowWrapper {
...
-            &self.window,
-        ) {
-            self.mouse_position = new_mouse_position;
-            if self.mouse_down && previous_position != self.mouse_position {
-                BRIDGE.queue_command(UiCommand::Drag(
-                    self.mouse_position.width,
-                    self.mouse_position.height,
-                ));
-            }
+        let physical_size = PhysicalSize::new(
```
j4qfrost: Why not just do LogicalSize::new here? Did that not compile?

jotacosta: I just assumed that the ` pub fn handle_pointer_motion(&mut self, x: i32, y: i32) {` receives x,y as PhysicalSize and that needs converting to LogicalSize. Therefore, I'm doing PhysicalSize::new and then converting with PhysicalSize.to_logical(1.0) .

Now, I agree with you. I am not sure this is the best way (and the 1.0 working suggest that LogicalSize and PhysicalSize are the same). But I didn't try LogicalSize to be honest.

j4qfrost: I personally can't test this, because I just realized now that I don't even have the mouse functionality for moving the cursor. As long as it works and is more readable, it should be good to commit.

jotacosta: Just tested on OSX, works.

Cargo.toml
```diff
@@ -7,16 +7,15 @@ build = "build.rs"
...
+default = ["embed-fonts"]
 embed-fonts = []
 
 [dependencies]
 euclid = "0.20.7"
 font-kit = "0.5.0"
 skribo = { git = "https://github.com/linebender/skribo" }
 lru = "0.4.3"
-skulpin = { git = "https://github.com/Kethku/skulpin", branch = "sdl2" }
+skulpin = {version = "0.9.0", features=["skulpin_sdl2"]}
```
Kethku: rumor has it we can add back the sdl2-static-link by overriding the sdl2 dependency. We should probably figure out how that works before merging this

src/window.rs
```diff
@@ -34,7 +38,6 @@ fn handle_new_grid_size(new_size: LogicalSize, renderer: &Renderer) {
     if new_size.width > 0 && new_size.height > 0 {
         let new_width = ((new_size.width + 1) as f32 / renderer.font_width) as u32;
         let new_height = ((new_size.height + 1) as f32 / renderer.font_height) as u32;
-        // Add 1 here to make sure resizing doesn't change the grid size on startup
         BRIDGE.queue_command(UiCommand::Resize {
             width: new_width,
```
Kethku: I believe this is still necessary. The conversion here isn't about dpi. Its for converting pixel sizes to grid sizes. So this change should be reverted or else window resize will seriously mess things up

src/window.rs
```diff
@@ -218,20 +221,18 @@ impl WindowWrapper {
 
     pub fn handle_pointer_motion(&mut self, x: i32, y: i32) {
         let previous_position = self.mouse_position;
-        if let Ok(new_mouse_position) = LogicalSize::from_physical_size_tuple(
```
Kethku: Again, mouse positions are handled in grid coordinates. So this conversion is still necessary.

Kethku: Oh I see. You are. My bad

src/window.rs
Kethku: The scaling factor is about the dpi scaling. This will work if you don't have any scaling enabled, but if the screen you are on has some sort of scaling, you'll run into issues. I believe we will have to create a skulpin window wrapper here and ask it for the scale factor

src/window.rs
```diff
@@ -293,24 +294,11 @@ impl WindowWrapper {
...
-                let physical_size = PhysicalSize::new(&self.window);
-                self.window
-                    .set_size(
-                        (physical_size.width as f32 * new_dpis.0 / self.previous_dpis.0) as u32,
-                        (physical_size.height as f32 * new_dpis.1 / self.previous_dpis.1) as u32,
-                    )
-                    .unwrap();
-                self.previous_dpis = new_dpis;
-            }
+        let sdl_window = Sdl2Window::new(&self.window);
```
Kethku: You use the name sdl_window to refer to 2 concepts. Above, its the actual sdl window and here its the sdl window wrapper. I'm curious if a better name could be used here to make sure its obvious that this is the wrapper?

Cargo.toml
j4qfrost: I usually add the dependency to [patch.crates-io]. That's how I fixed my build errors in my game project when integrating neovide.

src/window.rs
j4qfrost: wrapper_window?

src/window.rs
```diff
@@ -140,6 +142,8 @@ impl WindowWrapper {
             .expect("Failed to create window");
         info!("window created");
 
+        let window = Sdl2Window::new(&sdl_window);
```
Kethku: Maybe grouping window and skulpin_renderer in a local scope which returns the created window would help to show that the sdl2window isn't used again and is immediately dropped? I believe the compiler is doing that under the covers or else a borrow error would have showed up at creation of the WindowWrapper.

src/window.rs
Kethku: https://github.com/aclysma/skulpin/blob/master/skulpin-renderer-sdl2/src/lib.rs#L33

jotacosta: Yeah. You're right. We should ask the system what the scaling is and do that. 

On Linux I have "scaling" by xrandr into the desired resolution. And i think OSX does it 'transparent' to applications. 

We might have to test this a bit better, let me know if you need help on this.

src/window.rs
jotacosta: Looking back, 

I would use window as before for the ...subsystem and the sdl_window for the Sdl2Window::new.

This is a very good point, needs changing before accepted.

src/window.rs
Kethku: I can do the fix after this merges. But yes I believe this will need manual testing on all the platforms. I suspect there will be necessary upstream changes as well

Cargo.toml
Kethku: That makes sense. Can you specify features that way?

j4qfrost: Yep, it's pretty much exactly the same as specifying any other dependency to my knowledge.

j4qfrost: e.g.
```toml
[patch.crates-io]
# [patch.crates-io.sdl2-sys]
sdl2 = { git = "https://github.com/Rust-SDL2/rust-sdl2.git" }
# # path = "../rust-sdl2/sdl2-sys"
sdl2-sys = { git = "https://github.com/Rust-SDL2/rust-sdl2.git" } 
```

Kethku: very cool

Kethku:
@jotacosta whats your status? I may branch this to fix the few remaining issues and merge a separate PR if you can't get to this. I think it may fix a number of issues with dpi people are running into, so I'm anxious to get this in so we can start working out the remaining upstream bugs

jotacosta:
Sorry, I totally thought you were going to take it from here.

I can do it between tonight and tomorrow. I realise we now have conflicts, so I might have to merge master back into this and re-PR. 

- [ ] Revisit window and sdl_window name
- [ ] Try the new SDL2 linking in the .toml
- [ ] Ask the system what the scaling is?

Am I missing anything? Do you want to take over, or should i do these things?

Kethku:
No no. I can take it from here. I didn't want to step on your toes.

jotacosta:
Ok. So I'm on the assumption that you're going to take it from here. 

There are decisions on our discussion that might be more to do with how you want to add the SDL2.

Let me know if you need any help.

j4qfrost:
Got a bit distracted this week. Mind if I fix this up and merge it today? I want to start working on tests.

Kethku:
@j4qfrost I already rebased and such and will push to a new branch for collaboration. I got stuck on the skia safe feature configuration :(

Kethku:
New PR with the new in repo branch is here: https://github.com/Kethku/neovide/pull/240

Kethku:
Closing in favor of other PR

Issue #221, opened 2020-03-30
Macos Crash due to skulpin dropping object


j4qfrost:
```
libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
Abort trap: 6
```

This seems to happen because when ui_loop finishes something is dropped manually and then automatically drops, causing the MoltenVK to try to destroy a buffer that doesn't seem to exist. Here are the relevant bits of the crash report:
```
10  libc++.1.dylib                	0x00007fff7056454d std::__1::mutex::lock() + 29
11  libMoltenVK.dylib             	0x0000000131e7069f MVKDevice::destroyBuffer(MVKBuffer*, VkAllocationCallbacks const*) + 31
```

I'm not sure about the specifics about why this happens, but I am able to resolve the issue by disabling either this line https://github.com/Kethku/skulpin/blob/c0b194833f97b839bd44a30fc2f17851b14e1016/src/renderer/renderer.rs#L461 or this line https://github.com/Kethku/skulpin/blob/c0b194833f97b839bd44a30fc2f17851b14e1016/src/renderer/renderer.rs#L463 for the upstream skulpin dependency. 

The above "fix" may lead to memory leaks.

j4qfrost:
https://github.com/KhronosGroup/MoltenVK#memory-management

Maybe a better mind can make more sense of this than me.

Kethku:
I think the correct step for handling stuff like this bug is to first rebase on the master version of skulpin. As is I think the fork we are using (my private fork for the initial sdl2 support) is not sufficient and has diverged too much. @jotacosta has started some work on this over here: https://github.com/Kethku/neovide/issues/218 but if necessary I can take a quick stab to get it done sooner.

Further @j4qfrost I have invited you as a contributor on the project. I really appreciate the bug reporting and general contributions you've been doing.

j4qfrost:
Cool, thanks for the invite and kind words. I'll follow the issue. I took a look at the drop implementation for the most up-to-date version of skulpin, and it doesn't seem to be much different aside from the plugins dropping. Though I guess getting everyone on the same page is useful  in that we can start helping push skulpin development forward too.

jotacosta:
[...]

PR #231, opened 2020-04-02
fixes #226, and temporarly #221 based on j4qfrost

### LINKED ISSUE #226: Quit doesn't Quit
Continuing on #221, I tried @j4qfrost 's temporary solution to put process::exit after ui loop breaks and it seems to have alleviated the crashing problem however, (`:q` | `:qa`) doesn't quit Neovide, instead the window freezes. the animation loop seems to stop but the window doesn't exit anyway.

Reproduction (on Mojave)
1. Open new Neovide in an empty buffer
2. `:q`

exoticus:
This PR is all it takes to fix the ghost window issue discussed at #226 and temporarily mitigates the crashing issue #221 based on @j4qfrost suggestion of adding `process::exit`

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -1,3 +1,4 @@
+use std::sync::atomic::Ordering;
 use std::thread::sleep;
 use std::time::{Duration, Instant};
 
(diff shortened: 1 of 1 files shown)


j4qfrost:
@Kethku I think we need unit tests.

j4qfrost:
@exoticus thanks and sorry for the snippiness.

Kethku:
I think it makes sense to merge this even though it gets in the way of the embedding goals for now. Unit tests can come afterward.

j4qfrost:
Yeah, I had to get to my computer to merge them. I'll start working on tests now. One of the packages for input I'm trying isn't working so integration tests might be second pass @Kethku

PR #238, opened 2020-04-09
Shrink Binary Size

Kethku:


.vscode/launch.json
```diff
@@ -5,15 +5,16 @@
     "version": "0.2.0",
     "configurations": [
         {
-            "name": "(Windows) Launch",
-            "type": "cppvsdbg",
+            "name": "Debug",
```
Kethku, line 8: Probably not cleaned up yet. We should probably copy this instead of changing the existing one. I think these are probably changes to get debugging working on linux

.vscode/tasks.json
```diff
@@ -0,0 +1,18 @@
+{
+  // See https://go.microsoft.com/fwlink/?LinkId=733558 
+  // for the documentation about the tasks.json format
+  "version": "2.0.0",
+  "tasks": [
+    {
+      "type": "cargo",
```
Kethku, line 7: very cool

.vscode/launch.json
exoticus: yeah will write one to be cross platform. LLDB might not need much changes on linux too, have to check first though

Kethku: This article seems to indicate that cppvsdbg works better on windows but that clearly wont work on other platforms https://www.forrestthewoods.com/blog/how-to-debug-rust-with-visual-studio-code/

exoticus: @Kethku yeah we just need something that could give you step through code, and variable inspection. LLDB is decent, works in vscode and is cross platform too

src/renderer/caching_shaper.rs
```diff
@@ -258,24 +315,16 @@ impl CachingShaper {
         trace!("Font changed {:?} {:?}", &font_name, &base_size);
         self.font_name = font_name.map(|name| name.to_string());
         self.base_size = base_size.unwrap_or(DEFAULT_FONT_SIZE);
-        self.font_loader
-            .load(font_name.unwrap_or_default(), false, false);
+        self.font_set = FontSet::new(font_name, &mut self.font_loader);
```
exoticus, line 318: Changing font size is now instant, and changing to different font no longer reloads Symbols/Emojis etc

.vscode/launch.json
Kethku: No reason not to have both :) Vscode supports multiple configurations in this file

src/renderer/caching_shaper.rs
```diff
@@ -48,101 +44,150 @@ struct Asset;
...
 const DEFAULT_FONT_SIZE: f32 = 14.0;
 
-#[derive(new, Clone, Hash, PartialEq, Eq, Debug)]
-struct ShapeKey {
-    pub text: String,
-    pub bold: bool,
-    pub italic: bool,
+#[derive(Clone)]
+pub struct ExtendedFontFamily {
+    pub fonts: Vec<SkriboFont>
```
Kethku, line 49: Would it make sense to make this a hashmap with the properties as key?

src/renderer/caching_shaper.rs
```diff
@@ -48,101 +44,150 @@ struct Asset;
...
 }
 
-pub fn build_collection_by_font_name(
-    font_name: Option<&str>,
-    bold: bool,
-    italic: bool,
-) -> FontCollection {
-    let source = SystemSource::new();
+fn to_family(x: ExtendedFontFamily) -> FontFamily {
+    let mut y = FontFamily::new();
```
Kethku, line 76: More useful local variable names than x and y are preferable to ease reading

exoticus: yeah sorry forgot that there

src/renderer/caching_shaper.rs
```diff
@@ -48,101 +44,150 @@ struct Asset;
...
+    }
 }
 
-pub fn build_collection_by_font_name(
-    font_name: Option<&str>,
-    bold: bool,
-    italic: bool,
-) -> FontCollection {
-    let source = SystemSource::new();
+fn to_family(x: ExtendedFontFamily) -> FontFamily {
```
Kethku, line 75: This can be moved to the above impl block

src/renderer/caching_shaper.rs
exoticus: i didn't wanna deviate that much from the original FontFamily type, plus properties are already contained in the FontRef and we can reach them like [this](https://github.com/exoticusd/neovide/blob/9af54bf225cda844ecbd04209e1216c47d453c62/src/renderer/caching_shaper.rs#L66)

src/renderer/caching_shaper.rs
exoticus: for some reason i did fight a lot with the borrow checker for this but i can't remember why

.vscode/launch.json
exoticus: yeah sure why not. i initially didn't mean to remove your Windows config anyway idk why the hell i did that

src/renderer/caching_shaper.rs
```diff
@@ -67,16 +67,16 @@ impl ExtendedFontFamily {
...
         None
     }
-}
-
-fn to_family(x: ExtendedFontFamily) -> FontFamily {
-    let mut y = FontFamily::new();
 
-    for font in x.fonts {
-        y.add_font(font);
+    pub fn to_family(self) -> FontFamily {
```
Kethku, line 71: Maybe take self as a reference?

Kethku: probably a mutable reference

exoticus: > Maybe take self as a reference?

@Kethku so self here takes ownership, but it doesn't have implications so does it really differ from this in this case?

```rust
pub fn to_family(&self) -> FontFamily {
    let mut new_family = FontFamily::new();

    for font in &self.fonts {
        new_family.add_font(font.to_owned());
    }

    new_family
}
```

> probably a mutable reference

any particular reason?

Kethku: It was a dumb suggestion. I was misunderstanding the purpose of the function. Ignore

src/renderer/caching_shaper.rs
```diff
@@ -11,11 +11,12 @@ use skribo::{FontCollection, FontFamily, FontRef as SkriboFont, LayoutSession, T
 use skulpin::skia_safe::{Data, Font as SkiaFont, TextBlob, TextBlobBuilder, Typeface};
 
 use log::{trace, warn};
+use cfg_if::cfg_if as define;
 
 const STANDARD_CHARACTER_STRING: &str =
     "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890";
 
-cfg_if::cfg_if! {
+define! {
```
Kethku, line 19: Im not sure this improves understandability...

exoticus: :( it's like c ifdef, and looks nice™ too

Kethku: fair enough

Kethku:
I finally got to looking at this. I'm not sure what is going wrong in the branch, but my local build of this fork is having metric problems: 
image

I need to spend more time digging in, but I'm hesitant to merge this in the current state if it regress in this way.

Kethku:
I think it may be related to the fact that it isn't loading my configured font properly. I'm not sure what changed in the font discovery though...

pushqrdx:
@Kethku can you send me your font config because i think i know what's wrong

Kethku:
It's "Delugia\\ Nerd\\ Font". I can link to the actual config tomorrow

exoticus:
@Kethku yeah i see, it was [this](https://github.com/exoticusd/neovide/blob/ca7f984d306f20881f10a95e8c9362bec13e3aa6/src/renderer/caching_shaper.rs#L69) as i didn't handle if the font only had one variant like `Delugia Nerd Font`, only had Regular, so when bold/italic etc was getting added and aren't found nothing was added to the collection.

In the old implementation it automatically fell back to whatever found inside that font because of [`select_best_match`](https://github.com/Kethku/neovide/blob/a7a01c9247bbaa8f87f8d24a68766e3bb0cac132/src/renderer/caching_shaper.rs#L97)

Hopefully my latest commit fixed that though

Kethku:
Trying shortly

Kethku:
:( It still can't load my delugia nerd font.

exoticus:
@Kethku but i installed delugia and it worked, can you link me the font + the config so i can see what's causing that?

Kethku:
https://github.com/Kethku/vim-config
https://github.com/Kethku/vim-config/blob/master/fnl/dotfiles/module/settings.fnl#L57

My config is written in fennel (lispy lua) so I'll try again this evening. The delugia nerd font is from here: https://github.com/adam7/delugia-code/releases

exoticus:
@Kethku to confirm a suspicion can you try manually setting the font like that?
`set guifont=Delugia\ Nerd\ Font\ Book:h14`

Kethku:
font loading issue

Kethku:
Oh woops typo haha

Kethku:
font loading issue

Kethku:
This shows two separate issues:
1. The default font on windows has metrics problems
2. The system isn't loading the configured guifont because the font is wrong and the ligatures don't match.

It does appear to be loading the font size correctly, just not the font name

Kethku:
I will dig deeper in tonight

exoticus:
i believe it's a fontkit issue then, because it works fine on mac so for some reason select_family_by_name maybe doesn't load the font?
Hint: this doesn't look like it even loaded the font at all, because this is how neovide looks if no font were loaded (i can tell from the cursor being ahead of the text)

Kethku:
So yeah. I seem to recall that select_family_by_name was less preferable than the select_best_match because of this very reason. select_best_match will do a best effort matching of the font by name if it can't find an exact match.

This part of the codebase used select_family_by_name original but had to switch over before your changes.

As for the default font having cursor issues, that means that the default font (however it is selected) is incorrect because it isn't monospace which is required by vim's system.

exoticus:
@Kethku Consolas is monospace. So i am sure it's a loading issue then, but we need select_family because it saves us multiple file system calls if a font contains multiple variants. so instead of calling select_best_match 3 times for each variant, we load the whole thing and it becomes cheap loading it's variants [like this](https://github.com/Kethku/neovide/blob/187ead2e1fc9bcbf95ca398e10d3f5c9714c8e92/src/renderer/caching_shaper.rs#L60)

Kethku:
weirdly I don't think consolas is being loaded here because setting the guifont to consolas before this PR doesn't have the same metrics issues.

exoticus:
@Kethku yeah it's not loaded for sure, `select_family_by_name` should be working though,  so maybe you can fiddle a little bit deeper to see what seems to be the culprit? i do believe that the current implementation should work correctly on linux as it's on macos on my end. For some reason windows isn't though

Kethku:
And as confirmation, setting guifont to consolas in your branch works properly and removes the cursor gap

exoticus:
@Kethku so `select_family_by_name` works but not with Delugia for some weird reason?

Kethku:
Looks that way

Kethku:
I vaguely remember this as well. `select_family_by_name` might not handle spaces properly on windows...

exoticus:
@Kethku i'd check if in system fonts `Delugia` is exactly named `Delugia Nerd Font`, maybe something to do with space escaping? Maybe on windows it requires the extension, all these little quirks

exoticus:
@Kethku Something funny came to my head, it might be the backward slash `\` in the escaping as windows uses that as a path delimiter, so instead of loading Delugia Nerd Font it tries to look for \path\Delugia\Nerd\Font

Kethku:
maaaaybe... Guifont requires backslash and I'm pretty sure we escape the values in the guifont handling...

exoticus:
@Kethku  idk but i feel like that should be the culprit, because that explains why it runs on macos/linux normally as it uses forward `/` slash for path

Kethku:
Ok figured out some more bits. You were right and I was an idiot not to check my installed fonts. My desktop has the font installed by default, but my work laptop did not. My bad.

However the font setting isn't getting picked up on startup. So there is a bug still around where the font system isn't loading on startup but will load afterward.

Further the above investigation highlights that font fallback isn't working. The font system isn't using the default font if the configured font isn't found. This is a bug that needs fixed I think

Thanks again for the awesome investigation and sorry for nit picking these small issues

exoticus:
@Kethku i think i know the issue of not falling back as my current implementation assumes configuration to be the source of truth, however i believe it should be easily fixable by adding system defaults to the fallbacks

And you absolutely shouldn't be sorry for nit picking, that's why neovide is neat and polish, we shouldn't merge the changes unless they work for the better.

exoticus:
@Kethku could you checkout latest commit, it should resolve the issue. albeit needs to be refactored (very late at night so kinda butchered it)

Kethku:
This branch is working for me now! Awesome work.

src/renderer/caching_shaper.rs
```diff
@@ -48,101 +40,159 @@ struct Asset;
...
+
+        if let Some(handle) = &self.fonts.first() {
+            return Some(&handle.font);
+        }
+
+        None
+    }
+
+    pub fn to_family(self) -> FontFamily {
+        let mut new_family = FontFamily::new();
```
Kethku, line 75: I think this could be replaced by `FontFamily { fonts: self.fonts }` which would be faster than creating a new font family and adding each one.

src/renderer/caching_shaper.rs
```diff
@@ -48,101 +40,159 @@ struct Asset;
...
+impl FontLoader {
+    pub fn new() -> FontLoader {
+        FontLoader {
+            cache: LruCache::new(10),
+            source: SystemSource::new(),
+        }
+    }
+
+    fn get(&mut self, font_name: &str) -> Option<ExtendedFontFamily> {
+        if let Some(family) = self.cache.get(&String::from(font_name)) {
```
Kethku, line 99: This can be simplified with `cloned` https://doc.rust-lang.org/std/option/enum.Option.html#method.cloned

src/renderer/caching_shaper.rs
```diff
@@ -48,101 +40,159 @@ struct Asset;
...
+    }
+
+    fn load(&mut self, font_name: &str) -> Option<ExtendedFontFamily> {
+        let handle = match self.source.select_family_by_name(font_name) {
+            Ok(it) => it,
+            _ => return None,
+        };
+
+        let fonts = handle.fonts();
+        let mut family = ExtendedFontFamily::new();
```
Kethku, line 125: minor: maybe it would make sense to add a function `from_normal_font_family` as a "static" function to ExtendedFontFamily to simplify the logic here? Maybe not worth the effort.

Kethku:
I added a couple of minor comments, but in general I'm hugely impressed. This is a fantastic improvement in organization and is way more principled than the spaghetti that was here before. After you respond to the couple of comments I will merge this as soon as possible

src/renderer/caching_shaper.rs
exoticus: yeah the reason i did it like that is that skribo for some reason has fonts set to private... it sucks

src/renderer/caching_shaper.rs
exoticus: sweet

src/renderer/caching_shaper.rs
exoticus: nah, it was worth, more clean

src/renderer/caching_shaper.rs
Kethku: Ah I see. So this line means public only for the crate? I'm not familiar with Rust's access rules https://github.com/linebender/skribo/blob/master/src/collection.rs#L16

Kethku:
This is looking good to me. Anything else needed before I merge?

exoticus:
@Kethku i believe that's it

PR #240, opened 2020-04-10
progress toward switching to upstream skulpin

Kethku:
supersedes https://github.com/Kethku/neovide/pull/222


Kethku:
I can't figure out how to get cargo to allow upstream feature changes

PR #247, opened 2020-04-14
Fixes win32 input for Danish keyboard layout

R-Michelsen:
I've made a small change to the input handling to match that of neovim-qt which handles characters "[", "]", "@" and "\\" properly on my Danish keyboard layout which neovide didn't. 

Since I am largely unfamiliar with vim and its input model, I don't know if this breaks other layouts but I would assume not since neovim-qt uses this logic. In any case I thought I would open this PR in case you want to merge it.

The change is taken from [input_win32.cpp](https://github.com/equalsraf/neovim-qt/blob/master/src/gui/input_win32.cpp) lines 28-37 and is a minor change from what already existed in the codebase.

### DIFF
--- a/src/bridge/layouts/mod.rs
+++ b/src/bridge/layouts/mod.rs
@@ -64,20 +64,20 @@ fn append_modifiers(
     if result == "<" {
         result = "lt".to_string();
         special = true;
-    }
+    }    
 
     if shift {
         special = true;
         result = format!("S-{}", result);
     }
-    if ctrl {
+	if ctrl && !alt {
         special = true;
         result = format!("C-{}", result);
     }
-    if alt {
+    if alt && !ctrl {
         special = true;
         result = format!("M-{}", result);
-    }
+	}
     if cfg!(not(target_os = "windows")) && gui {
         special = true;
         result = format!("D-{}", result);


R-Michelsen:
I just noticed this seems to mess up clipboard functionality, (as is messed up in neovim-qt currently aswell) I will try to fix this then reopen the PR

R-Michelsen:
Actually, after more investigation believe the clipboard "issues" are present even in the master branch so I'll reopen the PR. I'm not sure they are issues really though, it may just be that I expect the app to paste the clipboard contents on CTRL+V but that might be easier to handle through vim mappings.

src/bridge/layouts/mod.rs
```diff
@@ -64,20 +64,20 @@ fn append_modifiers(
...
+	if ctrl && !alt {
         special = true;
         result = format!("C-{}", result);
     }
-    if alt {
+    if alt && !ctrl {
         special = true;
         result = format!("M-{}", result);
-    }
+	}
```
Kethku, line 80: This breaks C-M-{} bindings on keyboard layouts which allow such bindings

R-Michelsen: Hmm I see, thats interesting. I'll investigate further, as I noted I more or less stole this change from neovim-qt but perhaps they don't handle C-M-{} bindings either? In any case I'll have to further investigate why `[ ] @ \ ` and a few other keys weren't working for me in the first place.

Kethku: Interesting. Its very possible that I don't understand keyboard input in neovim yet either :(

R-Michelsen:
After some consideration, I think the best workaround for me is to fix the broken Alt+Gr keys in my init.vim like this:
```vim
" Fix broken Alt+Gr characters
inoremap <C-M-@> @
inoremap <C-M-£> £
inoremap <C-M-]> ]
inoremap <C-M-[> [
inoremap <C-M-\> \
```
Ultimately from what I can tell each editor has its own implementation that accommodates some set of keyboard layouts but probably not all. Therefore it's probably best to leave the implementation as is. Feel free to include the work-around somewhere in your docs if you want, but it doesn't seem like a frequent issue, maybe we aren't many Danish vim users or something 😃

Kethku:
Unfortunately I don't know what the "correct" solution is :( Its a hard problem which is made harder by the windowing system.

PR #248, opened 2020-04-14
Toggleable fullscreen (Alt+Enter)

R-Michelsen:
I had a few issues with the way the fullscreen behaviour worked so I made this PR to try and improve it.

I've improved the borderless fullscreen behaviour loosely based on https://discourse.libsdl.org/t/true-borderless-fullscreen-behaviour/24622/11. It works seamlessly on my multi-monitor Windows 10 setup. 

I also added toggleability with Alt+Enter.

### DIFF
--- a/src/bridge/layouts/mod.rs
+++ b/src/bridge/layouts/mod.rs
@@ -70,14 +70,14 @@ fn append_modifiers(
         special = true;
         result = format!("S-{}", result);
     }
-	if ctrl && !alt {
+    if ctrl {
         special = true;
         result = format!("C-{}", result);
     }
-    if alt && !ctrl {
+    if alt {
         special = true;
         result = format!("M-{}", result);
-	}
+    }
     if cfg!(not(target_os = "windows")) && gui {
         special = true;
         result = format!("D-{}", result);
--- a/src/bridge/layouts/mod.rs
+++ b/src/bridge/layouts/mod.rs
@@ -64,20 +64,20 @@ fn append_modifiers(
     if result == "<" {
         result = "lt".to_string();
         special = true;
-    }
+    }    
 
     if shift {
         special = true;
         result = format!("S-{}", result);
     }
-    if ctrl {
+	if ctrl && !alt {
         special = true;
         result = format!("C-{}", result);
     }
-    if alt {
+    if alt && !ctrl {
         special = true;
         result = format!("M-{}", result);
-    }
+	}
     if cfg!(not(target_os = "windows")) && gui {
         special = true;
         result = format!("D-{}", result);
--- a/src/window.rs
+++ b/src/window.rs
@@ -135,13 +135,13 @@ impl WindowWrapper {
 
         let window = video_subsystem
             .window("Neovide", logical_size.width, logical_size.height)
+            .position_centered()
             .allow_highdpi()
             .resizable()
             .vulkan()
             .build()
             .expect("Failed to create window");
         info!("window created");
-
 
         let skulpin_renderer = RendererBuilder::new()
             .prefer_integrated_gpu()
--- a/src/window.rs
+++ b/src/window.rs
@@ -5,7 +5,7 @@ use std::time::{Duration, Instant};
 use log::{debug, error, info, trace};
 use skulpin::sdl2;
 use skulpin::sdl2::event::{Event, WindowEvent};
-use skulpin::sdl2::keyboard::{Keycode, Mod};
+use skulpin::sdl2::keyboard::{Keycode};
 use skulpin::sdl2::video::{Window};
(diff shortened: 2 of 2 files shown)


src/bridge/layouts/mod.rs
```diff
@@ -64,20 +64,20 @@ fn append_modifiers(
...
         special = true;
-    }
+    }    
 
     if shift {
         special = true;
         result = format!("S-{}", result);
     }
-    if ctrl {
+	if ctrl && !alt {
```
Kethku, line 73: As much as it is a pain, I don't think its a good idea to have both sets of changes in this PR. I added a comment on the other PR about this code and we should iterate on them separately.

src/window.rs
```diff
@@ -206,6 +236,12 @@ impl WindowWrapper {
     pub fn handle_keyboard_input(&mut self, keycode: Option<Keycode>, text: Option<String>) {
         let modifiers = self.context.keyboard().mod_state();
 
+        if let Some(key) = keycode {
+            if key == Keycode::Return && modifiers.contains(Mod::LALTMOD) {
```
Kethku, line 240: This isn't very vim friendly. This should ideally be implemented as a setting or function inside of vim so that users can create their own key bindings rather than mandating that AltEnter has this function. Alt Enter can be bound inside of vim which would be overridden by this change.

src/window.rs
```diff
@@ -168,7 +171,44 @@ impl WindowWrapper {
             previous_dpis,
             transparency: 1.0,
             fullscreen: false,
+            cached_width: 0,
```
Kethku, line 174: Minor: Why not store width and height as a tuple as well?

src/bridge/layouts/mod.rs
R-Michelsen: I'll update the branch later to remove this change from the PR.

src/window.rs
```diff
@@ -168,7 +171,44 @@ impl WindowWrapper {
...
+            let raw_handle = self.window.raw();
+            let display_index = sdl2::sys::SDL_GetWindowDisplayIndex(raw_handle);
+
+            if let Ok(rect) = self.window.subsystem().display_bounds(display_index) {    
+                if self.fullscreen {
+                    // Set window back to resizable
+                    sdl2::sys::SDL_SetWindowResizable(raw_handle, sdl2::sys::SDL_bool::SDL_TRUE);
+
+                    // Use cached size and position
+                    self.window.set_size(self.cached_width as u32, self.cached_height as u32).unwrap();
```
Kethku, line 191: I like this and I think it is a marked improvement over the current experience. I do think it should be integrated with the setting system that already exists though. I don't feel comfortable merging until then.

src/window.rs
R-Michelsen: I agree it probably shouldn't be hardcoded, I'm not overly familiar with the codebase, but I'll investigate and see if I can make it work as a setting like the other init.vim settings

src/window.rs
R-Michelsen: I don't have a strong opinion on it, but it might be a bit ambiguous whether width or height comes first in a tuple.

src/window.rs
R-Michelsen: By integration do you mean an option to either have the old fullscreen mode or have the toggleable mode?

src/window.rs
Kethku: standard practice is width, height

R-Michelsen: Noted, I'll change that aswell then

src/window.rs
Kethku: Its already toggleable, but has bugs. It can be enabled or disabled by setting g:neovide_fullscreen to v:true or v:false. I'm suggesting integrating the bug fixes with that existing system.

R-Michelsen: Ah I see, if I understand you correctly, I'll just try to integrate the functionality with the existing setting so that users can just keybind a true/false toggle in their .vimrc

R-Michelsen:
I've reverted the keybind changes on this branch, and made the toggle functionality work with the `g:neovide_fullscreen` variable such that it can be used like:
```vim
function! ToggleFullscreen()
    if g:neovide_fullscreen == v:true
        let g:neovide_fullscreen=v:false
    else
        let g:neovide_fullscreen=v:true
    endif
endfunction
nnoremap <F4> :call ToggleFullscreen()<CR>
```

Unfortunately, <Alt>+<Return> cannot actually be mapped in vim as far as I can tell. However I do think integration with the vim variable makes more sense regardless.

exoticus:
@Kethku this doesn't work properly on macOS, status line is hidden

image

Issue #117, opened 2020-02-05
Add cursor animation options


kiuKisas:
The cursor animation is very eye-catching when switching from normal/insert mode to the bottom command line.
As stated in this issue: https://github.com/Kethku/neovide/issues/85 , being able to disable animation would be a great feature, but it might be even cooler to have a less flashy effect for this transition, like fade in or something.

What do you think about that?

Kethku:
Sounds good to me. I am definitely planning on multiple configuration options for the cursor. Its on the list once we figure out how to implement settings correctly. Its coming!

Kethku:
This is partially solved now by @jonvaldes solution. The framework is set, just need to add a way to disable the trail in various circumstances

Kethku:
This has been solved for a while now. We dont necessarily have a fade animation but the animation is configurable in a variety of ways

PR #252, opened 2020-04-18
Proper setting to disable cursor animation while in insert mode

### LINKED ISSUE #225: Cursor doesn't animate going to right
I have noticed that cursor animations doesn't work going to the right side whether while typing or simply hitting `l` all other directions animate correctly, i wonder if there's any reason for this? or is it a bug?

### LINKED ISSUE #253: Cursor trail size over 1 causes render issues
Setting neovide setting `cursor_trail_size` to a value over 1 will break cursor animation (because of an infinite loop)

out

exoticus:
So there's currently a hack in neovide to avoid animating the cursor while typing which relied on calculating whether the travel distance is one character to the right, it works but also disables any one char wide right motion, including hitting `l` in normal mode. A behavior which might be interpreted as a bug to many new comers (i too thought that)

Changes:
1. Adds a neovide setting `cursor_animate_in_insert_mode` _defaults to true_ to allow the user to selectively disable animations specifically in insert mode. Should fix #225 and probably #117
2. Fixes #253 a bug caused by cursor trail sizes over one, where setting any value over 1 will cause the cursor to behave weirdly (animation doesn't end), resulting in a broken cursor
3. Some whitespace/newlines house keeping

### DIFF
--- a/src/bridge/events.rs
+++ b/src/bridge/events.rs
@@ -115,6 +115,20 @@ pub enum WindowAnchor {
 }
 
 #[derive(Debug)]
+pub enum EditorMode {
+    // The set of modes reported will change in new versions of Nvim, for
+    // instance more sub-modes and temporary states might be represented as
+    // separate modes. (however we can safely do this as these are the main modes)
+    // for instance if we are in Terminal mode and even though status-line shows Terminal,
+    // we still get one of these as the _editor_ mode
+    Normal,
+    Insert,
+    Visual,
+    CmdLine,
+    Unknown(String),
+}
+
+#[derive(Debug)]
 pub enum RedrawEvent {
     SetTitle {
         title: String,
@@ -126,6 +140,7 @@ pub enum RedrawEvent {
         gui_option: GuiOption,
     },
     ModeChange {
+        mode: EditorMode,
         mode_index: u64,
     },
     MouseOn,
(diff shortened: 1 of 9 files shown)


src/renderer/cursor_renderer/mod.rs
```diff
@@ -102,17 +108,15 @@ impl Corner {
...
         if destination != self.previous_destination {
-            let travel_distance = destination - self.previous_destination;
-            let chars_travel_x = travel_distance.x / font_dimensions.x;
-
-            if travel_distance.y == 0.0 && (chars_travel_x - 1.0).abs() < 0.1 {
-                // We're moving one character to the right. Make movement immediate to avoid lag
-                // while typing
+            // 2 for insert mode as it's the only mode where
+            // the cursor might appear to lag behind
+            if !settings.animate_in_insert_mode && mode == 2 {
```
Kethku, line 119: I'm nervous about using the 2 constant here. Ideally we should search for the mode which has a name of Insert (or whatever the mode name is that gets reported). This index could change in the future breaking things.

What do you think?

src/editor/mod.rs
```diff
@@ -69,6 +71,7 @@ impl Editor {
             RedrawEvent::ModeInfoSet { cursor_modes } => self.cursor.mode_list = cursor_modes,
             RedrawEvent::OptionSet { gui_option } => self.set_option(gui_option),
             RedrawEvent::ModeChange { mode_index } => {
+                self.mode = mode_index;
```
Kethku, line 74: Maybe a more principled change, although somewhat larger, would be to store the mode info list in the editor, store the current mode also in the editor and set the mode object itself in the cursor object.

This way we have access to the current mode name, and it pushes the editor state up to the editor instead of storing that in the cursor like its done today.

src/renderer/cursor_renderer/mod.rs
exoticus: totally agree, i wanted to have the current mode as an editor property but this seemed like easy cheat till we have proper mode handling in the editor type. perhaps i should rework that then

src/editor/mod.rs
exoticus: will have to fiddle around more to get this done, it'll also benefit us in the long run if we ever come to doing more fancy stuff like status line theming etc

src/bridge/events.rs
```diff
@@ -114,6 +114,20 @@ pub enum WindowAnchor {
...
+    // The set of modes reported will change in new versions of Nvim, for
+    // instance more sub-modes and temporary states might be represented as
+    // separate modes. (however we can safely do this as these are the main modes)
+    // for instance if we are in Terminal mode and even though status-line shows Terminal,
+    // we still get one of these as the _editor_ mode
+    Normal,
+    Insert,
+    Visual,
+    CmdLine,
+    Unknown(String),
```
Kethku, line 128: God I love rust enums

PR #256, opened 2020-04-20
better fullscreen toggle and minor cleanup

Kethku:
Rework of the fullscreen toggle to remove unnecessary unsafe calls and use standard fullscreen on macos

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -6,7 +6,7 @@ use log::{debug, error, info, trace};
 use skulpin::sdl2;
 use skulpin::sdl2::event::{Event, WindowEvent};
 use skulpin::sdl2::keyboard::Keycode;
-use skulpin::sdl2::video::Window;
+use skulpin::sdl2::video::{FullscreenType, Window};
 use skulpin::sdl2::Sdl;
 use skulpin::{dpis, CoordinateSystem, PresentMode, Renderer as SkulpinRenderer, RendererBuilder};
 use skulpin::{LogicalSize, PhysicalSize};
@@ -177,11 +177,13 @@ impl WindowWrapper {
 
     pub fn toggle_fullscreen(&mut self) {
         if self.fullscreen {
-            if cfg!(not(target_os = "macos")) {
+            if cfg!(target_os = "windows") {
                 unsafe {
                     let raw_handle = self.window.raw();
                     sdl2::sys::SDL_SetWindowResizable(raw_handle, sdl2::sys::SDL_bool::SDL_TRUE);
                 }
+            } else {
+                self.window.set_fullscreen(FullscreenType::Off).ok();
             }
 
             // Use cached size and position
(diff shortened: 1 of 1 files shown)


src/window.rs
```diff
@@ -176,46 +176,44 @@ impl WindowWrapper {
...
                     // Set window to fullscreen
                     self.window.set_size(rect.width(), rect.height()).unwrap();
                     self.window.set_position(
                         sdl2::video::WindowPos::Positioned(rect.x()),
                         sdl2::video::WindowPos::Positioned(rect.y()),
                     );
-                    self.window.set_bordered(true);
                 }
+            } else {
+                self.window.maximize();
```
exoticus, line 221: So depending on the behavior on window this might need to be changed? does full screen on windows make either
- New desktop
- Or completely cover the whole screen even the taskbar? (and it becomes borderless)

Because if so then macOS has similar thing where it'll move window to a new Space and toggle screen wide full screen

exoticus: `window.maximize()` on the other hand just sets the window to cover all available space minus menubar (ala taskbar)

Just for feature consistency across systems

Kethku: fullscreen covers everything including the taskbar which is the desired functionality. I figured window.maximize() would swap to a new desktop on macos which is the equivalent functionality on that platform. Maybe I'm misunderstanding?

exoticus: @Kethku  no that would be then to achieve this
```rust
self.window.set_fullscreen(FullscreenType::Desktop).ok();
```
However you need to set it back to `Off` when going back to normal so
```rust
self.window.set_fullscreen(FullscreenType::Off).ok();
```

Kethku: I see. I will update to match this. Do you think linux machines would expect that method as well?

exoticus: does that work on window?, what if you use `set_fullscreen(FullscreenType::Desktop)` instead of all the current functionality, SDL docs specify that it should behave across platforms normally so perhaps does the job can you test it?

```rust
pub fn toggle_fullscreen(&mut self) {
        if self.fullscreen {
            self.window.set_fullscreen(FullscreenType::Off).ok();
            self.window
                .set_size(self.cached_size.0, self.cached_size.1)
                .unwrap();
            self.window.set_position(
                sdl2::video::WindowPos::Positioned(self.cached_position.0),
                sdl2::video::WindowPos::Positioned(self.cached_position.1),
            );
        } else {
            self.cached_size = self.window.size();
            self.cached_position = self.window.position();
            self.window.set_fullscreen(FullscreenType::Desktop).ok();
        }

        self.fullscreen = !self.fullscreen;
    }
```

Kethku: This doesn't work on windows because the app crashes when you manually swap to another window.

The original PR was a hackyish attempt to get the desired functionality without the fullscreen. I will special case it to windows though.

exoticus: @Kethku yeah i think special casing windows to fuzz out this behavior would be the right call

Kethku: How bout now. Does it work on your side now?

exoticus: @Kethku yup, excellant

unphased:
Hi, I would like to request if we could bring a SDL (or whatever is serving OpenGL or Metal) based borderless fullscreen mode back, as an alternative to use instead of native macOS fullscreen? 

- I don't want to rip Spaces out entirely from my macOS usage. it's useful for separating different modes of work (coding vs content consumption etc)
- a natively-fullscreened app yanks an app out of its Stage Manager group. Upon returning we have to do the usual Stage Manager faff to reincorporate it into the suitable project app window group. 
- fullscreen allows better focus but just as valid of a workflow is when we are still toggling back and forth to work on something but still need to maximize real estate. The unfortunate integration macOS does between native window fullscreen and Spaces means that the animations involved with Spaces are totally unavoidable. Furthermore when we have autoswoop, that is to say, this option: 
   image
   disabled, which is how I need to set it for other workflows, then simply re-targeting Neovide via task switcher won't even hop us back to the new full screen space that it has been put on. 

Thanks.

unphased:
> Hi, I would like to request if we could bring a SDL (or whatever is serving OpenGL or Metal) based borderless fullscreen mode back, as an alternative to use instead of native macOS fullscreen?

Hmm, so I will note though, that I believe using `frame = "none"` config (launch flag) and then using other means (e.g. hammerspoon) to control the positioning may do the trick at least for me here.

PR #258, opened 2020-04-20
Neovide can now open files... with drag/drop

### LINKED ISSUE #242: Drag/drop files into neovide window to open it
Drag file from system explorer and drop into neovide window to open the file. Possible to have this feature?

exoticus:
In order to open a file if there’s changes in current buffer, a user should have `set hidden` in his/her vimrc. This should fix #242

@Kethku i wonder why you stated that it isn't possible with sdl, just an oversight?

### DIFF
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -21,6 +21,7 @@ pub enum UiCommand {
         position: (u32, u32),
     },
     Drag(u32, u32),
+    FileDrop(String),
     FocusLost,
     FocusGained,
     Quit,
@@ -74,6 +75,9 @@ impl UiCommand {
                 .expect("Focus Gained Failed"),
             UiCommand::Quit => {
                 nvim.command("qa!").await.ok(); // Ignoring result as it won't succeed since the app closed.
+            },
+            UiCommand::FileDrop(path) => {
+                nvim.command(format!("e {}", path).as_str()).await.ok();
             }
         }
     }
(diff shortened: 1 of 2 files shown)


Kethku:
Must be macos only. The windows version doesn't work :l Good to have it where we can though

exoticus:
@Kethku weird that it doesn't, is there an error?

Kethku:
https://discourse.libsdl.org/t/sdl-2-0-drag-drop-issue/19144/4

No error. Just does nothing.

src/bridge/ui_commands.rs
```diff
@@ -74,6 +75,9 @@ impl UiCommand {
                 .expect("Focus Gained Failed"),
             UiCommand::Quit => {
                 nvim.command("qa!").await.ok(); // Ignoring result as it won't succeed since the app closed.
+            },
+            UiCommand::FileDrop(path) => {
+                nvim.command(format!("e {}", path).as_str()).await.ok();
```
Kethku, line 80: Maybe surface issue to the user if this fails? I suppose we should do the same with the other ones too...

exoticus:
> https://discourse.libsdl.org/t/sdl-2-0-drag-drop-issue/19144/4
> 
> No error. Just does nothing.

holy... since 2012

src/bridge/ui_commands.rs
exoticus: what about just adding `e file/path` to the command line and leave it to the user to execute it? i wonder if that's possible though

Kethku:
yeah for such a widely used piece of software its kinda a mess

src/bridge/ui_commands.rs
Kethku: I suspect that would be more work than folks would like. The goal is to be able to just drag a file on and get rolling. Adding another step would be less nice. I vote we just merge as is. Seems good to me

PR #262, opened 2020-04-23
Better symbol font for linux

jamespotz:
Instead of using unifont we can use noto sans mono for better unicode support

With `Unifont`
image

With `Noto Sans`
image

### DIFF
--- a/src/renderer/caching_shaper.rs
+++ b/src/renderer/caching_shaper.rs
@@ -22,7 +22,7 @@ define! {
         const SYSTEM_EMOJI_FONT: &str = "Segoe UI Emoji";
     } else if #[cfg(target_os = "linux")] {
         const SYSTEM_DEFAULT_FONT: &str = "Droid Sans Mono";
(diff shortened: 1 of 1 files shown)


exoticus:
@jamespotz can you please verify if it comes with most linux distros?

jamespotz:
Yes it comes with all the major linux distro.

Kethku:
I'm alright with this change, but I know exactly zero about linux fonts so I may let someone else make the call. If nobody weighs in though I will merge this and see if anybody yells ;)

j4qfrost:
I asked around and did some googling and it looks prevalent enough that few people would complain.

SirJson:
The problem with supporting Linux is that people will always complain. Unifont was at least some GPL project I have never heared before, no Google there and old enough that you should find it in Debian. I wish there would be an anwser like Segoe UI on Windows for Unicode symbols but in the end of the day Linux is not an Operating System...

PR #265, opened 2020-04-29
Update github actions config

j4qfrost:
This will make our builds look green as long as there is a codecov token. It also makes them robust to errors when building rust-skia in Macos as opposed to downloading the prebuilt binaries (the master branch currently downloads prebuilt).

In messing with the configurations, I discovered that an alternative to grcov, tarpaulin, doesn't require incremental builds being disabled and dead code being linked, which makes the upstream updates for core-foundation nonessential.


Kethku:
Whats the status here. Is this ready to merge or does it need more work?

j4qfrost:
There's still the issue that the tests fail on Windows, because they are dependent on neovim being installed. Another option is stubbing out neovim. For some reason the installation of neovim sends an error code. I'll look into it now.

j4qfrost:
Ready. I'll let you do the honors. @Kethku

j4qfrost:
@Kethku This should work now, but I've used my own dependencies to enforce that SDL version 0.33 and skulpin 0.27.2 are used.

Issue #272, opened 2020-05-06
Disabling "embed-fonts" breaks build


SirJson:
Just fell over this issue while trying to figuring out why special glphys fail to render again on windows.
If I disable the "embed-fonts" feature neovide refuses to build.

```shell
C:\dev\neovide master ✗
❯ cargo build --release --no-default-features
   Compiling neovide v0.5.0 (C:\dev\neovide)
error[E0599]: no method named `load_from_asset` found for type `&mut renderer::caching_shaper::FontLoader` in the current scope
   --> src\renderer\caching_shaper.rs:149:18
    |
149 |             self.load_from_asset(font_name)
    |                  ^^^^^^^^^^^^^^^ method not found in `&mut renderer::caching_shaper::FontLoader`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0599`.
error: could not compile `neovide`.

To learn more, run the command again with --verbose.
C:\dev\neovide master ✗
❯ vim .\Cargo.toml
C:\dev\neovide master ✗
❯ cargo build --release
   Compiling neovide v0.5.0 (C:\dev\neovide)
error[E0599]: no method named `load_from_asset` found for type `&mut renderer::caching_shaper::FontLoader` in the current scope
   --> src\renderer\caching_shaper.rs:149:18
    |
149 |             self.load_from_asset(font_name)
    |                  ^^^^^^^^^^^^^^^ method not found in `&mut renderer::caching_shaper::FontLoader`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0599`.
error: could not compile `neovide`.

To learn more, run the command again with --verbose.
```

The second try was to make sure that there is no cargo bug that disables the feature of every dependency or something like that. I just removed `embed-fonts` from the default-features list.

That output looks kinda strange to me tho, `embed-fonts` used to do exactly that and not introduce dependencies or something like that. 

I'll see if I can repair this but first I have to get familiar again with the source code.

j4qfrost:
https://github.com/Kethku/neovide/pull/312

PR #273, opened 2020-05-06
get_or_load() works now without embedFont

SirJson:
This is the fix for #272

Because get_or_load defines "asset" as parameter in his public api I decided to create an copy of the original get_or_load that prints a warning if someone is using this option but haven't enabled `embed-fonts` it yet.

This is still WIP because there are two points that are debatable where I'm kinda indifferent.

1. I think it is better to have one `get_or_load()`, but for some reason I could not get the cfg! macro to work. More if's aren't a silver bulet but in the long them I think this would be easier to maintain. If someone can update me on the syntax on that and agrees on with that change I will add that.

2. Do we want the use of `asset` be a warning or a debug message? The problem in the first place is that `embed-fonts` can't be a feature anymore with get_or_load exposing it. The clean solution would be to refactor the code so `asset` isn't exposed anymore or drop `embed-fonts` as a feature completely. Also here input and feedback is wanted and very much appreciated.

As for now at least this version did fix the build and produced a functioning binary.

### DIFF
--- a/src/renderer/caching_shaper.rs
+++ b/src/renderer/caching_shaper.rs
@@ -140,28 +140,21 @@ impl FontLoader {
         self.get(font_name)
     }
 
-    #[cfg(feature = "embed-fonts")]
     pub fn get_or_load(&mut self, font_name: &str, asset: bool) -> Option<ExtendedFontFamily> {
         if let Some(family) = self.get(font_name) {
             return Some(family);
         }
 
         if asset {
-            self.load_from_asset(font_name)
-        } else {
-            self.load(font_name)
-        }
-    }
-
-    #[cfg(not(feature = "embed-fonts"))]
-    pub fn get_or_load(&mut self, font_name: &str, asset: bool) -> Option<ExtendedFontFamily> {
-        if let Some(family) = self.get(font_name) {
-            return Some(family);
-        }
-
-        if asset {
-            warn!("Tried to load {} from assets but build didn't include embed-fonts feature",font_name);
-            None
+            #[cfg(not(feature = "embed-fonts"))]
+            {
+                warn!("Tried to load '{}' from assets but build didn't include embed-fonts feature", font_name);
+                None
+            }
+            #[cfg(feature = "embed-fonts")]
+            {
+                self.load_from_asset(font_name)
+            }
         } else {
             self.load(font_name)
         }
--- a/src/renderer/caching_shaper.rs
+++ b/src/renderer/caching_shaper.rs
@@ -140,6 +140,7 @@ impl FontLoader {
         self.get(font_name)
     }
 
+    #[cfg(feature = "embed-fonts")]
     pub fn get_or_load(&mut self, font_name: &str, asset: bool) -> Option<ExtendedFontFamily> {
         if let Some(family) = self.get(font_name) {
             return Some(family);
@@ -147,6 +148,20 @@ impl FontLoader {
 
         if asset {
             self.load_from_asset(font_name)
+        } else {
+            self.load(font_name)
+        }
+    }
+
+    #[cfg(not(feature = "embed-fonts"))]
+    pub fn get_or_load(&mut self, font_name: &str, asset: bool) -> Option<ExtendedFontFamily> {
+        if let Some(family) = self.get(font_name) {
+            return Some(family);
+        }
+
+        if asset {
+            warn!("Tried to load {} from assets but build didn't include embed-fonts feature",font_name);
+            None
         } else {
             self.load(font_name)
         }
@@ -195,7 +210,6 @@ pub fn build_collection_by_font_name(
             collection.add_family(family.to_normal_font_family());
         }
     }
-
     for font in &[EXTRA_SYMBOL_FONT, MISSING_GLYPH_FONT] {
         if let Some(family) = loader.get_or_load(font, true) {
             collection.add_family(family.to_normal_font_family());
--- a/src/renderer/caching_shaper.rs
+++ b/src/renderer/caching_shaper.rs
@@ -117,7 +117,7 @@ impl FontLoader {
     }
 
     #[cfg(feature = "embed-fonts")]
-    fn load_from_asset(&mut self, font_name: &str) -> Option<ExtendedFontFamily> {
+    pub fn load_from_asset(&mut self, font_name: &str) -> Option<ExtendedFontFamily> {
         let mut family = ExtendedFontFamily::new();
 
         Asset::get(font_name)
@@ -126,6 +126,15 @@ impl FontLoader {
 
         self.cache.put(String::from(font_name), family);
         self.get(font_name)
+    }
+
+    #[cfg(not(feature = "embed-fonts"))]
+    pub fn load_from_asset(&self, font_name: &str) -> Option<ExtendedFontFamily> {
+        warn!(
+            "Tried to load {} from assets but build didn't include embed-fonts feature",
+            font_name
+        );
+        None
     }
 
     fn load(&mut self, font_name: &str) -> Option<ExtendedFontFamily> {
@@ -140,24 +149,11 @@ impl FontLoader {
         self.get(font_name)
     }
 
-    pub fn get_or_load(&mut self, font_name: &str, asset: bool) -> Option<ExtendedFontFamily> {
+    pub fn get_or_load(&mut self, font_name: &str) -> Option<ExtendedFontFamily> {
         if let Some(family) = self.get(font_name) {
             return Some(family);
         }
-
-        if asset {
-            #[cfg(not(feature = "embed-fonts"))]
-            {
-                warn!("Tried to load '{}' from assets but build didn't include embed-fonts feature", font_name);
-                None
-            }
-            #[cfg(feature = "embed-fonts")]
-            {
-                self.load_from_asset(font_name)
-            }
-        } else {
-            self.load(font_name)
-        }
+        self.load(font_name)
     }
 }
 
@@ -190,7 +186,7 @@ pub fn build_collection_by_font_name(
         .chain(iter::once(SYSTEM_DEFAULT_FONT));
 
     for font_name in gui_fonts {
-        if let Some(family) = loader.get_or_load(font_name, false) {
+        if let Some(family) = loader.get_or_load(font_name) {
             if let Some(font) = family.get(properties) {
                 collection.add_family(FontFamily::new_from_font(font.clone()));
                 break;
@@ -199,15 +195,20 @@ pub fn build_collection_by_font_name(
     }
 
     for font in &[SYSTEM_SYMBOL_FONT, SYSTEM_EMOJI_FONT] {
-        if let Some(family) = loader.get_or_load(font, false) {
+        if let Some(family) = loader.get_or_load(font) {
             collection.add_family(family.to_normal_font_family());
         }
     }
-    for font in &[EXTRA_SYMBOL_FONT, MISSING_GLYPH_FONT] {
-        if let Some(family) = loader.get_or_load(font, true) {
-            collection.add_family(family.to_normal_font_family());
+
+    #[cfg(feature = "embed-fonts")]
+    {
+        for font in &[EXTRA_SYMBOL_FONT, MISSING_GLYPH_FONT] {
+            if let Some(family) = loader.load_from_asset(font) {
+                collection.add_family(family.to_normal_font_family());
+            }
         }
     }
+
 
     collection
 }


j4qfrost:
Not sure if this answers the first point, but this is how one get_or_load function would be possible, though it's still a bit messy:
```rust
    pub fn get_or_load(&mut self, font_name: &str, asset: bool) -> Option<ExtendedFontFamily> {
        if let Some(family) = self.get(font_name) {
            return Some(family);
        }

        if asset {
            #[cfg(not(feature = "embed-fonts"))]
            {
                warn!("Tried to load {} from assets but build didn't include embed-fonts feature", font_name);
                None
            } 

            #[cfg(feature = "embed-fonts")]
            {
                self.load_from_asset(font_name)
            }
        } else {
            self.load(font_name)
        }
    }
```
cfg! won't work, because you exclude the load_from_asset function if the embed-fonts feature is disabled.

For point 2, the message should definitely be a warning since it's probably something a user would want to be notified about.

> The problem in the first place is that embed-fonts can't be a feature anymore with get_or_load exposing it. The clean solution would be to refactor the code so asset isn't exposed anymore or drop embed-fonts as a feature completely.

Not sure what you mean by this.

Kethku:
We could take the stance that embedding fonts isn't something we should be doing, but I think we need to get font fallback working using os apis before we attempt such a change.

j4qfrost:
Are there any good case studies around where to write out OS specific code and where to avoid it? It seems like we are trading one level of complexity for a higher level of complexity. Implementing font fallback that is OS specific seems like the correct way to do things, but it seems like the "harder to maintain" way.

SirJson:
@j4qfrost Thanks for the snippet, my projects never had to a lot of `cfg` and I always thought you could use the macro also in those cases.

> The problem in the first place is that embed-fonts can't be a feature anymore with get_or_load exposing it. The clean solution would be to refactor the code so asset isn't exposed anymore or drop embed-fonts as a feature completely.

Fair enough sounds worse than it is. In a perfect world you wouldn't expose that you can embed fonts into your build or not. It would be handled inside the shaper and the caller doesn't have to worry about this detail. 

With exposing `asset` every time someone is calling `get_or_load()`,now or in the future, he has to remember `embed-fonts`. That design can lead to more complexity in other modules because if a future contributer calls it with asset == true and `embed-fonts` not included it will spam warnings. 

And I agree this should be an warning because it is invalid usage of the API but is it important to the caller of `get_or_load` if fonts are embedded or not? I think it shouldn't because the feature fixes a problem on Platforms that just don't provide everything needed.

j4qfrost:
@SirJson I looked at it closer. The counter against using warn over debug is that the user doesn't actually have control over whether the asset is used or not.

SirJson:
Yep, and that's why I was hoping on input because this is no black and white topic and really depends on what @Kethku wants. The most pragmatic version is removing the dual function version that I pushed first and replace that with the inner `cfg` if version and call it a day.

That just came up because the "boy scout rule" is something that is kinda drilled into me tough talking software design is streching it a little bit :)

j4qfrost:
@SirJson Sorry just looked at the code more. This would be a lot cleaner:
```rust
#[cfg(feature = "embed-fonts")]
fn load_from_asset(&mut self, font_name: &str) -> Option<ExtendedFontFamily> {
    let mut family = ExtendedFontFamily::new();

    Asset::get(font_name)
        .and_then(|font_data| Font::from_bytes(font_data.to_vec().into(), 0).ok())
        .map(|font| family.add_font(SkriboFont::new(font)));

    self.cache.put(String::from(font_name), family);
    self.get(font_name)
}

#[cfg(not(feature = "embed-fonts"))]
fn load_from_asset(&self, font_name: &str) -> Option<ExtendedFontFamily> {
    warn!(
        "Tried to load {} from assets but build didn't include embed-fonts feature",
        font_name
    );
    None
}

pub fn get_or_load(&mut self, font_name: &str) -> Option<ExtendedFontFamily> {
    if let Some(family) = self.get(font_name) {
        Some(family)
    } else if let Some(family) = self.load_from_asset(font_name) {
        Some(family)
    } else {
        self.load(font_name)
    }
}
```

SirJson:
Ok I like this even more because this already almost looks like the implemention of the perfect world version. But I don't know why this parameter was introduced in the first place so removing it might be dangerous.

j4qfrost:
Lol I tried something similar before and it failed, but then I took a closer look.
```rust
pub fn get_or_load(&mut self, font_name: &str) -> Option<ExtendedFontFamily> {
        if let Some(family) = self.get(font_name) {
            Some(family)
        }  else if let Some(family) = self.load(font_name) {
            Some(family)
        } else {
            self.load_from_asset(font_name)
        }
    }
```

That appears to work with the additional change here:

```rust
    for font in &[
        SYSTEM_SYMBOL_FONT,
        SYSTEM_EMOJI_FONT,
        EXTRA_SYMBOL_FONT,
        MISSING_GLYPH_FONT,
    ] {
        if let Some(family) = loader.get_or_load(font) {
            collection.add_family(family.to_normal_font_family());
        }
    }
```

I think I'll tackle implementing unit tests here. It would've gave me more initial encouragement to refactor.

SirJson:
```rust   
for font in &[
        SYSTEM_SYMBOL_FONT,
        SYSTEM_EMOJI_FONT,
        EXTRA_SYMBOL_FONT,
        MISSING_GLYPH_FONT,
    ] {
        if let Some(family) = loader.get_or_load(font) {
            collection.add_family(family.to_normal_font_family());
        }
    }```

Something like this was in my head the whole time, like "It shouldn't matter what is available it's a collection of fonts with n elements and everything is just data. If it is not included that group of fonts is empty and in the end it's just adding all groups together". 

Or in TL;DR what you have written above. But now with two PR's and the stuff that would be obsolete by the refactored version I'm confused what the way forward is. 

Also sorry but I can't answer as fast as you can. For one english is not my first language and and I'm shocked how much Rust you can forget even though I have worked on a audio project in Rust for over a year daily day and night (almost literally... I was breathing Rust) plus two Projects spearheaded at my workplace. I often read up on subjects again just to be sure...

j4qfrost:
Yeah, don't sweat it. I just sent in a commit your way that addresses a caching issue that I ran into with my previous changes. We can take our time to figure out the path forward. I just have a lot of time on my hands, so I respond quicker. I'm comfortable waiting for as long as you need to review the change. Rust has a lot of fun toys. It helps to have a compiler.

Kethku:
Just wanted to chime in that @exoticus spent a lot of time in this area managing startup speed as well, so any solution proposed here should pay attention to effects on startup on top of correctness. Other than that looks great. Feel free to pull me in to discussion if you think I can help in any way

j4qfrost:
@Kethku here's the additional relevant pull request https://github.com/SirJson/neovide/pull/1

This is mostly about code prettification, but a change I added accidentally put empty font families into the cache. Probably low impact, but I cleaned it up. One question I still have though is whether my implementation of defaulting to using load first, then falling back to load_from_asset is appropriate functionality.

Kethku:
The only weirdness is that today we embed (or maybe at one point did embed) custom fonts that will never be in the system font. There was one that was paired down to only include some special ui characters. In that case I guess it would always fall back to load_from_asset which is probably fine, just strange

j4qfrost:
When first looking at the code I thought it was a fun way to quickly inject a joke font to use in the editor.

SirJson:
I agree with that the special font for neovide is a "feature" feature especially if you extend it futher with common nerd font-like glpyhs. No editor I know of offers that as far as I know.

What I need to know to make the API safety proposal is do you make a clear cut with the "embed-fonts" feature flag or will the new system font get extra treatment? @Kethku

Kethku:
I think the super tiny system font would be embedded regardless if it doesn't effect load time too much. The biggest problem with embedding fonts before was the font loading performance on macos which was way worse than the other platforms. But if the custom font is small enough the benefits probably outway the costs.

Kethku:
Another option would be to load the fonts asynchronously which would fix the startup performance issue on macos... But they would still be pretty big so idk.

SirJson:
Ok TIL. But is this something across the board with all macs? I didn't notice a difference on my old Macbook Pro and that's already under powered in my eyes.

j4qfrost:
Sorry, I think I might have killed this PR with some confusion. I think the idea is good in that it fixes the issue. If we want to add a feature to allow for injecting fonts like I suggested, there should be a separate issue for that.

@SirJson There are some improvements you can make with your PR, that I added with my PR to your branch. Sorry again for the confusion. One thing to look out for is accidentally putting garbage in the cache, because lines like this https://github.com/j4qfrost/neovide/blob/b9b3a50ea42a26edbc0cea241055760529154ee5/src/renderer/caching_shaper.rs#L121 might cause an empty fontfamily to be put in the cache.

SirJson:
Don't worry, there are a lot of ways to approach this but in the end the fix for it won't be a rewrite.
I just thought you are so fast, by the time I finish my last thoughts you are done already with a good solution. 

All I wanted is to fix this small problem since some users might build neovide that way and maybe prevent the oversight which caused the build issue from happening again :) I'll have a look at it later and see that we can get our (our your?) changes approved.

exoticus:
@SirJson 
When i was writing this there was a couple of goals in mind as well as some limitations of the current font fallback system. The decisions i took was purely based on
+ Optimizing Neovide for better startup times _(down from 3 seconds to roughly 850ms on macOS, and nearly instantaneous on other platforms)_. 
In order to achieve that, i wrote the asset caching system so that we don't have to load especially large fonts like `Apple Color Emoji`, over and over again. And got rid of the embedded Noto fonts which were both big in size and consequently loading times
+ Not sacrificing too much niceties of embedded glyph fonts, for that i created a very lightweight `Missing Glyph` font that contains the `ln` character as well as other common nerd font glyphs

I tried to achieve these while keeping in mind that relying on font embedding and even font name embedded is a means to an end once prober font fallback detection is implemented. However, i do consider having embedded font loading is a nice feature even after full fallback is implemented, so that we can ship lightweight nerd glyphs and other extra symbols out of the box

SirJson:
Uh I'm not against embedded fonts, in fact I think they provide an advantage. Still when I try to reproduce user bugs around font rendering I always try both, with embedded fonts and without just as a routine. That is how I noticted the feature is / was broken.

I started this discussion because who ever wrote this didn't know about this feature or forgot it existed in the first place. That's totally normal and human, so I was thinking about how it probably happend and how it could be prevented without writing a big developer manual ;)

One solution can be to encapsulate the detail "embedded fonts" inside the glyph cacher and don't leak it into other modules or files. Because I know very well that this might not be possible (and because I honestly never really used the cfg macro and misunderstood that feature) I thought hey let's ask someone with more hands on experience with the internals if we can make it better.

And it looks like we can it is just my situation right now didn't allow me to look at it and finish the work here. If I remember correctly there might be some Linux Distribution (Arch Linux?) out there who prefers to build without embedded fonts and as long that is not fixed the build for that distro is broken as well. 

Considering the cause of the issue I understand that this is taking way too long, but you know life happens, especially now...

exoticus:
> who ever wrote this didn't know about this feature or forgot it existed in the first place. That's totally normal and human, so I was thinking about how it probably happend and how it could be prevented without writing a big developer manual ;)

Can you please elaborate which _feature_ you're talking about, and which code are you referring to by `this`.

if you mean the embedded font loader code i wrote. Then it changed nothing about glyph caching and is all about file system font caching. The reason Linux broke was #262 which changed the more broadly used `Unifont` with `Noto Sans Mono` which turned out to be missing on your side and other Linux distros...

Now the way Neovide masked missing fonts before was that it bundled `NotoCJK`, `FiraCode`, `NotoEmoji` and some other Symbol font and used them by default; However that came at the cost of both huge binary size and longer startup because of the way `Skribo` fallback system needs to load fallbacks into each font family category. So, For instance, on macOS a file system read call to the big _160Mb_ `Apple Color Emoji` font was made at least 3 times so that it's loaded into the `Normal/Italic/Bold` variants of whatever font is used (Fira Code by Default back then)

Here's a simple flow of what Neovide used to do VS what the current implementation does
`Startup > Retrieve user defined font if any > If none load the embedded FiraCode > Load emoji font from FS > add to FiraCode (Normal) > Load emoji font from FS > add to FiraCode (Italic) > Load emoji font from FS > add to FiraCode (Bold)`

The caching system just took the `FS` out of the above workflow so we load a font once and cache it in memory so it becomes pretty fast during the process of fallback list synthesis.

Note: Neovide did also repeat that process whenever you changed the font during runtime with the `CMDLine` resulting in a lag spike on macOS again because of Emoji, which is now none existant as it's just cached.

I hope this clarified things a little bit, and yeah it's a pity the whole situation RN with everyday life, i was getting started on exploring native fallback but then stuff happened

j4qfrost:
I'm going to go ahead and close this since I merged in my branch here https://github.com/Kethku/neovide/pull/312

PR #274, opened 2020-05-06
Mouse position macos

j4qfrost:


j4qfrost:
I'll leave this up for a day before merging it.

src/window.rs
```diff
@@ -284,8 +284,17 @@ impl WindowWrapper {
...
-        self.mouse_position = physical_size.to_logical(sdl_window_wrapper.scale_factor());
+        #[cfg(not(target_os = "macos"))]
+        {
+            let sdl_window_wrapper = Sdl2Window::new(&self.window);
+            self.mouse_position = physical_size.to_logical(sdl_window_wrapper.scale_factor());
+        }
+
+        #[cfg(target_os = "macos")]
+        {
+            self.mouse_position = physical_size.to_logical(1.0);
```
Kethku, line 295: What happens if the macos device is connected to a higher dpi display?

j4qfrost: I've switched resolutions on both my monitors, and it seems to work with the change. I'm not sure if my test is exhaustive. This may be a case of "works fine on my machine". I can wait for another Mac user to complain. I'll try to share this around for testing.

Kethku: I think dpi is different from resolution. This stuff has to do with the scaling factor. Unless you explicitly set the dpi, I don't think it will change by just changing the resolution

j4qfrost: I'll just let this hang while I dig into skulpin, then.

j4qfrost:
I'm going to close this to check if there are issues upstream with skulpin.

PR #276, opened 2020-05-07
Point skulpin at upstream & turn off Vulkan debug layer

vivlim:
This PR:
* picks up the fix for Kethku/neovide#147
* disables the vulkan debug layer. It was blocking me from running on my Mac with a broken (?) vulkan-sdk install. Neovide runs fine without it, and @aclysma mentioned on gitter that it shouldn't be on in shipping software & having it enabled has significant perf impact.

(@j4qfrost I noticed that neovide is currently using https://github.com/j4qfrost/skulpin, does my PR trample on something you're doing?)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -307,21 +307,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "cocoa"
-version = "0.20.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


j4qfrost:
Nope, I think we can safely switch to upstream for now. My change was in anticipation for updating the core libraries for doing test coverage. The issue lies somewhere else now, so I've put it on hold. I'm currently working on some clippy changes and adding husky, so I'll return to the issue tomorrow.

PR #277, opened 2020-05-07
Husky

j4qfrost:
There are some clippy suggestions that were made that I wasn't about, particularly the abs() changes:
https://github.com/j4qfrost/neovide/blob/7802b64c2c1520724d40360cc059270fec36cd26/src/renderer/cursor_renderer/mod.rs#L121
 and enforcing a smaller number of function parameters:
https://github.com/j4qfrost/neovide/blob/7802b64c2c1520724d40360cc059270fec36cd26/src/renderer/cursor_renderer/mod.rs#L237

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1306,7 +1306,7 @@ dependencies = [
 
 [[package]]
 name = "neovide"
-version = "0.5.0"
+version = "0.6.0"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -103,7 +103,7 @@ checksum = "da71fef07bc806586090247e971229289f64c210a278ee5ae419314eb386b31d"
 dependencies = [
  "proc-macro2 1.0.12",
  "quote 1.0.4",
- "syn 1.0.18",
+ "syn 1.0.19",
 ]
 
 [[package]]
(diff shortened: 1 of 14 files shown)


j4qfrost:
@Kethku one thing I'm also unsure about is whether to keep clippy in as a precommit. I'm not sure if that will help serve as an educational tool for writing better rust or deter people from submitting changes.

Kethku:
I'm not sure myself. Is there way to include it as some sort of warning? That way we can look at the results and get notified but not prevent build? I'm not sure if thats a thing for github actions

j4qfrost:
Including clippy as a feature for husky on precommit hooks will prevent the commit from happening if the code has some "bad" code. You can bypass them using --no-verify when committing, so if someone is really annoyed they can just do that. It looks like we can configure separate options to warn or deny https://crates.io/crates/clippy/0.0.94, but it doesn't touch the build process. If we want to include clippy in the github actions, then we could just write a hack to ignore the exit code.

SirJson:
Unless they ruined clippy it used to have pretty good sugestions, it was kinda like a virtual reviewer but this was last year. 

Didn't they release some statistics recently that showed only a "problem" was the `complexity` module because it was by far the most reported lint so far?

Could be because people tend to write too complex code or a sensitivity problem in clippy...

I would start with the defaults minus `complexity` or at least only the `correctness` module because that might prevent bugs later on.

Kethku:
@j4qfrost whats the status on this. Is it ready to merge, do you need something from me? Sorry I haven't been keeping up

j4qfrost:
I actually wanted to make a habit of getting someone to take a look before merging for quality and just to make sure not just one person has all knowledge. I think it's all ready though.

Issue #303, opened 2020-05-28
Ligatures broken after updating


kirjavascript:
updated to the latest version of neovide-git and appear to be having a few font problems

here's my guifont and some example non-ligatures;

![broken font stuff](https://i.imgur.com/UxuVb9F.png)

I'm using the otf-nerd-fonts-fira-code AUR package, it all worked fine before upgrading and I haven't changed anything in my vimrc

upon closer inspection, it appears to have been caused by this commit:  [1892073c8a222e6dfd30039e7f691440533deeb5](https://github.com/Kethku/neovide/commit/1892073c8a222e6dfd30039e7f691440533deeb5)

j4qfrost:
Merged.

PR #304, opened 2020-05-28
Allow font fallback

theotherjimmy:
When debugging a font issue, I noticed that fallbacks are not set correctly. Seems we dropped a `!` in a recent refactor.


kirjavascript:
can confirm this fixes #303

j4qfrost:
Oops that ones on me. We went from greater than length to is empty. This needs unit tests. Thanks for catching it.

PR #311, opened 2020-06-12
Windows right click menu #94

nganhkhoa:
- [x] Custom commands
- [x] Right click directory and folder
- [x] Remove registry
- [X] Admin checks // just open the registry, if fail, no admin
- [ ] Error reporting // Tokio async problem
- [x] Move to a new module?

Also:
- [x] Setup client in Neovim
- [x] Get channel Id
- [x] Setup custom commands through `command! call ... rpcnotify(<channel_id>, 'event_name')`

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -207,11 +207,11 @@ async fn start_process(mut receiver: UnboundedReceiver<UiCommand>) {
     info!("Neovide registered to nvim with channel id {}", neovide_channel);
 
     #[cfg(windows)]
-    nvim.command(&build_neovide_command(neovide_channel, 0, "NeovideRegisterRightClick", "reg_right_click"))
+    nvim.command(&build_neovide_command(neovide_channel, 0, "NeovideRegisterRightClick", "register_right_click"))
         .await.ok();
 
     #[cfg(windows)]
-    nvim.command(&build_neovide_command(neovide_channel, 0, "NeovideUnregisterRightClick", "unreg_right_click"))
+    nvim.command(&build_neovide_command(neovide_channel, 0, "NeovideUnregisterRightClick", "unregister_right_click"))
         .await.ok();
 
     nvim.ui_attach(width as i64, height as i64, &options)
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -195,8 +195,11 @@ async fn start_process(mut receiver: UnboundedReceiver<UiCommand>) {
         }).unwrap_or(0);
     info!("Neovide registered with channel {}", neovide_channel);
 
+    #[cfg(windows)]
     nvim.command(&build_neovide_command(neovide_channel, 0, "NeovideRegisterRightClick", "neovide.reg_right_click"))
         .await.ok();
+
+    #[cfg(windows)]
     nvim.command(&build_neovide_command(neovide_channel, 0, "NeovideUnregisterRightClick", "neovide.unreg_right_click"))
         .await.ok();
 
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -108,6 +108,18 @@ async fn drain(receiver: &mut UnboundedReceiver<UiCommand>) -> Option<Vec<UiComm
     }
 }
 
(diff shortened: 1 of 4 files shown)


j4qfrost:
I would definitely like to see this separated to another module. It makes the file much easier to parse.

nganhkhoa:
I just don't know which folder and the name for the module, will move to a seperate file later. Maybe `settings/registry.rs`?

j4qfrost:
I would be more verbose and go with ```settings/windows_registry.rs```, I'll try to give it a thorough review once you're done, and once I am back from my trip at the end of the week. Thanks for working on this.

nganhkhoa:
I moved the module to another file, I'm trying to get the admin check running, but following [this answer](https://stackoverflow.com/questions/8046097/how-to-check-if-a-process-has-the-administrative-rights/8196291#8196291) yields no result.

j4qfrost:
I have no experience with the admin check issue, but I found this thread https://users.rust-lang.org/t/how-do-i-determine-if-i-have-admin-rights-on-windows/35710/4 

Does that help? @nganhkhoa

src/bridge/mod.rs
```diff
@@ -153,6 +165,44 @@ async fn start_process(mut receiver: UnboundedReceiver<UiCommand>) {
...
+        "ui",
+        vec![],
+        vec![]
+    ).await.ok();
+
+    // parsing the client list to get the channel id for neovide
+    let neovide_channel: u64 =
+        nvim.list_chans().await.ok()
+        .and_then(|chans| {
+            let (_, id) = chans.iter().find_map(|chan| {
```
Kethku, line 181: This may be heretical, but I wonder if this could be simplified by just using a for loop. To my eyes these iter, find_map, as_map, find calls look pretty confusing.

src/bridge/mod.rs
```diff
@@ -108,6 +108,18 @@ async fn drain(receiver: &mut UnboundedReceiver<UiCommand>) -> Option<Vec<UiComm
     }
 }
 
+fn build_neovide_command(channel: u64, nargs: u64, command: &str, action: &str) -> String {
```
Kethku, line 111: Beautiful. This is going to be useful for other future commands

Kethku:
Can you go into more details about what the admin check is and what happens if it fails? It may be reasonable to just say it doesn't work if you don't register while in admin mode and add that explicitly to the docs. Other than that, this looks awesome to me, only a minor code readability question

nganhkhoa:
This is the current code I use to check elevation, it just like most example code on the internet. But I got a false result when running in admin. I will try to fix this, been busy these days.
```rust
pub fn is_elevated() -> bool {
    let mut token_handle: HANDLE = null_mut();
    let mut elevation = TOKEN_ELEVATION {
        TokenIsElevated: 0
    };
    let result: bool;
    unsafe {
        OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &mut token_handle);
        GetTokenInformation(token_handle, TokenElevation,
                            &mut elevation as *mut _ as *mut c_void,
                            4 /* sizeof(DWORD) */, null_mut());
        result = elevation.TokenIsElevated != 0;
        CloseHandle(token_handle);
    }
    result
}
```

Also, the code is now just a proof of concept, there is no NSTATUS check so I still need to make these checks.

nganhkhoa:
I tried to make a call to err_write using `_nvim`. But I don't know how to put it on await, because `tokio::task::spawn_blocking` creates a closure with no async. How can I call a nvim api to display error?

nganhkhoa:
I may just leave the PR like this. I cannot call `neovim_err_writeln`. I don't know how to work arround the issue.

```rust
    async fn handle_notify(
        &self,
        event_name: String,
        arguments: Vec<Value>,
        _neovim: Neovim<Compat<ChildStdin>>,
    ) {
        _neovim.err_writeln("").await.ok(); // freezes Neovide
        task::spawn_blocking(move || match event_name.as_ref() {
            // put _neovim.err_writeln() in here is not possible, due to async in blocking
        }
    }
```

Kethku:
Sorry for not getting to this sooner. I will take a look at it. Thank you so much for your hard work

PR #312, opened 2020-06-16
Build without embed fonts

j4qfrost:
@SirJson I think this will help move some issues forward faster, so I went ahead and created another branch for better visibility.

@exoticus Can you check the caching?


j4qfrost:
Continues https://github.com/Kethku/neovide/pull/273

PR #314, opened 2020-06-16
Allows for multiple texts to be entered at once

j4qfrost:


src/window.rs
```diff
@@ -439,8 +439,15 @@ pub fn ui_loop() {
                     ..
                 } => {
                     keycode = received_keycode;
+                    keytexts.push((keycode, None));
+                }
+                Event::TextInput { text, .. } => {
+                    if let Some(keycode) = keycode {
+                        if text.len() > 1 || keycode.name() != text.to_uppercase() {
```
Kethku, line 446: text.to_uppercase feels possibly like a hack to me... Can you explain what the goal is here? Maybe it would be better just to remove the check?

src/window.rs
```diff
@@ -439,8 +439,15 @@ pub fn ui_loop() {
                     ..
                 } => {
                     keycode = received_keycode;
+                    keytexts.push((keycode, None));
+                }
+                Event::TextInput { text, .. } => {
```
Kethku, line 444: Do we know if textinput always happens after the keycode event?

src/window.rs
j4qfrost: Yeah, I think removing the check would be best also and I already have locally.

j4qfrost: I don't think SDL has an is_character function.

src/window.rs
j4qfrost: It appears that way. Though I place this line here in order to catch things like backspace and enter. It causes a double up with characters.

Kethku: my only concern is that sometimes sdl works differently on other platforms. You're probably right, but this may come back to bite us later

j4qfrost:
@Kethku Aaaa shoot I don't think I pushed any workable fixes into this branch.

Kethku:
I screwed up my bad

Kethku:
I should have mentioned I was planning on merging before just pulling the trigger.

Kethku:
reverting and reopening

j4qfrost:
Np, I haven't thought of a really good fix for this problem yet.

Kethku:
I'll play around with it. I'm running into it with a weird firmware in my keyboard

j4qfrost:
I wonder if we can integrate something like PistonWindow, though there are flexibility tradeoffs.

Kethku:
So your idea is to use the pistonwindow windowing system or do what pistonwindow did?

PR #316, opened 2020-06-23
Fix underscores not displayed on every line

obreitwi:
With certain font-configurations, underscores are not drawn on every line.

Locally, setting `set guifont=Deja\ for\ Powerline\ Medium:h11` caused underscores to be hidden on two lines and then displayed on the following three:
example of underscores not displayed

Fixing the line height to the nearest integer seems to alleviate the issue.


Kethku:
Thanks! this looks like a good fix

PR #321, opened 2020-07-07
Allows for multiple texts to be entered at once

Kethku:
Reopening https://github.com/Kethku/neovide/pull/314 my bad :(


Kethku:
I'm gonna close this because it doesn't solve the problem yet.

Issue #313, opened 2020-06-16
Missing input events


roflolilolmao:
### Description
Rapidly inputting key strokes often results in missing strokes.

### Expected behaviour
All key strokes, be it by hand or macros, should get to the text editor.

### Steps to reproduce
Either mash your keyboard or use external macro software (such as AutoHotKey).

For example, mashing 'aoeu' together results in:
> uuuu

In gvim, this would result in:
> aoeuaoeuouaeaoeu

### Environment
Windows 10 build 1909, using AutoHotKey and a mecanical keyboard with N-Key rollover support.

### Plausible cause
After a quick look through the code base, I noticed that, in the `ui_loop` function in `src/window.rs`, the variables `keycode` and `keytext` are set to the last `Event::KeyDown` or `Event::TextInput` of the `event_pump` iterator, ignoring the first such events of the loop.

j4qfrost:
@Kethku Is this caused by the animation?

Kethku:
I don’t think so. It’s likely either a race condition or something weird about the keyboard handling. Extra weirdly I don’t really encounter the described frequency. So I’m not sure what is going on here

j4qfrost:
@roflolilolmao You were right https://github.com/Kethku/neovide/pull/314

roflolilolmao:
> I don’t think so. It’s likely either a race condition or something weird about the keyboard handling. Extra weirdly I don’t really encounter the described frequency. So I’m not sure what is going on here

This might be bad wording on my end: I don't actually encounter it that much by regular typing but my AutoHotKey macros work about 10% of the time. Since they are important to my workflow, I'm not really sure I can switch to neovide just yet, which is a shame because it looks really good.

j4qfrost:
@roflolilolmao Can you check that my pull request works for you?

roflolilolmao:
I have tried compiling it myself this afternoon but LLVM was not being cooperative so I gave up. I'll give it another shot tomorrow. Thanks

j4qfrost:
Were you using just the binary release before? When you're ready to try again, ping me on gitter.

j4qfrost:
Or you can use this https://github.com/j4qfrost/neovide/suites/805035516/artifacts/8725885

roflolilolmao:
On #314, by hitting several keys at once or using a macro, I get all strokes. However, space and some shifted keys are duplicated or otherwise output weirdly.

[...]

PR #322, opened 2020-07-07
A possibly better fix for #313

Kethku:
This stores keycode and textinput information in a vector of pairs so that subsequent keycode and textinput events are kept together irrespective of the order in which they are pressed.

This seems to fix the issue of missing key presses but does not always get the ordering correct. This may be a limitation of the sdl keyboard input features.

@j4qfrost any thoughts? Sorry again for prematurely merging the other branch.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -77,7 +77,7 @@ pub fn window_geometry() -> Result<(u64, u64), String> {
                 .map(|dimension| {
                     dimension
                         .parse::<u64>()
-                        .or_else(|_| Err(invalid_parse_err.as_str()))
+                        .map_err(|_| invalid_parse_err.as_str())
                         .and_then(|dimension| {
                             if dimension > 0 {
                                 Ok(dimension)
--- a/src/window.rs
+++ b/src/window.rs
@@ -440,7 +440,6 @@ pub fn ui_loop() {
                     keycode: received_keycode,
                     ..
                 } => {
-                    info!("Keycode: {:?}", &keycode);
                     // If keycode has a value, add it to the list as the new keycode supercedes
                     // this one.
                     if keycode.is_some() {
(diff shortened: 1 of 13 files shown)


j4qfrost:
So the issue that was blocking my changes is that text would double up. I'm having trouble tracing how the extra text gets deduped.

Kethku:
My branch fixes that problem. I think the doubling would occur when events for both the keycode and the keytext got send separately from eachother. This gets around that by pairing them up. The assumption being that keycode and keytext will always come in pairs.

src/window.rs
```diff
@@ -438,9 +440,31 @@ pub fn ui_loop() {
...
                     keycode = received_keycode;
+
+                    if keytext.is_some() {
+                        keyboard_inputs.push((keycode, keytext));
+                        keycode = None;
+                        keytext = None;
+                    }
+                }
+                Event::TextInput { text, .. } => {
+                    if keytext.is_some() {
```
j4qfrost, line 456: This is where I'm afraid the double happens.

src/window.rs
```diff
@@ -438,9 +440,31 @@ pub fn ui_loop() {
                     keycode: received_keycode,
                     ..
                 } => {
+                    if received_keycode.is_some() {
```
j4qfrost, line 443: Also doubling up here?

src/window.rs
Kethku: id be interested in seeing how it works on non windows platforms. It doesn't double up on windows

src/window.rs
Kethku: this logic says if there is already a keycode then we can assume that the new keycode is a different key. So we send that keycode to the vector to get processed and then store the new keycode in the temp variable to be paired with a textinput if it becomes available.

src/window.rs
j4qfrost: Grabbing now.

src/window.rs
Kethku: So in practice no doubling up happens

j4qfrost: Wait I think there's a typo then. You want to check keycode and not received_keycode right?

Kethku: I'll add some comments and double check

src/window.rs
j4qfrost: There isn't any double up.

src/window.rs
Kethku: yup you're totally right. Doh

Kethku: Fixed and pushed. I also reformatted it slightly for dryness and readability

j4qfrost:
Looks good.

Kethku:
I'm going to test it just a bit more, (the ordering thing is still bugging me) but then I'll probably merge

PR #343, opened 2020-08-28
Winit

j4qfrost:
This is still missing some features provided by SDL, but it more or less works. I want to take a look and see if there can be any more refactoring that can be done. This might be good to reference in case there might be some areas we potentially step on each others' toes.


Kethku:
@j4qfrost I'm actually super interested in merging this, but we should set the default back to sdl2. I tested locally and its VERY promising, just need some fiddling to work out the bugs. I vote we merge soon and iterate together via PRs (or whatever depending on what time you're willing to commit). This is amazing work. Thank you so much!

Kethku:
https://github.com/Kethku/neovide/commit/60e1fe06148465fc23361d72e0d6468ac5d307b0 you can cherry pick this to add back icon support

Issue #86, opened 2020-01-30
Line ordering issues when scrolling under heavy load


jonvaldes:
Example. Notice the line numbers:
![image](https://user-images.githubusercontent.com/48644/73441923-224ce500-4354-11ea-9f0f-9d28d9b77f3d.png)

I'm just scrolling up/down while my computer is hammered by other processes consuming all available CPU power. The internal nvim buffers seem to preserve ordering correctly. All editing seems to work with the correct line ordering. This looks like just a rendering issue.

Edit: I should note this seems to happen for lines that come into view when scrolling. Once a line is in view, it remains as it is until it goes out of view. Sometimes, moving the cursor over the line will make the correct characters appear.

Kethku:
I believe this is highly related to https://github.com/Kethku/neovide/issues/41. Frustratingly I also think it is a race condition that I'm struggling to track down. This issue started when I swapped to using async and nvim-rs which fixed a ton of other bugs but introduced this one... 

Weirdly it's preserved when scrolling which seems to indicate to me that either the editor structure is missing a redraw event (or parsing it incorrectly) or it is parsing the clear portions of grid lines wrong. This one has been bothering me :/

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jonvaldes:
I've been toying around with this, and I don't know if it completely fixes this issue, but this change seems to at least make it better (at least for me):

```
 async fn handle_current_commands(receiver: &mut UnboundedReceiver<UiCommand>, nvim: &Neovim<Compat<ChildStdin>>) -> bool {
     if let Some(commands) = drain(receiver).await {
-        let (resize_list, other_commands): (Vec<UiCommand>, Vec<UiCommand>) = commands
-            .into_iter()
-            .partition(|command| command.is_resize());
-        if let Some(resize_command) = resize_list.into_iter().last() {
-            resize_command.execute(&nvim).await;
-        }
-
-        for ui_command in other_commands.into_iter() {
-            ui_command.execute(&nvim).await;
+        for command in commands {
+            command.execute(&nvim).await;
         }
```

I also have no idea why you were splitting off the resize events, so this might introduce other issues around that...

Kethku:
Interesting. To be honest I don't remember why I split those off. Could you try that for a while and let me know here if you run into it again?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Kethku:
[...]

PR #344, opened 2020-08-30
Nvim-rs: Fix https://github.com/Kethku/neovide/issues/86

KillTheMule:
Hey, 

so I finally came around trying to fix this issue for real. This is somewhat rough, but neovide builds and seems to work, so what I'd like to know at this point is: Does this fix https://github.com/Kethku/neovide/issues/86 indeed? I have no way to test, so I depend on someone else reproducing the issue and seeing if this is fixed indeed. If so, I'll work out the details in nvim-rs, and I'll send a PR for neovide when it's done.

Note: When quitting neovide with this PR, there's a panic somwhere (I guess from the `method_handler` task).  I don't think it matters at this point though :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ derive-new = "0.5"
 rmpv = "0.4.4"
 rust-embed = { version = "5.2.0", features = ["debug-embed"] }
 image = "0.22.3"
-nvim-rs = { git = "https://github.com/kethku/nvim-rs", features = [ "use_tokio" ] }
+nvim-rs = { git = "https://github.com/KillTheMule/nvim-rs", branch = "serial", features = [ "use_tokio" ] }
 tokio = { version = "0.2.9", features = [ "blocking", "process", "time" ] }
 async-trait = "0.1.18"
 lazy_static = "1.4.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1427,7 +1427,7 @@ dependencies = [
 [[package]]
 name = "nvim-rs"
 version = "0.1.1-alpha.0"
-source = "git+https://github.com/kethku/nvim-rs#3abe58a50eaaccccc15bef7260f89020e280b382"
+source = "git+https://github.com/KillTheMule/nvim-rs?branch=serial#26f12b147bf21b743948893c4eb36355302c4c17"
 dependencies = [
  "async-trait",
  "futures 0.3.4",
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -20,7 +20,7 @@ derive-new = "0.5"
 rmpv = "0.4.4"
 rust-embed = { version = "5.2.0", features = ["debug-embed"] }
 image = "0.22.3"
-nvim-rs = { git = "https://github.com/kethku/nvim-rs", features = [ "use_tokio" ] }
+nvim-rs = { git = "https://github.com/KillTheMule/nvim-rs", branch = "serial", features = [ "use_tokio" ] }
```
j4qfrost, line 23: Get this merged upstream first. It's bad practice to change the dependencies to another remote for maintenance reasons and security reasons.

j4qfrost: Just realized that you are upstream. lol

Kethku: I'll test it up in a bit

Kethku:
This appears to work. I'm still testing though as the failure before was intermittent

Kethku:
Further testing reveals that the test_read_initial_values test stalls with your branch and prints a reported exception to the console on this line: https://github.com/KillTheMule/nvim-rs/blob/serial/src/neovim.rs#L216. I think this is due to your branch, but I'm double checking.

Kethku:
Yup definitely throws and does not throw on my hack branch. Not sure whats going on

KillTheMule:
Can you show how you adapted the test? Maybe you're just not running the futures correctly :) I'll investigate that, but can you answer my original question if that fixes #86? The main binary does build, so you should be able to test that, right? Or maybe I've misunderstood something?

Kethku:
It appears to work other than that. You're right I likely adapted the futures incorrectly.

I'm in the middle of a pretty large refactor of the project, so I may have to put this on pause for a bit while I settle that out. Is that ok with you?

KillTheMule:
> I'm in the middle of a pretty large refactor of the project, so I may have to put this on pause for a bit while I settle that out. Is that ok with you?

Of course! Now that I know that this works as expected, I'm going to polish and rework stuff into something I can release. I'll try to make a PR for neovide then to switch you over to a release (crossing fingers here :))

Kethku:
@KillTheMule I just finished the above mentioned refactor, so I'm going to reopen this PR and take a look at the changes to see what I can incorporate.

Kethku:
Closing because I think the current workaround is sufficient.

PR #348, opened 2020-09-01
To fix can not switch IME issue.

uhuntu:


j4qfrost:
@Kethku is it possible to get rid of these lines completely? What errors pop up on Windows if they aren't there?

PR #350, opened 2020-09-02
To fix can not switch IME issue.

uhuntu:


uhuntu:
I think I found the root cause. https://github.com/rust-lang/cargo/issues/7914

So the command line would be:

```
cargo run -Z features=itarget
```

PR #357, opened 2020-09-09
Add window maximization

j4qfrost:
This is a bit of a quick hack. The clean way would involve deglobalizing EDITOR.


Kethku:
my multigrid branch deglobalizes the bridge, editor, and renderer. It leaves the redraw scheduler and settings as globals since their influence is broad and largely immutable (or at least lockless)

For now this unblocks the user, but I kinda feel like using the same command creation logic that was introduced for the context menu integration work that was merged recently would be a more vim-y way to solve this problem.

What do you think?

j4qfrost:
Yeah, I was going for that approach initially, then I realized just how much I would have to change to make sending the command play nice, I can wait. I need to take a look at your multigrid changes.

Kethku:
I'm actively working on it, its just gonna be a bit

Kethku:
FYI still working on this. I want to at least squash most of the bugs before I merge cause there seems to be a lot of folks running from master

j4qfrost:
I know you don't mean it that way, but I'm sorry lol.

Kethku:
I 100% was not saying that as a poke at you!

Kethku:
superseded by https://github.com/Kethku/neovide/pull/442/files

PR #358, opened 2020-09-10
font-kit = "0.10.0"

uhuntu:


src/renderer/mod.rs
```diff
@@ -175,6 +175,7 @@ impl Renderer {
             self.surface = None;
         }
 
+        #[allow(deprecated)]
```
j4qfrost, line 178: Don't do this.

.gitignore
```diff
@@ -1,3 +1,4 @@
 /target
 **/*.rs.bk
 *.log
+Cargo.lock
```
j4qfrost, line 4: @Kethku I am curious why we keep Cargo.lock. It does help give more stable builds.

Kethku: Cargo lock makes builds way more consistent. I would prefer to keep this in unless a cogent argument can be made not to

src/renderer/mod.rs
Kethku: I'm curious about why this was added. Any details @uhuntu?

Kethku:
@uhuntu I appreciate updates like this, but I'm interested in the motivation. Is there a missing feature or something?

src/renderer/mod.rs
uhuntu: When I commit something, it will say
```
error: use of deprecated item 'skia_safe::core::types::Budgeted::YES': use Yes
   --> src/renderer/mod.rs:180:28
    |
180 |             let budgeted = Budgeted::YES;
    |                            ^^^^^^^^^^^^^
    |
    = note: `-D deprecated` implied by `-D warnings`

error: aborting due to previous error

error: could not compile `neovide`.
```

.gitignore
uhuntu: Yes, I understand now.

uhuntu:
> @uhuntu I appreciate updates like this, but I'm interested in the motivation. Is there a missing feature or something?

I can not reproduce now when I kept Cargo.lock

uhuntu:
But one possible reason is https://github.com/linebender/skribo/commit/759d664a53c444155dbc08fe1e4912ecdf83f9e2

src/renderer/mod.rs
pragmatrix: `...: use Yes`

use `Budgeted::Yes` to omit the warning.

PR #359, opened 2020-09-14
Upgrade sdl2-sys to 0.34.3

uhuntu:


j4qfrost:
This isn't completely necessary. 0.34 covers 0.34.3 and future patches. If you really want bleeding edge features, you can do ```cargo update```

j4qfrost:
@Kethku should we consider ignoring the lock file? Maybe having unstable builds will motivate us to keep the project more up-to-date and easier to predict with regard to bug tracking.

j4qfrost:
@uhuntu bumping dependencies isn't very descriptive as a pull request title. Highlighting the specific features or bugs would be more productive in the pull request description. We could ask and wait for your response, but it's much smoother if you volunteer that information when you make the request. Glad for your interest, but please be more verbose. Thanks.

Kethku:
I don't think ignoring the lock is worth it. The benefits outway the nonexistent (?) cost imho. But I'm definitely interested in any arguments folks might have against.

Kethku:
Re dependency bumps, I agree. Its awesome when an upstream dependency bumps versions and makes a change that effects us, but without a description, its hard to know what its actually doing or what the benefit to the project is.

j4qfrost:
> I don't think ignoring the lock is worth it. The benefits outway the nonexistent (?) cost imho. But I'm definitely interested in any arguments folks might have against.

The old lock file actually blocked an update to 0.34.3, which contained my fix for CFLAGS not included in 0.34.0. This is a Linux specific issue. I didn't notice this until I went back and rebuilt with a new lock file. It looks like the lock file prevented an update to skulpin, 0.4.0 contained the fix.

I think that if we want to keep the lock file we should cherry pick libraries that are most prone to updates and use the full version. This will prevent overnight updates from breaking things and make sure that the lock file doesn't restrict dependencies to specific versions.

uhuntu:
@j4qfrost @Kethku I updated the description of the PR

uhuntu:
> This isn't completely necessary. 0.34 covers 0.34.3 and future patches. If you really want bleeding edge features, you can do `cargo update`

After I cargo update, I fall into another issue:
```
hunt@hunt-OptiPlex-7071:~/work/neovide$ cargo build
error: multiple packages link to native library `fontconfig`, but a native library can be linked only once

package `servo-fontconfig-sys v4.0.6`
    ... which is depended on by `servo-fontconfig v0.4.0`
    ... which is depended on by `font-kit v0.6.0`
    ... which is depended on by `neovide v0.6.0 (/home/hunt/work/neovide)`
links to native library `fontconfig`

package `servo-fontconfig-sys v5.1.0`
    ... which is depended on by `servo-fontconfig v0.5.1`
    ... which is depended on by `font-kit v0.10.0`
    ... which is depended on by `skribo v0.1.0 (https://github.com/linebender/skribo#a35c876b)`
    ... which is depended on by `neovide v0.6.0 (/home/hunt/work/neovide)`
also links to native library `fontconfig`
```

Looks like it related to https://github.com/Kethku/neovide/pull/358 also

j4qfrost:
@uhuntu at this point I would defer to @Kethku. I think there needs to be a discussion about tracking dependencies overall rather than simply patching things as they come. We may need to switch away from using git for dependencies.

uhuntu:
After I run ```cargo update -p sdl2-sys```, then things go better https://github.com/Kethku/neovide/compare/master...uhuntu:sdl2-sys

PR #360, opened 2020-09-14
Upgrade sdl2-sys to 0.34.3

uhuntu:
This will include the fix: https://github.com/Rust-SDL2/rust-sdl2/pull/1027
That actuall update SDL2 to 2.0.12
And it will fix the IME not showing issue in Linux
Also refer to https://github.com/Rust-SDL2/rust-sdl2/pull/1025 also

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1938,9 +1938,9 @@ dependencies = [
 
 [[package]]
 name = "sdl2-sys"
-version = "0.34.0"
+version = "0.34.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "575db20c45cce6565d4ece46509288ded53f5571dc89e4af165a562eb0eeec26"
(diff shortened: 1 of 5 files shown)


Kethku:
Ok, with the extra details about what is getting fixed, it makes sense to bump the versions in the Cargo.toml. My view of Cargo.lock files (And I'm very willing to be persuaded out of this view) is that they are a way to make builds consistent across runs without worrying about package updates breaking things. Given the absolutely massive number of dependencies that we currently build for the project, I believe they are a necessary evil. However when we want to make an actual change to a package dependency version or what have you, the way it should be done is by updating the cargo.toml to make the change, not just by calling cargo update to update the Cargo.toml file.

It sounds like you have a specific package update in mind with this change, so would it be possible to encode that package update into the cargo.toml versions rather than just via the lock file? I really appreciate your tenacity with this fix. Sorry its been somewhat muddled.

uhuntu:
I recommited the patch, and run ```cargo fmt``` also.

Kethku:
Thanks so much again, and sorry for the back and forth.

PR #361, opened 2020-09-15
Winit Refactor, Window Management, Modifier

j4qfrost:
I merged in master and made significant changes to the winit windowing, adding window management and rearranging how events are handled for cleanliness and granularity. Window management may not be leveraged in upstream in the future but from this I'll probably be forking way off the path. The modifier usability is fixed, so you get capital letters and other stuff. The rendering still feels a bit off. 

This is still a draft, because it contains some SDL window changes that may or may not be pleasing with the current state of Neovide. These changes were made in order to potentially support multi-window and for more granularity, but it turned out to be more difficult to leverage the events for multi-window with SDL. Thoughts would be appreciated.


j4qfrost:
Closing in favor of newer draft. #417

PR #347, opened 2020-08-31
Multigrid

Kethku:
Work in progress multigrid support. Renders works now which is awesome. There are some remaining bugs and the occasional render issue which is concerning.

Remaining work:
- [x] Setting to enable and disable multigrid extension (this may not be possible at runtime)
- [x] Grid based cursor positioning
- [x] Clip offscreen windows
- [x] Blurred background of floating windows
- [x] Animated grid positioning
- [x] General code cleanliness
- [x] Perf testing
- [x] Windows appear in top left corner before being re-positioned
- [x] No way to disable multigrid support
- [x] Popup windows prevent neovide from responding to ui commands
- [x] No way to disable window animations
- [x] No way to disable or configure window transparency

PR #364, opened 2020-09-22
glfw

uhuntu:
This is my first try to implement by glfw


Cargo.toml
```diff
@@ -7,17 +7,18 @@ build = "build.rs"
 description = "A simple GUI for Neovim."
 
 [features]
-default = ["sdl2"]
+default = ["glfw"]
```
j4qfrost, line 10: Leave this as "sdl2" and simply run with features=glfw.

Cargo.toml
```diff
@@ -7,17 +7,18 @@ build = "build.rs"
...
+glfw = ["skulpin/skulpin_glfw"]
 winit = ["skulpin/skulpin_winit"]
 
 [dependencies]
 euclid = "0.20.7"
 font-kit = "0.6.0"
 skribo = { git = "https://github.com/linebender/skribo" }
 lru = "0.4.3"
-skulpin = { git = "https://github.com/aclysma/skulpin" }
+skulpin = { git = "https://github.com/uhuntu/skulpin", branch = "glfw-windows" }
```
j4qfrost, line 21: Get this validated with upstream first.

src/bridge/layouts/mod.rs
```diff
@@ -123,6 +132,36 @@ pub fn produce_neovim_keybinding_string(
...
+    keycode: Option<Keycode>,
+    keytext: Option<String>,
+    modifiers: Option<Mod>,
+) -> Option<String> {
+    let mut shift = false;
+    let mut ctrl = false;
+    let mut alt = false;
+    let mut gui = false;
+    if let Some(modifiers) = modifiers {
+        shift = modifiers.contains(Mod::Shift);
```
j4qfrost, line 146: This is usually handled using bit shifts, so there isn't a need for an entire vector to handle the state. Is that how the modifier is implemented?

src/bridge/ui_commands.rs
```diff
@@ -3,7 +3,9 @@ use nvim_rs::compat::tokio::Compat;
 use nvim_rs::Neovim;
 use tokio::process::ChildStdin;
 
+#[cfg(not(feature = "glfw"))]
```
j4qfrost, line 6: This could possibly wait for #347 to be merged in.

src/bridge/ui_commands.rs
```diff
@@ -56,6 +65,7 @@ impl UiCommand {
                         .expect("Mouse Input Failed");
                 }
             }
+            #[cfg(not(feature = "glfw"))]
```
j4qfrost, line 68: This is pretty important to not include.

j4qfrost:
Good, start but it's still got a ways away. I would start by getting your skulpin branch approved upstream first.

Cargo.toml
uhuntu: https://github.com/aclysma/skulpin/pull/72

src/bridge/layouts/mod.rs
uhuntu: Actually I did not completely get what you mean ? I just reference the code from both sdl2 and winit

src/bridge/ui_commands.rs
uhuntu: I know, for now, the branch is mostly prove the theory, I will try to implement the rest of them in spare time.

src/bridge/layouts/mod.rs
j4qfrost: I double checked. This is fine. I thought this was self-implemented.

j4qfrost:
I gave aclysma a heads up for your changes for glfw. Get that merged and implement scrolling, then I will run through it and do a final check.

src/bridge/ui_commands.rs
uhuntu: https://github.com/Kethku/neovide/pull/364/commits/a97fb234ea7eeefc5c8f0f3b89f54b8f227d6da4

src/bridge/ui_commands.rs
uhuntu: https://github.com/Kethku/neovide/pull/364/commits/a97fb234ea7eeefc5c8f0f3b89f54b8f227d6da4

src/window/glfw.rs
```diff
@@ -273,7 +273,7 @@ pub fn initialize_settings() {
         .contains(&String::from("--noIdle"));
 
     SETTINGS.set(&WindowSettings {
-        refresh_rate: 60,
+        refresh_rate: 120,
```
j4qfrost, line 276: Do you have a reason for changing this? Not trying to micromanage, just curious.

uhuntu: Yes, you find the root point, I realized that glfw seems performance lower than sdl2, so I just increase that to 120, then seems performance improved. And I am not sure if it has any side effect, stay tuned.

Kethku: Interesting. I will dig into this more once a full PR is ready. I suspect there  are minor differences in the way each windowing system queues a new frame.

src/bridge/ui_commands.rs
uhuntu: https://github.com/Kethku/neovide/pull/364/commits/2c3cde00d3fe4c9d54d8ef75a6afe0cad154f7c8

Cargo.toml
uhuntu: https://github.com/uhuntu/neovide/commit/5a924dcfbc26b1d413ce52b1ce1495032b89b6a4

uhuntu: https://github.com/aclysma/skulpin/issues/68#issuecomment-699685177

aclysma:
Regarding upstreaming GLFW support in skulpin, I left some comments here: https://github.com/aclysma/skulpin/issues/68#issuecomment-699685177

I'm happy to accept a PR for a GLFW implementation as long as it includes an example that works reasonably well on windows/mac/linux. I prototyped an implementation and example but it does not meet that requirement (on macOS anyways) as resizing the window makes the window contents permanently black.

While I don't want to upstream that implementation/example in it's broken state, I did upstream the changes that would be necessary to continue working on it out-of-tree. It's really quite small and it's very possible the bug is just bad example code!
 * Implementation: https://github.com/aclysma/skulpin/blob/6e54523f230d86f57c8a96612173aba584a343fd/skulpin-renderer-glfw/src/lib.rs
 * Example code: https://github.com/aclysma/skulpin/blob/6e54523f230d86f57c8a96612173aba584a343fd/examples/glfw_renderer_only.rs
 * It's in this branch: https://github.com/aclysma/skulpin/tree/glfw

uhuntu:
As winit should be final choice, I would like to close this PR for this moment.

PR #380, opened 2020-10-09
Issue 365

j4qfrost:
https://github.com/Kethku/neovide/issues/365 Solves this issue, but I'm wondering if there is a better way.


Kethku:
Seems reasonable to me

Kethku:
This was fixed in a different pr

PR #390, opened 2020-10-21
check font stretch property when selecting

NBonaparte:
Solves an issue I had where it was selecting a font with the correct weight and style but wrong "stretch".

### DIFF
--- a/src/renderer/caching_shaper.rs
+++ b/src/renderer/caching_shaper.rs
@@ -73,7 +73,10 @@ impl ExtendedFontFamily {
                 let font = &handle.font;
                 let properties = font.properties();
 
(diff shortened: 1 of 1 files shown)


Kethku:
Very interesting. Do you have any data as to what stretch means?

NBonaparte:
Stretch refers to whether the font is condensed/expanded. In this case, I set the font to Iosevka, but it selected Iosevka Expanded for bold fonts, which caused the bold text to be cut off at the end.

neovide_bold_expanded

Kethku:
Gotcha. This makes perfect sense to me!

Kethku:
Gotcha. This makes perfect sense to me!

Issue #50, opened 2020-01-27
Enable Remote Neovim Process Connection


Kethku:
It would be cool to be able to connect to a remote neovim process such as one in wsl or over the network.

KillTheMule:
That should be pretty easy using [tcp](https://github.com/KillTheMule/nvim-rs/blob/master/src/create/tokio.rs#L44). I tried investigating using UDS, but there does not seem to be a library providing those as `AsyncWrite/AsyncRead`ers.

Kethku:
Do you mean UDP? Just trying to follow. I think tcp is probably the way to go as I don't want to have to track lost packets

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

KillTheMule:
No, [UDS](https://lib.rs/crates/uds_windows). UDP is unsuitable indeed, but afaiu UDS is just like a unix socket, just on windows. No matter, doesn't look like I can support it anyways.

Kethku:
Oh I see! Yes that makes sense.

awlayton:
I am not able to compile on Windows to try, but can you not simply run `wsl nvim --embed` instead of `nvim --embed` to attach to the wsl one?

Kethku:
This is a very good question to which I have no answer

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Kethku:
I will attempt it shortly. I think it might work

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Kethku:
Blocked on https://github.com/Kethku/neovide/issues/97

BatmanAoD:
The trivial approach of hardcoding `wsl` as the command and `nvim` as the first argument doesn't seem to work for me. I've printed the debug representation of `cmd` just as a sanity-check:

```
     Running `target\debug\neovide.exe`
Command { std: "wsl" "nvim" "--embed", kill_on_drop: false }
```

A window appears briefly, but then disappears, and the app exits without showing any error output. The exit code is 0.

Kethku:
wsl seems to work now.

BatmanAoD:
What is the command-line argument for it? Do I need to have neovim installed on the non-WSL side?

Kethku:
Reopening to remind myself to respond when I get back to the computer

BatmanAoD:
Is there a command line argument in the 0.6.0 release for this?

Kethku:
I believe it is. 

[...]

PR #404, opened 2020-12-09
Fix --wsl option

calvinkosmatka:
Hi,
This PR addresses the WSL issues mentioned in #50. It works locally, but could probably use some more testing. Let me know what you think of the changes!

Edit: spoke too soon, release build didn't work. Removing `#![cfg_attr(not(debug_assertions), windows_subsystem = "windows")]` makes it work for some reason.

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -29,16 +29,14 @@ fn set_windows_creation_flags(cmd: &mut Command) {
 
 #[cfg(windows)]
 fn platform_build_nvim_cmd(bin: &str) -> Option<Command> {
-    if !Path::new(&bin).exists() {
-        return None;
-    }
-
     if env::args().any(|arg| arg == "--wsl") {
         let mut cmd = Command::new("wsl");
         cmd.arg(bin);
         Some(cmd)
-    } else {
+    } else if Path::new(&bin).exists() {
         Some(Command::new(bin))
+    } else {
+        None
     }
 }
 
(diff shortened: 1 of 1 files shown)


calvinkosmatka:
Ok, `nvim.eval("has(\"nvim-0.4\")").await`  returns an
```rust
Err(DecodeError(InvalidMessage(NotAnArray(Integer(PosInt(84))))))
```
consistently with windows_subsystem =  "windows", causing neovide to exit with the "Neovide requires version 0.4 or higher" message, and a 0 exit code. nvim_rs indicates [here](https://github.com/KillTheMule/nvim-rs/blob/master/src/error.rs#L50) that InvalidMessage errors are indicative of neovim bugs, but I'm not sure if that's the right path to go down here.

calvinkosmatka:
After a bit of research into WSL io redirection issues, I guessed that using inherited Stdio for WSL's stderr was causing the rpc problems when running the release build. I changed it to used piped stderr and it worked perfectly. @Kethku I think this is _actually_ ready for review/further testing now.

A couple specific questions I have:

- Do you think I should move the WSL command building portion into a separate function? Maybe `wsl_build_nvim_cmd`?
- Does how I cfg-gated inherited vs piped stderr seem reasonable (particularly for non-Windows)?

Kethku:
@calvinkosmatka I just merged a big change which changes pretty much every file in the project. It cleans up a bunch of issues and makes the architecture cleaner, but results in open PRs needing to be basically rebuilt on top of it.

Kethku:
I'm happy to port it if you'd like, or you can do it. Up to you :)

calvinkosmatka:
It looks like the parts of bridge/mod.rs that I touch here stayed mostly intact, so I think I'll be able to rebase without too much of an issue. I'll give it a try a little later today!

calvinkosmatka:
@Kethku A little trickier than I thought, but it's working now!

Kethku:
Ok I'll take a look soon. Thanks for the work!

Kethku:
Looks great. Thank you so much!

Kethku:
https://github.com/Kethku/neovide/wiki/Configuration#wsl the documentation for the feature has been added here. I'll elaborate a little in the readme too!

keepitsane:
How would one go about using Node and Python providers from WSL rather then the Windows ones when using the `--wsl` option. I have my providers set to the WSL side path, but when starting Neovide using wsl it seems to default to the providers from the Windows side. Thanks in advance!

PR #406, opened 2020-12-11
added version info and exit on unknown command line option

tarunzoot:
the issue arises mainly because `-h ` or `--version` quits the `neovim` instance immediately. Since the user already have `neovim` installed, so echoing neovim version back is not logical when user supplies `[-v | --version]`, instead `[-v | --version]`  should print `neovide`'s version and help. Other thing is to check if we can access `neovim` API before executing any commands on `neovim` instance, because then we can be sure we are accessing a valid instance.

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -132,6 +132,11 @@ async fn start_process(mut receiver: UnboundedReceiver<UiCommand>) {
             .await
             .unwrap_or_explained_panic("Could not locate or start the neovim process");
 
+    if nvim.get_api_info().await.is_err() {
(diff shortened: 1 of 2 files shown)


j4qfrost:
Alright looks good, I'll merge this after you ping me back that you saw this message.

First, the name of your branch shouldn't include "#", it's not a big deal, just annoying because I have to enclose the branch in quotes when I switch to it. Second, this is probably any issue on our side that we don't have a proper changelog, but it would be nice to have a brief description about what you changed or what small functionality you changed, but again no big deal.

Do you want to add more descriptions for the help section? Also I imagine that once we get the project to 0.7, we will start to add more verbose version descriptions.

Good start.

Issue #402, opened 2020-12-08
[macOS] neovide process disown from terminal?


mitnk:
I'm new to neovide. When I build neovide with `cargo build --release`, and running it (ie. `./target/release/neovide`) from the shell (terminal), it does not disown with the terminal. The neovide spawned the window, but the command in shell does not return until neovide terminate. Meaning closing the terminal window (or Ctrl-C) also closes the neovide window.

For now I have to define a shell function like this: 
```sh
function nvim() {
    nohup neovide $@ > /dev/null &
}
```
So that when I type `nvim foo.txt` and closing the terminal tab does not closing the window of enovide too.

Maybe we need a script like [mvim](https://github.com/macvim-dev/macvim/blob/master/src/MacVim/mvim) in MacVim?

tarunzoot:
a quick way to do that is to put the following code in `src -> main.rs` after `#[cfg(target_os = "macos")] {`, and creating an alias `alias nvim=neovide --disown`
```
#[cfg(not(debug_assertions))]
        if std::env::args().find(|f| f == "--disown").is_some() {
            if let Ok(curr_exe) = std::env::current_exe() {
                assert!(std::process::Command::new(curr_exe)
                    .args(std::env::args().skip(1).filter(|f| f != "--disown"))
                    .spawn()
                    .is_ok());
                std::process::exit(0);
            }
        }
```

j4qfrost:
This looks good to me. Feel free to make a pull request. I have been a bit tied up recently, but I can merge the changes this weekend.

mitnk:
Consider to make this `--disown` option default? Since neovide is an GUI app, not depends on the shell/terminal, right?

Kethku:
Yeah I agree. On mac machines we should just do this by default. On windows this is the normal behavior. Not sure about linux

tarunzoot:
okay, so to do that the modified code is
file -> `main.rs`
```rust
#[cfg(not(debug_assertions))]
        if std::env::args().find(|f| f == "--disowned").is_none() {
            if let Ok(curr_exe) = std::env::current_exe() {
                assert!(std::process::Command::new(curr_exe)
                    .args(std::env::args().skip(1))
                    .arg("--disowned")
                    .spawn()
                    .is_ok());
                std::process::exit(0);
            } else {
                eprintln!("error in disowning process, cannot obtain the path for the current executable, continuing without disowning...");
            }
        }
```

[...]

PR #411, opened 2020-12-18
macos neovide disown from terminal #402

tarunzoot:
nevoid can now be run on macOS without spawning a new terminal, it will disown itself on the first run.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -36,7 +36,7 @@ fn main() {
     {
         // incase of app bundle, we can just pass --disowned option straight away to bypass this check
         #[cfg(not(debug_assertions))]
-        if std::env::args().find(|f| f == "--disowned").is_none() {
+        if !std::env::args().any(|f| f == "--disowned") {
             if let Ok(curr_exe) = std::env::current_exe() {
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -34,6 +34,21 @@ fn main() {
 
     #[cfg(target_os = "macos")]
     {
+        // incase of app bundle, we can just pass --disowned option straight away to bypass this check
+        #[cfg(not(debug_assertions))]
+        if std::env::args().find(|f| f == "--disowned").is_none() {
```
keithsim-msft, line 39: minor nitpick: to me using `!any()` rather than `find().is_none()` makes more sense as we wouldn't be depending on the option being none for the boolean value

Kethku, line 39: minor nitpick: to me using `!any()` rather than `find().is_none()` makes more sense as we wouldn't be depending on the option being none for the boolean value

Kethku:
I think this makes sense to merge. One thing I'm curious about, how does this effect startup performance? I suspect it might effect it a bit, but I wonder by how much?

PR #416, opened 2020-12-25
Layouts refactor

tim-harding:
Refactoring the layouts module for readability as I begin to comb through and understand the project. This is a totally optional PR, but hopefully I've done a decent enough job of compressing and clarifying the code that it will be worthwhile to merge for future maintenance.


Kethku:
Apologies for the massive rewrite. I'm more than happy to take refactoring prs though so long as you're ok with discussion! Always looking for collaborators

PR #417, opened 2020-12-26
Winit multiwindow manager

j4qfrost:
Merry Christmas! At least it still is in PST. This draft has some of my initial ideas for multiple windows.I had to throw things together a bit haphazardly, because I had been away from the project a while. It runs with ```--features=winit --no-default-features```, but is broken for sdl


src/bridge/layouts/winit.rs
```diff
@@ -0,0 +1,164 @@
+use super::unsupported_key;
```
Kethku, line 1: Layouts got moved to the window folder which is where they should have been to start, I just had them in the bridge for historical reasons. So these files aren't necessary anymore

src/bridge/ui_commands.rs
```diff
@@ -112,25 +112,25 @@ impl UiCommand {
                 if unregister_rightclick() {
                     let msg = "Could not unregister previous menu item. Possibly already registered or not running as Admin?";
                     nvim.err_writeln(msg).await.ok();
-                    error!("{}", msg);
```
Kethku, line 115: I'm curious about this change. Wouldn't we want to log these as errors to the console if the registration fails?

src/renderer/mod.rs
```diff
@@ -59,11 +60,11 @@ pub struct Renderer {
     pub font_width: f32,
     pub font_height: f32,
     pub window_regions: Vec<WindowDrawDetails>,
-    pub batched_draw_command_receiver: Receiver<Vec<DrawCommand>>,
+    pub batched_draw_command_receiver: Option<Arc<Receiver<Vec<DrawCommand>>>>,
 }
 
 impl Renderer {
-    pub fn new(batched_draw_command_receiver: Receiver<Vec<DrawCommand>>) -> Renderer {
```
Kethku, line 66: Minor tidbit, but I think the default version of new should take the draw command receiver as I see the window as requiring it to function normally. But clearly you need a version of the renderer which doesn't have it hooked up to start, so I think we should have a separate new, maybe "new_unhooked" which constructs the renderer without the draw command channel? What do you think?

src/settings/mod.rs
```diff
@@ -12,6 +12,7 @@ use nvim_rs::Neovim;
 use parking_lot::RwLock;
 pub use rmpv::Value;
 use tokio::process::ChildStdin;
+#[cfg(windows)]
```
Kethku, line 15: I like this change

src/window/keyboard.rs
```diff
@@ -1,81 +1,81 @@
...
-        result = format!("D-{}", result);
-    }
-
-    if special {
-        result = format!("<{}>", result);
-    }
-
-    result
-}
+use log::error;
```
Kethku, line 1: I assume this is diffed due to newlines? I reformatted my machine mid PR and things got a little weird :l

src/window/mod.rs
```diff
@@ -101,20 +101,13 @@ pub fn create_window(
 ) {
     let (width, height) = window_geometry_or_default();
 
-    let renderer = Renderer::new(batched_draw_command_receiver);
```
Kethku, line 104: I'm curious about the moving of the size here. Can you give some details?

Kethku:
@j4qfrost this pretty out of date. Are you still planning on coming back to this or should I close it as stale?

j4qfrost:
I don't mind closing it, I think even while I was working on it the overall
project had a different trajectory. I've been distracted by other things,
will be back eventually. Hope you guys are well.

PR #418, opened 2020-12-26
Keyboard refactor

tim-harding:
This refactoring does not add or change any features. Rather, it improves code reuse in the keyboard layouts code and makes the key code remapping more terse. It also aids discoverability by changing some large tuples related to modifier keys to structs. The number of `format!()` calls has been reduced in the `neovim_keybinding_string` function to minimize allocations. This is very much an optional PR, but hopefully you will find the refactorings useful enough for maintainability to be worth merging.


Kethku:
Thanks for the PR I'll try to look at it shortly. It might take me a bit, but I want to get to it :)

Kethku:
looked over it. Looks great. Thanks so much!

PR #419, opened 2020-12-27
Springy cursor

tim-harding:
I've created an alternative cursor movement for your consideration. The cursor smear is very fun and got my interested in this project. I thought I'd offer another interpretation. It uses a spring-damper setup rather than a lerp on the corners. It also adds volume preservation on the squash and stretch. 

Here are some examples of the new movement style. Critical damping gives a more subdued feel:
https://user-images.githubusercontent.com/13814470/103178181-65019080-4835-11eb-9dfe-acfab602cfb3.mp4

Underdamping gives a snappier, more playful feel with a bit of overshoot:
https://user-images.githubusercontent.com/13814470/103178200-911d1180-4835-11eb-8e9c-577211d1ea6c.mp4

As an aside, the codebase has great discoverability, even with few comments. I love how easy it was to add new global settings.


src/renderer/cursor_renderer/mod.rs
```diff
@@ -22,6 +22,10 @@ pub struct CursorSettings {
     antialiasing: bool,
     animation_length: f32,
     animate_in_insert_mode: bool,
+    springloaded: bool,
```
Kethku, line 25: Would it make sense to have a cursor_trail enum similar to the VfxMode rather than a boolean here? I feel like that way it leaves the door  open for future trails.

Alternatively I could see an argument that this is just a better implementation of the trail animation... I don't really know without playing with it a bit more.

tim-harding: That's a good idea. I'll have a look into it in the next couple of days.

src/renderer/cursor_renderer/mod.rs
```diff
@@ -290,16 +318,54 @@ impl CursorRenderer {
...
-                let corner_animating = corner.update(
-                    &settings,
-                    font_dimensions,
-                    center_destination,
-                    dt,
-                    !settings.animate_in_insert_mode && in_insert_mode,
-                );
+            if settings.springloaded {
+                // Note: implicit euler integration becomes unstable for
+                // spring constant values greater than about 1024.
```
LakeByTheWoods, line 323: Velocity Verlet integration might give some extra stability.

Kethku:
@tim-harding ping? Do you plan on coming back to this? I may close it as stale if not given that there are merge conflicts.

tim-harding:
Closing as stale. Please forgive the abandonment.

PR #423, opened 2020-12-30
Smooth scrolling

Kethku:
Implements smooth scrolling by positioning window textures inside of the window bounds according to a scrolled position which animates into place rather than jumping immediately. Implements and further details about the implementation can be found there https://github.com/Kethku/neovide/issues/62

### DIFF
--- a/src/bridge/events.rs
+++ b/src/bridge/events.rs
@@ -926,6 +926,7 @@ pub fn parse_redraw_event(event_value: Value) -> Result<Vec<RedrawEvent>> {
             "win_hide" => Some(parse_win_hide(event_parameters)?),
             "win_close" => Some(parse_win_close(event_parameters)?),
             "msg_set_pos" => Some(parse_msg_set_pos(event_parameters)?),
+            "win_viewport" => Some(parse_win_viewport(event_parameters)?),
             "cmdline_show" => Some(parse_cmdline_show(event_parameters)?),
             "cmdline_pos" => Some(parse_cmdline_pos(event_parameters)?),
             "cmdline_special_char" => Some(parse_cmdline_special_char(event_parameters)?),
--- a/src/bridge/events.rs
+++ b/src/bridge/events.rs
@@ -216,6 +216,13 @@ pub enum RedrawEvent {
         scrolled: bool,
         separator_character: String,
     },
+    WindowViewport {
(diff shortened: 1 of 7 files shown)


Kethku:
@tim-harding you mentioned interest in this feature. Would you be willing to look over the changes here and weigh in? I think this is pretty good, but another opinion would be great!

tim-harding:
I tried checking out the branch the other day and didn't see the smooth scroll happening, but perhaps I'm just missing a setting somewhere. I'll have another look and see if I can get it working.

szbergeron:
@tim-harding you may need to enable `--multiGrid`, I ran into that just now and had to debug.

Kethku:
yes multiGrid is required for now. You can set it via an environment variable as well https://github.com/Kethku/neovide/wiki/Configuration#multiGrid

Kethku:
I keep it behind a flag because there are a bunch of upstream bugs with respect to mouse input and completion menu placement that I would like to have fixed before this is enabled by default. Which is a mad bummer because the multigrid stuff is really awesome, but it is what it is

szbergeron:
@Kethku is there any way this could take in precise scroll events on linux systems? It should be available through xinput2 or native wayland protocols. Not sure if the windowing lib you're using supports those yet or if more groundwork would be needed. This would allow scrolling fractions of a line, and should make more elaborate animations easier (fling, bounce, etc)

Kethku:
Yes and no. Technically, I think I could respond to scroll events more precisely and allow the screen to stay at partial scroll positions. This becomes complicated though when the cursor is at the top or bottom of the screen. What happens when the scroll moves the cursor off the top by half a line? How do we render the text below the bottom of the screen in this half on half off state because there is an extra line than the viewport size to render, we don't have anything to draw there.

In general, my goal with neovide is not to implement features that don't exist in normal vim, but rather to augment the features in vim with animations that make it easier to track what is going on. So scroll values which stay at fractions of a line, are less valuable to me than animating the text into new scroll positions based on the internal vim buffer position. I'm not sure if this distinction makes sense?

I'm not against making the scrolling more natural though! I'm totally up for more discussion. This is just my first stab at a more humane scrolling experience :)

szbergeron:
Good plan, this definitely improves things on that front at least :)

Issue #436, opened 2021-01-05
Floating point errors accumulate after multiple scroll operations with dpi scaling


duvallj:
Here is a short gif demonstrating the issue:
![2021-01-04-neovide-scrolling-smaller](https://user-images.githubusercontent.com/10897431/103595883-97ad2800-4eca-11eb-966e-e052b5153c37.gif)
Basically, it seems the previously-rendered text gets cut-and-pasted over and over again when scrolling, and is slightly off due to floating point errors, resulting in entire lines being off when the effect compounds.

This could be a Skia limitation, but is it be possible to work around it?

Kethku:
Woah. This is wild. Mind sharing what font you're using? feels like maybe the font calculations are wrong. I don't see how this could be floating errors that quickly.

Kethku:
But also I rewrote all of the rendering just recently, so who knows.

duvallj:
(Ah sorry for the late response) `guifont` isn't set, so it should be using the default for Windows, Consolas.

I mention floating point errors because while I was poking around the rendering code trying to fix #424 starting from <https://github.com/Kethku/neovide/commit/8580ca065007c3e63ada85d5b20c2b2af593220c>, I noticed that [pixel_width and pixel_height](https://github.com/Kethku/neovide/blob/8580ca065007c3e63ada85d5b20c2b2af593220c/src/renderer/rendered_window.rs#L55) calculated from the font have to be cast to integers, while [the actual scrolling code](https://github.com/Kethku/neovide/blob/8580ca065007c3e63ada85d5b20c2b2af593220c/src/renderer/rendered_window.rs#L478) used the raw floats. It's just a hypothesis but I think it's at least plausible.

Kethku:
I think https://github.com/Kethku/neovide/pull/438 might fix this issue, but I'm not sure because I can't really reproduce the problem on my machine.

Kethku:
It fixes a ton of other issues with regard to dpi scaling though so maybe it fixes this as well?

Issue #424, opened 2020-12-30
Blurry font, possibly because of scaling?


kirawi:
![Capture](https://user-images.githubusercontent.com/67773714/103329052-17e61000-4a29-11eb-8736-f3a836b31be6.PNG)
![Capture1](https://user-images.githubusercontent.com/67773714/103329057-1a486a00-4a29-11eb-893e-01eaa9f97045.PNG)

Neovide vs. Neovim-qt

I'm using Windows 10, with a custom scaling size of 125% (inside 'Advanced scaling settings'), and building neovide from `c63b054`.

Kethku:
If you temporarily set the scaling to 100% does it look better? I think I messed up dpi scaling on master with my most recent refactor

kirawi:
I'll check in a few hours, but on my main desktop that has no scaling, it looks absolutely fine.

Kethku:
Yeah looks like scaling is screwed up. I'm investigating

PR #438, opened 2021-01-06
Fix dpi scaling

Kethku:
This is an attempt at fixing the dpi scaling of the renderer.


Kethku:
@duvallj would you be up for trying this branch? It fixes dpi scaling on my machine

duvallj:
@Kethku sure! Just tried it now in fact, and can confirm it works on my machine as well to fix #424

Doesn't fix #436 yet, though..

Kethku:
Im gonna merge this because I think its an improvement, but I'm going to think some more about how to fix the other issue

PR #462, opened 2021-02-04
Add an icon to the SDL2 window

lf-:


lf-:
I've deleted the code crime breaking this on big endian. It should be good now.

Issue #388, opened 2020-10-20
'{' in insert mode act like in normal mode.


GGCristo:
When I try to write '{' it behave like '{' in normal mode, this mean that it leaves Insert mode and go one paragraphs backward.
Edit : Others keys like '\' doesn't work either.

nilsmartel:
I have the same problem.

Using a MacBook Pro 2019 with German keyboard Layout.

The `"` character works fine for me, but neither `{`, nor `}` can be entered into text

nilsmartel:
just checked, the combination option/alt ⌥ + any other key seems to fail, no characters get inserted, the editor jumps into normal mode

keithsim-msft:
What keyboard layout do you folks use?

Kethku:
what keyboard layout do you folks use? This feels like a layout issue

Kethku:
Currently we have infrastructure for specifying more keyboard layouts here: https://github.com/Kethku/neovide/blob/master/src/bridge/layouts/sdl2.rs but today we only have qwerty mostly because I haven't spent the time to add more

The hope is that we can add common layouts and then let users configure them via a global vim variable. Its not the nicest solution, but as far as I can tell its the best we've got with Rust windowing systems.

GGCristo:
As you would have expected, I use the Spanish layout.

nilsmartel:
Thanks for the quick answer, it's German Mac Layout for me

nilsmartel:
Only thing I could figure out so far is, that characters like `{` don't even get propagated to that handler function

nilsmartel:
to be continued

nilsmartel:
I did some science and figured out that English keyboard layouts (which are heaven to work with for programming tasks, you wouldn't believe just how bad German is at this actually) get's recognised like this:

```
keycode: LeftBracket
keytext: "{"
``` 

where in the German layout the key get's recognised as

```
keycode: Num8
keytext: "{"
```

Now, here's why this is fantastic news:

I think the most straightforward thing to do is not creating individual handlers for all layouts, but a simple function `fn (KeyCode, KeyText) -> (KeyCode, KeyText)` that can be used to unify across all keyboards at this point.

a simple implementation might be
```rust
fn unify_layouts(code: KeyCode, text: &str) -> (KeyCode, &str) {
 match (code, text) {
   (_, "{") => (LeftBracket, "{")
  ...
   (code, text) => (code, text)
 }
}
```

nilsmartel:
[...]

PR #463, opened 2021-02-04
Add vim setting for ISO layout compatibility (sdl2)

ref209:
As discussed in #388, I have created a new vim setting to differentiate between ANSI and ISO layout to handle the Right Alt/Alt Gr key properly. I added an optional setting `neovide_iso_layout` that can be set to true when necessary.

At the moment, the fix only applies to sdl2. Winit doesn't differentiate between left and right Alt and I don't know how to fix this problem without heavily modifying how the input is handled. 

Because this is potentially affecting many users, I opened a PR with only the sdl2 fix. That way, at least, will be solved for some people. 

Let me know if you have any suggestion.

### DIFF
--- a/src/window/sdl2/layouts/mod.rs
+++ b/src/window/sdl2/layouts/mod.rs
@@ -1,16 +1,17 @@
 mod qwerty;
 
-use crate::window::keyboard::Modifiers;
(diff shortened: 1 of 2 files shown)


Kethku:
@ref209 would you be willing to make a documentation change in the wiki to fix document what this setting does?

PR #464, opened 2021-02-05
Disable the default-on SDL hint to disable compositing on X11

### LINKED ISSUE #370: Firefox incompatible
I'm using Firefox 80.0.1, neovim 0.5.0 and Linux 5.4.60 (Manjaro). When I use neovide, Firefox gui hangs, especially with github pages - don't know why. No problem with neovim from terminal and neovim-qt.

lf-:
SDL tells the window manager to disable compositing for its windows by
default in order to somewhat improve rendering performance.  This is
unfortunate for a text editor one would be using concurrently with other
programs, because KDE misbehaves with its global taskbar if this is
done, and it also may freeze or otherwise break the display of running
Firefox windows. This misbehaviour affects all SDL windows including
e.g. games such as Factorio.

There is a global switch to ignore all applications making this request
in the KDE compositor settings called "Allow applications to block
compositing", as well as an option in the window rules dialog to do it
for individual windows by class, but we should try to do the right thing
by default. The user can decide to block compositing per-window with
these rules if they wish to preserve the previous behaviour.

Fixes #370.

### DIFF
--- a/src/window/sdl2/mod.rs
+++ b/src/window/sdl2/mod.rs
@@ -380,6 +380,17 @@ impl Sdl2WindowWrapper {
     }
 }
 
+fn allow_compositing() {
+    // This fixes any vestiges of https://github.com/Kethku/neovide/issues/370
+    // which is an issue where KDE and perhaps others misbehave when compositing is forced off for
+    // a window, as is done by default with SDL2. Since this just sets a hint that will be ignored
+    // if unsupported, it is OK to leave in for all platforms.
+    //
+    // This is `sdl2_sys::SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR`
+    let name = "SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR";
+    sdl2::hint::set(name, "0");
+}
+
 pub fn start_loop(
     window_command_receiver: Receiver<WindowCommand>,
     ui_command_sender: TxUnbounded<UiCommand>,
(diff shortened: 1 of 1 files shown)


j4qfrost:
Can you make this a feature that the user can switch on and off and add it for Linux targets? Also I'm wondering if this should be pushed upstream seeing as how this would probably hit other applications using rust SDL as well.

lf-:
@j4qfrost It is, in my view, a per application question as to whether it should be enabled. Poking around GitHub code search, pretty much nobody except games seems to use this X11 hint to disable compositing. The specific group of programs whose usability will be substantially affected by disabling compositing is programs where the user will regularly use them while also using other stuff on the computer, which means if it were made default, that would mess up a few games where the performance outweighs the potential usability issue. This effectively restricts where it is desirable to default to regular compositing behaviour to mostly desktop apps like neovide that happen to use SDL for handling windowing.

I can make it configurable, but such a configuration setting would only be useful to people who it makes a substantial performance difference to, whose compositors honour the hint (since the behaviour after this PR is the default behaviour for X11 windows and if a compositor was ignoring the hint, this PR is a no-op) and the WM doesn't have built in configuration. I believe that the latter two requirements effectively restrict it to mutter, xfwm and picom. KDE/KWin users can turn off compositing per-window if they wish using the rules system.

I think one thing that perhaps the upstream `rust-sdl2` should do is reexport the constant or add a function to set the hint so we aren't grabbing a stringly typed thing out of sdl2-sys and putting it in our source.

j4qfrost:
Gotcha, well there's a detailed explanation for posterity, so if anyone ever has a reason to change things it'll be easy. I'll merge it.

PR #471, opened 2021-02-11
Update README.md

nilsmartel:


Kethku:
Makes good sense. thanks!

PR #482, opened 2021-03-01
Fix 100% cpu usage when multigrid is enabled

jonvaldes:
Found the culprit in the multigrid-related high CPU usage issue. Just a small logic oversight


Kethku:
Beauty. I actually committed a similar change to the develop branch.

Side note, I am working on trying to get back to winit as I think its the solution for https://github.com/Kethku/neovide/issues/445 and I have a develop branch now which uses an opengl backend taken from the skia repo rather than skulpin because I think it may have better driver support than vulkan for weird niche operating systems that neovim users seem to favor.

Kethku:
This change was integrated in a different PR. I appreciate you taking a look

PR #498, opened 2021-03-18
start with `set termguicolors`

### LINKED ISSUE #287: Passing `set termguicolors` command to nvim
Hi! Thanks for the application. It looks promising. I found a little annoyance that took some time to investigate. I used `neovim-qt` and everything was fine. When I switched to `neovide`, everything worked smoothly except the colors were weird. After digging the issue, I realized that I have to add `set termguicolors` to `init.vim` to make colors nice. But why `neovim-qt` does not need it? `Neovim-qt` passes ` --cmd set termguicolors` in command line when running `nvim`.

Could you make `neovide` to pass the option, so one shouldn't update neovim config to make a colorscheme look the same as in neovim-qt?

p00f:
Closes #287 

https://user-images.githubusercontent.com/36493671/111667971-3e9f3000-883b-11eb-8e59-b7fa3d9f0643.mp4

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -112,13 +112,7 @@ pub fn build_neovide_command(channel: u64, num_args: u64, command: &str, event: 
 
 pub fn create_nvim_command() -> Command {
     let mut cmd = build_nvim_cmd();
-
-    cmd.arg("--embed")
-        .args(SETTINGS.neovim_arguments.iter().skip(1));
-    cmd.arg("--cmd")
-        .args(SETTINGS.neovim_arguments.iter().skip(1));
-    cmd.arg("set termguicolors")
-        .args(SETTINGS.neovim_arguments.iter().skip(1));
+    cmd.args(&["--embed", "--cmd", "set termguicolors"]).args(SETTINGS.neovim_arguments.iter().skip(1));
 
 
 
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -115,6 +115,12 @@ pub fn create_nvim_command() -> Command {
 
     cmd.arg("--embed")
         .args(SETTINGS.neovim_arguments.iter().skip(1));
+    cmd.arg("--cmd")
+        .args(SETTINGS.neovim_arguments.iter().skip(1));
+    cmd.arg("set termguicolors")
+        .args(SETTINGS.neovim_arguments.iter().skip(1));
+
+
 
     #[cfg(not(debug_assertions))]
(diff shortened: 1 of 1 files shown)


p00f:
it doesn't work when you add an argument `./target/release/neovide Cargo.toml` - it opens `Cargo.toml` and a new file named `set termguicolors`

p00f:
re https://github.com/Kethku/neovide/issues/287#issuecomment-626416687, nvim-qt doesn't let you pass --cmd arguments

p00f:
I AM SO SORRY I should've looked at the tokio docs first.  Arguments work properly now
(filename and additional arguments, like `./target/release/neovide Cargo.toml --cmd set\ exapndtab`)

p00f:
https://user-images.githubusercontent.com/36493671/111739901-ce7bc300-88a9-11eb-8547-420b36835612.mp4

Kethku:
I wonder if the correct way to solve this would be to call the set variable function rather than passing an argument to the neovim executable? We do something similar here: https://github.com/Kethku/neovide/blob/main/src/bridge/mod.rs#L241

p00f:
that is fine too, as long as it happens before loading pulgins/config files. Themes etc need it defined before load

p00f:
I'm not sure if calling the set option function is done before loading config, because neovim would have to be already started before that

p00f:
> I'm not sure if calling the set option function is done before loading config, because neovim would have to be already started before that, right?

nvm it works, i get "yay" with this `init.lua`:
```lua
local api = vim.api

if vim.o.termguicolors then
    api.nvim_err_writeln("yay")
else
    api.nvim_err_writeln("nay")
end
--api.nvim_command("packadd packer.nvim")
--vim.o.termguicolors = true
--vim.g["conjure#mapping#doc_word"] = "K"
--
--local packer = require("packer")
--local packages = require("packages")
--
--packer.startup(function()
--    for _, value in pairs(packages) do
--        packer.use(value)
--    end
--end)
--packer.use_rocks({ "lunajson", "http" })
--vim.cmd([[
--    command! -complete=file -nargs=* DebugC lua require "dap_setup".start_c_debugger({<f-args>}, "gdb")
--]])
--vim.cmd([[
--    command! -complete=file -nargs=* DebugRust lua require "dap_setup".start_c_debugger({<f-args>}, "gdb", "rust-gdb")
--]])
```

p00f:
i'll create another pr.

p00f:
#500

Kethku:
Closing in favor of https://github.com/Kethku/neovide/pull/500

PR #500, opened 2021-03-19
Set `termguicolors` to true

### LINKED ISSUE #287: Passing `set termguicolors` command to nvim

p00f:
Closes #287, #498

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -241,6 +241,9 @@ async fn start_neovim_runtime(
     nvim.set_option("lazyredraw", Value::Boolean(false))
         .await
         .ok();
+    nvim.set_option("termguicolors", Value::Boolean(true))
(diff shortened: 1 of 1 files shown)


Kethku:
Thanks a bunch for the contribution! Looks good to me

PR #517, opened 2021-04-07
Fix clippy, fix font load test

p00f:
I've removed unused imports, prefixed an unused variable with an underscore and didn't do anything to an unused `pub` function. Also fixed
```
warning: panic message is not a string literal
 --> src/error_handling.rs:5:12
  |
5 |     panic!(explanation.to_string());
  |            ^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(non_fmt_panic)]` on by default
  = note: this is no longer accepted in Rust 2021
help: add a "{}" format string to Display the message
  |
5 |     panic!("{}", explanation.to_string());
```
After this, there's only the one warning about the unused function ( `build_neovide_command` in `src/bridge/mod.rs` )

### DIFF
--- a/src/bridge/handler.rs
+++ b/src/bridge/handler.rs
@@ -44,7 +44,7 @@ impl Handler for NeovimHandler {
     ) {
         trace!("Neovim notification: {:?}", &event_name);
 
-        let _ui_command_sender = self.ui_command_sender.clone();
+        let ui_command_sender = self.ui_command_sender.clone();
         let redraw_event_sender = self.redraw_event_sender.clone();
         task::spawn_blocking(move || match event_name.as_ref() {
             "redraw" => {
--- a/src/bridge/handler.rs
+++ b/src/bridge/handler.rs
@@ -44,7 +44,9 @@ impl Handler for NeovimHandler {
     ) {
         trace!("Neovim notification: {:?}", &event_name);
 
+        #[cfg(windows)]
         let ui_command_sender = self.ui_command_sender.clone();
+
         let redraw_event_sender = self.redraw_event_sender.clone();
         task::spawn_blocking(move || match event_name.as_ref() {
             "redraw" => {
--- a/src/bridge/handler.rs
+++ b/src/bridge/handler.rs
@@ -44,7 +44,7 @@ impl Handler for NeovimHandler {
     ) {
         trace!("Neovim notification: {:?}", &event_name);
 
-        let ui_command_sender = self.ui_command_sender.clone();
+        let _ui_command_sender = self.ui_command_sender.clone();
         let redraw_event_sender = self.redraw_event_sender.clone();
         task::spawn_blocking(move || match event_name.as_ref() {
             "redraw" => {
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -91,7 +91,7 @@ fn build_nvim_cmd() -> Command {
     }
 }
 
-pub fn build_neovide_command(channel: u64, num_args: u64, command: &str, event: &str) -> String {
+pub fn _build_neovide_command(channel: u64, num_args: u64, command: &str, event: &str) -> String {
     let nargs: String = if num_args > 1 {
         "+".to_string()
     } else {
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -91,6 +91,7 @@ fn build_nvim_cmd() -> Command {
     }
 }
 
+#[cfg(windows)]
 pub fn build_neovide_command(channel: u64, num_args: u64, command: &str, event: &str) -> String {
     let nargs: String = if num_args > 1 {
         "+".to_string()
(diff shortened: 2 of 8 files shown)


p00f:
I ignored the `test_load` test in `src/renderer/fonts/font_loader.rs` because it fails when I don't have DejaVu Serif on linux. Also, I had to prepend an underscore to `build_neovide_command` in `src/bridge/mod.rs` to make clippy pass.

All the pre-commit hooks work now, for me

p00f:
I don't know if testing for DejaVu is a good assumption, only 20 packages listed on https://archlinux.org/packages/extra/any/ttf-dejavu/ require it explicitly, out of which 9 are testing dependencies and _only 11 need it at runtime_

p00f:
I'm learning rust, so apologies if it is overreaching :sweat_smile:

j4qfrost:
It would probably be a good idea to find a better default don't to test against rather than remove the test altogether.

p00f:
I don't know if one exists

src/bridge/handler.rs
```diff
@@ -44,7 +44,7 @@ impl Handler for NeovimHandler {
     ) {
         trace!("Neovim notification: {:?}", &event_name);
 
-        let ui_command_sender = self.ui_command_sender.clone();
+        let _ui_command_sender = self.ui_command_sender.clone();
```
Kethku, line 47: if this isn't used, the variable should just be removed I think

src/bridge/mod.rs
```diff
@@ -91,7 +91,7 @@ fn build_nvim_cmd() -> Command {
     }
 }
 
-pub fn build_neovide_command(channel: u64, num_args: u64, command: &str, event: &str) -> String {
+pub fn _build_neovide_command(channel: u64, num_args: u64, command: &str, event: &str) -> String {
```
Kethku, line 94: Similarly here. I would double check though, its possible that this just isn't used for this platform but is used on a different platform. In which case you'd get the warning on some builds but not others. Not sure how to deal with that issue

src/error_handling.rs
```diff
@@ -2,7 +2,7 @@ use log::error;
 
 fn show_error(explanation: &str) -> ! {
     error!("{}", explanation);
-    panic!(explanation.to_string());
+    panic!("{}", explanation.to_string());
```
Kethku, line 5: This feels dumb. Do you know of a way to disable a given clippy suggestion? Or otherwise can you give some details as to why this was suggested? It doesn't make sense to me.

src/renderer/fonts/font_loader.rs
```diff
@@ -230,6 +230,7 @@ mod test {
     }
 
     #[test]
+    #[ignore]
```
Kethku, line 233: I agree with @j4qfrost I think this should either be left enabled or swapped to use a different font rather than deleting the test? What do you think

Kethku:
I reviewed the PR and generally think it looks good. I added a couple of thoughts. Thanks so much for the contribution!

src/renderer/fonts/font_loader.rs
j4qfrost: I looked it up, Monospace is installed for all Linux systems.

src/error_handling.rs
p00f:
```
warning: panic message is not a string literal
 --> src/error_handling.rs:5:12
  |
5 |     panic!(explanation.to_string());
  |            ^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(non_fmt_panic)]` on by default
  = note: this is no longer accepted in Rust 2021
help: add a "{}" format string to Display the message
  |
5 |     panic!("{}", explanation.to_string());
  |            ^^^^^
help: or use std::panic::panic_any instead
  |
5 |     std::panic::panic_any(explanation.to_string());
```
I didn't want to use `panic_any` ([here](https://doc.rust-lang.org/std/panic/fn.panic_any.html)) becaus this would be surely be string

src/bridge/mod.rs
p00f: Yep its usage was guarded under `#[cfg(windows)]` so I'll change it back

p00f: Guess I should not rely on lsp references

src/renderer/fonts/font_loader.rs
p00f: Cool, I'll change it to monospace

p00f: Why were you checking for DejaVu Serif though? Shouldn't you have been checking for DejaVu Mono?

src/bridge/handler.rs
p00f: Its usage was also guarded under `#[cfg(windows)]`

p00f:
@Kethku can you take a look?

src/error_handling.rs
Kethku: Makes good sense!

p00f:
I've not yet understood when clippy marks things as warnings/errors

p00f:
Clean now!

PR #541, opened 2021-04-12
Don't call `.vulkan()` when building the SDL window on macOS

Benjamin-Davies:
This appears to fix some (not all) of the problems that are mentioned in #466


Kethku:
This looks good. I also tested on windows and it appears that this line isn't necessary on any platform. I looked into it further and it seems like this implements vulkan support in sdl2, but we use the raw window handle to do this, so it shouldn't be necessary. Thanks for the PR

PR #542, opened 2021-04-12
Fix MacOS sdl2 bug

### LINKED ISSUE #538: Doesn't work in MacOS Big Sur 11.2.3
截屏2021-04-12 上午5 42 33
截屏2021-04-12 上午5 42 45

Miloas:
fix #538 

it seems a upstream issue https://github.com/Rust-SDL2/rust-sdl2/pull/1030


Kethku:
looks great to me. Thanks for looking into this!

PR #557, opened 2021-04-13
Remote TCP

### LINKED ISSUE #50: Enable Remote Neovim Process Connection

Benjamin-Davies:
Allows the user to connect to a remote instance of Neovim over TCP using the `--remote-tcp` flag.

Because rust does not have full support for polymorphism, I created the enum type `TxWrapper` to allow either `ChildStdin` or `WriteHalf<TcpStream>` to be used for the connection stream, while avoiding a complicated mess of type parameters. To do this I had to turn some of our implicit dependencies (ie. dependencies of dependencies) into explicit dependencies, but there are no new dependencies. I also adapted some of the functions from `nvim_rs::create::tokio` to work with this new general type.

I have also tried to make it clear in the readme that, without SSH port forwarding or similar, TCP sockets are not secure.

I think that this fixes #50

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1577,6 +1577,7 @@ dependencies = [
  "euclid",
  "flexi_logger",
  "font-kit",
+ "futures 0.3.12",
  "image",
  "lazy_static",
  "log",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1585,6 +1585,7 @@ dependencies = [
  "neovide-derive",
  "nvim-rs",
  "parking_lot 0.10.2",
+ "pin-project",
  "rand",
  "rmpv",
  "rust-embed",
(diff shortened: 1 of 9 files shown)


Kethku:
I did a pass and it looks good by inspection. I will try to test it out tonight. Thank you so much for the great contributions

Kethku:
Works beautifully. Awesome addition, thanks so much

jmlucjav:
Hi,

I was trying to set this up in Win10 and could not connect to nvim. This was the first time I was trying neovide so it might by my bad, so just trying to rule out issues....is this supposed to run on win, or maybe hasn't been tested?

As a side note, I had to use 0.6.0, 0.7.0 wouldn't even start for me in a brand new (although heavily customized) win10...

thanks!

thanks!

jmlucjav:
duh, I should have noticed...

rashil2000:
@Benjamin-Davies Is there a way to close the Neovide window, without terminating the remote server too? Currently, both `:q` and clicking the X button close the server. Ideally I'd like to only close the Neovide client.

PR #573, opened 2021-04-19
Fix warnings

Benjamin-Davies:
Fix some small problems that appeared as warnings in the `cargo build` output, and were causing the pre-commit git hook (which I had disabled locally) to fail. These were mostly just unused code and improper attribute order.

Of note is `Sdl2WindowWrapper.transparency`, which was set whenever `WindowSettings.transparency` is out of sync with the window's actual opacity, but was never read from so I removed it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,6 +14,10 @@ jobs:
       run: |
         cargo fmt --all -- --check
 
+    - name: Lint with Clippy
+      run: |
+        cargo clippy --all -- -D warnings
+
     - name: Install Neovim
       run: |
         choco install -y neovim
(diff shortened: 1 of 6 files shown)


j4qfrost:
Strange, clippy should've caught these.

Kethku:
@j4qfrost I disabled the git hook because it was confusing people and making development hard. So at this point they tend to sneak in.

j4qfrost:
@Kethku gotcha, yeah normally I just run commits with --no-verify when I'm rapid-firing. Maybe there should be a clippy check for merges. It would reduce the number of formatting PRs.

j4qfrost (approved):

PR #584, opened 2021-04-23
Bump skia-safe to 0.39.1

shaunsingh:
Compiling for apple silicon/m1 only works on 0.39.1


haringsrob:
I tested this `cargo build --release --target aarch64-apple-darwin` and could build successfully with this change.

Kethku:
Beauty. Thanks for the upgrade

PR #586, opened 2021-04-24
Remove RAlt as a modifier in non-iso layouts.

vinicius-toshiyuki:
In ABNT2 brazilian portuguese keyboard layout RAlt (or AltGr in ABNT2)
does not work the same as LAlt and should not be used as a modifier in
Neovim. It is used to enable typing special characters in combination
with Shift, such as ?`´~^ºª etc, either as the only way for typing them
or as a faster way combining with other keys.
Using Neovim in xterm, kitty, gnome-terminal etc has the expected
behavior, as using in other places in the system as well (e.g. browsers)

### DIFF
--- a/src/window/sdl2/layouts/mod.rs
+++ b/src/window/sdl2/layouts/mod.rs
@@ -7,11 +7,10 @@ pub use qwerty::handle_qwerty_layout;
 
 impl From<Mod> for Modifiers {
     fn from(mods: Mod) -> Modifiers {
-        let iso_layout = SETTINGS.get::<WindowSettings>().iso_layout;
         Modifiers {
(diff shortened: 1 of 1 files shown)


Kethku:
I'm hesitant to merge this. Maybe a better strategy would be to rename the iso setting option to something like ignore right alt?

As a side note, this will be fixed once upstream changes in winit have completed, but they may be a ways out yet

vinicius-toshiyuki:
To be fair, I didn't expect it to be merged, the changes I made work for my layout but I don't know about other layouts, so maybe leaving RAlt as it is makes more sense. I just wanted to let you know about this "issue" more than anything and also pushed my changes so someone with the same problem as me can solve it if needed.
If that's going to be fixed anyway and I have a working version I guess there's no need to rush, I'll just wait

PR #590, opened 2021-04-25
Version and help without opening the program

ProgramingIsTheFuture:
Now we can see the version and use the help command without opening the neovide GUI. 
If i need to confirm my neovide version it will open the neovide and i dont want it. I just want to see the version.


Kethku:
I think this makes sense. Thanks for the change

PR #597, opened 2021-04-26
fix: print --version, --help when requested and exit immediately

srevinsaju:


Kethku:
I merged a separate version of this. Thanks though!

PR #602, opened 2021-04-28
Fix typo in Arch Linux AUR installation.

not-bailey:
Git clones "neovide-git" folder from the AUR, but cd changes the current directory to "neovide" instead of "neovide-git".


Lord-Valen (approved):

Kethku:
Given the consensus that this is needed, I'm going to approve. Thanks for the fix! Was this referenced in any issues?

PR #611, opened 2021-04-30
Fix build

lf-:
The build was broken by skulpin shifting around its feature flags as
part of the rewrite for 0.12. I have set it to the last pre 0.12
version from the registry (independently confirmed there were no
unreleased changes in the skulpin git repo after that and before the
rewrite).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,16 +40,36 @@ dependencies = [
 
 [[package]]
 name = "andrew"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


lf-:
Actually, I have /absolutely/ no idea why cargo decided to ignore the perfectly good version in the lockfile. Maybe something to do with `cargo install` being weird?

Kethku:
Thanks for the catch. I'm working on moving away from skulpin in the opengl branch, but its blocked on a segmentation fault I can't quite track down on linux. So we are still using it for now.

PR #621, opened 2021-05-03
Add distance dependent animation length

JeWe37:
One of the primary reasons I wanted to try out Neovide is its cursor animations that are supposed to improve cursor tracking. I have found that to be true for me and it helps me locate it quite a bit, but I wasn't quite happy with how it felt.

The issue was that the animation length was always the same no matter if you were just typing along and therefore only moving a relatively short distance with each keystroke or if you were going all the way from the top of the view to the bottom. In one case typing just felt sluggish, in the other I could barely see the trail anymore.

To fix this I created this simple pull request that really only modifies the cursor renderer slightly. It adds an option `g:neovide_cursor_distance_length_adjust` that lets the user choose to have their animation times increased logarithmically as travel distance increases.


src/renderer/cursor_renderer/mod.rs
```diff
@@ -86,6 +90,11 @@ impl Corner {
             self.t = 0.0;
             self.start_position = self.current_position;
             self.previous_destination = destination;
+            self.length_multiplier = if settings.distance_length_adjust {
+                (destination - self.current_position).length().log10()
```
Kethku, line 94: Interesting. Can you explain the log term?

JeWe37:
That's solely so that the multiplier doesn't get too large. If I had just let it scale linearly for instance with my setting for the animation time(0.035) it would have taken about 0.07 for just regular typing but 6.3s(!) to go from top to bottom. log seemed like a reasonable choice, though other functions might work as well(roots etc.).

src/renderer/cursor_renderer/mod.rs
JeWe37: That's solely so that the multiplier doesn't get too large. If I had just let it scale linearly for instance with my setting for the animation time(0.035) it would have taken about 0.07 for just regular typing but 6.3s(!) to go from top to bottom. log seemed like a reasonable choice, though other functions might work as well(roots etc.).

Kethku: Makes good sense to me. I will try this out on my local machine and get it merged tonight. Great idea!

JeWe37:
While I was at it... I also fixed the cursor randomly jumping to the command line when switching modes. This does not have an option as I would consider the current behavior a bug.

I did however also add an option to disable trails when actually explicitly going to the command line. I personally find it useful to keep it on, but both behaviors seem like something someone could want.

This also fixes #117 though there may be other cursor options that one might want to add in the future that are not explicitly stated there.

Note that I consider my fix for the jumping of the cursor when switching modes to be a hack. I do not know if you want to merge that. It does work and I know I will use it, however I cannot guarantee it won't cause problems with some plugins or the like/become a burden down the road.

Kethku:
I actually think the fix for the cursor switching modes makes sense to me. I'm going to merge it as is, but if you have any ideas beyond this for refactorings or w/e I'm all ears.

PR #631, opened 2021-05-07
Version check

j4qfrost:
This includes some clippy stuff and the eglCurrentDisplay patch as well.

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -22,6 +22,7 @@ use crate::settings::*;
 use crate::window::window_geometry_or_default;
 pub use events::*;
 use handler::NeovimHandler;
+use regex::Regex;
 pub use tx_wrapper::{TxWrapper, WrapTx};
 pub use ui_commands::UiCommand;
 
@@ -182,13 +183,14 @@ async fn start_neovim_runtime(
         close_watcher_running.store(false, Ordering::Relaxed);
     });
 
-    if let Ok(Value::Integer(correct_version)) = nvim.eval("has(\"nvim-0.4\")").await {
-        if correct_version.as_i64() != Some(1) {
-            error!("Neovide requires version 0.4 or higher");
+    if let Ok(output) = nvim.command_output("version").await {
+        let re = Regex::new(r"NVIM v0.[4-9]\d*.\d+").unwrap();
+        if !re.is_match(&output) {
+            error!("Neovide requires nvim version 0.4 or higher. Download the latest version here https://github.com/neovim/neovim/wiki/Installing-Neovim");
             std::process::exit(0);
         }
     } else {
-        error!("Neovide requires version 0.4 or higher");
+        error!("Neovide requires nvim version 0.4 or higher. Download the latest version here https://github.com/neovim/neovim/wiki/Installing-Neovim");
         std::process::exit(0);
     };
 
--- a/src/renderer/cursor_renderer/mod.rs
+++ b/src/renderer/cursor_renderer/mod.rs
@@ -21,7 +21,7 @@ const DEFAULT_CELL_PERCENTAGE: f32 = 1.0 / 8.0;
 const STANDARD_CORNERS: &[(f32, f32); 4] = &[(-0.5, -0.5), (0.5, -0.5), (0.5, 0.5), (-0.5, 0.5)];
 
 #[derive(SettingGroup)]
-#[setting_prefix = "keyboard"]
+#[setting_prefix = "cursor"]
 #[derive(Clone)]
 pub struct CursorSettings {
     antialiasing: bool,
--- a/src/renderer/cursor_renderer/mod.rs
+++ b/src/renderer/cursor_renderer/mod.rs
@@ -20,8 +20,9 @@ const DEFAULT_CELL_PERCENTAGE: f32 = 1.0 / 8.0;
 
 const STANDARD_CORNERS: &[(f32, f32); 4] = &[(-0.5, -0.5), (0.5, -0.5), (0.5, 0.5), (-0.5, 0.5)];
 
-#[setting_prefix = "cursor"]
-#[derive(Clone, SettingGroup)]
+#[derive(SettingGroup)]
+#[setting_prefix = "keyboard"]
+#[derive(Clone)]
 pub struct CursorSettings {
     antialiasing: bool,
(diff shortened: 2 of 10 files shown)


src/renderer/fonts/caching_shaper.rs
```diff
@@ -15,7 +15,7 @@ cfg_if! {
     if #[cfg(target_os = "windows")] {
         pub const SYSTEM_DEFAULT_FONT: &str = "Consolas";
     } else if #[cfg(target_os = "linux")] {
-        pub const SYSTEM_DEFAULT_FONT: &str = "Ubuntu";
+        pub const SYSTEM_DEFAULT_FONT: &str = "Mono";
```
j4qfrost, line 18: I didn't mean for these changes to make it in. I was working on something else.

src/bridge/mod.rs
```diff
@@ -182,13 +183,14 @@ async fn start_neovim_runtime(
         close_watcher_running.store(false, Ordering::Relaxed);
     });
 
-    if let Ok(Value::Integer(correct_version)) = nvim.eval("has(\"nvim-0.4\")").await {
-        if correct_version.as_i64() != Some(1) {
-            error!("Neovide requires version 0.4 or higher");
+    if let Ok(output) = nvim.command_output("version").await {
```
Kethku, line 186: I'm slightly suspicious of this change. neovim defines options for all of the major versions, so any version greater than 0.4 should cause "has("nvim-0.4")" to return true.

Is this a fix for something I missed? If not, I think I should track down the documentation that indicates this behavior and add it as a comment.

src/bridge/mod.rs
```diff
@@ -182,13 +183,14 @@ async fn start_neovim_runtime(
...
+        let re = Regex::new(r"NVIM v0.[4-9]\d*.\d+").unwrap();
+        if !re.is_match(&output) {
+            error!("Neovide requires nvim version 0.4 or higher. Download the latest version here https://github.com/neovim/neovim/wiki/Installing-Neovim");
             std::process::exit(0);
         }
     } else {
-        error!("Neovide requires version 0.4 or higher");
+        error!("Neovide requires nvim version 0.4 or higher. Download the latest version here https://github.com/neovim/neovim/wiki/Installing-Neovim");
```
Kethku, line 193: Good call giving further instructions

src/renderer/cursor_renderer/mod.rs
```diff
@@ -20,8 +20,9 @@ const DEFAULT_CELL_PERCENTAGE: f32 = 1.0 / 8.0;
 
 const STANDARD_CORNERS: &[(f32, f32); 4] = &[(-0.5, -0.5), (0.5, -0.5), (0.5, 0.5), (-0.5, 0.5)];
 
-#[setting_prefix = "cursor"]
-#[derive(Clone, SettingGroup)]
+#[derive(SettingGroup)]
+#[setting_prefix = "keyboard"]
```
Kethku, line 24: correct me if I'm wrong, but I believe this represents a breaking change

src/renderer/mod.rs
```diff
@@ -18,8 +18,9 @@ use crate::editor::{Colors, DrawCommand, Style, WindowDrawCommand};
 use crate::settings::*;
 use cursor_renderer::CursorRenderer;
 
-#[setting_prefix = "window"]
-#[derive(Clone, SettingGroup)]
+#[derive(SettingGroup)]
+#[setting_prefix = "keyboard"]
```
Kethku, line 22: This seems definitely incorrect. Maybe a merge error?

src/renderer/cursor_renderer/mod.rs
j4qfrost: It behaves the same when I run it. Clippy was complaining about it and saying that the previous line would be counted as an error in the future.

src/window/keyboard/settings.rs
```diff
@@ -1,8 +1,9 @@
 use super::KeyboardLayout;
 use crate::settings::FromValue;
 
+#[derive(SettingGroup)]
```
Kethku, line 4: Interesting to split these up. Was that a clippy suggestion?

src/renderer/mod.rs
j4qfrost: Yes. I think so lol.

Kethku:
Good stuff. I appreciate the maintenance work

src/renderer/mod.rs
j4qfrost: I think I was just brainlessly copy and pasting.

j4qfrost:
Thanks, I'll make the fixes and push.

src/window/keyboard/settings.rs
j4qfrost: It was more about the order of the derivation. It wanted SettingsGroup to be resolved before the prefix was used.

src/bridge/mod.rs
j4qfrost: While trying to build a snap, I ran into an issue when using the latest neovim version from git. The repo was marked with version 0.5.0, which wasn't caught by has.

src/renderer/cursor_renderer/mod.rs
Kethku: to be specific, setting_prefix is changed from cursor to keyboard here. Thats what I was pointing out.

src/renderer/mod.rs
Kethku: Ok cool. Just making sure to capture, window was changed to keyboard here.

src/window/window_wrapper/renderer.rs
```diff
@@ -51,6 +51,9 @@ impl SkiaRenderer {
         gl::load_with(|s| windowed_context.get_proc_address(&s));
 
         let interface = skia_safe::gpu::gl::Interface::new_load_with(|name| {
+            if name == "eglGetCurrentDisplay" {
```
j4qfrost, line 54: I think the PR should be ready, though I this fix should be studied a bit more.

j4qfrost:
@Kethku I'll continue my research with eglGetCurrentDisplay. I may need to make an upstream commit. The snap should be ready to merge too. I would need a review for the new github workflow though. If you would like to have admin access to the snapcraft package, just send me your email.

PR #652, opened 2021-05-14
Update Cargo Resolver Version

ck3d:
Update cargo resolver to version 2 to avoid unwanted feature activation.

The package sdl2-sys has features bundled and static-link for target
linux activated, since the cargo resolver version 1 aggregates all
possible features. The newer version avoids this behaviour.

See also:
https://doc.rust-lang.org/cargo/reference/resolver.html#feature-resolver-version-2


Kethku:
This has been done in the opengl branch, but I will merge this for those still on main

ck3d:
Thanks a lot!

PR #654, opened 2021-05-15
Fix linux default font

p00f:
@Kethku @j4qfrost 

Are there assumptions about the existence of a specific font on linux elsewhere? I'll fix them all before marking this as ready for review

### DIFF
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -15,7 +15,7 @@ cfg_if! {
     if #[cfg(target_os = "windows")] {
         pub const SYSTEM_DEFAULT_FONT: &str = "Consolas";
     } else if #[cfg(target_os = "linux")] {
-        pub const SYSTEM_DEFAULT_FONT: &str = "Ubuntu";
+        pub const SYSTEM_DEFAULT_FONT: &str = "monospace";
     } else if #[cfg(target_os = "macos")] {
         pub const SYSTEM_DEFAULT_FONT: &str = "Menlo";
     }


p00f:
Because it panics when I don't have the Ubuntu font installed (unwrapping `None`)

j4qfrost:
I found that "Mono" works.

p00f:
as in "Ubuntu Mono" ?

j4qfrost:
This issue has been going back and forth a lot. I think a way to alleviate this issue is by creating more package distributions like snap or nix. 

> as in "Ubuntu Mono" ?

No, just "Mono"

p00f:
"monospace" is reasonable assumption because it's not actually a font but whatever `fontconfig` has configured as your monospace font.

Kethku:
Weird. I would have expected that not to work... If that is a general purpose solution though, that would be ideal

j4qfrost:
Have you tested this behavior?

> "monospace" is reasonable assumption because it's not actually a font but whatever `fontconfig` has configured as your monospace font.

p00f:
Yeah, just change the monospace font in your fontconfig then check the font settings in firefox

Kethku:
https://github.com/Kethku/neovide/blob/42aee72351e1add0933455f82a608120cebd5c87/src/renderer/fonts/caching_shaper.rs#L70 the font is resolved by name via this skia function. You very well may be right that this works just fine, but skia may be doing something strange to find fonts by name.

j4qfrost:
I prefer to leave the solution as simple as possible.


> Yeah, just change the monospace font in your fontconfig then check the font settings in firefox

With these steps it would probably just be easier to just refer to the doc and install noto-fonts or something.

Kethku:
Or embed the noto fonts. Apparently it has full coverage of every possible glyph. To me the extra executable size seems worth it

Kethku:
And that way we don't have to depend on any preinstalled system fonts

p00f:
> With these steps it would probably just be easier to just refer to the doc and install noto-fonts or something.

You just have to edit 2-3 words and run `fc-cache`

p00f:
> Or embed the noto fonts. Apparently it has full coverage of every possible glyph. To me the extra executable size seems worth it

Didn't you do this earlier?

Kethku:
I did. Somebody complained that the file size was too high, but honestly I'm not sure I agree that the file size being larger is worth the cost of the brittle default config

j4qfrost:
Well if it brings the issue count down, then that's a win for me.

p00f:
That font _is_ huge though if you want all glyphs 
image

Kethku:
True. We could do a good middle ground by just embedding the monospace one. Then the fallback solution I outlined in the fallback issue could take it from there.

p00f:
If you are embedding something, can you embed Fira Code instead? It has the most programming glyphs and ligatures

Kethku:
done. I will work on that

PR #655, opened 2021-05-15
Opengl

Kethku:
Ok. Opengl branch seems to be the future. Remaining issues I know of:
1. *DONE* The cursor is frequently slightly off from the grid. I believe this stems from the font hinting work
2. *DONE and introduces actual font fallback solving https://github.com/Kethku/neovide/issues/444 once https://github.com/Kethku/neovide/pull/701 merges* Fallback is busted. I think we will have to swap to https://docs.rs/swash/0.1.2/swash/index.html rather than rustybuzz because there is a good example for hall font fallback might work in the docs (plus its way faster, and written in rust rather than ported)
3. There is a bug where dragging selections past the bottom of the window causes it to crash.
4. *DONE once https://github.com/Kethku/neovide/pull/701 merges* Font shaping sometimes uses different glyph widths than we expect. I think we need to manually place glyphs in a grid rather than relying on the font shaping.
5. Cursor freezes sometimes https://github.com/Kethku/neovide/issues/640

I think problem 1 can be ignored for merge. 3 isn't a huge deal so long as its fixed after merge. 2 however is a big deal because it means main will no longer have any font fallback. So moving forward I'm going to try to get swash working and then take a look at implementing the fallback algorithm outlined in the docs.

4 is a hard problem. To do it correctly while still supporting ligatures we will have to draw glyphs at expected grid positions except when they are merged by the font shaping. In that case I propose relying on the font shaping to position the glyph at the expected advance compared to the previous glyph and then continue on. This would not handle shaping of non ligature type substitutions, but honestly we don't support scripts like that well anyway today. I'm pretty ok punting on that for the time being.


Kethku:
I have merged swash support. Next up, I need to work on font fallback.

j4qfrost:
@Kethku for snap I can either give you a key, or give you admin authority to the snapcraft package.

Kethku:
Honestly I don't know anything about snap. I dont really know what its for. So I will defer to your judgement for which you think is a better choice.

j4qfrost:
Snap is just a popular package manager. For the CI to work you need to add SNAPCRAFT_TOKEN to the secrets. I can generate a token for you next week. I don't have access to my Ubuntu machine atm.

Kethku:
Problem 2 listed above is now fixed. I used swash to cluster and select fonts from the fallback list manually. Appears to work, but the positioning of double width glyphs is broken.

I'm becoming convinced that font selection and fallback should happen at the editor level. This moves font selection to the editor's responsibility rather than the renderer. Luckily there is no graphics context required in the caching shaper, so this should be an easy change even though this would be introducing a skia dependency to the editor.

Kethku:
I've now tested using skia's built in font selection given a character and it can properly select a font from the systems installed font list and load it dynamically. Getting very close

I've also added an issue to the list, I've noticed that occasionally the window freezes and prints warning about creating a stencil buffer. This must be fixed before merging.

Kethku:
I just merged font fallback into this branch. Now just the crash on drag remains. I have a hunch that the stencil buffer issue has been fixed, but if I encounter it again, I found a flag in skia configuration which avoids the use of stencil buffers entirely

Issue #622, opened 2021-05-04
Remember Last Window Dimensions.


PyGamer0:
Neovide should remember the last window dimension and open the new window with the same dimensions.
Could be a setting like `g:neovide_remember_dimensions`

Kethku:
I think this is a reasonable request. I would be happy to take a PR for this.

bbli:
I can try to take a stab at this. Any recommendations on what files to look at?

PyGamer0:
@bbli Pretty sure this is the place neovide get its window dimensions. https://github.com/Kethku/neovide/blob/main/src/window/mod.rs#L63

bbli:
@PyGamer0 thanks, that definitely pushed me in the right direction. @Kethku , I couldn't find in the codebase a "settings" file that is written to. If there is, can you point me to it, and if not, do you have any guidelines for where to write to in the user's filesystem?

PyGamer0:
> settings" file that is written to

 @bbli do you mean neovide settings? They are stored in user's init.vim.

> where to write to in the user's filesystem

On windows there is an environment variable TEMP. Don't know about other OS.

bbli:
> They are stored in user's init.vim.
Ohh no, I am not talking about setting the boolean g:neovide_remember_dimensions variable, but rather the window's width and height. Because my solution to having neovide remember the dimension is to write to a file upon exit, and on startup read from that file.

In any case, I probably will submit a pull request tomorrow, so perhaps looking over that code will clear things up, and we can take it from there.

Pangoraw:
I had taken a try at this issue and by default, I stored the file in the `stdpath("data")` path return by neovim, this is where I have seen other package write to. See [the start of my implementation](https://github.com/Pangoraw/neovide/commit/890c4022596762eae373a8e80f5344e850a0dac6#diff-1b0002da06aecc299358a022e976fac801bb3db88b77817e1a1aee77dbc09f3eR281-R285).

Kethku:
Oh very interesting. I didn't realize that was an standard place for this sort of data. Very cool!

I wonder if we should build a standard way for neovide to store data from run to run. Maybe use serde and serialize out a structure as json?

bbli:
> stored the file in the `stdpath("data")` path
Ohh ok, thanks for letting me know! It causes some complications that I talk about in my pull request #665 .

> if we should build a standard way for neovide to store data from run to run
[...]

PR #665, opened 2021-05-18
Initial Implementation for #622

bbli:
Hey there!

So I tried to call neovim to get the standard path as suggested by Pangoraw in #622 , but because of the way this code is structured, it causes the following problem:
1. We would need to extract a neovim handle from "start_bridge" and pass it on to "create_window", since both functions call "window_geometry_or_default". That is sort of ugly.
2. If we instead just make the call to neovim just inside "start_bridge" and writing to the SETTINGS global variable, "start_bridge" would need to be synchronized with "create_window". But that's not good as we don't want to block.

So, I just hardcoded the respective paths for unix and windows. If neovim does decide to change what `:call stdpath("data")` evaluates to, we can just change it with a simple commit.

I used serde as @Kethku suggested, but as of now the code assumes the neovide-settings.txt file only contains the last window size, rather than being some sort of general json structure. I think once that is done, we can just read the settings file in main, and pass the json object on to "start_bridge" and "create_window"?(Since right now I reread the file in each of those functions) 

(All changes in the 2 renderer files was because of rustfmt.)
(Also, as an aside, is there a way for me to temporarily turn off CI when making commits? I would like to make a bunch of smaller commits as I work, and then squash when I finally finish.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1516,6 +1516,8 @@ dependencies = [
  "regex",
  "rmpv",
  "rust-embed",
+ "serde",
+ "serde_json",
  "skia-safe",
  "swash",
  "tokio",
@@ -2196,6 +2198,20 @@ name = "serde"
 version = "1.0.126"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ec7505abeacaec74ae4778d9d9328fe5a5d04253220a85c4ee022239fc996d03"
+dependencies = [
+ "serde_derive",
+]
+
+[[package]]
+name = "serde_derive"
+version = "1.0.126"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "963a7dbc9895aeac7ac90e74f34a5d5261828f79df35cbed41e10189d3804d43"
+dependencies = [
+ "proc-macro2 1.0.26",
+ "quote 1.0.9",
+ "syn 1.0.72",
+]
 
 [[package]]
 name = "serde_json"
@@ -2868,6 +2884,7 @@ dependencies = [
  "parking_lot 0.11.1",
  "percent-encoding",
  "raw-window-handle",
+ "serde",
  "smithay-client-toolkit",
  "wayland-client",
  "winapi 0.3.9",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,8 @@ default = []
 embed-fonts = []
 
 [dependencies]
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
 neovide-derive = { path = "neovide-derive" }
 euclid = "0.20.7"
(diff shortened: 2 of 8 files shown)


Kethku:
Thanks so much for digging into this! I'm a bit swamped mid week but will try to get to this this weekend to review.

A couple thoughts:
1. you can turn off the cargo checks by using the -n argument when doing git commands. I have removed the hooks in the opengl branch (but if you swap I think you have to manually remove them from the .git folder)
2. current development is targeting the opengl branch. Any chance you could rebase on that and have the PR set to merge to opengl?
3. Given that the solution requires calling a data path in neovim and then accessing that data. I wonder if the correct way to solve this is expose some mechanism for setting and retrieving the size of the window in neovim itself and then writing a plugin to do the save and restore that way. If necessary we could ship that plugin with neovide somehow, but that way the code that is really neovim specific can live in neovim. What do you think?

Regardless, I will take a look at the PR and give some more in depth thoughts soon. Thanks again!

bbli:
Yeah no problem. I wanted to learn more about neovide's internals, and as it turns out, this was indeed a good first issue for doing so!

> you can turn off the cargo checks by using the -n argument when doing git commands. 

k, thanks for the tip.

> Any chance you could rebase on that and have the PR set to merge to opengl?

Alright, I will try to get a working commit in before the weekend so you can review it. 

> expose some mechanism for setting and retrieving the size of the window in neovim itself and then writing a plugin to do the save and restore that way.

Hmm, can neovim do serialization, or do you just want to get `stdpath`? The stdpath is a soft dependency though, as even if neovim decides to change the location(which I think is unlikely), neovide will continue working.

bbli:
Alright, the executable works now on the opengl branch, though some tests are failing from missing a "font_fix" import. Also, just wondering, why is `INITIAL_DIMENSIONS` type (u64,u64), when it just gets converted later to u32 for `PhysicalSize<u32>`/`LogicalSize<u32>`?

Kethku:
I don't think theres a particular rhyme or reason for why we would use u32 vs u64 other than history. Its possible that sdl2 used u64 for window sizes or something similar. As for how to get the data into the right place I'll have to think a bit more. I generally try to send that kind of data via ui commands or similar in order to communicate between the different components, but its become somewhat clunky...

bbli:
> a particular rhyme or reason 
haha, ok. 

> As for how to get the data into the right place I'll have to think a bit more
Yeah, no probs. Let me know when you want to pick this back up!

Kethku:
Looks like when I deleted the opengl branch, this PR got disabled. Not sure why. Maybe this needs to be rebased on main now?

bbli:
Kk, I am a bit busy atm but will get around to it by the end of the week. 

So any updates on what you think the best way is to handle the save location + how to load the persistent setting variables into neovide :)?

PR #668, opened 2021-05-21
Fix 577

### LINKED ISSUE #577: when guicursor set to blinking, the cursor starts in 'off' mode before blinking
EDIT: you can ignore this initial post, please look at the 3rd message: https://github.com/Kethku/neovide/issues/577#issuecomment-823240031

Hi,

When I hold e.g. `l` and let the keyrepeat kick in, the cursor is not visible during the hold time.
If I try this in neovim shell or gvim/vim the cursor is visible.

I know I shouldn't be holding movement keys.. but it happens from time to time :)
(I haven't set any of the neovide specific settings)

PyGamer0:
Fixes #577

### DIFF
--- a/src/renderer/cursor_renderer/blink.rs
+++ b/src/renderer/cursor_renderer/blink.rs
@@ -75,8 +75,8 @@ impl BlinkStatus {
         }
 
         match self.state {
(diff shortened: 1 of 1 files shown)


Kethku:
Awesome. Thanks for the fix!

PR #671, opened 2021-05-22
Add instructions to install Neovide with Scoop to README.md

kiedtl:


Kethku:
I'm slightly hesitant. Given that neovide is a single compiled executable, I wonder if the pre compiled installation instructions should lead to just downloading the exe rather than using scoop.

Maybe I'm just not familiar with scoop in general...

kiedtl:
`scoop install` is essentially a wrapper for "download executable, move to $PATH, and add desktop icons if any". It just simplifies the process into one convenient script and makes uninstallation easy, IMO.

(And yeah, scoop does download precompiled executables instead of building from source, if that's what you were asking.)

j4qfrost:
@kiedtl is someone already uploading to scoop? Ideally the CI should be pushing exes to scoop.

kiedtl:
Scoop directly downloads executables. There is no work necessary on part of the software maintainer, see [the Neovide manifest](https://github.com/lukesampson/scoop-extras/blob/master/bucket/neovide.json#L26) in scoop's extras bucket. When a Neovide update is pushed, Scoop's maintainers use scripts like [these](https://github.com/lukesampson/scoop/blob/master/bin/checkver.ps1#L5) to automatically update the manifest. Cool, huh? :)

Kethku:
I'm ok with something like this going in then. Could you target the opengl branch though? Currently it is serving as a develop branch until I can get my act together and fix the remaining issues with it.

Kethku:
I manually added the change. Thanks for the PR!

PR #672, opened 2021-05-22
Use interactive path for neovim resolution

### LINKED ISSUE #647: Neovide not reading .~/bashrc nor ~/.profile when using --wsl
When running with the --wsl option, startup dotfiles like ~/.bashrc or ~/.profile are not read, causing to not have them by default on neovide.

anirudhb:
This PR does two things:
1) It uses `bash -ic` to call `which` in WSL, which allows .bashrc env to be taken into account (resolves #647)
2) It adds a startup command to neovim to get the interactive PATH and set it, so that other things don't break (e.g. LSP which depends on the server being in the PATH)


Kethku:
Looks good to me! Thanks for the contribution!

PR #680, opened 2021-05-24
Implement CLI parsing with clap

ErikMayrhofer:
This is a draft of integrating clap into neovide. I tried to stick to the existing architecture as much as possible, yet it could be that I overlooked or misunderstood something, so comments and criticism are greatly appreciated.

### Summary
* Created a file cmd_line.rs which has a `CmdLineSettings` struct. The function `handle_command_line_arguments` does the clap-thingy, builds a `CmdLineSettings` and sticks it into the `SETTINGS`.
* `SETTINGS.neovim_args` was removed and put into `CmdLineSettings` and the appropriate changes were made to the rest of the files.
* The `init_logger` function was moved to `main.rs` (It could probably be moved to a separate file completely...) and it reads `verbosity` and `log_to_file` from `CmdLineSettings` instead of taking them as a parameter.
* The primitive command-handling from `main.rs` was replaced with a call to `handle_command_line_arguments` mentioned above.
* `cargo fmt` made changes to `cursor_renderer/mod.rs` and `renderer/mod.rs`.
* Changed usages of `std::env::args()` to the new `CmdLineSettings` api.

### Currently implemented CLI arguments

```
Neovide 0.7.0
keith <keith@the-simmons.net>
A simple GUI for Neovim.

USAGE:
    neovide [FLAGS] [OPTIONS] [files]... [-- <neovim_args>...]

FLAGS:
        --disowned     Disown the process. (only on macos)
    -h, --help         Prints help information
        --log          Log to a file
        --maximized    Maximize the window
        --multiGrid    Enable Multigrid
    -V, --version      Prints version information
    -v                 Set the level of verbosity
        --wsl          Run in WSL

OPTIONS:
        --geometry <geometry>        Specify the Geometry of the window
        --remote-tcp <remote_tcp>    Connect to Remote TCP

ARGS:
    <files>...          Specify the Geometry of the window
    <neovim_args>...    Specify Arguments to pass down to neovim
```

I also would propose to change `--multiGrid` to `--multi-grid` or `--multigrid` to adhere to a unified unix-y style.

### DIFF
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -2,11 +2,12 @@ use crate::settings::*;
 
 use clap::{App, Arg};
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct CmdLineSettings {
     pub verbosity: u64,
     pub log_to_file: bool,
     pub neovim_args: Vec<String>,
+    pub neovim_bin: Option<String>,
     pub files_to_open: Vec<String>,
 
     pub disowned: bool,
@@ -20,6 +21,7 @@ pub struct CmdLineSettings {
 impl Default for CmdLineSettings {
     fn default() -> Self {
         Self {
+            neovim_bin: None,
             verbosity: 0,
             log_to_file: false,
             neovim_args: vec![],
@@ -96,7 +98,14 @@ pub fn handle_command_line_arguments() {
 
     let matches = clapp.get_matches();
 
+    /*
+     * Integrate Environment Variables as Defaults to the command-line ones.
+     *
+     * NEOVIM_BIN
+     * NeovideMultiGrid || --multiGrid
+     */
     SETTINGS.set::<CmdLineSettings>(&CmdLineSettings {
+        neovim_bin: std::env::var("NEOVIM_BIN").ok(),
         neovim_args: matches
             .values_of("neovim_args")
             .map(|opt| opt.map(|v| v.to_owned()).collect())
@@ -108,7 +117,7 @@ pub fn handle_command_line_arguments() {
             .map(|opt| opt.map(|v| v.to_owned()).collect())
             .unwrap_or_default(),
         maximized: matches.is_present("maximized"),
-        multi_grid: matches.is_present("multi_grid"),
+        multi_grid: std::env::var("NeovideMultiGrid").is_ok() || matches.is_present("multi_grid"),
         remote_tcp: matches.value_of("remote_tcp").map(|i| i.to_owned()),
         disowned: matches.is_present("disowned"),
         wsl: matches.is_present("wsl"),
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -1,5 +1,7 @@
 use crate::settings::*;
 
+use clap::{App, Arg};
+
 #[derive(Clone)]
 pub struct CmdLineSettings {
     pub verbosity: u64,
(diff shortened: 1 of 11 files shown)


ErikMayrhofer:
I just noticed that opening specific files is now broken, I will try to fix it. Oopsie

ErikMayrhofer:
I also noticed that there are still a few places where the command-line arguments are pulled directly from `std::env`

- [ ]  `main.rs`: Inside of macos-specific code. 
- [ ] `src/window.mod.rs` to parse `--geometry`
- [ ] `src/bridge/mod.rs` inside of windows-wsl code
- [ ] `src/bridge/mod.rs` to attach to --remote-tcp
- [ ] `src/bridge/mod.rs` to handle --multi-grid
- [ ] `src/window/winit/mod.rs` to handle `--maximized`
- [ ] `src/window/sdl2/mod.rs` to handle `--maximized`

I will try to move those to the new system, but as I am not yet extremely well versed in the inner workings of neovide, I would recommend reviewing those places... I would hate to break things.

src/cmd_line.rs
```diff
@@ -0,0 +1,72 @@
...
+            .map(|opt| opt.map(|v| v.to_owned()).collect())
+            .unwrap_or_default(),
+        verbosity: matches.occurrences_of("verbosity"),
+        log_to_file: matches.is_present("log_to_file"),
+        files_to_open: matches
+            .values_of("files")
+            .map(|opt| opt.map(|v| v.to_owned()).collect())
+            .unwrap_or_default(),
+        maximized: matches.is_present("maximized"),
+        multi_grid: matches.is_present("multi_grid"),
```
Kethku, line 66: multigrid also sets via an environment variable. Is that handled here or where it is used?

Kethku:
Super great work. I'm very excited :) This looks like a huge improvement to the organization and maintainability. Bravo

I will take a look soon and try to get this merged.

src/cmd_line.rs
ErikMayrhofer: I did think of also refactoring the environment variables but really wanted to get your opinion on how to integrate them. Should they just be used as "defaults" for command-line arguments? And if yes should every cli-argument be replaceable through an environment variable? 

For now I simply didn't touch them, so it's still handled in [`src/bridge/mod.rs` ](https://github.com/ErikMayrhofer/neovide/blob/778f4b8cbcf501b1609950d163d4f771a45278c4/src/bridge/mod.rs#L273)

ErikMayrhofer:
Thanks for your nice words!

src/cmd_line.rs
Kethku: I think a strategy where some environment variables are defined as defaults is a good way to go. That way its just handled in one place rather than specialized handling wherever its needed. What do you think?

Kethku:
Ok this looks super great. However, current development is happening in the opengl branch because the two branches have diverged pretty significantly. Is it possible for you to rebase your PR on top of opengl and change it to target that?

Its not great having the second development branch, but I've been trying to get it merged for a couple weeks, theres just been some blockers which have been a pain.

ErikMayrhofer:
Please do not merge yet, I am still testing out if things work correctly after the rebase...

src/cmd_line.rs
ErikMayrhofer: Yup, that seems like a reasonable way.

Just so that I make sure we are on the same page: 
* environment variables exist for some parameters
* environment variables are overwritten by command line parameters so `NeovideMultiGrid=false ./neovide --multiGrid` would result in `CmdLineSettings.multiGrid` being `true`
* this whole logic is handled in cmd_line.rs or a similar place and included in this PR in its final state

Kethku: Yes. That sounds perfect

ErikMayrhofer:
Is it just me or does the current opengl branch not compile the tests? 
Initially I thought I messed something up with the rebase, but then I checked out the current upstream/opengl and it seems like [`src/renderer/fonts/font_loader.rs`](https://github.com/Kethku/neovide/blob/a81e9657c3adce4660aac8c8c3a3d7aca2dd5d00/src/renderer/fonts/font_loader.rs#L87) still uses `font_kit` in its tests... Which is a dependency that has been removed from the `Cargo.toml` in 9579b8a3140df10efa5bb8546e575f096d9968bc

Kethku:
No it absolutely does not compile tests. Don't worry about it. I will fix them later. Sorry I should have mentioned that

ErikMayrhofer:
So, as I see it that should be about it. I didn't find anything that I could have broken, but the last judgement should be up to your keen eyes.

I also introduced an alias for the environment variable NeovideMultiGrid called `NEOVIDE_MULTI_GRID` as pascal cased environment variables seem uncanny to me XD. 
For now, I kept backwards compatibility for `--multiGrid` and `NeovideMultiGrid` but at some point, these probably could be deprecated and replaced with (at least in my opinion) more canonical `--multigrid` and `NEOVIDE_MULTI_GRID`

Kethku:
Merged. Thanks again for the great contribution. This has been on my list for a while

PR #681, opened 2021-05-25
Added trackpad support to scrolling/smoothscrolling in OpenGL branch

shaunsingh:
Fixes https://github.com/Kethku/neovide/issues/679.

### DIFF
--- a/src/window/window_wrapper/mod.rs
+++ b/src/window/window_wrapper/mod.rs
@@ -329,7 +329,7 @@ impl GlutinWindowWrapper {
                         ..
                     },
                 ..
-            } => self.handle_mouse_wheel(0, (lpos.y as i32)/10),
+            } => self.handle_mouse_wheel(0, (lpos.y as f32)/10),
             Event::WindowEvent {
                 event:
                     WindowEvent::MouseInput {
--- a/src/window/window_wrapper/mod.rs
+++ b/src/window/window_wrapper/mod.rs
@@ -329,7 +329,7 @@ impl GlutinWindowWrapper {
                         ..
                     },
                 ..
-            } => self.handle_mouse_wheel(0.0, (lpos.y as f32)/10000000000000000000000000000000000000000f32),
+            } => self.handle_mouse_wheel(0.0, (lpos.y as f32)/1000000000000000000000000000000000000000f32),
             Event::WindowEvent {
                 event:
                     WindowEvent::MouseInput {
--- a/src/window/window_wrapper/mod.rs
+++ b/src/window/window_wrapper/mod.rs
@@ -317,11 +317,12 @@ impl GlutinWindowWrapper {
             Event::WindowEvent {
                 event:
                     WindowEvent::MouseWheel {
-                        delta: MouseScrollDelta::PixelDelta(x, y),
+                        delta: MouseScrollDelta::LineDelta(x, y),
+                        delta: MouseScrollDelta::PixelDelta(lpos),
                         ..
                     },
                 ..
-            } => self.handle_mouse_wheel(x as i32, y as i32),
+            } => self.handle_mouse_wheel(lpos.x as i32, lpos.y as i32),
 
             Event::WindowEvent {
                 event:
--- a/src/window/window_wrapper/mod.rs
+++ b/src/window/window_wrapper/mod.rs
@@ -329,7 +329,7 @@ impl GlutinWindowWrapper {
                         ..
                     },
                 ..
-            } => self.handle_mouse_wheel(0.0, (lpos.y as f32)/100000000000000000000000000000000000000f32),
+            } => self.handle_mouse_wheel(0.0, (lpos.y as f32)/900000000000000000000000000000000000000f32),
             Event::WindowEvent {
                 event:
                     WindowEvent::MouseInput {
--- a/src/window/window_wrapper/mod.rs
+++ b/src/window/window_wrapper/mod.rs
@@ -329,7 +329,7 @@ impl GlutinWindowWrapper {
                         ..
                     },
                 ..
-            } => self.handle_mouse_wheel(0.0, (lpos.y as f32)/10000000000000000000000000000000000000f32),
+            } => self.handle_mouse_wheel(0.0, (lpos.y as f32)/10000000000000000000000000000000000000000f32),
             Event::WindowEvent {
                 event:
                     WindowEvent::MouseInput {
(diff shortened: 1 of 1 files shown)


shaunsingh:
Note: I divided the input value by 10 as it was very fast at the default setting. Maybe make this value configurable in the future? 

I also disabled horizontal scrolling as it was a bit jarring, but can replace it if you would like

shaunsingh:
Here is a little recording of the feature in action (with multigrid enabled). Let me know if I missed something

https://user-images.githubusercontent.com/71196912/119426892-1408a200-bcd8-11eb-868f-99dca82f2b8d.mov

src/window/window_wrapper/mod.rs
```diff
@@ -322,7 +322,14 @@ impl GlutinWindowWrapper {
                     },
                 ..
             } => self.handle_mouse_wheel(x as i32, y as i32),
-
+            Event::WindowEvent {
+                event:
+                    WindowEvent::MouseWheel {
+                        delta: MouseScrollDelta::PixelDelta(lpos),
```
shaunsingh, line 328: `PixelDelta` gets us the trackpad information

src/window/window_wrapper/mod.rs
```diff
@@ -322,7 +322,14 @@ impl GlutinWindowWrapper {
...
+                        ..
+                    },
+                ..
+            } => self.handle_mouse_wheel(0, (lpos.y as i32)/10),
```
shaunsingh, line 332: The value given is (x, y). Horizontal scroll is disabled (hence the 0), Y is divided by 10 as the default was quite fast (jumping around 200-300 lines at a time)

src/window/window_wrapper/mod.rs
```diff
@@ -322,7 +322,14 @@ impl GlutinWindowWrapper {
...
             } => self.handle_mouse_wheel(x as i32, y as i32),
-
+            Event::WindowEvent {
+                event:
+                    WindowEvent::MouseWheel {
+                        delta: MouseScrollDelta::PixelDelta(lpos),
+                        ..
+                    },
+                ..
+            } => self.handle_mouse_wheel(0, (lpos.y as f32)/10),
```
shaunsingh, line 332: Cast lpos.y to f32 instead of i32 (draft), resolves ignoring small inputs (hopefully)

j4qfrost:
@shaunsingh you might find running ```cargo fmt``` cleans up your format issues.

shaunsingh:
I apologize for my inexperience in rust, thanks for the help

shaunsingh:
@j4qfrost the issue I'm running into is that I would like to give a float value as an input while the function signature requires an integer. While integer values work well enough for regular scrollwheels, small inputs are ignored on the trackpad 

Would you happen to know how I could fix this

j4qfrost:
@shaunsingh It appears that there would be nothing wrong with you changing the function signature for handle_mouse_wheel, so go ahead and do that. I think it was set up that way when we were using a different library, but it'd just be easier to to switch to floating point.

src/window/window_wrapper/mod.rs
```diff
@@ -225,7 +225,7 @@ impl GlutinWindowWrapper {
         self.mouse_down = false;
     }
 
-    pub fn handle_mouse_wheel(&mut self, x: i32, y: i32) {
+    pub fn handle_mouse_wheel(&mut self, x: f32, y: f32) {
```
j4qfrost, line 228: I see you went ahead and did that. lol

shaunsingh:
Yeah lol, I thought there would be consequences for regular mouse input, but I forgot the scroll wheel should work just fine

j4qfrost:
Try changing to 1000?

shaunsingh:
its at 100000000000000000000000000000000000000 (which is the highest an f32 value can go iirc) and its still extremely fast 

I think what's happening is that my MacBook has scroll acceleration built into the trackpad, so its more sensitive than it actually is. If someone tries it on a linux or windows -based OS maybe its a better experience?

shaunsingh:
As far as I can tell, the difference between 10 and an obscenely large number is barely anything. Its definitely there, but I think what I need to do is change the input threshold for neovim scrolling, not dividing the original input value

shaunsingh:
I've tried about everything at this point. The scrolling itself works like a charm its just too quick. If anyone has any ideas how how we could make the scrolling slower I would love to hear them.

shaunsingh:
I verified with another friend on macOS, looks like its fine on his end. It might just be something in my configuration. If anyone else with a trackpad could test this pr that would be great

src/window/window_wrapper/mod.rs
```diff
@@ -321,8 +321,15 @@ impl GlutinWindowWrapper {
                         ..
                     },
                 ..
-            } => self.handle_mouse_wheel(x as i32, y as i32),
-
+            } => self.handle_mouse_wheel(x as f32, y as f32),
+            Event::WindowEvent {
+                event:
+                    WindowEvent::MouseWheel {
+                        delta: MouseScrollDelta::PixelDelta(lpos),
```
Kethku, line 328: nitpick: I try not to have abbreviations in variable names. For example its not clear to me what l in lpos stands for. Maybe this should be logical_position?

shaunsingh: It does stand for logical_position, ill change it back

Kethku: Other than that, this looks great! Were you able to figure out the fast scrolling issue?

Kethku: How does it feel if instead of dividing the logical_position.y by 10 you call this function: https://doc.rust-lang.org/std/primitive.f32.html#method.signum and send that to handle_mouse_wheel?

shaunsingh: The issue seems to be related to macOS trackpad acceleration. It works fine for my friend on Linux. I can try playing around with two more ideas but I doubt it will fix the issue

shaunsingh: As far as I can tell dividing by 10 has no effect not the scroll speed, so it can be removed.

Kethku: I'm happy to merge as is because its better than nothing. Just let me know if and when you finish fiddling with it!

j4qfrost: @Kethku Request: squash and merge

shaunsingh:
I think its good to go! I tried playing around with some values locally and it didn't make much of a difference.

shaunsingh:
Just so its not cluttering up the list of pending issues, I think we can close https://github.com/Kethku/neovide/issues/679 as well

Kethku:
Thanks a bunch @shaunsingh

Issue #566, opened 2021-04-16
how to make nvim server not to shutdown when a neovide client closes connection


LXYYY:
Hello, I am using neovide with remote tcp mode. When I start `nvim --headless --listen localhost:6666` in a remote machine, I can connect to it with neovide just as introduced. But I wonder if there is a way to leave remote nvim still running, when I close neovide in my end? So far, the remote nvim automatically stops, when neovide is closed.

Kethku:
This is an interesting point. I think this should be fixable

rashil2000:
Is there any update on this? This is really essential :(

Kethku:
No update. I'm happy to take a PR though. It shouldn't be too hard to adjust

PR #687, opened 2021-05-26
Fix for #566

bbli:
This commit will allow users to keep the neovim server running(if neovide started with `--remote-tcp`) by closing the neovide window, and quit both neovide and neovim if the user uses any variant of `:q`.

### DIFF
--- a/src/window/window_wrapper/mod.rs
+++ b/src/window/window_wrapper/mod.rs
@@ -105,10 +105,14 @@ impl GlutinWindowWrapper {
         self.windowed_context.window().set_title(&self.title);
     }
 
-    pub fn handle_quit(&mut self) {
-        self.ui_command_sender
-            .send(UiCommand::Quit)
-            .expect("Could not send quit command to bridge");
+    pub fn handle_quit(&mut self, running: &Arc<AtomicBool>) {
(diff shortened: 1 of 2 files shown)


Kethku:
Sweet! I'm probably going to wait to merge this for the cli PR as the person who did that has already rebased a couple times. But after that, this is going in

src/window/window_wrapper/mod.rs
```diff
@@ -104,10 +104,15 @@ impl GlutinWindowWrapper {
         self.windowed_context.window().set_title(&self.title);
     }
 
-    pub fn handle_quit(&mut self) {
-        self.ui_command_sender
-            .send(UiCommand::Quit)
-            .expect("Could not send quit command to bridge");
+    pub fn handle_quit(&mut self, running: &Arc<AtomicBool>) {
+        let prefix = "--remote-tcp";
+        if let None = std::env::args().find(|arg| arg.starts_with(prefix)) {
```
Kethku, line 109: Could you change this to access the setting using the new infrastructure that just got merged? An example can be found here: https://github.com/Kethku/neovide/blob/opengl/src/bridge/mod.rs#L144

bbli: kk, it's done.

Kethku:
Perfect. Thanks so much for the contribution!

PR #689, opened 2021-05-27
Add additional information about Neovim

mikkovedru:


Kethku:
This makes sense to add. Could you change your PR to target the opengl branch though?

mikkovedru:
@Kethku 
Done. Hopefully, I did everything correctly as this is my first time using github's graphical interface to fix merge conflicts.

Kethku:
Hmmm. Weird that it included a bunch of other changes... I'm just gonna cherry pick your change instead. Sorry for the confusion.

Kethku:
Thanks for the PR!

src/bridge/mod.rs
```diff
@@ -35,7 +35,11 @@ fn set_windows_creation_flags(cmd: &mut Command) {
 fn platform_build_nvim_cmd(bin: &str) -> Option<Command> {
     if SETTINGS.get::<CmdLineSettings>().wsl {
         let mut cmd = Command::new("wsl");
-        cmd.arg(bin);
+        cmd.args(&[
```
j4qfrost, line 38: What is this change for?

src/bridge/events.rs
```diff
@@ -117,7 +117,7 @@ pub enum WindowAnchor {
     SouthEast,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Debug, Clone)]
```
j4qfrost, line 120: Was this suggested by fmt?

src/bridge/mod.rs
Kethku: I think this was a merge error

src/bridge/events.rs
Kethku: I think this was a merge error

src/bridge/mod.rs
j4qfrost: It's quite curious. That command looks like it would set PATH to an empty string. Was curious if wsl had some special properties.

j4qfrost: Aaa nvm, it prints error to null.

mikkovedru:
> Hmmm. Weird that it included a bunch of other changes... I'm just gonna cherry pick your change instead. Sorry for the confusion.

Yeah, I don't know what happened... I edited the PR, changed the target from master branch to opengl branch and a bunch of other commits showed up (only one of which actually touched the same readme file I modified).

Glad to help! :-)

PR #694, opened 2021-05-28
Basic Frameless window support

### LINKED ISSUE #67: Frameless Window
Can we have frameless window support.

I am using a modified version of [neovim-qt](https://github.com/khalidchawtany/neovim-qt/commit/9f2af46da7c4013d7cecce68db146ee8ef047bce) in which I remove the window frame to make it clean. 
It is draggable resizable  and everything.

PyGamer0:
This commit will add a basic frameless window option. It is invoked by:
```
neovide --frameless
```
This window is not resizeable nor movable as of now.


src/cmd_line.rs
```diff
@@ -32,6 +33,7 @@ impl Default for CmdLineSettings {
             remote_tcp: None,
             multi_grid: false,
             maximized: false,
+            frameless: true,
```
Kethku, line 36: I'm guessing we don't want frameless to default to true. Could you change this to false?

src/cmd_line.rs
```diff
@@ -69,6 +71,11 @@ pub fn handle_command_line_arguments() {
                 .long("multiGrid")
                 .help("Enable Multigrid"),
         )
+        .arg(
+            Arg::with_name("frameless")
+            .long("frameless")
+            .help("Enables Frameless Neovide")
```
Kethku, line 77: I think maybe a better help message would be "Removes the window frame. Note, might make the window not resizable"

src/cmd_line.rs
```diff
@@ -124,5 +131,6 @@ pub fn handle_command_line_arguments() {
         disowned: matches.is_present("disowned"),
         wsl: matches.is_present("wsl"),
         geometry: matches.value_of("geometry").map(|i| i.to_owned()),
+        frameless: !matches.is_present("frameless"),
```
Kethku, line 134: I'm confused by this. If you have a frameless commandline argument, does that set frameless to false?

If thats true, does frameless remove the frame?

Kethku (commented):
Just a couple quick questions. Thanks so much for looking into this!

src/cmd_line.rs
PyGamer0: if `frameless` is `false`;  then the window **is frameless**; that is why I put a not there

src/cmd_line.rs
PyGamer0: ok then

src/cmd_line.rs
PyGamer0: the same reason is why the default value is `true`

Kethku: That is counterintuitive to me. I expected that frameless being false would mean there is a frame. Maybe I'm confused?

PyGamer0:
I have changed the frameless variable to default to false. @Kethku hopefully the code now makes sense.

j4qfrost:
Looks just about ready to merge. @PyGamer0 it would be helpful to add TODOs in sections where we can add code to make the window resizable or movable. We could add a little square that auto-hides for moving the window for example.

Kethku:
I'll just make an issue to track that work :)

Arelav:
Just for coming from Google as me.
You need to configure env variables.
If you run Neovide GUI not from terminal (for example from Alfred)
```
launchctl setenv NEOVIDE_FRAMELESS true
launchctl setenv NEOVIDE_FRAME none
```

bassamsdata:
@Arelav Hey, thank you so much, I was searching for a long time for this solution thanks a lot. It worked perfectly. 🙏

bassamsdata:
@Arelav 
Sorry to mention you again, but a just small question, did the command-v (paste) works with you? I tried a lot but no luck.
I have set this `let g:neovide_input_use_logo = v:true` but seems to not work.
Thanks

Arelav:
> @Arelav
> Sorry to mention you again, but a just small question, did the command-v (paste) works with you? I tried a lot but no luck.
> I have set this `let g:neovide_input_use_logo = v:true` but seems to not work.
> Thanks

I usually use shared clipboard  and use just `p` to paste from yank or clipboard as well. I can’t try just now.

bassamsdata:
Ah ok. I'm now using Ctrl-R to paste but it's a bit meh. thanks for the reply

PR #695, opened 2021-05-29
Fix some shift+key mappings

meluskyc:
Fix two bad keymappings in the Qwerty config. 

See my comment in the issue here: [https://github.com/Kethku/neovide/issues/691#issuecomment-850093364](https://github.com/Kethku/neovide/issues/691#issuecomment-850093364)


Kethku:
Ah good call. Thanks for the fix

PR #706, opened 2021-06-05
Natural Horizontal scrolling

shaunsingh:
Fixes: https://github.com/Kethku/neovide/issues/582, 

Re-enables horizontal trackpad scrolling and swaps the scroll direction. This was actually the source of the "jarring" feeling I was experiencing in my original PR. I also raised the scroll threshold to -2.0 and 2.0 so its *slightly* less sensitive. 

Tested on macOS and arch linux with MacBook trackpad, Magic Mouse, and a "conventional" scrolling razer mouse.

### DIFF
--- a/src/window/window_wrapper/mod.rs
+++ b/src/window/window_wrapper/mod.rs
@@ -236,8 +236,8 @@ impl GlutinWindowWrapper {
         }
 
         let vertical_input_type = match y {
-            _ if y > 2.0 => Some("up"),
-            _ if y < -2.0 => Some("down"),
+            _ if y > 1.8 => Some("up"),
+            _ if y < -1.8 => Some("down"),
             _ => None,
         };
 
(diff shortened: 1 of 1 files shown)


shaunsingh:
This only fixes the sensitivity issue (how much you need to move to do the initial scroll), neovim itself still scrolls too fast. As far as I understand theres no way around it, we would need to implement sub-pixel scrolling for that.

src/window/window_wrapper/mod.rs
```diff
@@ -252,8 +252,8 @@ impl GlutinWindowWrapper {
         }
 
         let horizontal_input_type = match y {
```
Kethku, line 254: just noticed, I think this y should be x. Actually its a bit odd to have a match where the actual value isn't used.

Kethku:
```suggestion
        let horizontal_input_type = match x {
```

PR #708, opened 2021-06-06
Add environment variables for window options

PyGamer0:
This pr adds environment variables for `--frameless`, `--maximized`. Similar to `--multigrid` env var

### DIFF
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -123,7 +123,9 @@ pub fn handle_command_line_arguments() {
             .values_of("files")
             .map(|opt| opt.map(|v| v.to_owned()).collect())
             .unwrap_or_default(),
-        maximized: matches.is_present("maximized"),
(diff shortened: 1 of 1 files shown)


src/cmd_line.rs
```diff
@@ -123,14 +123,18 @@ pub fn handle_command_line_arguments() {
             .values_of("files")
             .map(|opt| opt.map(|v| v.to_owned()).collect())
             .unwrap_or_default(),
-        maximized: matches.is_present("maximized"),
+        maximized: matches.is_present("maximized")
+            || std::env::var("NeovideMaximized").is_ok()
```
Kethku, line 127: nit: I don't think we should have the camel case versions for this. The command line argument change made a good point that we shouldn't use camel case for environment variables.

Could you remove those? Other than that, this is a great change!

PR #711, opened 2021-06-08
M1 Runner

j4qfrost:
After this gets merged in, @Kethku either you can give me admin access to the repo to set up the runner or we can hop on a voice call and get things set up. I can get the snap craft set up as well.


Kethku:
Sounds great to me. I'll try to block off some time maybe this sunday or something to finish it up. Thanks so much for setting it up!

Issue #714, opened 2021-06-12
Multigrid is broken with latest neovim-git


last-partizan:
I'm using latest neovim#master, with my patch to fix menu rendering in multigrid mode and neovide#main

```
neovide-git 0.7.0.r24.g51fa8c0-1
neovim-git 0.4.0.r3810.gd09b8ad71-1
```

But when i trigger completion menu (or command completion menu), it doesn't appears, and i see this in terminal:

```
 ~ > export RUST_BACKTRACE=full
 ~ > /usr/bin/neovide --multiGrid
Ignored client type property: "methods"
Ignored client type property: "attributes"
ERROR [neovide::error_handling] Could not parse event from neovim: invalid event format
thread 'tokio-runtime-worker' panicked at 'Could not parse event from neovim: invalid event format', src/error_handling.rs:5:5
stack backtrace:
   0:     0x556054a9cf70 - <unknown>
   1:     0x556054941dff - <unknown>
   2:     0x556054a9c7b1 - <unknown>
   3:     0x556054a9c115 - <unknown>
   4:     0x556054a9b46a - <unknown>
   5:     0x556054ab9288 - <unknown>
   6:     0x556054ab91fc - <unknown>
   7:     0x556054ab91ad - <unknown>
   8:     0x5560548f2fb0 - <unknown>
   9:     0x5560549cef27 - <unknown>
  10:     0x556054a01c80 - <unknown>
  11:     0x556054ad52d4 - <unknown>
  12:     0x556054accbf1 - <unknown>
  13:     0x556054ad0354 - <unknown>
  14:     0x556054acac0a - <unknown>
  15:     0x7f5ddca6a259 - start_thread
  16:     0x7f5ddc84f5e3 - __GI___clone
  17:                0x0 - <unknown>
```

last-partizan:
```
nvim --version
NVIM v0.5.0-dev+1398-gd09b8ad71-dirty
```

nvim version is actually this, (for some reason PKGBUILD shows it as 0.4 ...)

last-partizan:
https://github.com/neovim/neovim/commit/edb5864a29ec0d1049c4674afb75c8279301446c

probably this commit is related, they added z-index to ui_call_win_float_pos call.

Kethku:
This has been fixed in the opengl branch. Let me find the commit

Kethku:
[...]

PR #715, opened 2021-06-12
fix: Fix parsing win_float_pos_arguments #714

last-partizan:
This fixes event handling from latest neovim, which added z_index to win_float_pos_arguments.

### DIFF
--- a/src/bridge/events.rs
+++ b/src/bridge/events.rs
@@ -200,6 +200,7 @@ pub enum RedrawEvent {
         anchor_row: f64,
         anchor_column: f64,
         focusable: bool,
+        z_index: i64,
     },
     WindowExternalPosition {
         grid: u64,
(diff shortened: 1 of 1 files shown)


src/bridge/events.rs
```diff
@@ -200,6 +200,7 @@ pub enum RedrawEvent {
         anchor_row: f64,
         anchor_column: f64,
         focusable: bool,
+        z_index: i64,
```
last-partizan, line 203: @Kethku z_index is integer, so i'm using i64 here. But, anchor_row and column also integers, but they using f64.

Should i also change this to f64? Why are they using f64?

last-partizan:
btw, `cargo fmt -- --check` was giving warning in pre-commit hook, but they are unrelated to my changes, so i skipped pre-commit checks.

src/bridge/events.rs
seandewar: These values are floats in the spec.
This is so the position of floating windows does not need to be exactly aligned with the character grid in GUIs.

Kethku: @seandewar thats correct. We made that change not that long ago

Kethku:
I'm happy to merge this to main, but an equivalent change has been made in the opengl branch. So in the not too distant future its going to get clobbered.

But I'm happy to merge this in the meantime.

PR #722, opened 2021-06-14
Snapcraft Stable

j4qfrost:
This will build stable releases for snap so I can push them to the stable channel.


Kethku:
So we should likely wait for the token before merging this correct?

j4qfrost:
Yep, it was something just learned from looking at the workflow build. The token I gave you was for nightly builds. Right now builds are being made for nightly and dev channels, so they're are available on snap.

PR #724, opened 2021-06-14
Remove duplicate `distance_length_adjust`  from  `src/renderer/cursor_renderer/mod.rs`

ihasdapie:
I just saw the newly pushed font fallback features in the opengl branch so I tried building it. Noticed that it didn't compile because `duplicate_length_adjust` was specified twice (probably by accident) in `src/renderer/cursor_renderer/mod.rs` ... so I deleted the extra line.
Not sure if it's worth a pull request, but here it is anyways.

### DIFF
--- a/src/renderer/cursor_renderer/mod.rs
+++ b/src/renderer/cursor_renderer/mod.rs
@@ -48,7 +48,6 @@ impl Default for CursorSettings {
             animate_in_insert_mode: true,
             animate_command_line: true,
             trail_size: 0.7,
-            distance_length_adjust: false,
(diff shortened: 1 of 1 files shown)


PyGamer0 (changes requested):
I dont think we want the `distance_length_adjust` to be `true` by default; can you change it to false

j4qfrost:
This isn't necessary. The clippy fix has already been applied.

pencilcheck:
I just checkout opengl branch and still getting this error.

Kethku:
Opengl branch has been merged to main. It is no longer being used for development

PR #736, opened 2021-06-20
integrate new keyboard api

Kethku:
Integrates the keyboard api from upstream winit. These haven't merged yet, but by integrating the branches, we can try it out on various platforms.

Yes, this significantly changes keyboard handling and may send different keyboard inputs to neovim. This could result in some key bindings not working like they did before.


Kethku:
Merging this as its strictly better than without it.

Issue #723, opened 2021-06-14
Fonts not rendering on certain places.


PyGamer0:
![image](https://user-images.githubusercontent.com/64531844/121843807-947f5900-cd00-11eb-9fcf-88c0d264b34e.png)

As seen in the above screenshot; in some places the font is not rendering. (latest main branch)
My current guifont is:
```viml
set guifont=Iosevka\ Custom,DejaVuSansMono\ NF:h16
```
if i change it to
```viml
set guifont=DejaVuSansMono\ NF:h16
```
it works
but if i change it to
```viml
set guifont=Iosevka\ Custom:h16
```
it stops working

last-partizan:
Yes, i'm facing similar issue:

with
```
set guifont=JetBrainsMono\ Nerd\ Font,DejaVu\ Sans\ Mono,Iosevka,DejaVu\ Sans,Droid\ Sans\ Fallback:h24
```
![image](https://user-images.githubusercontent.com/301015/121955710-f5427c00-cd68-11eb-9722-2db9ad18a2be.png)


with
```
set guifont=Iosevka:h24
```
![image](https://user-images.githubusercontent.com/301015/121955820-0e4b2d00-cd69-11eb-8701-8b7f71b2879a.png)

anirudhb:
I'm facing a similar issue on the OpenGL branch on the last word in a line (when there is no highlighted word after it):
![image](https://user-images.githubusercontent.com/9709011/122626893-8cf1e400-d061-11eb-83ad-55a430c4bd0d.png)

Kethku:
The opengl branch is no longer in development. Please use main.

tpict:
I'm seeing something similar with a fresh build from the main branch. The git status indicator in my shell prompt (something like `master• ❱`) is invisible, but occupies the space it would normally. Interestingly, if I place a cursor over the space where a character should be, it is drawn:

<img width="432" alt="image" src="https://user-images.githubusercontent.com/12588098/122657509-f2ed7280-d118-11eb-8ee0-fbea716d99fd.png">

anirudhb:
I believe the issue seems to be that the size of the text region is computed incorrectly, as currently, Neovide assumes that all characters are the same width (when e.g. ligatures and emoji/special characters) can make the text longer than a simple `len(text) * avg_width` (which is how it's currently computed [here](https://github.com/Kethku/neovide/blob/main/src/renderer/mod.rs#L98-L103)

Kethku:
[...]

PR #737, opened 2021-06-21
Don't drop entire blob when shaping control characters

anirudhb:
Fixes the issue where a control character in a text run would cause the entire blob to be dropped. Now it will just skip the control characters.

This issue can appear at the end of a line when there is e.g. virtual text after the newline character

@PyGamer0 @last-partizan @tpict Please try out this branch and see if it fixes your issues!

Ref #723

### DIFF
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -235,7 +235,7 @@ impl CachingShaper {
             });
 
             if glyph_data.is_empty() {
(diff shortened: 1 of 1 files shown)


Kethku:
I'm just gonna merge this because I think its correct. Thanks so much for investigating!

Kethku:
Actually nvm. My fix is a superset of this one. Reverting. Even so thanks so much for investigating! I appreciate it.

PR #746, opened 2021-06-23
Fix macOS .dmg builds

shaunsingh:
(fixes https://github.com/Kethku/neovide/issues/741)

# Pull Requests

Please note that we accept pull requests from anyone, but that does not mean it will be merged.

This PR introduces a new method of building the .dmg file, which should clear apple's quarantine process and publish it to GitHub artifacts. The method itself is based off of the one used for [Alacritty'(https://github.com/alacritty/alacritty), a terminal emulator also written in rust 

* No

## Notes 
I don't have much experience with GitHub actions/artifacts, so it would be great if someone could help me out with publishing the dmg build as an artifact. The icon itself is the vim icon from [here](https://macosicons.com/vim), as it matches the new macOS 11 icon styling, but I can use any icon you would like.

image

## Checklist 

- [x] Setup Makefile
- [x] Setup app templete 
- [x] Get .dmg to build 
- [ ] Get .dmg to publish (either as a nightly release or artifact)
- [x] Add changes to M1 build as well

Note that there are issues with code signing M1 builds, this should resolve that. Currently the app still crashes on start, but I believe thats related to a code issue, not a building issue.


IBS27:
@shaunsingh when will this pull request be completed?

Kethku:
Please be patient. We are all working on this in our spare time. It will get done when it gets done

shaunsingh:
@IBS27 Sorry for the delay, I've been terribly sick the past week or so, but I promise ill resolve this issue as soon as I get better

Currently the only two issues are 
1. Uploading the produced .dmg build to GitHub actions 
2. Fixing the crashing issue https://github.com/Kethku/neovide/issues/447, which wasn't resolved as we thought.

IBS27:
No problem @shaunsingh. Get well soon!

Kethku:
@shaunsingh absolutely no pressure. Hope you feel better soon.

j4qfrost:
I'll take a look this weekend. When I downloaded the latest dmg build from actions, it works just fine. Maybe this is indicative of a linking issue with opengl on some systems. The icon is gone though. M1 builds might be failing because people are running old cargo toolchains? Builds are working just fine on the CI.

.github/workflows/upload_asset.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
+
+# Assure parameters are correct.
+if [ $# -lt 2 ]; then
+    echo "Usage: upload_asset.sh <FILE> <TOKEN>"
+    exit 1
+fi
+
+repo="Kethku/neovide"
```
j4qfrost, line 9: This should be a variable. We shouldn't be hardcoding the repo.

.github/workflows/build.yml
```diff
@@ -107,21 +107,13 @@ jobs:
...
+    - name: Make App
+      run: make dmg
 
-    - name: Create .dmg file
+    - name: Upload Application
       run: |
-        hdiutil create Neovide-uncompressed.dmg -volname "Neovide" -srcfolder target/release/bundle/osx
-        hdiutil convert Neovide-uncompressed.dmg -format UDZO -o Neovide.dmg
-
-    - uses: actions/upload-artifact@v1
```
j4qfrost, line 121: Why do you want to replace this with a script? The action works fine doesn't it? Sorry I'm out of the loop.

extra/osx/Neovide.app/Contents/Info.plist
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
j4qfrost, line 1: I thought cargo bundle handled generating a lot of the properties in this file.

Kethku:
If opengl is the issue, it should be possible to make a metal build for macos without too much difficulty

ahmed-rezk-dev:
Does this build work with Intel?

j4qfrost:
@Kethku just looked at the linked issue. Thought it was another problem, but it's just a permission issue. We need to set up codesigning. The app itself should work though. I have no problem running the builds, so I don't know how to repro the issue.

shaunsingh:
Sorry it took a while, I'm back now and feeling better 

> If opengl is the issue, it should be possible to make a metal build for macos without too much difficulty

Although I'm interested in making a metal build in the future, openGL isn't the issue in this case 

> Does this build work with Intel?

It does, but I advise against using it. Its quite janky and as @j4qfrost mentioned cargo can handle most of the stuff I put manually, so I'll revert those changes

> Why do you want to replace this with a script? The action works fine doesn't it? Sorry I'm out of the loop.

I originally thought that the crashing issue was due to cargo's build-dmg process, but as you can see turns out its not. I think I'll close this PR for now, and open a new one quickly correcting the icon issue with cargo. 

> just looked at the linked issue. Thought it was another problem, but it's just a permission issue. We need to set up codesigning. The app itself should work though. I have no problem running the builds, so I don't know how to repro the issue.

Are you referring to https://github.com/Kethku/neovide/issues/447? The issue shouldn't be with codesigning (if you ctrl-rightclick and hit open, it attempts to open the application). The warning message you get when you first try to open the app is only solvable by notarizing the binary, which requires a $99/year. dev account 

Either way, notarizing the binary (or codesigning it differently) wouldn't resolve the issue we have. If you build the .dmg yourself (anything you build locally is considered secure by apple only for your machine) it still results in the crashing behavior. If you open the app via `open -a <full path>/Neovide.app` it works fine, same if you manually run the unix executable (./neovide), leading me to believe its a pathing issue.

Issue #704, opened 2021-06-05
Set WM_CLASS or WM_INSTANCE 


Federico-Ciuffardi:
I would like to be able to change the WM_CLASS or WM_INSTANCE properties of the noevide window (see the [link](https://www.x.org/docs/ICCCM/icccm.pdf) for more information on those properties).

Something like [st](http://manpages.org/stterm) `-c` flag.

Kethku:
Can you give more details? Is this linux specific? What are WM_CLASS and WM_INSTANCE for?

Federico-Ciuffardi:
Yes, it is Linux specific. Actually it's [xorg](https://www.x.org/wiki/) specific (used by most linux window managers). 

I use it to set a specific class (`notes` for example) for a vim instance with my notes. Then I set up a shortcut that shows me the window class=`notes` and if it doesn't find a window with that class, it opens a vim instance with my notes and the class set to `notes`. This allows me to access my notes quickly and also avoid multiple instances. A window like this is called a [scratchpad](https://i3wm.org/docs/userguide.html#_scratchpad).

TimB87:
So, [winit](https://docs.rs/winit/0.16.0/winit/os/unix/trait.WindowBuilderExt.html#tymethod.with_class) supports defining a class. I feel like the correct way to implement a this feature request we would need to actually set the class, and make it overrideable via a clap argument, e.g. `-c, --class`.

In order to be correct about things, we need to consider the same thing for waylands `app_id`.

last-partizan:
https://github.com/rust-windowing/winit/blob/c916eb613734b389d7add7efec7f698ca739bae2/src/platform_impl/linux/x11/window.rs#L301

Looks like we don't need special argument, this could be done by using variable `RESOURCE_NAME`.

@Federico-Ciuffardi could you try it?

PR #747, opened 2021-06-23
this adds app_id attribute to neovide

TimB87:
Hi!

This is a little feature for Linux users that are on wayland only desktops. Over here, we don't have `class` or `instance` attributes, but an `app_id`.
This should not bring any breaking changes into the codebase.

Further information on the topic is over [here](https://wayland-book.com/xdg-shell-basics/xdg-toplevel.html)


src/window/window_wrapper/mod.rs
```diff
@@ -232,6 +233,8 @@ pub fn start_loop(
         .with_maximized(SETTINGS.get::<CmdLineSettings>().maximized)
         .with_decorations(!SETTINGS.get::<CmdLineSettings>().frameless);
 
+    let winit_window_builder = winit_window_builder.with_app_id("Neovide".to_string());
```
Kethku, line 236: Does this need a cfg to make sure it only runs when on unix machines?

TimB87: Good catch! I haven't tested on any other platform to be honest, and I think this should be true.
Same would be for the `class` and `instance` vars, I'll try to cover them too!

TimB87: After a quick review, I think it might not be enough to check for just 'unix', instead it might need a finer selection between x11 and wayland environments..
I can try to implement it, but no guarantees given 😅

TimB87:
Updated for now. While this shouldn't introduce any problems, this is cutting short. In another PR, it would be great to actually include logic to decide between x11 and wayland features of glutin and adjust the window building to the correct one present.

Kethku:
@TimB87 would you be wiling to update #704 with what you think would be necessary?

PR #751, opened 2021-06-25
Remove <S- when holding shift

### LINKED ISSUE #750: Hitting the + button sends `<S-+>`, which appears to behave differently from just `+`
**Describe the bug**
Neovide appears to be sending `<S-+>` instead of `+`.

Normally this is not a problem. Other symbols that require shift seem to work fine (e.g. `*` sends `<S-*>` but works as expected)

**To Reproduce**
1. Use a keyboard layout where "+" requires you to hold shift
2. Press +
3. Instead of the normal `+` behavior, my cursor moves down one screen.

**Expected behavior**
I jump to the beginning of the next line.

 - OS: Arch Linux
 - Neovide Version 4159c47ff4f30073b92b9d63fc6ab70e07b74b6d
 - Neovim Version v0.4.4

**Please run `neovide --log` and paste the results here:**

If you want the whole thing, let me know, but this line stood out to me:
```
TRACE [neovide::bridge::ui_commands] Keyboard Input Sent: <S-+>
```

**Additional context**
I have some free time right now and am messing with the Neovide source code as I write this :grin:.

I added a quick hack like this and it fixed the problem:
```rust
// src/bridge/ui_commands.rs
            UiCommand::Keyboard(input_command) => {
                trace!("Keyboard Input Sent: {}", input_command);
                if &input_command == "<S-+>" {
                    nvim.input("+").await.expect("Input failed");
                } else {
                    nvim.input(&input_command).await.expect("Input failed");
                }
            }
```

But I'm wondering if we should just not wrap shifted keys with `<S-` at all. Will continue playing around and update this if I find any good ideas.

Resonious:
Fixes #750 

This PR makes the window wrapper no longer send `<S-{key}>` when the shift key is held.

See the linked issue for details. I don't know much about how nvim handles raw terminal input, but Neovide seems to behave perfectly fine sending `A` instead of `<S-A>`.

Maybe! This change is pretty closely related to windowing and input, and I only have an Arch Linux device to test on right now. Would be awesome if somebody could hop on Windows/macOS and try holding the shift button.

### DIFF
--- a/src/window/window_wrapper/keyboard_manager.rs
+++ b/src/window/window_wrapper/keyboard_manager.rs
@@ -86,16 +86,15 @@ impl KeyboardManager {
     }
 
     fn format_keybinding_string(&self, special: bool, text: &str) -> String {
-        let special = special || self.shift || self.ctrl || self.alt || self.logo;
+        let special = special || self.ctrl || self.alt || self.logo;
 
         let open = or_empty(special, "<");
-        let shift = or_empty(self.shift, "S-");
         let ctrl = or_empty(self.ctrl, "C-");
         let alt = or_empty(self.alt, "M-");
         let logo = or_empty(use_logo(self.logo), "D-");
         let close = or_empty(special, ">");
 
-        format!("{}{}{}{}{}{}{}", open, shift, ctrl, alt, logo, text, close)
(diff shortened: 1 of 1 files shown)


Kethku:
I think this is a great idea so I merged the change into https://github.com/Kethku/neovide/pull/752

PR #752, opened 2021-06-25
Keyboard deadkey handling

Kethku:
# Pull Requests

Supports handling dead keys more effectively

Reworks the keyboard input text code to handle deadkeys better

No this fixes broken input but should leave qwerty input pretty much the same

### DIFF
--- a/src/window/window_wrapper/keyboard_manager.rs
+++ b/src/window/window_wrapper/keyboard_manager.rs
@@ -4,14 +4,6 @@ use glutin::keyboard::Key;
 use crate::bridge::UiCommand;
 use crate::channel_utils::LoggingTx;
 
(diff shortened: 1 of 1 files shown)


clason:
I am sorry to say that this PR made the situation _worse_ for me: dead keys that work on main (like `é` and `è` no longer work. 

(This is on macOS Intel, German keyboard layout.)

clason:
I looked at the trace log, and neither the key `´` nor the subsequent key `e` (which should compose to `é`) show up as input events.  (The  `´` key corresponds to the `=` key on the US layout and is -- apparently correctly treated by neovide as -- a dead key.)

(Note that this combination doesn't involve a `shift` key, or any modifier.)

PR #753, opened 2021-06-25
Fix sporadic cursor jumps (a bit hacky)

### LINKED ISSUE #341: Option to disable cursor animation when jumping to/from command line
Just tried out Neovide today; it's awesome! I'm really fond of the cursor animation, however I find the jumps to and from the command line quite jarring. For example I have `<C-s>` mapped to `:wa<cr>` and this pretty much results in a streak of light across my screen whenever I hit save. (along with other mapped commands)

Cheers!

### LINKED ISSUE #743: Debounce Cursor Moves When Jumping To Message/CommandLine
**Describe the bug**
mouse scrolling animation is triggered randomly when typing. This is so distracting.

https://user-images.githubusercontent.com/21097695/123016096-81971380-d3e7-11eb-8c6c-d33c400245aa.mp4

 - OS: Windows 10
 - Neovide Version [built from source]
 - Neovim Version [0.5.0]

**Please run `neovide --log` and paste the results here:**
[neovide_rCURRENT.log](https://github.com/Kethku/neovide/files/6698055/neovide_rCURRENT.log)

Resonious:

Fixes #743
Fixes #341 (well, non-optionally)

Hello, me again :grin:. My cursor jumps to the bottom line quite often, especially with MultiGrid turned on. Sometimes just pressing "i" is enough to get a white streak across my screen.

This PR is a bit of a hack, but it completely resolves the problem for me.

This PR relies on the following assumptions:
- that the gutter window is always ID 3, and
- that intentionally moving the cursor to the gutter always lands you on column 1.

In my case, both of those assumptions hold, so I will leave this up for discussion!

### DIFF
--- a/src/editor/mod.rs
+++ b/src/editor/mod.rs
@@ -383,6 +383,26 @@ impl Editor {
     }
 
     fn set_cursor_position(&mut self, grid: u64, grid_left: u64, grid_top: u64) {
+        // The gutter is window ID 3. Automatic command execution sometimes sends the
+        // cursor to the gutter unexpectedly, causing confusing trail effects.
+        let cursor_sent_to_gutter = self.cursor.parent_window_id != 3 && grid == 3;
+
+        // When the user presses ":" to type a command, the cursor is sent to the gutter
+        // in position 1 (right after the ":"). In that case, it's probably intentional
+        // and OK to show a trail.
+        let gutter_probably_intentional = grid_left == 1;
+
+        // Keep cursor unchanged if we suspect it was unintentionally sent to the gutter.
+        if cursor_sent_to_gutter && !gutter_probably_intentional {
+            trace!(
+                "Cursor unexpectedly sent to gutter ({} {} {})",
+                grid,
+                grid_left,
+                grid_top
+            );
+            return;
+        }
+
         self.cursor.parent_window_id = grid;
         self.cursor.grid_position = (grid_left, grid_top);
     }
--- a/src/editor/mod.rs
+++ b/src/editor/mod.rs
@@ -242,6 +242,7 @@ impl Editor {
         } else {
             let window = Window::new(
                 grid,
+                WindowType::Editor,
                 width,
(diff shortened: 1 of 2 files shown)


Kethku:
Again, great idea! Some thoughts:

As is I think its likely that this should just get merged. With the newer version of neovim, this problem got worse not better, so fixing it for real is a good plan.

Beyond this trick though we can do a little better https://github.com/neovim/neovim/blob/b86b410a923ca0f7a43a747f528411d2f0311b2a/runtime/doc/ui.txt#L584-L593 the neovim multigrid ui extension explicitly calls msg_set_pos in order to show a message. So I think it should be possible to pipe a flag from the message, through to the editor in order to drive the filter you wrote.

Alternatively, theres an entire extension for drawing messages rather than using the grid system. I have avoided doing something like that with the hope that leaving it as is would make neovide more consistent with other guis/terminal but if we need to dip into that in order to fix this bug, I'm more than willing.

Kethku:
Just as a follow up, I think the data is already piped to the editor https://github.com/Kethku/neovide/blob/main/src/editor/mod.rs#L216 we would just need to set a flag on the particular rendered window that gets created with it and then not move the cursor when it jumps to that position.

Kethku:
And side note: thanks so much for the sponsorship! It makes the project tick and keeps me focused on this particular project rather than other random side efforts. It means a lot

Resonious:
Thanks! That makes sense, thanks for the insight. I wasn't sure how to naturally discern the difference between the message buffer and a regular one.

I've worked out a quick commit for this PR that removes the need for a literal `3`. The `1` is still there, but it's at least half as "hacky" :grin:

Resonious:
And on the sponsorship, of course! I have been using Neovide at work everyday for awhile now and it felt wrong to not give anything back.

Although here I am contributing anyway.

Kethku:
Thanks again for the fix! I think this is a great approach

PR #760, opened 2021-06-26
Add labels and fix feature req template

kilpkonn:
Noticed that you need to manually add bug and feature labels to issues.  
This should fix this problem. 

Also Github didn't show me the feature request template option so I guess this header is needed.
Now hopefully works correctly


shaunsingh:
Thanks for the help, looks like I missed the header earlier.

PR #767, opened 2021-06-27
Make dead keys work on macOS

smolck:
If a dead key is entered, `key_event.text` is `None`, but `key_event.text_with_all_modifiers()` is `Some`. Kind of clunky, but this PR checks for that and uses `key_event.text_with_all_modifiers()` accordingly so dead keys can be inputted.

### DIFF
--- a/src/window/window_wrapper/keyboard_manager.rs
+++ b/src/window/window_wrapper/keyboard_manager.rs
@@ -18,7 +18,6 @@ fn use_logo(_: bool) -> bool {
     false
 }
 
-
 #[cfg(not(target_os = "macos"))]
 fn use_alt(alt: bool) -> bool {
     alt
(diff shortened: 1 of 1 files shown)


clason:
Can confirm that this works fully (modified input, control mappings, dead keys) for macOS and German keyboard layout.

src/window/window_wrapper/keyboard_manager.rs
```diff
@@ -166,11 +165,14 @@ impl KeyboardManager {
...
-                                let key_text = if self.alt && cfg!(target_os = "macos") {
-                                    key_event.text_with_all_modifiers()
-                                } else {
-                                    key_event.text
-                                };
+                                let is_dead_key = key_event.text_with_all_modifiers().is_some()
+                                    && key_event.text.is_none();
+                                let key_text =
+                                    if (self.alt || is_dead_key) && cfg!(target_os = "macos") {
+                                        key_event.text_with_all_modifiers()
```
smolck, line 172: I thought of doing something like `let text_with_mods = key_event.text_with_all_modifiers()` at the start of the `else` and using that so `text_with_all_modifiers()` isn't called twice, but I'm not sure if that's necessary?

Kethku: yeah I'm pretty sure it doesn't matter too much. Im gonna leave it as its more concise this way

Kethku:
Thanks so much for looking into this. I'm gonna merge this and then go back to the non qwerty input issue and double check with the crowd to see if things are working for them

Issue #709, opened 2021-06-06
guifont not properly bolding the font


AnsonH:
I'm using the [MesloLGS NF][1] font with font weight of bold (`:b`) as configured in `init.vim`:
```
set guifont=MesloLGS\ NF:h14.5:b 
```

However, it seems like the bold font weight is not applying to the text of the file. The name of the file in the tabline ("init.vim" in red background) is in bold font weight though. This issue is present in **v0.5.0** as well as in **v0.7.0**.

![Bold font weight not applying](https://user-images.githubusercontent.com/57580593/120919697-8a040480-c6ed-11eb-9753-41b9326d2bd0.png)

To prove that my font is properly installed, here's a screenshot from my Windows Terminal showing what the expected results should look like:

![Windows Terminal](https://user-images.githubusercontent.com/57580593/120919796-0565b600-c6ee-11eb-8da3-0ad2366b8259.png)

Here's some useful information about my environment:
- OS: Windows 10 version 20H2
- neovim version: v0.5.0-dev+1325-gdab6b08a1
- Color scheme: [sonokai][2]

Thank you very much!

[1]: https://github.com/romkatv/powerlevel10k#manual-font-installation
[2]: https://github.com/sainnhe/sonokai

PyGamer0:
related to #457

gshpychka:
@PyGamer0 it's a bug, though, not an enhancement.

Kethku:
It's fine either way. The difference is unimportant for bugs like this which don't break core functionality. We will get to it eventually

gshpychka:
You say it's unimportant, but bugs like these make the tool unsuitable for actual use, since the way the font looks is so important for a text editor. 

I do understand that there are more broken places that need to be fixed, though.

Kethku:
The unimportant bit is whether this is a bug or an enhancement. The issue is important.

Fixing it would not be overly difficult but touches some code that changed recently to enable font fallback. If you're interested in pushing this forward, I can give some hints for what it might look like. Let me know and I'll give some more details about the parts that should change.

gshpychka:
>The unimportant bit is whether this is a bug or an enhancement. The issue is important.

I realize now that I misread your previous comment, disregard.

smolck:
> Fixing it would not be overly difficult but touches some code that changed recently to enable font fallback. If you're interested in pushing this forward, I can give some hints for what it might look like. Let me know and I'll give some more details about the parts that should change.

[...]

PR #769, opened 2021-06-28
feat(fonts): add support for :b and :i in guifont

smolck:
Ref: #709

* Feature

(EDIT: I'm gonna put feature although it could also be seen as a fix . . . technicalities . . . )

* No

### DIFF
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -42,8 +42,16 @@ impl CachingShaper {
         let font_key = self
             .options
             .as_ref()
-            .map(|options| options.fallback_list.first().unwrap().clone().into())
-            .unwrap_or(FontKey::Default);
+            .map(|options| FontKey {
+                italic: options.italic,
+                bold: options.bold,
+                font_selection: options.fallback_list.first().unwrap().clone().into(),
+            })
+            .unwrap_or(FontKey {
+                italic: true,
+                bold: true,
+                font_selection: FontSelection::Default,
+            });
 
         self.font_loader
             .get_or_load(&font_key)
@@ -134,21 +142,32 @@ impl CachingShaper {
 
             // Add guifont fallback list
             if let Some(options) = &self.options {
-                font_fallback_keys.extend(
-                    options
-                        .fallback_list
-                        .iter()
-                        .map(|font_name| font_name.into()),
-                );
+                font_fallback_keys.extend(options.fallback_list.iter().map(|font_name| FontKey {
+                    italic: options.italic,
+                    bold: options.bold,
+                    font_selection: font_name.into(),
+                }));
             }
             // Add default font
-            font_fallback_keys.push(FontKey::Default);
+            font_fallback_keys.push(FontKey {
+                italic: true,
+                bold: true,
+                font_selection: FontSelection::Default,
+            });
 
             // Add skia fallback
-            font_fallback_keys.push(cluster.chars()[0].ch.into());
+            font_fallback_keys.push(FontKey {
+                italic: true,
+                bold: true,
+                font_selection: cluster.chars()[0].ch.into(),
+            });
 
             // Add last resort
-            font_fallback_keys.push(FontKey::LastResort);
+            font_fallback_keys.push(FontKey {
+                italic: true,
+                bold: true,
+                font_selection: FontSelection::LastResort,
+            });
 
             let mut best = None;
             // Use the cluster.map function to select a viable font from the fallback list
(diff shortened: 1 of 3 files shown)


src/renderer/fonts/font_loader.rs
```diff
@@ -46,36 +46,46 @@ pub struct FontLoader {
 }
 
 #[derive(Debug, Hash, PartialEq, Eq, Clone)]
-pub enum FontKey {
+pub struct FontKey {
+    // TODO(smolck): Could make these private and add constructor method(s)?
+    // Would theoretically make things safer I guess, but not sure . . .
+    pub bold: bool,
+    pub italic: bool,
```
smolck, line 53: Thoughts on the TODO here?

src/renderer/fonts/caching_shaper.rs
```diff
@@ -42,8 +42,16 @@ impl CachingShaper {
         let font_key = self
             .options
             .as_ref()
-            .map(|options| options.fallback_list.first().unwrap().clone().into())
-            .unwrap_or(FontKey::Default);
+            .map(|options| FontKey {
+                italic: options.italic,
```
kilpkonn, line 46: Could `TryFrom` trait be used here?

src/renderer/fonts/caching_shaper.rs
```diff
@@ -42,8 +42,16 @@ impl CachingShaper {
...
+                bold: options.bold,
+                font_selection: options.fallback_list.first().unwrap().clone().into(),
+            })
+            .unwrap_or(FontKey {
```
kilpkonn, line 50: Maybe `.unwrap_or(FontKey::Default);`

kilpkonn (commented):
Not rly a collaborator here, but some things I noticed :D

src/renderer/fonts/caching_shaper.rs
smolck: This is kind of what my [TODO above](https://github.com/Kethku/neovide/pull/769#discussion_r659818432) is about . . . it would in theory be cleaner to impl `Default` for `FontKey`, but that can only be used here; below it would require adding a method like `.with_font_selection(selection)` which I'm not sure how I feel about.

I don't really have any strong opinions on which way is better, I'm happy to add a method or two on `FontKey`, I just feel like this way is simpler.

src/renderer/fonts/caching_shaper.rs
smolck: I guess it could, but it would only be used here and so I'm not sure it's worth the added complexity (and moving the logic out of this file thus breaking things up that could stay together).

src/renderer/fonts/caching_shaper.rs
kilpkonn: I think default option is used at least in 2 places so yeah, I'd recommend to implement it to avoid duplicate logic, even if it's really small. :p
But maybe @Kethku can say more which he prefers...

src/renderer/fonts/font_loader.rs
Kethku: eh. I don't really care that much. Its only us consuming these types atm so I'd just leave the todo and move on. Looks good

src/renderer/fonts/caching_shaper.rs
```diff
@@ -134,21 +142,32 @@ impl CachingShaper {
...
+                font_fallback_keys.extend(options.fallback_list.iter().map(|font_name| FontKey {
+                    italic: options.italic,
+                    bold: options.bold,
+                    font_selection: font_name.into(),
+                }));
             }
             // Add default font
-            font_fallback_keys.push(FontKey::Default);
+            font_fallback_keys.push(FontKey {
+                italic: false,
```
Kethku, line 153: Should default and skia fallback use options.italic and options.bold? I think that would maybe help the system pick a font which matches a little better

smolck: I have no idea, I just figured "default" would mean no extra options like that set?

src/renderer/fonts/caching_shaper.rs
Kethku: I agree with smolck. This is explicit about whats happening which I like

src/renderer/fonts/caching_shaper.rs
Kethku: I think in this case its better to set them from the options. Default here means fall back to this when the current glyph isn't present in the configured fallback list.

Kethku: I think this is the only remaining work for this PR. Thanks so much for taking this on!

smolck: > I think in this case its better to set them from the options. Default here means fall back to this when the current glyph isn't present in the configured fallback list.

Pushed the change, look good?

> I think this is the only remaining work for this PR. Thanks so much for taking this on!

No problem, thank you for the detailed info on how to do this! Next up is https://github.com/Kethku/neovide/issues/333#issuecomment-869234020 I think, hopefully I can get that working too.

PR #770, opened 2021-06-28
doc(PULL_REQUEST_TEMPLATE): use checklist & comment

smolck:
Just an update to the PR template that I think might be an improvement; feel free to close if not, of course.


shaunsingh:
Thanks, I think the checklist looks much better as well. I would also like to say I love your work on on uivonim.

PR #782, opened 2021-07-02
feat(keyboard_manager): allow <S-Tab> mappings

smolck:
See https://github.com/Kethku/neovide/issues/445#issuecomment-872053369

This may not be the *best* way to go about this, but it does work; open to suggestions if there's a better way though.

- [ ] Fix
- [x] Feature
- [ ] Codestyle
- [ ] Refactor
- [ ] Other

- [x] No

### DIFF
--- a/src/window/window_wrapper/keyboard_manager.rs
+++ b/src/window/window_wrapper/keyboard_manager.rs
@@ -38,7 +38,7 @@ fn or_empty(condition: bool, text: &str) -> &str {
     }
 }
 
-fn is_control_key(key: Key<'static>) -> Option<&str> {
+fn is_control_key(shift: bool, key: Key<'static>) -> Option<&str> {
     match key {
         Key::Backspace => Some("BS"),
         Key::Escape => Some("Esc"),
@@ -64,6 +64,7 @@ fn is_control_key(key: Key<'static>) -> Option<&str> {
         Key::End => Some("End"),
         Key::PageUp => Some("PageUp"),
         Key::PageDown => Some("PageDown"),
+        Key::Tab if shift => Some("Tab"),
         _ => None,
     }
 }
@@ -157,7 +158,9 @@ impl KeyboardManager {
                         if key_event.state == ElementState::Pressed {
                             // Determine if this key event represents a key which won't ever
                             // present text.
-                            if let Some(key_text) = is_control_key(key_event.logical_key) {
+                            if let Some(key_text) =
+                                is_control_key(self.shift, key_event.logical_key)
+                            {
                                 let keybinding_string =
                                     self.format_keybinding_string(true, true, key_text);
 


Kethku:
This looks good, but I think there are likely other keybindings which should be similarly handled. For instance, shift + space likely doesn't work right now.

smolck:
Right, and this should allow for adding those other ones with just an extra line each; shift-tab was the only one I could think of when I threw this together yesterday. Do you want me to add shift+space to this PR too?

Kethku:
Nah its all good. I'll add them at a later time.

I can't help but feel like these layers of fixes mean we need further changes upstream. I don't think our use case is all that unique.

Specifically knowing if a given modifier was consumed when producing a particular text character. That way we can determine whether we should include the modifier in the resulting key text.

Another option would be to modify neovim to allow modifiers even if the key text used it up. Although think that would be a breaking change for some configurations.

src/window/window_wrapper/keyboard_manager.rs
```diff
@@ -157,7 +158,9 @@ impl KeyboardManager {
                         if key_event.state == ElementState::Pressed {
                             // Determine if this key event represents a key which won't ever
                             // present text.
-                            if let Some(key_text) = is_control_key(key_event.logical_key) {
+                            if let Some(key_text) =
+                                is_control_key(self.shift, key_event.logical_key)
```
last-partizan, line 162: I don't think this is a good idea, we're basically replacing shift with tab.

But, to make it right - we need some refactoring.

```rs
if key_event.state == ElementState::Pressed {
   ...
}
```

What all code in this block is doing, is - formatting keybinding, if it looks like keybinding.

Most of the code - if's, and calling `format_keybinding_string` with different args.

This needs to be refactored into something like this:

```rs
if let Some(keybinding) = self.maybe_format_keybinding(key_event) {
  ... send keybinding
}
```

`maybe_format_keybinding` should use guard clauses to return None for keybindigs which should be ignored, and Some keybinding for valid cases.

@smolck what do you think?

@Kethku do you like this idea?

Kethku: I'm thinking about this. I don't have a good model in my head for what needs to happen here, so I'm going to hold off a bit and come back to it. I think we need more information from upstream in order to really do this correctly.

src/window/window_wrapper/keyboard_manager.rs
```diff
@@ -64,6 +64,7 @@ fn is_control_key(key: Key<'static>) -> Option<&str> {
         Key::End => Some("End"),
         Key::PageUp => Some("PageUp"),
         Key::PageDown => Some("PageDown"),
+        Key::Tab if shift => Some("Tab"),
```
last-partizan, line 67: Why do we need this `if shift` check?

I tested variant without, and it works fine. Mabe i missed something?

last-partizan:
Closing this, issues is already fixed.

PR #783, opened 2021-07-02
Update rust-skia to support aarch64

shaunsingh:
- [x] Fix
- [ ] Feature
- [ ] Codestyle
- [ ] Refactor
- [ ] Other

- [x] No

Tested locally on a chromebook machine, seems to work fine. Fixes https://github.com/Kethku/neovide/issues/766. I would just double check with @Shatur before merging 

Side note: I think quite a few of the dependencies are out of date now. Is it possible to add dependabots for rust? if not, I might have to manually take a look at them one day

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2276,7 +2276,7 @@ dependencies = [
 
 [[package]]
 name = "skia-bindings"
-version = "0.39.1"
+version = "0.40.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "751462e70a68391c38c76d5e14ea24ed4bedd7db4b9bcefffa179f63bddf5c86"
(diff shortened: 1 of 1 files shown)


Kethku:
I'm inclined to merge this regardless as it probably doesn't break anything.

Shatur:
@Kethku @shaunsingh Can't build this. The error is exactly the same. I guess we should update the version inside the `.toml` file too.

shaunsingh:
@Shatur Whoops, forgot about that. Fixed in https://github.com/Kethku/neovide/blob/main/Cargo.toml, I think it should work now

Shatur:
@shaunsingh @Kethku Now the build is broken for all platforms, you forgot to update the checksums. See #798.

Issue #771, opened 2021-06-29
Incorrect window size when using font scaling under Gnome/Xwayland


last-partizan:
I use 4k display, with font scaling set to 1.6, when neovide opens new window, i don't see statusline, it's somewhere below window border. When i resize window, it will correct itself, and i see statusbar.

When i set scaling to 1.0, neovide starts with correct window size and statusbar is visible.

**To Reproduce**
1. In Gnome, use tweak-tools to set scaling to something bigger than 1.0
2. export WINIT_UNIX_BACKEND=x11
3. Run neovide

**Screenshots**
- [Incorrect size with 1.6 scaling](https://user-images.githubusercontent.com/301015/123742528-772bbc80-d8b4-11eb-8d83-af5579a1a4fb.png)
- [Correct size with 1.0 scaling](https://user-images.githubusercontent.com/301015/123742570-8b6fb980-d8b4-11eb-85ca-5f14beb47c8e.png)

**Desktop (please complete the following information):**
 - ArchLinux (Gnome/Wayland)
 - neovide-git 0.7.0.r61.gbc1009b-1
 - neovim-git 0.4.0.r3826.g141647cfb-1

**Additional context**
I use WINIT_UNIX_BACKEND=x11 to fix window decorations #728, and looks like it cannot resize properly in this case. Without this, on wayland session - size is correct.

[neovide_rCURRENT.log](https://github.com/Kethku/neovide/files/6731300/neovide_rCURRENT.log)

last-partizan:
It happens when using Xorg session too, so that's probably not related to Xwayland specifically, but to Xorg in general.

Kethku:
I've encountered this issue on windows as well. I'm hopeful that in progress work to get the size to actually match the passed geometry, and equivalent work to get the start menu to start correctly will address this problem as well.

last-partizan:
![image](https://user-images.githubusercontent.com/301015/124360165-ff052400-dc30-11eb-9975-174b5f5bb175.png)


I've added some debug `println` into handle_new_grid_size, and this is what i get on start:

wayland:
```
Renderer font 8x18
Send resize 100x50
```

x11:
```
Renderer font 8x18
Send resize 160x80
```

If i change window size just a little, next update is correct for both cases (100x50).

Oh! i found why it was happening. First handle_grid_size was using logical, and second was using physical.

![image](https://user-images.githubusercontent.com/301015/124360410-4b049880-dc32-11eb-8032-3688c4f5dc61.png)


[...]

PR #784, opened 2021-07-03
Fix incorrect initial scaling

last-partizan:
Also, I replaced inverse scaling (like 1 / scaling) with natural scaling (like 1 ... 2, what returned by window.scale_factor())
I think easier to work with natural scaling factor and multiplication, instead of inverse scaling factor and division.

It looks fine, and works as expected, but please, review carefully parts where i changed scaling.

Also, instead of passing pair of width/height to handle_new_grid_size i pass LogicalSize. So it's impossible to pass incorrect value.

(Probably, this could be splitted into "just fix" and "refactoring", but i hope it's small enough and can be merged as is)

- [x] Fix
- [ ] Feature
- [ ] Codestyle
- [x] Refactor
- [ ] Other

- [x] No

Refs #771

### DIFF
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -255,7 +255,7 @@ impl Renderer {
 
     #[allow(clippy::needless_collect)]
     pub fn draw_frame(&mut self, root_canvas: &mut Canvas, dt: f32, scaling: f32) -> bool {
-        trace!("Drawing Frame");
+        trace!("Drawing Frame at {} scale", scaling);
         let mut font_changed = false;
 
         let draw_commands: Vec<_> = self
@@ -277,7 +277,7 @@ impl Renderer {
         root_canvas.save();
 
         root_canvas.reset_matrix();
-        root_canvas.scale((1.0 / scaling, 1.0 / scaling));
+        root_canvas.scale((scaling, scaling));
 
         if let Some(root_window) = self.rendered_windows.get(&1) {
             let clip_rect = root_window.pixel_region(self.font_width, self.font_height);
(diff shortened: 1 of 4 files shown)


last-partizan:
Also, i changed this file, for commit to pass pre-commit hooks, but that's completely unrelated to this issue. Should i create another PR with that fix?

```diff
diff --git a/src/editor/grid.rs b/src/editor/grid.rs
index 8990da2..3a835cb 100644
--- a/src/editor/grid.rs
+++ b/src/editor/grid.rs
@@ -243,7 +243,7 @@ mod tests {
         character_grid.characters = vec![grid_cell.clone(); context.area];

         // RUN FUNCTION
-        character_grid.resize(width, height);
+        character_grid.resize((width, height));

         assert_eq!(character_grid.width, width);
         assert_eq!(character_grid.height, height);
```

Kethku:
Beauty. Thanks for looking into this!

Kethku:
> 
> 
> Also, i changed this file, for commit to pass pre-commit hooks, but that's completely unrelated to this issue. Should i create another PR with that fix?
> 
> ```diff
> diff --git a/src/editor/grid.rs b/src/editor/grid.rs
> index 8990da2..3a835cb 100644
> --- a/src/editor/grid.rs
> +++ b/src/editor/grid.rs
> @@ -243,7 +243,7 @@ mod tests {
>          character_grid.characters = vec![grid_cell.clone(); context.area];
> 
>          // RUN FUNCTION
> -        character_grid.resize(width, height);
> +        character_grid.resize((width, height));
> 
>          assert_eq!(character_grid.width, width);
>          assert_eq!(character_grid.height, height);
> ```

I think this is a good change regardless. I don't mind batching them up.

kabouzeid:
Thank you for fixing this! ❤️

PR #791, opened 2021-07-04
Scroll Fix

Kethku:
Fixes https://github.com/Kethku/neovide/issues/755 by removing the image copy performance optimization which I'm not convinced did all that much.

- [x] Fix
- [ ] Feature
- [ ] Codestyle
- [ ] Refactor
- [ ] Other

- [x] No

### DIFF
--- a/src/renderer/rendered_window.rs
+++ b/src/renderer/rendered_window.rs
@@ -1,6 +1,6 @@
 use std::collections::VecDeque;
 
-use skia_safe::canvas::{SaveLayerRec, SrcRectConstraint};
+use skia_safe::canvas::SaveLayerRec;
 use skia_safe::gpu::SurfaceOrigin;
 use skia_safe::{
     image_filters::blur, BlendMode, Budgeted, Canvas, Color, Image, ImageInfo, Paint, Point, Rect,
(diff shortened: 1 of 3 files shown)


last-partizan:
> which I'm not convinced did all that much.

Can we measure performance impact?

I just tried it, and without multigrid it feels almost the same (maybe just a little sluggish), but with multigrid scroll performance is noticeable worse.

src/renderer/rendered_window.rs
```diff
@@ -372,42 +372,6 @@ impl RenderedWindow {
...
-                left,
-                right,
-                rows,
-                cols,
-            } => {
-                let scrolled_region = Rect::new(
-                    (left * renderer.font_width) as f32 * scaling,
-                    (top * renderer.font_height) as f32 * scaling,
-                    (right * renderer.font_width) as f32 * scaling,
-                    (bot * renderer.font_height) as f32 * scaling,
```
last-partizan, line 387: Problem here is definitely with this scaling.

And i think we can somehow fix it, without removing this optimization.

Here is example with scale factor 1.6
image

Here is example with scale factor 2:
image

No artifacts!

But i noticed something else interesting, when using `--multiGrid`, there is no artifacts, with either scroll factor.

PR #793, opened 2021-07-05
Refactor scaling code

last-partizan:
Basic idea here is to remove scaling from as much app as possible.

Let's make only fonts aware of scaling, everything else is expected to just works.

What do you think? Is it good idea? Could this give us any problems in the future?

The basics are already working, it scales as i expect it to. No artifacts.

## What needs to be adressed before merge:

- [x] Fira Code font problem.
    It is not mac-specific, to reproduce you need to have `Fira Code` installed, and no `guifont` in vimrc. Then neovide tries to load Fira Code as default font, and with scaling greater than 1, bug happens.

-------------------

- Fix
- Refactor

- No

### DIFF
--- a/src/window/window_wrapper/mod.rs
+++ b/src/window/window_wrapper/mod.rs
@@ -13,7 +13,7 @@ use std::{
 
 use glutin::{
     self,
-    dpi::{LogicalSize, PhysicalSize},
+    dpi::PhysicalSize,
     event::{Event, WindowEvent},
     event_loop::{ControlFlow, EventLoop},
     window::{self, Fullscreen, Icon},
@@ -25,8 +25,9 @@ use glutin::platform::unix::WindowBuilderExtUnix;
 
 use super::{handle_new_grid_size, settings::WindowSettings};
 use crate::{
-    bridge::UiCommand, channel_utils::*, cmd_line::CmdLineSettings, editor::WindowCommand,
-    redraw_scheduler::REDRAW_SCHEDULER, renderer::Renderer, settings::SETTINGS,
+    bridge::UiCommand, channel_utils::*, cmd_line::CmdLineSettings, editor::DrawCommand,
+    editor::WindowCommand, redraw_scheduler::REDRAW_SCHEDULER, renderer::Renderer,
+    settings::SETTINGS,
 };
 use image::{load_from_memory, GenericImageView, Pixel};
 use keyboard_manager::KeyboardManager;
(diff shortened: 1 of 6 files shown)


src/window/window_wrapper/mod.rs
```diff
@@ -57,22 +56,7 @@ impl GlutinWindowWrapper {
         let window = self.windowed_context.window();
         if self.fullscreen {
             window.set_fullscreen(None);
-
-            // Use cached size and position
-            window.set_inner_size(LogicalSize::new(
-                self.cached_size.width,
-                self.cached_size.height,
-            ));
-            window.set_outer_position(LogicalPosition::new(
```
Kethku, line 66: Looks like this got removed. Was this intentional? I don't remember off the top of my head what the state is for cached window size

src/window/window_wrapper/mod.rs
```diff
@@ -166,52 +150,49 @@ impl GlutinWindowWrapper {
 
     pub fn draw_frame(&mut self, dt: f32) {
         let window = self.windowed_context.window();
-        let scaling = window.scale_factor();
-        let new_size = window.inner_size();
+        let scale_factor = 1.0; // window.scale_factor();
```
Kethku, line 153: Is the scale factor here supposed to be commented out?

src/window/window_wrapper/mod.rs
last-partizan: Oh, that's fullscreen fix from earlier PR, probably i forgot to rebase.

src/window/window_wrapper/mod.rs
last-partizan: That's temporary. I'll remove it alltogether and use PhysicalSize.

last-partizan:
I think it's ready for review and testing.

Bet let's not rush with merging. This is big refactoring and it needs to be reviewed carefully.

vivlim:
The fonts are messed up in retina scaling mode on my mac:
image

I get this message with any retina resolution set
```
INFO [neovide::window::window_wrapper] window created (scale_factor: 2, font_size: 17x35)
```

If I switch to 1x scaling then the font is fine and I get this
```
INFO [neovide::window::window_wrapper] window created (scale_factor: 1, font_size: 8x18)
```

last-partizan:
@vivlim what fonts are you using?

vivlim:
Fira Code, but this also reproduces if I bypass my nvim config entirely.

last-partizan:
@vivlim i was able to reproduce it on mac, with empty config. But as soon as i set guifont to something like

```
set guifont=JetBrainsMono\ Nerd\ Font:h15
```
or
```
set guifont=monospace
```

Issue is gone.

Could you show your exact guifont line with `Fira Code`, and how do you installed it?

last-partizan:
@vivlim please check again, it must work now as expected.

vivlim:
Yep, that is working great now.

Kethku:
I just looked, things seem a little weird on windows. Almost as though the scaling is larger than expected... Capturing my findings here so that I can dig in more outside of work hours

At 100% scale, the font looks perfect. And in fact matches the expected size
image

However at 150% scale, the neovide text looks too small by maybe 30% or so 
image

At 200% it seems worse
image

And 250% even further
image

I think this means that on windows the scaling isn't taken into account as the size of the text doesn't seem to change. I'm wondering if winit isn't reporting a change in scale.

A related issue is that the window changes size drastically by the scaling factor eg the window increases in size 250% at scaling 250% when moved from a 100% scaling window.

last-partizan:
If you're changing scale and not restarting neovide, this is expected. Right now scaling is requested only once at the app start.

Kethku:
Oh gotcha. I think this is probably important to change for setups with multiple different scale factors. But I can make that change after this merges. If thats the case, I think this works! Great stuff

last-partizan:
I have few other refactorings in mind, and can look into dymanic scaling as well after merge.

Also, i have no other ideas what else to fix or check in this branch, so let's merge it.

last-partizan:
Please, use `Squash and merge`, and commit message in conventional commits format.

PR #798, opened 2021-07-06
Fix broken Cargo.lock

last-partizan:
@shaunsingh first time you updated only versions in lockfile, second time you updated only versions in Cargo.toml, i reverted both commits, and updated versions and checksums in one commit with.

```
# 1. Update version in Cargo.toml
# 2. Update lockfile
cargo update -p skia-safe
```

- [x] Fix
- [ ] Feature
- [ ] Codestyle
- [ ] Refactor
- [ ] Other

- [x] No

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2276,9 +2276,9 @@ dependencies = [
 
 [[package]]
 name = "skia-bindings"
-version = "0.39.1"
+version = "0.40.2"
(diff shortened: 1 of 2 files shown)


last-partizan:
Also, i've added `caret` for version, so we don't need to touch it anymore when updating minor versions.

shaunsingh:
Sorry about that, I forgot the checksums again. Thanks for taking a look

PR #801, opened 2021-07-06
Fix build m1

last-partizan:
Attempt to fix m1 build

- [x] Fix
- [ ] Feature
- [ ] Codestyle
- [ ] Refactor
- [ ] Other

- [x] No

### DIFF
--- a/.github/workflows/snap.yaml
+++ b/.github/workflows/snap.yaml
@@ -1,6 +1,8 @@
 name: Snap
 
-on: [push]
+on:
+  push:
+    branches: [ main ]
 
 jobs:
   snap:
@@ -9,12 +11,12 @@ jobs:
     steps:
     - name: Check out Git repository
       uses: actions/checkout@v2
-    
+
     - uses: snapcore/action-build@v1
       env:
         SNAPCRAFT_BUILD_ENVIRONMENT_MEMORY: 6G
       id: snapcraft
-    
+
     - uses: actions/upload-artifact@v2
       with:
         name: snap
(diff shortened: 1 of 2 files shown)


.github/workflows/snap.yaml
```diff
@@ -1,6 +1,8 @@
 name: Snap
 
-on: [push]
+on:
+  push:
+    branches: [ main ]
```
last-partizan, line 5: @Kethku i was fixing build for m1, when i noticed this. Bulding and publishing snaps on each push.

That's probably not how it supposed to work. I changed this to publish snaps only for main branch.

Kethku: I think you're likely correct. Good catch

Kethku:
I have no reason to assume this doesn't fix things, but I really don't have the expertise to know either way. @last-partizan I'll defer to your judgement. THis looks good to me.

last-partizan:
Great, merging thins.

And it will publish snaps after merge.

j4qfrost:
This was my bad, the on push was there because I was testing my changes.

last-partizan:
Btw, pipeline completed successfully, but latest release here: https://snapcraft.io/neovide is still from `13 June 2021`.

Any ideas why?

shaunsingh:
I believe that release is still at the stable version, there should be a flag to install the latest version published by the actions 

I believe it was `--beta`

j4qfrost:
I have to manually add the snaps in. Not sure there's a way around that.
Stable releases should be added manually anyway.

PR #806, opened 2021-07-07
refactor(startup): Use SETTINGS to store window geometry

last-partizan:
While i was researching neovim startup sequence, i refactored it a little.

- Removed multiple calls to functions getting window size
- Now there is only one call, when parsing command line arguments
- After this value is used from SETTINGS

I think this way it is better. @Kethku what do you think?

---------------------

- Refactor

- No

Btw, i suppose no one objects i'm using just lists instead of checkboxes in this template? Github is treating checkboxes as "tasks", and i don't like having "2 of 7 tasks" on my PRs :)

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -18,7 +18,6 @@ use tokio::runtime::Runtime;
 
 use crate::channel_utils::*;
 use crate::settings::*;
-use crate::window::window_geometry_or_default;
 use crate::{cmd_line::CmdLineSettings, error_handling::ResultPanicExplanation};
 pub use events::*;
 use handler::NeovimHandler;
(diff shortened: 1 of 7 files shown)


shaunsingh:
> Btw, i suppose no one objects i'm using just lists instead of checkboxes in this template? Github is treating checkboxes as "tasks", and i don't like having "2 of 7 tasks" on my PRs :)

Off topic but I'm thinking of reverting that as well. The checkboxes were added a while back but I didn't realize it gets picked up as a task

Kethku:
@last-partizan I think this is a great change. My only suggestion would be to move the geometry related functions into their own file. But its up to you.

last-partizan:
Hmmm.

That's only one "get_window_geometry", and it's only parsing command line argument. But, wait a minute.

I see settings/from_value.rs, and there is parsing options, but for simple types.

Can we do something like `impl FromValue from WindowGeometry`? That would be great. I'll try it!

last-partizan:
Idea was great, but i haven't been able to make it work. Or it's not possible to do it the way i wanted.

Maybe another time.

But i splitter parsing function into separate file and it looks better now.

Kethku:
Looks good to me. Merging.

Issue #387, opened 2020-10-19
Incosistent font size with other neovim GUIs


kishikaisei:
So basically Neovide renders font smaller than nvim-qt, or goneovim:
![image](https://user-images.githubusercontent.com/1894123/96498915-fb6d7080-124c-11eb-8918-ba33faabb77b.png)
Goneovim, nvim-qt and Neovide respectively.

felipec:
Just installed neovide and this is literally the first thing I noticed.

Kethku:
Not sure why this would be happening given that we are delegating to skia to handle the end result font rendering and the font size is directly taken from the guifont variable.

felipec:
The DPI setting? I am using 120, not 96.

Kethku:
Possible, but not sure. I think I can see a font size difference on my machine as well. In the meantime, neovide declares a global neovide variable on startup, so you could look for that and conditionally set the font size in your config until we figure out whats going on

Kethku:
I believe this was fixed by @last-partizan

last-partizan:
It isn't fixed yet, but we're closer to fix than before, i'll take a look :)

![image](https://user-images.githubusercontent.com/301015/125034457-45c3a580-e099-11eb-8a3c-1cc3c808ee66.png)

last-partizan:
Btw, we're assuming fonts are in "pixels", but maybe they're actually in "points"? That explains difference, we need to convert our scaling factor to account for this conversion.

last-partizan:
Anyone interested please check `fix-scaling` branch.

PR #814, opened 2021-07-09
fix(fonts): Fix font scaling (points to pixels conversion)

last-partizan:
Fixes font scaling from #387

Fonts are actually in points, not pixels. So we need to convert them.

image

- Fix

- No

### DIFF
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -316,7 +316,7 @@ impl CachingShaper {
 fn points_to_pixels(value: f32) -> f32 {
     // Fonts in neovim are using points, not pixels.
     //
-    // Skia docs is incorrectly statint it uses points, but uses pixels:
+    // Skia docs is incorrectly stating it uses points, but uses pixels:
     // https://api.skia.org/classSkFont.html#a7e28a156a517d01bc608c14c761346bf
     // https://github.com/mono/SkiaSharp/issues/1147#issuecomment-587421201
     //
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -30,7 +30,8 @@ pub struct CachingShaper {
 }
 
 impl CachingShaper {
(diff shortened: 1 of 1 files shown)


src/renderer/fonts/caching_shaper.rs
```diff
@@ -311,3 +312,18 @@ impl CachingShaper {
         self.blob_cache.get(&key).unwrap()
     }
 }
+
+fn points_to_pixels(value: f32) -> f32 {
+    // Fonts in neovim are using points, not pixels.
+    //
+    // Skia docs is incorrectly statint it uses points, but uses pixels:
```
last-partizan, line 319:
```suggestion
    // Skia docs is incorrectly stating it uses points, but uses pixels:
```

src/renderer/fonts/caching_shaper.rs
```diff
@@ -311,3 +312,18 @@ impl CachingShaper {
...
+
+fn points_to_pixels(value: f32) -> f32 {
+    // Fonts in neovim are using points, not pixels.
+    //
+    // Skia docs is incorrectly stating it uses points, but uses pixels:
+    // https://api.skia.org/classSkFont.html#a7e28a156a517d01bc608c14c761346bf
+    // https://github.com/mono/SkiaSharp/issues/1147#issuecomment-587421201
+    //
+    // So, we need to convert points to pixels.
+    let pixels_per_inch = 96.0;
```
Kethku, line 324: This feels suspicious to me. Pixels per inch are dependent on the monitor dpi settings aren't they?

last-partizan: I think yes, but here we only care about converting pt to px, so it's unrelated. Scaling according to DPI happens later, when we multiply by scale_factor from winit.

I'll add prefix 'default' here to avoid confision, and small note.

last-partizan: Probably better only comment, prefix would make variable names too lonk.

Kethku:
Awesome find btw. I wouldn't have caught this. This looks good to me.

last-partizan:
Merged!

PR #816, opened 2021-07-10
Remember window dimensions #622

bbli:
I couldn't figure out how to reopen the previous pull request(#665), so I just made a new one lol. Anyways, after the refactor done in #806, it looks like I can just check for the previous window setting while setting up global DEFAULT_WINDOW_GEOMETRY variable?

- [x] Feature

- [x] No

### DIFF
--- a/src/settings/window_geometry.rs
+++ b/src/settings/window_geometry.rs
@@ -12,13 +12,13 @@ pub struct WindowGeometry {
 }
 
 #[cfg(unix)]
-const SETTINGS_PATH_EXTENSION: &str = ".local/share/nvim/neovide-settings.json";
+const SETTINGS_PATH: &str = ".local/share/nvim/neovide-settings.json";
 #[cfg(windows)]
-const SETTINGS_PATH_EXTENSION: &str = "AppData/Local/nvim-data/neovide-settings.json";
+const SETTINGS_PATH: &str = "AppData/Local/nvim-data/neovide-settings.json";
 
 fn neovim_std_datapath() -> PathBuf {
     let mut settings_path = dirs::home_dir().unwrap();
-    settings_path.push(SETTINGS_PATH_EXTENSION);
+    settings_path.push(SETTINGS_PATH);
     settings_path
 }
 
@@ -26,10 +26,10 @@ pub fn try_to_load_last_window_size() -> Result<WindowGeometry, String> {
     let settings_path = neovim_std_datapath();
     let serialized_size = std::fs::read_to_string(&settings_path).map_err(|e| e.to_string())?;
 
-    let saved_window_size: WindowGeometry =
+    let deserialize_size: WindowGeometry =
         serde_json::from_str(&serialized_size).map_err(|e| e.to_string())?;
-    log::debug!("Loaded Window Size: {:?}", saved_window_size);
-    Ok(saved_window_size)
+    log::debug!("Loaded Window Size: {:?}", deserialize_size);
+    Ok(deserialize_size)
 }
 pub const DEFAULT_WINDOW_GEOMETRY: WindowGeometry = WindowGeometry {
     width: 100,
@@ -50,9 +50,9 @@ pub fn maybe_save_window_size(window_size: PhysicalSize<u32>, renderer: &Rendere
     };
 
     let settings_path = neovim_std_datapath();
-    let se = serde_json::to_string(&saved_window_size).unwrap();
-    //log::debug!("Saved Window Size: {}", se);
-    std::fs::write(settings_path, se).unwrap();
+    let serialized_size = serde_json::to_string(&saved_window_size).unwrap();
+    log::debug!("Saved Window Size: {}", serialized_size);
+    std::fs::write(settings_path, serialized_size).unwrap();
 }
 
 pub fn parse_window_geometry(geometry: Option<String>) -> Result<WindowGeometry, String> {
--- a/src/settings/window_geometry.rs
+++ b/src/settings/window_geometry.rs
@@ -26,9 +26,10 @@ pub fn try_to_load_last_window_size() -> Result<WindowGeometry, String> {
     let settings_path = neovim_std_datapath();
     let serialized_size = std::fs::read_to_string(&settings_path).map_err(|e| e.to_string())?;
 
-    let de: WindowGeometry = serde_json::from_str(&serialized_size).map_err(|e| e.to_string())?;
-    log::debug!("Loaded Window Size: {:?}", de);
-    Ok(de)
+    let saved_window_size: WindowGeometry =
+        serde_json::from_str(&serialized_size).map_err(|e| e.to_string())?;
+    log::debug!("Loaded Window Size: {:?}", saved_window_size);
+    Ok(saved_window_size)
 }
 pub const DEFAULT_WINDOW_GEOMETRY: WindowGeometry = WindowGeometry {
     width: 100,
--- a/src/settings/window_geometry.rs
+++ b/src/settings/window_geometry.rs
@@ -2,61 +2,49 @@ use crate::renderer::Renderer;
 use crate::settings::SETTINGS;
 use crate::window::WindowSettings;
 use glutin::dpi::PhysicalSize;
+use serde::{Deserialize, Serialize};
 use std::path::PathBuf;
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 pub struct WindowGeometry {
     pub width: u64,
     pub height: u64,
 }
 
-pub fn try_to_load_last_window_size() -> Result<WindowGeometry, String> {
+#[cfg(unix)]
+const SETTINGS_PATH_EXTENSION: &str = ".local/share/nvim/neovide-settings.json";
+#[cfg(windows)]
+const SETTINGS_PATH_EXTENSION: &str = "AppData/Local/nvim-data/neovide-settings.json";
+
+fn neovim_std_datapath() -> PathBuf {
     let mut settings_path = dirs::home_dir().unwrap();
-    settings_path.push(".local/share/nvim/neovide-settings.txt");
+    settings_path.push(SETTINGS_PATH_EXTENSION);
+    settings_path
+}
+
+pub fn try_to_load_last_window_size() -> Result<WindowGeometry, String> {
+    let settings_path = neovim_std_datapath();
     let serialized_size = std::fs::read_to_string(&settings_path).map_err(|e| e.to_string())?;
 
-    let de: PhysicalSize<u32> =
-        serde_json::from_str(&serialized_size).map_err(|e| e.to_string())?;
+    let de: WindowGeometry = serde_json::from_str(&serialized_size).map_err(|e| e.to_string())?;
     log::debug!("Loaded Window Size: {:?}", de);
-    Ok(WindowGeometry {
-        width: de.width as u64,
-        height: de.height as u64,
-    })
+    Ok(de)
 }
-#[cfg(unix)]
-fn neovim_std_datapath() -> PathBuf {
-    let mut settings_path = dirs::home_dir().unwrap();
-    settings_path.push(".local/share/nvim/neovide-settings.txt");
-    settings_path
-}
-#[cfg(windows)]
-fn neovim_std_datapath() -> PathBuf {
-    let mut settings_path = dirs::home_dir().unwrap();
-    settings_path.push("AppData/Local/nvim-data/neovide-settings.txt");
-    settings_path
-}
-
-const PRIVATE_WINDOW_GEOMETRY: WindowGeometry = WindowGeometry {
+pub const DEFAULT_WINDOW_GEOMETRY: WindowGeometry = WindowGeometry {
     width: 100,
     height: 50,
 };
 
-lazy_static! {
-    pub static ref DEFAULT_WINDOW_GEOMETRY: WindowGeometry = try_to_load_last_window_size()
-        .or::<String>(Ok(PRIVATE_WINDOW_GEOMETRY))
-        .unwrap();
-}
-
 pub fn maybe_save_window_size(window_size: PhysicalSize<u32>, renderer: &Renderer) {
-    let saved_window_size = if SETTINGS.get::<WindowSettings>().remember_dimension {
-        PhysicalSize::<u32> {
-            width: (window_size.width as f32 / renderer.font_width as f32) as u32,
-            height: (window_size.height as f32 / renderer.font_height as f32) as u32,
+    let saved_window_size = if SETTINGS.get::<WindowSettings>().remember_window_size {
+        WindowGeometry {
+            width: (window_size.width as f32 / renderer.font_width as f32) as u64,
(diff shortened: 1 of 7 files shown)


src/window/settings.rs
```diff
@@ -8,6 +8,7 @@ pub struct WindowSettings {
     pub fullscreen: bool,
     pub iso_layout: bool,
     pub scroll_dead_zone: f32,
+    pub remember_dimension: bool,
```
last-partizan, line 11:
```suggestion
    pub remember_window_size: bool,
```

src/settings/window_geometry.rs
```diff
@@ -1,16 +1,73 @@
...
+    settings_path
+}
+#[cfg(windows)]
+fn neovim_std_datapath() -> PathBuf {
+    let mut settings_path = dirs::home_dir().unwrap();
+    settings_path.push("AppData/Local/nvim-data/neovide-settings.txt");
+    settings_path
+}
+
+const PRIVATE_WINDOW_GEOMETRY: WindowGeometry = WindowGeometry {
```
last-partizan, line 39: I think it's better to leave this as `DEFAULT_WINDOW_GEOMETRY`, and make it private.

And instead of lazy static, make function `get_default_window_geometry`, and use it in settings.

src/settings/window_geometry.rs
```diff
@@ -1,16 +1,73 @@
...
+use std::path::PathBuf;
+
+#[derive(Debug, Clone, Copy)]
 pub struct WindowGeometry {
     pub width: u64,
     pub height: u64,
 }
 
-pub const DEFAULT_WINDOW_GEOMETRY: WindowGeometry = WindowGeometry {
+pub fn try_to_load_last_window_size() -> Result<WindowGeometry, String> {
```
last-partizan, line 13: I think it's better to rename this to `maybe_load_saved_window_size`.

src/settings/window_geometry.rs
```diff
@@ -1,16 +1,73 @@
...
+    let mut settings_path = dirs::home_dir().unwrap();
```
last-partizan, line 14: This should use neovim_std_datapath

src/settings/window_geometry.rs
```diff
@@ -1,16 +1,73 @@
...
+
+pub fn maybe_save_window_size(window_size: PhysicalSize<u32>, renderer: &Renderer) {
+    let saved_window_size = if SETTINGS.get::<WindowSettings>().remember_dimension {
+        PhysicalSize::<u32> {
+            width: (window_size.width as f32 / renderer.font_width as f32) as u32,
+            height: (window_size.height as f32 / renderer.font_height as f32) as u32,
+        }
+    } else {
+        PhysicalSize::<u32> {
+            width: PRIVATE_WINDOW_GEOMETRY.width as u32,
```
last-partizan, line 58: I don't think we should save default values.

If remember_window_size is false, this function should do nothing.

src/settings/window_geometry.rs
```diff
@@ -1,16 +1,73 @@
...
 
+lazy_static! {
+    pub static ref DEFAULT_WINDOW_GEOMETRY: WindowGeometry = try_to_load_last_window_size()
+        .or::<String>(Ok(PRIVATE_WINDOW_GEOMETRY))
+        .unwrap();
+}
+
+pub fn maybe_save_window_size(window_size: PhysicalSize<u32>, renderer: &Renderer) {
+    let saved_window_size = if SETTINGS.get::<WindowSettings>().remember_dimension {
+        PhysicalSize::<u32> {
```
last-partizan, line 52: Better make this WindowGeometry.

last-partizan (changes requested):

src/settings/window_geometry.rs
```diff
@@ -1,16 +1,73 @@
...
+    log::debug!("Loaded Window Size: {:?}", de);
+    Ok(WindowGeometry {
+        width: de.width as u64,
+        height: de.height as u64,
+    })
+}
+#[cfg(unix)]
+fn neovim_std_datapath() -> PathBuf {
+    let mut settings_path = dirs::home_dir().unwrap();
+    settings_path.push(".local/share/nvim/neovide-settings.txt");
```
last-partizan, line 29: Let's store this in const outside of function, so we don't have to repeat function two times.

And let's use '.json' as extension.

src/settings/window_geometry.rs
bbli: The problem with that is that after the user changes the "remember_window_size" variable back to false, neovide will still use the previously saved window setting instead of the default.

last-partizan: Oh, i see.

And we probably can't check that variable here, becouse remember_window_size is not initialized at the moment of reading cmdline settings?

Maybe it's better to move it to env variable for the start, and then if it works fine, remove variable and just do that by default. Other apps just do it that way.

Let's wait what @Kethku has to say about this.

src/settings/window_geometry.rs
bbli: Yeah sounds good. I think I chose to do that because previously the parsing window geometry function was called in multiple places in the codebase.

src/settings/window_geometry.rs
bbli: It has different semantics from `maybe_save_window_size` though. There, we either save or not save, whereas in `try_to_load_last_window_size` we always load, except that the loading may fail

src/settings/window_geometry.rs
bbli: ohh yeah, thanks for catching that!

src/settings/window_geometry.rs
bbli: The issue with that approach is if the user changes `remember_window_size` back to false, they will still get the previously saved window size instead of the default upon restart.

src/settings/window_geometry.rs
bbli: Kk

src/settings/window_geometry.rs
last-partizan: But it's better to make it behave just as `maybe_save_window_size`, check the flag, maybe load.

bbli: Ohh so while that would be ideal, the flag won't be set before this function gets called, as you have mentioned in our discussion below.

src/settings/window_geometry.rs
```diff
@@ -1,16 +1,63 @@
...
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 pub struct WindowGeometry {
     pub width: u64,
     pub height: u64,
 }
 
+#[cfg(unix)]
+const SETTINGS_PATH_EXTENSION: &str = ".local/share/nvim/neovide-settings.json";
+#[cfg(windows)]
+const SETTINGS_PATH_EXTENSION: &str = "AppData/Local/nvim-data/neovide-settings.json";
```
Kethku, line 17: Should we use something like the dirs package https://docs.rs/dirs/3.0.2/dirs/fn.data_dir.html? It adds a dependency (whats another one right 😂) but at least this way we don't have to have platform specific strings...

Kethku: An a related point, why for unix is it nvim when on windows its nvim-data?

src/settings/window_geometry.rs
```diff
@@ -1,16 +1,63 @@
...
+use serde::{Deserialize, Serialize};
+use std::path::PathBuf;
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 pub struct WindowGeometry {
     pub width: u64,
     pub height: u64,
 }
 
+#[cfg(unix)]
```
Kethku, line 14: The feature of a persistent neovide settings file is interesting. Couple of questions (none blocking this):

1. Should it be user readable and documented? Maybe this could be a more approachable way/place to specify multigrid/other features which can't be represented in the neovim config?
2. It feels more general purpose than just window size and sort fits somewhere in the settings module. Would it make sense to expose in a more general way for the rest of the components?

src/settings/window_geometry.rs
```diff
@@ -1,16 +1,63 @@
...
+    let mut settings_path = dirs::home_dir().unwrap();
+    settings_path.push(SETTINGS_PATH_EXTENSION);
+    settings_path
+}
+
+pub fn try_to_load_last_window_size() -> Result<WindowGeometry, String> {
+    let settings_path = neovim_std_datapath();
+    let serialized_size = std::fs::read_to_string(&settings_path).map_err(|e| e.to_string())?;
+
+    let de: WindowGeometry = serde_json::from_str(&serialized_size).map_err(|e| e.to_string())?;
```
Kethku, line 29: nit: although it makes sense that de represents deserialized value. I try to avoid single or double character variable names. Its not that much more effort to write it out, and I think it makes the codebase more approachable.

src/settings/window_geometry.rs
bbli: nvim_datapath

According to the nvim help manual, that's what they chose the paths to be lol.

Look like using the dirs package will produce a separate save location compared to the manual. But like I said in the previous pull request, it's just a soft dependency, so I'm fine with either method.

src/settings/window_geometry.rs
bbli: kk will fix

src/settings/window_geometry.rs
bbli: > Should it be user readable and documented?
> which can't be represented in the neovim config
Ohh, so sorta as a way to set "defaults" for the command line arguments? Depending on the most common usecase, perhaps put this into a separate file. Do we want "defaults" to take priority or "last adjusted settings"?

> It feels more general purpose than just window size
> Would it make sense to expose in a more general way for the rest of the components?
Ohh yeah definitively!  Without more concrete things to save, I am not sure how to get started programming this in general though. An issue that may arise is having access to all nesscary objects at time of exit. Right now, I'm relying on the fact that the `window_wrapper` object is conveniently in scope during the exit.

Kethku: Ok I think it makes sense to just merge this as is. Was just brainstorming a bit

src/settings/window_geometry.rs
Kethku: Welp looks like theres not really a better way to do this, so just merging as is

src/settings/window_geometry.rs
Kethku: I think I'm fine with this behavior until we restructure the app to wait for the neovim connection before popping the window.

Kethku:
@last-partizan I think this is fine to merge, but I'll defer to your active review. Let me know if you think this is ready

src/settings/window_geometry.rs
```diff
@@ -1,16 +1,64 @@
...
+        }
+    } else {
+        WindowGeometry {
+            width: DEFAULT_WINDOW_GEOMETRY.width as u64,
+            height: DEFAULT_WINDOW_GEOMETRY.height as u64,
+        }
+    };
+
+    let settings_path = neovim_std_datapath();
+    let se = serde_json::to_string(&saved_window_size).unwrap();
```
last-partizan, line 53: It's better to use full variable name here.

src/settings/window_geometry.rs
```diff
@@ -1,16 +1,64 @@
...
+    //log::debug!("Saved Window Size: {}", se);
```
last-partizan, line 54:
```suggestion
    log::debug!("Saved Window Size: {}", se);
```

Maybe remove both, or leave both log:debug? no need to leave it commented.

src/settings/window_geometry.rs
```diff
@@ -1,16 +1,64 @@
...
+use std::path::PathBuf;
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 pub struct WindowGeometry {
     pub width: u64,
     pub height: u64,
 }
 
+#[cfg(unix)]
+const SETTINGS_PATH_EXTENSION: &str = ".local/share/nvim/neovide-settings.json";
```
last-partizan, line 15: This is not "extension", it's just settings_path.

```suggestion
const SETTINGS_PATH: &str = ".local/share/nvim/neovide-settings.json";
```

last-partizan (changes requested):

src/settings/window_geometry.rs
bbli: ohh yeah, I should have caught this after Kethku mentioned it above.

src/settings/window_geometry.rs
bbli: Kk

src/settings/window_geometry.rs
bbli: Hmm sorta? I already have a variable called `settings_path` defined in
https://github.com/Kethku/neovide/pull/816/files#diff-ccd3e532d23ff949faa2bcac192d9441254cd1dbeedb5aebff2214f824f8def0R19-R23

which appends the above string to the home directory to create the absolute path.

last-partizan: I think that's fine.

last-partizan: (i mean, having local settings_path variable and global SETTINGS_PATH with same name, so, change it)

bbli: ohh ok. Yea, sure let me do that.

last-partizan (approved):
Looks good to me.

@Kethku i think this can be merged.

clason:
Hmm, that doesn't seem to work for me on macOS; windows always open at default size, and `neovide-settings.json` are always `{"width":100,"height":50}`

Am I doing something wrong? (And is Neovim's config directory the right place? Shouldn't neovide have its own data directory?)

clason:
Also, is Neovim's data directory the right place? Shouldn't neovide have its own data directory? And consistently use these settings for configuration (Font, VFX etc.)?

I feel like this new config file was a bit snuck in under the radar here ;)

bbli:
Ohh yeah, forgot to put documentation of this in the README :sweat_smile: . Remembering the window dimensions is turned off by default. If you add `let neovide_remember_window_size = 1` in your vimrc/init.vim, it should work.

bbli:
@Kethku , should I write this up in the README or the wiki?

Kethku:
I think probably the wiki. We can point people in that direction if needed

Kethku:
@clason I'm interested in this feedback. Are you saying neovide shouldn't create a persistence file, should only create a persistence file if the setting is set, or should create the file and allow users to customize it?

clason:
I'm saying:

* neovide _should_ create the file
* neovide should create the file in its _own_ data directory (rather than neovim's)
* neovide should (optionally) use this file for configuration instead of `[g]init.vim`

The last point is maybe a bit controversial as it mixes storage and config, but 

* since it's JSON it should be safe to write as well as read
* you can use it for manually setting the geometry if the persistent setting is not set (and if it is, you are _asking_ for neovide to override the geometry)

bbli:
> The last point is maybe a bit controversial as it mixes storage and config
Hmm yeah, perhaps have two files then, one for storage and for configuration? 

Also @Kethku , do I need permission to edit the wiki? It says "Permission denied" when I try to push

Kethku:
@bbli I just sent you an invite. Let me know if that doesn't work

bbli:
yup, works fine now

PR #817, opened 2021-07-10
Dynamic scale factor updates

last-partizan:
This is draft attemmt to support `ScaleFactorChanged` event.

I use target branch `fix-scaling`, to show only relevant changes in this branch. I'll change it to main when `fix-scaling` is merged.

Here i refactored CachingShaper to always include `options`, and not `Option<FontOptions>`. And removed Option from few other parts, becouse i was lost in `unwrap`s and `map`s. And this is probably works as expected.

Font size is changing when i change scaling in display settings, but window size does not.

Please take a look at this, try it, maybe you have some ideas for how to make window scale too.

Please, pay close attention to any rust-specific features i try to use here, (like pointers, and dereferencing), i usually work with python, and here i write code with help from langserver, rust-compiler and google.

- Fix
- Refactor

- No

### DIFF
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -43,8 +43,7 @@ impl CachingShaper {
 
     fn current_font_pair(&mut self) -> Arc<FontPair> {
         let default_key = FontKey::default();
-
-        let font_key = self.options.as_font_key();
+        let font_key = FontKey::from(&self.options);
 
         if let Some(font_pair) = self.font_loader.get_or_load(&font_key) {
             return font_pair;
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -154,8 +154,8 @@ impl CachingShaper {
 
             // Add default font
             font_fallback_keys.push(FontKey {
-                italic,
-                bold,
+                italic: self.options.italic || italic,
+                bold: self.options.bold || bold,
                 font_selection: FontSelection::Default,
             });
 
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -28,8 +28,7 @@ pub struct CachingShaper {
 }
 
 impl CachingShaper {
-    pub fn new(device_scale_factor: f32) -> CachingShaper {
-        let scale_factor = points_to_pixels(device_scale_factor);
+    pub fn new(scale_factor: f32) -> CachingShaper {
         let options = FontOptions::default();
         let font_size = options.size * scale_factor;
         CachingShaper {
@@ -58,9 +57,9 @@ impl CachingShaper {
         self.options.size * self.scale_factor
     }
 
-    pub fn update_scale_factor(&mut self, device_scale_factor: f32) {
-        trace!("Device scale_factor changed: {:.2}", device_scale_factor);
-        self.scale_factor = device_scale_factor * points_to_pixels(device_scale_factor);
+    pub fn update_scale_factor(&mut self, scale_factor: f32) {
+        trace!("scale_factor changed: {:.2}", scale_factor);
+        self.scale_factor = scale_factor;
         self.reset_font_loader();
     }
 
@@ -282,26 +281,4 @@ impl CachingShaper {
 
         self.blob_cache.get(&key).unwrap()
     }
-}
-
-fn points_to_pixels(value: f32) -> f32 {
-    // Fonts in neovim are using points, not pixels.
-    //
-    // Skia docs is incorrectly stating it uses points, but uses pixels:
-    // https://api.skia.org/classSkFont.html#a7e28a156a517d01bc608c14c761346bf
-    // https://github.com/mono/SkiaSharp/issues/1147#issuecomment-587421201
-    //
-    // So, we need to convert points to pixels.
-    //
-    // In reality, this depends on DPI/PPI of monitor, but here we only care about converting
-    // from points to pixels, so this is standard constant values.
-    let pixels_per_inch = 96.0;
-    let points_per_inch = 72.0;
-    // On macos points == pixels
-    #[cfg(target_os = "macos")]
-    let points_per_inch = 96.0;
-
-    let pixels_per_point = pixels_per_inch / points_per_inch;
-
-    value * pixels_per_point
 }
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -12,8 +12,6 @@ use unicode_segmentation::UnicodeSegmentation;
 use super::font_loader::*;
 use super::font_options::*;
 
-const DEFAULT_FONT_SIZE: f32 = 14.0;
-
 #[derive(new, Clone, Hash, PartialEq, Eq, Debug)]
 struct ShapeKey {
     pub cells: Vec<String>,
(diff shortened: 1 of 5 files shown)


src/renderer/fonts/font_loader.rs
```diff
@@ -54,6 +54,16 @@ pub struct FontKey {
     pub font_selection: FontSelection,
 }
 
+impl FontKey {
+    pub fn default() -> FontKey {
```
smolck, line 58: There’s a trait for this, see https://doc.rust-lang.org/std/default/trait.Default.html:

```rust
impl Default for FontKey {
    fn default() -> Self {
      // …
    }
}
```

smolck:
> Here i refactored CachingShaper to always include options, and not Option<FontOptions>.

Why did you do this? I’m not familiar with the codebase enough to know why it was an Option type in the first place, so not sure if it’s really unnecessary or not . . .

> And removed Option from few other parts, becouse i was lost in unwraps and maps. And this is probably works as expected.

If they really are unnecessary, that’s fine. But in some cases having a `None` value is a valid thing, in which case you’ll want to leave it as Option. But I haven’t really studied the code at this point so this may not be relevant here.

last-partizan:
>> Here i refactored CachingShaper to always include options, and not Option.
>
> Why did you do this? I’m not familiar with the codebase enough to know why it was an Option type in the first place, so not sure if it’s really unnecessary or not . . .

>>  And removed Option from few other parts, becouse i was lost in unwraps and maps. And this is probably works as expected.
> 
> If they really are unnecessary, that’s fine. But in some cases having a None value is a valid thing, in which case you’ll want to leave it as Option. But I haven’t really studied the code at this point so this may not be relevant here.


Having default values instead of Option makes a lot of code unnecessary, here is good example:

```diff
-        self.options
-            .as_ref()
-            .map(|options| options.size)
-            .unwrap_or(DEFAULT_FONT_SIZE)
-            * self.scale_factor
+        self.options.size * self.scale_factor
```

And if unwrapped value has optional field, that's one more level of unwraps or maps.

src/renderer/fonts/font_loader.rs
last-partizan: Thanks!

src/renderer/fonts/caching_shaper.rs
```diff
@@ -68,26 +56,30 @@ impl CachingShaper {
...
+
+    pub fn update_scale_factor(&mut self, device_scale_factor: f32) {
+        println!("Update scale_factor: {}", device_scale_factor);
+        self.scale_factor = device_scale_factor * points_to_pixels(device_scale_factor);
+        self.reset_font_loader();
     }
 
     pub fn update_font(&mut self, guifont_setting: &str) -> bool {
-        let new_options = FontOptions::parse(guifont_setting, DEFAULT_FONT_SIZE);
+        println!("Updating font: {}", guifont_setting);
```
Kethku, line 69: use a trace here rather than println

src/renderer/fonts/caching_shaper.rs
```diff
@@ -68,26 +56,30 @@ impl CachingShaper {
...
-            true
-        } else {
-            false
-        }
+        true
+    }
+
+    fn reset_font_loader(&mut self) {
+        let font_size = self.options.size * self.scale_factor;
+        println!("Using font_size: {:.2}", font_size);
```
Kethku, line 79: same here. Trace rather than log

src/renderer/fonts/caching_shaper.rs
last-partizan: That's temporary code until this is draft. I'll remove all prints after this is working as expected.

src/renderer/mod.rs
```diff
@@ -87,12 +87,24 @@ impl Renderer {
...
-        }
+        self.shaper.update_font(guifont_setting);
+        self.update_font_dimensions();
+    }
+
+    fn update_font_dimensions(&mut self) {
+        let (font_width, font_height) = self.shaper.font_base_dimensions();
+        self.font_width = font_width;
+        self.font_height = font_height;
+        println!(
```
Kethku, line 104: also here. Trace rather than println (note: i realize this is probably for testing, but is likely useful for debugging purposes, so marking these in the review)

src/window/window_wrapper/mod.rs
```diff
@@ -159,6 +165,7 @@ impl GlutinWindowWrapper {
         let previous_size = self.saved_inner_size;
 
         if previous_size != current_size {
+            println!("Updating grid size: {:#?}", current_size);
```
Kethku, line 168: and here: trace rather than println

Kethku:
```suggestion
            trace!("Updating grid size: {:#?}", current_size);
```

src/renderer/mod.rs
Kethku:
```suggestion
        trace!(
```

.github/workflows/build.yml
```diff
@@ -163,11 +167,15 @@ jobs:
...
+    - name: Archive Linux artifact
+      run: |
+        tar czvf ./target/release/neovide.tar.gz ./target/release/neovide
 
     - uses: actions/upload-artifact@v1
       with:
-        name: neovide-linux
-        path: ./target/release/neovide
+        name: neovide-linux.tar.gz
+        path: ./target/release/neovide.tar.gz
```
last-partizan, line 178: Hmm, i don't understand why are this changes here, i just synced branch with master :(

Please ignore this, i'll resolve it later.

clason:
Just to mention this here explicitly: the `fix-scaling` PR "broke" HiDPI (at least macOS Retina) screens; it looks like the internal pixel doubling is applied twice -- the font is now huge! (But renders fine at that large size, hence the scare quotes around "broke".)

I assume correct handling of Retina screens is part of this PR?

last-partizan:
@clason this PR does not touch scaling itself, only dynamic changing of scale factor.

after `fix-scaling` font size was expected to get bigger, becouse we converted points to pixels. What is your font size for neovide?

on linux before `fix-scaling` my font-size was 14 (but actually that was 11pt).

maybe on mac-os points and pixels are the same. If your font size was consistent with other apps, i'll make another PR to fix that.

clason:
Ah, I see, then this was my misconception. I don't configure Neovide at all, so it must be using some defaults (which ones?) that are less than reasonable on a HiDPI screen ;)

last-partizan:
You could run `neovide --log`, and then `grep 'scale' neovide_rCURRENT.log`

It should look like this:
```
INFO [neovide::window::window_wrapper] window created (scale_factor: 1.6041666666666667, font_size: 18x37)
```

(that's my 11pt, multiplied by 1.3 and with some extra height added)

clason:
```
INFO [neovide::window::window_wrapper] window created (scale_factor: 2, font_size: 23x46)
```
(macOS, external monitor at 2560x1440)

.github/workflows/build.yml
shaunsingh: Linux/windows builds are now distriubuted in a .tar.gz format as changed in https://github.com/Kethku/neovide/pull/822

last-partizan: That was becouse target branch was fix-scaling instead of main. Resolved.

last-partizan:
I replaced all "println" with "trace", and fixed bug with incorrect scaling.

@Kethku please check how it works for you.

Kethku:
This works perfectly on my machine. Merging.

Kethku:
Thanks for pushing this forward! Your changes have made a big difference for the stability of the app

PR #822, opened 2021-07-13
Put Windows artefact in zip archive

kradalby:
This PR adds an extra step to the Windows pipeline to put the Windows artefact in a `.zip` archive. 

A lot of places (specially companies) has proxies that outright blocks everything ending in `.exe` and this is a neat way to make it a bit easier to distribute from Github to those places.

- [ ] Fix
- [ ] Feature
- [ ] Codestyle
- [ ] Refactor
- [x] Other

- [x] Yes, please list breaking changes
- [ ] No

This change is potentially breaking if there exist update processes that links directly to the artefact, alternatively this could be an addition :)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,7 @@ jobs:
       run: |
         cargo build --release
         
-    - name: Zip Windows artifact
+    - name: Archive Windows artifact
       run: |
         powershell Compress-Archive ./target/release/neovide.exe ./target/release/neovide.zip 
 
@@ -167,11 +167,15 @@ jobs:
     - name: Build Release
       run: |
         cargo build --release
+        
+    - name: Archive Linux artifact
+      run: |
+        tar czvf ./target/release/neovide.tar.gz ./target/release/neovide
 
     - uses: actions/upload-artifact@v1
       with:
-        name: neovide-linux
-        path: ./target/release/neovide
+        name: neovide-linux.tar.gz
+        path: ./target/release/neovide.tar.gz
 
   build-m1:
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,11 +48,15 @@ jobs:
     - name: Build Release
       run: |
         cargo build --release
+        
+    - name: Zip Windows artifact
+      run: |
+        powershell Compress-Archive ./target/release/neovide.exe neovide.zip 
 
     - uses: actions/upload-artifact@v1
       with:
-        name: neovide-windows.exe
-        path: ./target/release/neovide.exe
+        name: neovide-windows.zip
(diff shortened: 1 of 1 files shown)


Kethku:
Is it worth doing the same for the other artifacts? Also although I'm sympathetic to the cause of working around company restrictions, I can't help but feel that this puts a minor but real extra burden on maintenance.

@j4qfrost @shaunsingh @last-partizan any of you have thoughts?

kradalby:
I agree that it adds another layer, it might be worth doing for Linux, my impression is that it is a quite common practive these days:
- [syncthing](https://github.com/syncthing/syncthing/releases/tag/v1.18.0) (or the [macOS app (https://github.com/syncthing/syncthing-macos/releases/tag/v1.17.0-1))
- [bat](https://github.com/sharkdp/bat/releases/tag/v0.18.2)
- [starship](https://github.com/starship/starship/releases/tag/v0.55.0)

j4qfrost:
Zipping up the file shouldn't be too big a hassle. I don't use a Windows workflow, but I think it's a lot better in the long term to provide a zip and tarball. I don't see any breakages off the top of my head. We could still provide the other download options alongside the zipped files.

Kethku:
Ok that makes sense to me. We've gotten requests to do similar for full releases before, just not actions.

If we're going to merge this I think it makes sense to put the mac and linux assets in tar.gz as well since the asymmetry feels odd as is

kradalby:
Give me a second, I will update the PR.

kradalby:
I have added a step, archiving linux into a `.tar.gz` archive. I left macOS as-is, I would say that `.dmg` is the desired format to publish already.

Is there a good way I can have the pipeline tested?

j4qfrost:
@kradalby I've looked into emulating the github actions pipeline, and it doesn't look like there's a good way to do that. Usually I just make a bunch of commits and just watch the runners. We squash and merge most of the time anyway, so there isn't too much clutter.

kradalby:
Ah ok, are any of you happy to give it a go? It is based on pipelines from other projects, I cant see any obvious typos right now, but in case.

j4qfrost:
@kradalby Sorry someone, took out the on PR hook for the github actions, so it didn't run automatically. @Kethku do you remember if there was a reason for this?

Kethku:
I think it may have been an attempt to stop the pipeline from running for snap builds that probably got over zealous

shaunsingh:
> any of you have thoughts?

I personally believe we should keep the .dmg format for macOS, we might run into issues with package managers in the future, especially homebrew which is annoying about this stuff. 

looks good for windows though

Kethku:
Okeydokey then this looks good to me. Merging.

PR #824, opened 2021-07-14
Fix Icon Metadata

shaunsingh:
Fixes https://github.com/Kethku/neovide/issues/823

- [x] Fix
- [ ] Feature
- [ ] Codestyle
- [ ] Refactor
- [ ] Other

- [x] No


Kethku:
good catch!

PR #828, opened 2021-07-15
fix(macos): Fix scaling on mac-os

last-partizan:
On MacOS points and pixels are the same.

@clason please check this

- Fix

- No


clason:
Yes, that restores the old behavior with a more reasonable font size. (It still looks more like 12 or even 13pt than 11pt, but that may be as desired by default.)

last-partizan:
Default is 14pt. 11pt is my custom configuration

clason:
Right, that checks out, then.

PR #833, opened 2021-07-18
Refactor keyboard manager

last-partizan:
This refactoring introduces no changes in behaviour, only moves some code:

- maybe_get_keybinding moved out of handle_event, this allows to call `send(UiCommand::Keyboard ..)` only once, amd make things easier to understand.
- moved private functions to botton of the file, this allows us to see importand parts first and details later.

- Codestyle
- Refactor

- No


Kethku:
This is a reasonable change imho. Thanks

PR #834, opened 2021-07-18
Make waylands app_id and x11s class overrideable

TimB87:
- [ ] Fix
- [x] Feature
- [ ] Codestyle
- [ ] Refactor
- [ ] Other

- [x] No

This adds two new command line options to neovide, nemely `-a, --appid <wayland_app_id>` and `-c, --class <x11_class>`, to make both overrideable by the user.
This tries to implement the feature that has been asked for in #704

many thanks for @last-partizan for helping me through this :)

### DIFF
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -17,8 +17,8 @@ pub struct CmdLineSettings {
     pub multi_grid: bool,
     pub maximized: bool,
     pub frameless: bool,
-    pub wayland_app_id: std::string::String,
-    pub x11_class: std::string::String,
+    pub wayland_app_id: String,
+    pub x11_wm_class: String,
 }
 
 impl Default for CmdLineSettings {
@@ -37,7 +37,7 @@ impl Default for CmdLineSettings {
             maximized: false,
             frameless: false,
             wayland_app_id: "Neovide".to_string(),
-            x11_class: "Neovide".to_string(),
+            x11_wm_class: "Neovide".to_string(),
         }
     }
 }
@@ -112,14 +112,12 @@ pub fn handle_command_line_arguments() -> Result<(), String> {
         )
         .arg(
             Arg::with_name("wayland_app_id")
-                .long("appid")
-                .short("a")
+                .long("wayland_app_id")
                 .takes_value(true)
         )
         .arg(
             Arg::with_name("x11_class")
-                .long("class")
-                .short("c")
+                .long("x11_class")
                 .takes_value(true)
         );
 
@@ -156,7 +154,7 @@ pub fn handle_command_line_arguments() -> Result<(), String> {
             .value_of("wayland_app_id")
             .unwrap_or(&"Neovide".to_owned().to_string())
             .to_string(),
-        x11_class: matches
+        x11_wm_class: matches
             .value_of("x11_class")
             .unwrap_or(&"Neovide".to_owned().to_string())
             .to_string(),
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -157,7 +157,7 @@ pub fn handle_command_line_arguments() -> Result<(), String> {
         x11_wm_class: matches
             .value_of("x11_wm_class")
             .unwrap_or("neovide")
-            .to_string()
+            .to_string(),
     });
     Ok(())
 }
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -36,8 +36,8 @@ impl Default for CmdLineSettings {
             multi_grid: false,
             maximized: false,
             frameless: false,
-            wayland_app_id: "Neovide".to_string(),
-            x11_wm_class: "Neovide".to_string(),
+            wayland_app_id: String::from("Neovide"),
+            x11_wm_class: String::from("Neovide"),
         }
     }
 }
@@ -112,12 +112,12 @@ pub fn handle_command_line_arguments() -> Result<(), String> {
         )
         .arg(
             Arg::with_name("wayland_app_id")
-                .long("wayland_app_id")
+                .long("wayland-appid")
                 .takes_value(true)
         )
         .arg(
-            Arg::with_name("x11_class")
-                .long("x11_class")
+            Arg::with_name("x11_wm_class")
+                .long("x11-class")
                 .takes_value(true)
         );
 
@@ -152,11 +152,11 @@ pub fn handle_command_line_arguments() -> Result<(), String> {
         geometry: parse_window_geometry(matches.value_of("geometry").map(|i| i.to_owned()))?,
         wayland_app_id: matches
             .value_of("wayland_app_id")
-            .unwrap_or(&"Neovide".to_owned().to_string())
+            .unwrap_or_default()
             .to_string(),
         x11_wm_class: matches
-            .value_of("x11_class")
-            .unwrap_or(&"Neovide".to_owned().to_string())
+            .value_of("x11_wm_class")
+            .unwrap_or_default()
             .to_string(),
     });
     Ok(())
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -117,7 +117,7 @@ pub fn handle_command_line_arguments() -> Result<(), String> {
         )
         .arg(
             Arg::with_name("x11_wm_class")
-                .long("x11-class")
+                .long("x11-wm-class")
                 .takes_value(true)
         );
 
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -2,6 +2,7 @@ use crate::settings::*;
 
 use clap::{App, Arg};
 
+
 #[derive(Clone, Debug)]
 pub struct CmdLineSettings {
     pub verbosity: u64,
@@ -76,7 +77,11 @@ pub fn handle_command_line_arguments() {
             .long("frameless")
             .help("Removes the window frame. NOTE: Window might not be resizable after this setting is enabled.")
         )
-        .arg(Arg::with_name("wsl").long("wsl").help("Run in WSL"))
+        .arg(
+            Arg::with_name("wsl")
+                .long("wsl")
+                .help("Run in WSL")
+        )
         .arg(
             Arg::with_name("remote_tcp")
                 .long("remote-tcp")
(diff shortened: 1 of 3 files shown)


src/cmd_line.rs
```diff
@@ -17,6 +17,8 @@ pub struct CmdLineSettings {
     pub multi_grid: bool,
     pub maximized: bool,
     pub frameless: bool,
+    pub wayland_app_id: std::string::String,
```
last-partizan, line 20:
```suggestion
    pub wayland_app_id: String,
```

You don't need that std::string ..., it's already included in https://doc.rust-lang.org/std/prelude/index.html

src/cmd_line.rs
```diff
@@ -17,6 +17,8 @@ pub struct CmdLineSettings {
...
+    pub x11_class: std::string::String,
```
last-partizan, line 21:
```suggestion
    pub x11_class: String,
```

src/cmd_line.rs
```diff
@@ -101,6 +109,18 @@ pub fn handle_command_line_arguments() -> Result<(), String> {
                 .takes_value(true)
                 .last(true)
                 .help("Specify Arguments to pass down to neovim"),
+        )
+        .arg(
+            Arg::with_name("wayland_app_id")
+                .long("appid")
```
last-partizan, line 115:
```suggestion
                .long("wayland-app-id")
```

I'd suggest using this format, and dropping short version. `-a` is too ambiguous.

src/cmd_line.rs
```diff
@@ -101,6 +109,18 @@ pub fn handle_command_line_arguments() -> Result<(), String> {
...
+                .short("a")
+                .takes_value(true)
+        )
+        .arg(
+            Arg::with_name("x11_class")
+                .long("class")
```
last-partizan, line 121:
```suggestion
                .long("x11-wm-class")
```

I'd sugges using this, and dropping short argument too.

src/cmd_line.rs
```diff
@@ -34,6 +36,8 @@ impl Default for CmdLineSettings {
             multi_grid: false,
             maximized: false,
             frameless: false,
+            wayland_app_id: "Neovide".to_string(),
+            x11_class: "Neovide".to_string(),
```
last-partizan, line 40:
```suggestion
            x11_wm_class: "Neovide".to_string(),
```

src/window/window_wrapper/mod.rs
```diff
@@ -222,7 +222,13 @@ pub fn start_loop(
         .with_decorations(!SETTINGS.get::<CmdLineSettings>().frameless);
 
     #[cfg(target_os = "linux")]
-    let winit_window_builder = winit_window_builder.with_app_id("Neovide".to_string());
+    let winit_window_builder = winit_window_builder
+        .with_app_id(SETTINGS.get::<CmdLineSettings>().wayland_app_id.to_string());
```
last-partizan, line 226: Try to get rid of this `to_string()` conversion. Make `wayland_app_id` of that type which is expected in with_app_id call.

src/window/window_wrapper/mod.rs
```diff
@@ -222,7 +222,13 @@ pub fn start_loop(
...
+    #[cfg(target_os = "linux")]
+    let winit_window_builder = winit_window_builder.with_class(
+        SETTINGS.get::<CmdLineSettings>().x11_class.to_string(),
+        SETTINGS.get::<CmdLineSettings>().x11_class.to_string(),
```
last-partizan, line 230: with_class takes two args, `class` and `instance`, i think we should customize only `class`, and leave other as default "neovide".

Also, i think you can avoid having to prefix both calls with `#[cfg(target_os = "linux")]`, if you put this calls inside block, like this

```
#[cfg(target_os = "linux")]
{
  let winit_...
  let winit_...
}
```

last-partizan (changes requested):

src/window/window_wrapper/mod.rs
TimB87: this block does not work for me

last-partizan:
```
    #[cfg(target_os = "linux")]
    let winit_window_builder = winit_window_builder
        .with_app_id(SETTINGS.get::<CmdLineSettings>().wayland_app_id)
        .with_class(
            SETTINGS.get::<CmdLineSettings>().x11_wm_class,
            "Neovide".to_string(),
        );
```

Try this.

TimB87: seems like this makes it ignore `--x11_class` :thinking:

last-partizan: image

are you sure? works for me

TimB87: compiling right now, i missed some arguments in `cmd_line.rs`

src/cmd_line.rs
```diff
@@ -101,6 +109,16 @@ pub fn handle_command_line_arguments() -> Result<(), String> {
                 .takes_value(true)
                 .last(true)
                 .help("Specify Arguments to pass down to neovim"),
+        )
+        .arg(
+            Arg::with_name("wayland_app_id")
+                .long("wayland-appid")
```
last-partizan, line 115:
```suggestion
                .long("wayland-app-id")
```

src/cmd_line.rs
```diff
@@ -132,6 +150,14 @@ pub fn handle_command_line_arguments() -> Result<(), String> {
         wsl: matches.is_present("wsl"),
         frameless: matches.is_present("frameless") || std::env::var("NEOVIDE_FRAMELESS").is_ok(),
         geometry: parse_window_geometry(matches.value_of("geometry").map(|i| i.to_owned()))?,
+        wayland_app_id: matches
+            .value_of("wayland_app_id")
+            .unwrap_or_default()
```
last-partizan, line 155:
```suggestion
            .unwrap_or(String::from("neovide"))
```

Make sure default value is "neovide", lowercase. As by default it matches executable name.

src/cmd_line.rs
```diff
@@ -132,6 +150,14 @@ pub fn handle_command_line_arguments() -> Result<(), String> {
...
+            .to_string(),
+        x11_wm_class: matches
+            .value_of("x11_wm_class")
+            .unwrap_or_default()
```
last-partizan, line 159:
```suggestion
            .unwrap_or(String::from("neovide"))
```

src/window/window_wrapper/mod.rs
```diff
@@ -222,7 +222,12 @@ pub fn start_loop(
         .with_decorations(!SETTINGS.get::<CmdLineSettings>().frameless);
 
     #[cfg(target_os = "linux")]
-    let winit_window_builder = winit_window_builder.with_app_id("Neovide".to_string());
+    let winit_window_builder = winit_window_builder
+        .with_app_id(SETTINGS.get::<CmdLineSettings>().wayland_app_id)
+        .with_class(
+            SETTINGS.get::<CmdLineSettings>().x11_wm_class,
+            "Neovide".to_string(),
```
last-partizan, line 229:
```suggestion
            "neovide".to_string(),
```

Kethku:
Looks good to me!

TimB87:
in it's current state it doesn't allow changing the class, though. It did when I created the PR. Changing the appid works just fine, maybe somebody with just X11 can test it again?

TimB87:
Should be fixed now :) even works with just one block in mod.rs as @last-partizan suggested.

src/window/window_wrapper/mod.rs
```diff
@@ -222,7 +222,10 @@ pub fn start_loop(
         .with_decorations(!SETTINGS.get::<CmdLineSettings>().frameless);
 
     #[cfg(target_os = "linux")]
-    let winit_window_builder = winit_window_builder.with_app_id("Neovide".to_string());
+    let winit_window_builder = winit_window_builder
+        .with_app_id(SETTINGS.get::<CmdLineSettings>().wayland_app_id)
+        .with_class("neovide".to_string(), SETTINGS.get::<CmdLineSettings>().x11_wm_class
```
last-partizan, line 227: https://docs.rs/winit/0.20.0-alpha4/winit/platform/unix/trait.WindowBuilderExtUnix.html#tymethod.with_class

you made "class" static, and now changing "instance" instead.

TimB87: https://docs.rs/winit/0.25.0/winit/window/struct.WindowBuilder.html#method.with_class
with the newer version of winit this seems to have switched places

last-partizan: Oh, i see, great :)

TimB87: It had me too :D
I just double checked, Kethku/winit `new-keyboard-all` branch reports it's based on version 0.24.0, and this already has the same structure as 0.25.0 https://docs.rs/winit/0.24.0/winit/window/struct.WindowBuilder.html#method.with_class :+1:

src/cmd_line.rs
```diff
@@ -34,6 +36,8 @@ impl Default for CmdLineSettings {
             multi_grid: false,
             maximized: false,
             frameless: false,
+            wayland_app_id: String::from("neovide"),
+            x11_wm_class: String::from("neovide"),
```
TimB87, line 40: Is there anything that speaks against exchanging those two to `String::new()`?

last-partizan: `String::new()` does not take arguments.

last-partizan (approved):
Looks good to me.

@Kethku take a look, and i think we can merge it.

src/cmd_line.rs
TimB87: Does it need to in our case? We set it via unwrap_or.. later?

TimB87: I mean I tested it before asking the question and it works fine, and seems logical to me to not set it twice, but I am also a novice, so this might be a silly question :) Sorry!

Btw, thats why I wanted to try `unwrap_or_default()` instead of `unwrap_or("neovide")` earlier :)

last-partizan: Hmm, makes sense. Let's use String::new.

TimB87 (commented):
As of now, this should be all you can think off on how to manipulate the appid/class of the neovide window

Kethku:
Looks like a good change to me. @TimB87 would you be willing to add to the docs explaining the change?

TimB87:
> Looks like a good change to me. @TimB87 would you be willing to add to the docs explaining the change?

Sure, I can do that tomorrow :)

Issue #445, opened 2021-01-11
Non Qwerty\En US Input


Kethku:
Right now keyboard layouts are handled by manually listing out bindings given key presses and modifiers. This is very broken.

Luckily there is movement here which should fix this issue. https://github.com/rust-windowing/winit/pull/1788

Once this lands and looks stable enough. I'm going to pull in some version of winit which uses it, and rewrite our keyboard input handling to use that. Most of the hard work was already done thanks to @j4qfrost, so now its a waiting game. Looks like its moving very quickly though so I'm hopeful we will have motion on this soon.

For now I'm going to comb through our issues looking for any that have to do with keyboard input handling and close them as this should fix them or at least make a fix possible.

pickfire:
Note that IME support is different from keyboard layout, IME uses an external program unlike using a different layout/variant.

Kethku:
Agreed. However IME support in sdl2 is not good either. I think at this point supporting multiple windowing systems is not the way to go, so I'm going to be moving to switch to winit only once the upstream changes land. At that point, we can work with them to get IME support working for real on the relevant platforms.

Kethku:
https://github.com/Kethku/neovide/pull/506 a PR has been contributed targetting the new winit api. It currently only works on windows, but once upstream support for linux and mac lands, we should be in very good shape.

Kethku:
https://github.com/rust-windowing/winit/pull/1890 looks like an upstream macos PR has been created. Very excited

p00f:
I've set my window manager to swap escape and caps lock but it doesn't work in neovide (they are not swapped), will this fix my issue too?

pickfire:
> I've set my window manager to swap escape and caps lock but it doesn't work in neovide (they are not swapped), will this fix my issue too point_rightpoint_left ?

I did the same but IIRC I don't have this issue, I only have an issue with dvorak keys when used with other combination keys, I even swap my left ctrl and left alt but I think I don't have an issue with that either. Maybe I did but I don't recall.

Kethku:
I believe this would fix both of your problems @p00f @pickfire. At the very least it would unblock addressing those issues.

Kethku:
https://github.com/rust-windowing/winit/pull/1788 PR one of 3 has merged! Shouldn't be long now!!!

Pebor:
12 days ago 
![image](https://user-images.githubusercontent.com/32438954/117420096-89752400-af1d-11eb-9847-54a8bf4f6509.png)
[...]

PR #836, opened 2021-07-19
fix(keyboard): Fix shift+tab handling

last-partizan:
This fixes Shift+Tab mappings (now it works).

This is simpler than variant from @smolck in https://github.com/Kethku/neovide/pull/782

And it works for me. @smolck please take a look, i think shift check in is_control_key isn't required.

Refs #445 

- Fix

- No

### DIFF
--- a/src/window/window_wrapper/keyboard_manager.rs
+++ b/src/window/window_wrapper/keyboard_manager.rs
@@ -188,6 +188,7 @@ fn is_control_key(key: Key<'static>) -> Option<&str> {
         Key::End => Some("End"),
         Key::PageUp => Some("PageUp"),
         Key::PageDown => Some("PageDown"),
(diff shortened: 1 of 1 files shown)


Kethku:
I am not against merging this, but I have sent a message to the creator of the new keyboard api via this comment https://github.com/rust-windowing/winit/pull/1890#issuecomment-885789537 which explains what I think is needed to solve this for real.

Kethku:
One potential strategy we could use is to modify the code to only include modifiers if the two forms of keyboard input (with and without modifiers) matches each other.

Without doing that we will have to special case characters which will always fail for some keyboard layouts.

last-partizan:
Let's merge this to fix this specific issue, and then refactor it when we have idea how to do it right.

PR #842, opened 2021-07-26
Fix to ensure provided gui font is valid

### LINKED ISSUE #818: When guifont is set Neovide becomes slow and unstable
**Describe the bug**
If I don't set `guifont` in my `init.lua` then the text in neovide is too small. After reading #3, I added  `vim.opt.guifont = 'Consolas:h32'` to my `init.lua`. This however causes Neovide to become very slow and occasionally crash.

**To Reproduce**
1. Be on archlinux running the sway window manager
2. Install neovide like so `paru -S neovide-git`
3. Add  `vim.opt.guifont = 'Consolas:h32'` to the `init.lua`
4. Open neovide
5. Experience slowness, maybe even a crash, lack of visual effects

**Expected behavior**
Normal neovide speed and visuals but with a larger font

**Screen record**
[normal good performance](https://imgur.com/a/QCfSJH2) without `guifont` set
[guifont set, degraded performance, crashes randomly](https://imgur.com/a/AGR24Wh)

**Desktop details**
 - OS: Arch Linux 5.12.14-arch1-1
 - Window Manager: [Sway](https://github.com/swaywm/sway) 1.6.1
 - Neovide Version: 0.7.0
 - Neovim Version: v0.5.0-dev+1434-g24e0c16fd
 - GPU: Radeon RX 5700 XT

`neovide --log` output:
```
Ignored client type property: "methods"
Ignored client type property: "attributes"
```

lucas-miranda:
Providing a wrong font was causing neovide to retry every time to load an
inexistent font, instead just trying to load once at it's setup should
be enough

Maybe this can fix #818

- [X] Fix

- [X] No

### DIFF
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -66,8 +66,16 @@ impl CachingShaper {
     pub fn update_font(&mut self, guifont_setting: &str) {
         trace!("Updating font: {}", guifont_setting);
 
-        self.options = FontOptions::parse(guifont_setting);
-        self.reset_font_loader();
+        let options = FontOptions::parse(guifont_setting);
+        let font_key = FontKey::from(&options);
+
+        if let Some(_) = self.font_loader.get_or_load(&font_key) {
+            trace!("Font updated to: {}", guifont_setting);
+            self.options = options;
+            self.reset_font_loader();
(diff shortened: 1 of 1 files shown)


Kethku:
Good catch! Thanks for the fix

Issue #826, opened 2021-07-14
Receiving key signals that should not be received


Empathy117:
When using `alt + tab` or `win + tab` to switch window it just type the tab to the file.
I think it should ignore key combinations.

PyGamer0:
which version of neovide?

Empathy117:
> which version of neovide?

0.7.0 on linux

Empathy117:
![Peek 2021-07-14 22-05](https://user-images.githubusercontent.com/60363116/125636700-f0c133cb-f488-4257-813c-bb2e6d68fe21.gif)

And this what happen, I'm using Win + 1 to Switch to window 1 and use Win + tab to Switch back ...etc

PyGamer0:
are you on main or using releases? i think this is foxed on main

Empathy117:
> are you on main or using releases? i think this is fixed on main

Yes, I'm using on AUR pakcage before, and I build from source code from main branch now, but it still matter....

masmullin2000:
On Arch Linux running AUR neovide-git (updated today, July 15th), I also experiencing something like this

Rather than receiving key signals upon entering the window, I experience key signals once I start typing. (the <D-char>... eg I type j and receive a <D-j>).

I should also note that my init.vim does not seem to be loaded once I enter the window.

I can fix by mouse clicking the neovide window and everything is back to normal.

This issue is very easy for me to reproduce, but very difficult to describe.  Let me know if I can assist.

Kethku:
Could folks try main? I believe this has been fixed for real now

mimullin-bbry:
It's not fixed for me.  Neovide recompiled for me this morning (august 2nd, 2021).

Note: I am running gnome 40 with pop-shell extension.

I do not know how to capture proper logs for this issue.  Let me know how I can help please.

PR #845, opened 2021-07-29
Fix to sending key signals when window focus lost

lucas-miranda:
When window focus has just been lost, key pressed events aren't ignored

I tried to track issue #826, I've found that input are been handled as normal when window lost focus, instead of been ignored, and, because of that, handling unnecessary events (such as Alt+Tab, Super+other keys, etc).

I don't have tested with every combination, but I think it solves the problem, as I have myself been facing this issue while using neovide.

- [X] Fix

- [X] No


Kethku:
this makes sense to me! Looks like a good change.

Issue #485, opened 2021-03-02
Neovide not disowned from MinGW bash shell even with `--disowned`


JakubKoralewski:
I'm running an emulated bash shell on Windows 10 (MinGW Git Bash; created a separate issue involving a similar problem with MinGW https://github.com/Kethku/neovide/issues/484 - in 484 neovide didnt even start on MinGW but it does work when built from source). I built from source (5c6d84a4f1cbec26ed72f41a0c1f058aaaaeffd8) 

#### How the problem manifests?
When opening neovide from the MinGW shell the neovide process is not disowned. Pressing Ctrl+C inside the terminal doesn't close the launched Neovide GUI, but closing the Neovide GUI lets go of the terminal block. Starting with `neovide --disowned` does not fix the issue.

#### What I think is the cause?
I think you shouldn't do conditional compilation for Windows, but rather a runtime check on Windows to first make sure user is not using an emulated msys/cygwin/mingw shell? idk

Related to https://github.com/Kethku/neovide/issues/402

PR #849, opened 2021-07-31
refactor: Refactor window module

last-partizan:
- Split platform-dependent code into separate function
- Remove wrapper function `create_window`
- Move handle_new_grid_size where it belongs

- Codestyle
- Refactor

- No (It introduces no changes at all).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -124,41 +124,11 @@ fn main() {
     #[cfg(not(test))]
     init_logger();
 
-    #[cfg(target_os = "macos")]
-    {
-        // incase of app bundle, we can just pass --disowned option straight away to bypass this check
-        #[cfg(not(debug_assertions))]
-        if !SETTINGS.get::<CmdLineSettings>().disowned {
-            if let Ok(curr_exe) = std::env::current_exe() {
-                assert!(std::process::Command::new(curr_exe)
-                    .args(std::env::args().skip(1))
-                    .arg("--disowned")
-                    .spawn()
-                    .is_ok());
-                return;
-            } else {
-                eprintln!("error in disowning process, cannot obtain the path for the current executable, continuing without disowning...");
-            }
-        }
+    #[cfg(target_os = "windows")]
+    windows_fix_dpi();
 
-        use std::env;
-        if env::var_os("TERM").is_none() {
-            let mut profile_path = dirs::home_dir().unwrap();
-            profile_path.push(".profile");
-            let shell = env::var("SHELL").unwrap();
-            let cmd = format!(
-                "(source /etc/profile && source {} && echo $PATH)",
-                profile_path.to_str().unwrap()
-            );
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -230,3 +200,48 @@ pub fn init_logger() {
...
+    // incase of app bundle, we can just pass --disowned option straight away to bypass this check
+    #[cfg(not(debug_assertions))]
+    if !SETTINGS.get::<CmdLineSettings>().disowned {
+        if let Ok(curr_exe) = std::env::current_exe() {
+            assert!(std::process::Command::new(curr_exe)
+                .args(std::env::args().skip(1))
+                .arg("--disowned")
+                .spawn()
+                .is_ok());
+            return;
```
last-partizan, line 224: Oh, wait a minute. This was expected to stop execution, but now it doesn't. I'll fix it.

src/main.rs
```diff
@@ -230,3 +200,47 @@ pub fn init_logger() {
...
+    // incase of app bundle, we can just pass --disowned option straight away to bypass this check
+    #[cfg(not(debug_assertions))]
+    if !SETTINGS.get::<CmdLineSettings>().disowned {
+        if let Ok(curr_exe) = std::env::current_exe() {
+            assert!(std::process::Command::new(curr_exe)
+                .args(std::env::args().skip(1))
+                .arg("--disowned")
+                .spawn()
+                .is_ok());
+            std::process::exit(0);
```
last-partizan, line 224: Now it calls exit() to stop process.

But i stil does not understand how it works?

Looks like, if process is called without `--disowned`, then it is called again with `--disowned`?

So, it's actually more like internal `flag`, not `option` to be used by user? And it always disowns process on macOS.

last-partizan: Actually, this code should be platform-independent.

If found this issue https://github.com/neovide/neovide/issues/402, and for linux i have to do the same (use setsid --fork to fork process, so it would detach itself from terminal).

@Kethku In windows you say it is done by default?

https://github.com/neovide/neovide/issues/485

I found another issue, where it does not.

I think, right way to do this - is disown by default on all platforms, and rename flag to `--nofork` (just like gvim). This also will be less confusing to users (they will not try to use `--disowned` to disown process).

But that's for another PR i guess.

Kethku: I agree with this, and to be honest I don't fully understand this code either as it was contributed and I just scanned it for errors and tested.

I agree that for most cases it makes sense for disowned to be the default. My only concern is if somebody configures neovide to be the default git editor which waits for the application to finish, but like you say, this can be handled by just configuring the git editor to use --nofork

Kethku:
As per usual, this is a great fix. I appreciate the changes!

src/main.rs
last-partizan: Nice, i'll make another PR to refactor this.

PR #852, opened 2021-08-02
refactor: Refactor disowning process.

### LINKED ISSUE #485: Neovide not disowned from MinGW bash shell even with `--disowned`

last-partizan:
Now process will be disowned on all platforms by default. Can be disabled by `--nofork`. Using `-v` implies `--nofork`.

Closes #485

- Feature (probably)
- Refactor

- Yes (--disowned is changed to --nofork)

### DIFF
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -10,7 +10,7 @@ pub struct CmdLineSettings {
     pub neovim_bin: Option<String>,
     pub files_to_open: Vec<String>,
 
-    pub disowned: bool,
+    pub nofork: bool,
     pub geometry: WindowGeometry,
     pub wsl: bool,
     pub remote_tcp: Option<String>,
@@ -29,7 +29,7 @@ impl Default for CmdLineSettings {
             log_to_file: false,
             neovim_args: vec![],
             files_to_open: vec![],
-            disowned: false,
+            nofork: false,
             geometry: DEFAULT_WINDOW_GEOMETRY,
             wsl: false,
             remote_tcp: None,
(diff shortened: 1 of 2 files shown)


src/cmd_line.rs
```diff
@@ -146,7 +146,7 @@ pub fn handle_command_line_arguments() -> Result<(), String> {
             || std::env::var("NeovideMultiGrid").is_ok()
             || matches.is_present("multi_grid"),
         remote_tcp: matches.value_of("remote_tcp").map(|i| i.to_owned()),
-        disowned: matches.is_present("disowned"),
+        nofork: matches.is_present("nofork") || matches.is_present("verbosity"),
```
Kethku, line 149: Can you explain the verbosity term? Maybe add a comment? Its not obvious to me why its added

Kethku:
This looks good to me! Just a simple question

src/cmd_line.rs
last-partizan: verbosity is not added, it was already there.

what is added, is using `-v -vv -vvv ...` implies `--nofork`, becouse verbosity without nofork is useless (as process detaches from terminal and we don't see logs)

last-partizan: Do you thin comment in this place?

It is explained in `help` for `verbosity` itself. Maybe message could be improved, any suggestions?

Kethku: Nah its all good. This makes sense

PR #861, opened 2021-08-06
add better touch scrolling

Kethku:
Fixes touch scrolling by keeping track of a virtual scroll position and only sending a scroll event when that scroll position crosses a the line size boundary.

- [x] Fix
- [ ] Feature
- [ ] Codestyle
- [ ] Refactor
- [ ] Other

- [x] No

### DIFF
--- a/src/window/settings.rs
+++ b/src/window/settings.rs
@@ -7,7 +7,6 @@ pub struct WindowSettings {
     pub transparency: f32,
     pub fullscreen: bool,
     pub iso_layout: bool,
-    pub scroll_dead_zone: f32,
     pub remember_window_size: bool,
 }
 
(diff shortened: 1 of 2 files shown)


Kethku:
@shaunsingh could you test this with a mac touchpad? I think this fixes the issues you were having with scrolling on those devices.

shaunsingh:
@Kethku Can confirm this fixes the issues I was having with scrolling speed earlier. Feels much better now

last-partizan:
I just tested it too, and it works great.

Old scrolling with touchpad was too fast, new is better :)

last-partizan:
But still no scrolling on touch display, maybe i'll investigate this later.

PR #862, opened 2021-08-07
fix(cmd_line): Renamed --multiGrid to --multigrid

last-partizan:
BREAKING CHANGE: Renamed --multiGrid to --multigrid
BREAKING CHANGE: Removed env var NeovideMultiGrid

- Other

- Yes, it renames --multiGrid to --multigrid and removes NeovideMultiGrid env var

Wiki should be updated after merge. I'll handle it.


Kethku:
Seems reasonable to me.

PR #863, opened 2021-08-07
Refactor startup

last-partizan:
I tried to delay window creation until we have guifont, but that way was too obscure, and i tried something different.

Actually, we don't care when window is created, what we do care about - is when window is resized to it's desired size.

So, this PR moves setting initial size into draw_frame, behind `first_render_pass` boolean flag. And then, instead of resizing grid to window size, resizes window to grid size.

Startup process now looks like this:
```
cargo r -- -vvvv --geometry 80x40 2> >(egrep -i '(Resize|First|font upd|dimens|matched)')
...
TRACE [neovide::channel_utils] redraw_event Resize { grid: 1, width: 80, height: 40 }
TRACE [neovide::renderer::fonts::caching_shaper] Font updated to: JetBrainsMono Nerd Font:h11
TRACE [neovide::renderer] Updating font dimensions: 29x14
TRACE [neovide::renderer::fonts::caching_shaper] Font updated to: JetBrainsMono Nerd Font:h11
TRACE [neovide::renderer] Updating font dimensions: 29x14
TRACE [neovide::window::window_wrapper] Grid matched saved size, skip update.
```

One problem i haven't fixed yet, is this black line at the bottom. 
image

But now we can see startup screen :)

I'll refactor it a little, before it's ready for merge. Suggestions welcomed.

- Fix
- Refactor

- No

### DIFF
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -16,8 +16,8 @@ use crate::windows_utils::{
 pub enum UiCommand {
     Quit,
     Resize {
-        width: u32,
-        height: u32,
+        width: u64,
+        height: u64,
     },
     Keyboard(String),
     MouseButton {
--- a/src/editor/mod.rs
+++ b/src/editor/mod.rs
@@ -236,7 +236,6 @@ impl Editor {
     }
 
     fn resize_window(&mut self, grid: u64, width: u64, height: u64) {
-        trace!("editor resize {}", grid);
         if let Some(window) = self.windows.get_mut(&grid) {
             window.resize((width, height));
         } else {
--- a/src/window/window_wrapper/mod.rs
+++ b/src/window/window_wrapper/mod.rs
@@ -52,7 +52,9 @@ pub struct GlutinWindowWrapper {
     mouse_manager: MouseManager,
     title: String,
     fullscreen: bool,
+    first_render_pass: bool,
     saved_inner_size: PhysicalSize<u32>,
+    saved_grid_size: (u64, u64),
     ui_command_sender: LoggingTx<UiCommand>,
     window_command_receiver: Receiver<WindowCommand>,
 }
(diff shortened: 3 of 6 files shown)


src/bridge/ui_commands.rs
```diff
@@ -16,8 +16,8 @@ use crate::windows_utils::{
 pub enum UiCommand {
     Quit,
     Resize {
-        width: u32,
-        height: u32,
+        width: u64,
+        height: u64,
```
last-partizan, line 20: This and related type chanes - to minimize type conversions when possible.

src/editor/mod.rs
```diff
@@ -236,7 +236,6 @@ impl Editor {
     }
 
     fn resize_window(&mut self, grid: u64, width: u64, height: u64) {
-        trace!("editor resize {}", grid);
```
last-partizan, line 239: This is removed, becouse Resize event is already looged by channel_utils.

src/window/window_wrapper/mod.rs
```diff
@@ -162,41 +164,45 @@ impl GlutinWindowWrapper {
...
-        let previous_size = self.saved_inner_size;
-
-        if previous_size != current_size {
-            trace!("Updating grid size: {:#?}", current_size);
-            self.saved_inner_size = current_size;
-            self.handle_new_grid_size(current_size);
-            self.skia_renderer.resize(&self.windowed_context);
-        }
 
         if REDRAW_SCHEDULER.should_draw() || SETTINGS.get::<WindowSettings>().no_idle {
```
last-partizan, line 168: I moved this code to the upper part of draw_frame, so font loading could be triggered before resizing.

src/window/window_wrapper/mod.rs
```diff
@@ -162,41 +164,44 @@ impl GlutinWindowWrapper {
...
+        if self.saved_inner_size != new_size {
+            self.saved_inner_size = new_size;
+            self.handle_new_grid_size(new_size);
+            self.skia_renderer.resize(&self.windowed_context);
         }
     }
 
-    fn handle_new_grid_size(&self, new_size: PhysicalSize<u32>) {
-        if new_size.width > 0 && new_size.height > 0 {
-            // Add 1 here to make sure resizing doesn't change the grid size on startup
```
last-partizan, line 193: This `+1` magick i removed too. It was doing nothing to prevent grid resize on startup.

src/window/window_wrapper/mod.rs
```diff
@@ -162,41 +164,44 @@ impl GlutinWindowWrapper {
...
-            let width = ((new_size.width + 1) / self.renderer.font_width as u32) as u32;
-            let height = ((new_size.height + 1) / self.renderer.font_height as u32) as u32;
-            self.ui_command_sender
-                .send(UiCommand::Resize { width, height })
-                .ok();
+    fn handle_new_grid_size(&mut self, new_size: PhysicalSize<u32>) {
+        let (width, height) = self.renderer.to_grid_size(new_size);
```
last-partizan, line 196: I was thinking, maybe we can leave only saved_grid_size and remove saved_inner_size, but my attempts to do so failed.

So i'm leaving both for now. Maybe we come up with better ideas later :)

last-partizan:
(I rebased/force-pushed branch, and not it could be merged without squash)

src/window/window_wrapper/mod.rs
```diff
@@ -52,7 +52,9 @@ pub struct GlutinWindowWrapper {
     mouse_manager: MouseManager,
     title: String,
     fullscreen: bool,
+    first_render_pass: bool,
```
Kethku, line 55: would it make sense to combine first_render_pass and saved_grid_size into a single option which starts out as None and changes to Some((u64, u64)) after a render has been made? I didn't fully walk through the logic here so I may be missing something.

Kethku:
This is a big improvement imho. Only a minor possible simplification.

src/window/window_wrapper/mod.rs
last-partizan: Well, i changed it and it works. But i don't know if it made code better or worse :)

I think it's good enough :)

PR #865, opened 2021-08-08
Refactor grid conversions

last-partizan:
This PR introduces new type: `Dimensions`, for values like font_dimensions and other sizes. And combines paired values into single argument.

- Refactor

- No

### DIFF
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -1,4 +1,5 @@
 use crate::settings::*;
+use crate::utils::Dimensions;
 
 use clap::{App, Arg};
 
@@ -11,7 +12,7 @@ pub struct CmdLineSettings {
     pub files_to_open: Vec<String>,
 
     pub nofork: bool,
-    pub geometry: WindowGeometry,
+    pub geometry: Dimensions,
     pub wsl: bool,
     pub remote_tcp: Option<String>,
     pub multi_grid: bool,
--- a/src/main.rs
+++ b/src/main.rs
@@ -20,6 +20,7 @@ mod error_handling;
 mod redraw_scheduler;
 mod renderer;
 mod settings;
+mod utils;
 mod window;
 pub mod windows_utils;
 
--- a/src/renderer/cursor_renderer/mod.rs
+++ b/src/renderer/cursor_renderer/mod.rs
@@ -228,8 +228,7 @@ impl CursorRenderer {
 
     pub fn update_cursor_destination(
         &mut self,
-        font_width: u64,
-        font_height: u64,
+        (font_width, font_height): (u64, u64),
         windows: &HashMap<u64, RenderedWindow>,
     ) {
         let (cursor_grid_x, cursor_grid_y) = self.cursor.grid_position;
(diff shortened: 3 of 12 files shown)


Kethku:
I merged the pr before this one. Ping me when you think this is ready

src/window/utils.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+impl From<Dimensions> for (u32, u32) {
+    fn from(dimensions: Dimensions) -> Self {
+        (dimensions.width as u32, dimensions.height as u32)
+    }
+}
+
+impl From<Dimensions> for (i32, i32) {
```
last-partizan, line 57: It is mostly ready, i'll review it carefully one more time later today, and test this build for a day to catch bugs.

But, i worried that there is this duplication, almost the same `impl From` for u64, u32 and i32. @Kethku maybe there is some better way to write this? i googled, and it can be done using macros, but it looks to obscure for me.

last-partizan (commented):
I think it's ready for review now.

I tried to use `Dimensions` only when it gives benefits over simple tuple, like - it uses multiplication or division.

In other parts - i just replaced two arguments with single tuple.

src/window/utils.rs
lucas-miranda: I could help writing a macro to avoid this impl repetition

last-partizan: Let's try it after this is merged.

Or point me into some docs about how to write it. I found some examples on stackoverflow, but they were not clear enough, so i decided not to bother.

lucas-miranda: I've made an [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=43ef2f0c10f1a2006977562d30bd426a) how it could be achieved to tuples

But there is some good docs to see using [examples](https://doc.rust-lang.org/reference/macros-by-example.html) and quick refs is needed (to see [token types](https://cheats.rs/#tooling-directives) and [repetition](https://cheats.rs/#macros-attributes))
If you want to see more, there is a good concise [guide](https://danielkeep.github.io/tlborm/book/README.html) to see some tips and tricks

src/window/utils.rs
```diff
@@ -0,0 +1,83 @@
+use glutin::dpi::PhysicalSize;
+use serde::{Deserialize, Serialize};
+use std::ops::{Div, Mul};
+
+// Maybe this should be independent from serialization?
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Serialize, Deserialize)]
+pub struct Dimensions {
```
last-partizan, line 7: Is window/utis good place for this file? maybe it should be `src/utils/dimensions`? I really like src/utils/.

Is Dimensions good name for this struct?

lucas-miranda: I think it should be nice to move to `src/utils` as it could be useful out of window module and I like the existence of a `src/utils` also

Dimensions should be good (or maybe Dimension, just to avoid using plural as a type name?), sometimes I use Size to describe something like that also

Kethku (approved):
This looks good to me. Thanks again as always for taking on these refactorings. They make an outsized difference for the maintenance and understandability of the code

Kethku:
Feel free to merge w/e

PR #866, opened 2021-08-08
Upgrade to tokio 1.1.1

KillTheMule:
This upgrades neovide to use tokio 1.1.1. I'm mainly sending it hoping you can test neovide  and hopefully reproduce the original problem with nvim-rs's current master. And maybe you could also test https://github.com/KillTheMule/nvim-rs/pull/26 ( you'd need to adjust the nvim-rs entry in Cargo.toml accordingly) and hopefully report that that does fix the problem :D Thanks!


Kethku:
closing in favor of https://github.com/neovide/neovide/pull/867

PR #867, opened 2021-08-08
Swap to unforked nvim-rs

Kethku:
- Refactor

- No

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ rust-embed = { version = "5.2.0", features = ["debug-embed"] }
 image = "0.22.3"
 nvim-rs = { git = "https://github.com/KillTheMule/nvim-rs", branch = "master", features = ["use_tokio"] }
 tokio = { version = "1.1.1", features = ["full"] }
-tokio-util = "*"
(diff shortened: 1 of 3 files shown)


Kethku:
This should merge after https://github.com/KillTheMule/nvim-rs/pull/26 just in case

Cargo.toml
```diff
@@ -26,8 +26,9 @@ derive-new = "0.5"
 rmpv = "0.4.4"
 rust-embed = { version = "5.2.0", features = ["debug-embed"] }
 image = "0.22.3"
-nvim-rs = { git = "https://github.com/kethku/nvim-rs", features = ["use_tokio"] }
-tokio = { version = "0.2.9", features = ["blocking", "process", "time", "tcp"] }
+nvim-rs = { git = "https://github.com/KillTheMule/nvim-rs", branch = "master", features = ["use_tokio"] }
+tokio = { version = "1.1.1", features = ["full"] }
+tokio-util = "*"
```
KillTheMule, line 31: You'll probably want to put in a real semver number here, to prevent unexpected breakage :)

Issue #844, opened 2021-07-27
let font "preferred styles" take precedece over "styles"


realToadtoad:
**Describe the bug**
Some fonts with both Oblique and Italic styles will use the "Preferred Styles" tag in order to distinguish the two styles, though both of them are marked as "Italic" under "Styles". [(Why this is done)](https://github.com/fontforge/designwithfontforge.com/blob/gh-pages/en-US/Font_Info_%26_Metadata.md#family-naming) An example of a font that does this is [Victor Mono](https://github.com/rubjo/victor-mono). This leads to the Oblique variant often being used over the italic variant, despite other programs like VSCode and Neovim-qt handling this properly.

**To Reproduce**
1. Install [Victor Mono](https://github.com/rubjo/victor-mono) with all styles.
2. Set guifont to Victor Mono
3. Install a theme that supports italics, such as [tokyonight.nvim](https://github.com/folke/tokyonight.nvim)
4. Write a comment, which should be italic.

**Expected behavior**
Comments appear as italic font (which in Victor Mono is partially cursive).

**Screenshots**
Current State
![Screenshot_20210727_100327](https://user-images.githubusercontent.com/35282290/127197301-14c25943-d951-4810-9cce-8212cef6010e.png)

Expected Behavior (simulated in Neovide by deleting oblique fonts. issue still applies with the full Nerd Font version)
![Screenshot_20210727_100754](https://user-images.githubusercontent.com/35282290/127197671-784c0aca-5fc3-4492-b876-f55c4d7fe0b6.png)

Neovim-qt
![Screenshot_20210727_101001](https://user-images.githubusercontent.com/35282290/127197954-8aaa4c40-e47a-443e-a3aa-f1e6ca87449f.png)

VSCode
![Screenshot_20210727_101133](https://user-images.githubusercontent.com/35282290/127198137-6173da77-bbb4-4b15-8991-5b467a49be43.png)

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Neovide Version: 0.7.0
 - Neovim Version: 0.5.0

**Please run `neovide --log` and paste the contents of the `.log` file here:**
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/6887470/neovide_rCURRENT.log)

**Additional context**
A problem both with the current state and fixing the bug as stated above is that there's, as far as I know, no way to set which font to use for the italic font. Including this will help with other fonts that have multiple slanted styles and users that have preferences of the font (such as cursive vs. non-cursive).

[...]

PR #869, opened 2021-08-09
Add support to oblique style

lucas-miranda:
toggle to use oblique slant style when rendering italic text

as stated at #844, some fonts has an oblique and an italic variant, neovide was supposedly using the wrong preferred style, but with my tests this isn't happening (not anymore at least)

and a suggestion was made to be able to change which variant to use when rendering italic text, which I think is pretty fine and nice to have

so this pull request introduces `g:neovide_font_use_italic_as_oblique` to be able to toggle between italic or oblique
if accepted, I'll update wiki accordly

I've made some additions to ensure a redraw occurs when the option changes it's value (as it's a font change, right?), but I'm not entirely sure if I did it at the right place, if someone could verify that, please (the refered [section](https://github.com/neovide/neovide/pull/869/commits/965075d7bb73fe969be307495e291ba8d68ef889#diff-d683eee04fe76db84204abb0d0ac423596f368fd20f604ad52266fec2a21c2e8R367-R371))

- Feature

- No


src/renderer/fonts/font_loader.rs
```diff
@@ -119,12 +120,17 @@ impl FontLoader {
...
+        let font_style = FontStyle::new(
+            {
+                if font_key.bold {
+                    Weight::BOLD
+                } else {
+                    Weight::NORMAL
+                }
+            },
+            Width::NORMAL,
+            font_key.slant,
```
last-partizan, line 132: If we already replaced italic from boolean to Slant type, maybe it would make sense to also replace bold from boolean to Weight?

src/renderer/mod.rs
```diff
@@ -360,6 +364,12 @@ impl Renderer {
 
         root_canvas.restore();
 
+        let font_use_italic_as_oblique = { SETTINGS.get::<FontSettings>().use_italic_as_oblique };
```
last-partizan, line 367: I really don't like idea of checking every frame for option change. There must be better way, @Kethku am i right?

I already see DrawCommand::FontChanged, which is sent on guifont change, maybe similar approach can be used for this?

src/renderer/fonts/font_loader.rs
lucas-miranda: yeah, I think it makes sense, just to avoid inconsistencies

src/renderer/mod.rs
lucas-miranda: I don't like it also, to be honestly, I really tried to do another approach trying to do a single redraw request only when the setting is changed, but after many many tries through Editor and Renderer, I just couldn't find a way using what already exists

last-partizan: If we couldn't find better way, i'd suggest putting this into separate function, like `is_font_options_changed`.

But let's wait for ideas from @Kethku , and then i'll try to think something.

last-partizan: Another idea to think about, this is actually `font option`, we're already parsing guifont setting, and i found this in `:help guifont`

```
	For the Win32 GUI					*E244* *E245*
	- takes these options in the font name:
		hXX - height is XX (points, can be floating-point)
		wXX - width is XX (points, can be floating-point)
		b   - bold
		i   - italic
		u   - underline
		s   - strikeout
		cXX - character set XX.  Valid charsets are: ANSI, ARABIC,
		      BALTIC, CHINESEBIG5, DEFAULT, EASTEUROPE, GB2312, GREEK,
		      HANGEUL, HEBREW, JOHAB, MAC, OEM, RUSSIAN, SHIFTJIS,
		      SYMBOL, THAI, TURKISH, VIETNAMESE ANSI and BALTIC.
		      Normally you would use "cDEFAULT".

	  Use a ':' to separate the options.
```

So, those font options is gui-dependent, and we could try to use something like `o` to stand for oblique instead of italic.

Take a look at FontOptions::parse, i think this is exact place where this option should be defined, as it will allow us to drop this new FontSettings, and we already have code for detectiong changes in guifont.

lucas-miranda: > Another idea to think about, this is actually `font option`, we're already parsing guifont setting, and i found this in `:help guifont`
> 
> ```
> 	For the Win32 GUI					*E244* *E245*
> 	- takes these options in the font name:
> 		hXX - height is XX (points, can be floating-point)
> 		wXX - width is XX (points, can be floating-point)
> 		b   - bold
> 		i   - italic
> 		u   - underline
> 		s   - strikeout
> 		cXX - character set XX.  Valid charsets are: ANSI, ARABIC,
> 		      BALTIC, CHINESEBIG5, DEFAULT, EASTEUROPE, GB2312, GREEK,
> 		      HANGEUL, HEBREW, JOHAB, MAC, OEM, RUSSIAN, SHIFTJIS,
> 		      SYMBOL, THAI, TURKISH, VIETNAMESE ANSI and BALTIC.
> 		      Normally you would use "cDEFAULT".
> 
> 	  Use a ':' to separate the options.
> ```
> 
> So, those font options is gui-dependent, and we could try to use something like `o` to stand for oblique instead of italic.
> 
> Take a look at FontOptions::parse, i think this is exact place where this option should be defined, as it will allow us to drop this new FontSettings, and we already have code for detectiong changes in guifont.

that should be ideal, I think, I didn't know that we could "break" compatibility with neovim gui options
and maybe I've misinterpreted that this options are to explicit state to use a bold at everywhere and not to use the font when a bold style occurs, it should be the right place to have a slant option (to italic or oblique) for sure

last-partizan: Hmm, you're right, this option is about explicit font choice, not "when bold occurs".

last-partizan: Renderer.draw_frame is called by GlutinWindowWrapper.draw_frame, and when font_changed, it calls `handle_new_grid_size`.

Does changing this option requires grid resize?

lucas-miranda: Definitely not, it should only triggers a redraw at every window, just to update sections to use new style (and reusing blob cache at CachingShaper` to avoid reshaping unaffected parts)

I think, maybe, shouldn't it be triggered at `Editor` and call a `window.redraw()` at every window?
Just as guifont does when it's changed, because this is the exactly behavior which should be achieved

I don't really know how to triggers a redraw within `Renderer` itself in a way that doesn't checks every frame, maybe it could be done if `FontSettings` could implement `SettingGroup` to send a redraw command to renderer when it's value is updated, but maybe it's too far to just a single style option

Kethku: Problem is changing the guifont changes the grid sizing. I believe we currently try to preserve the window size as best as possible which triggers a grid resize. In the past we hadn't triggered redraws on font change, and that resulted in fonts being preserved between changes (big text in the command line, old font being kept around, etc)

As far as where to preserve these types of options, I have two points:
1. reading a font is just an index into a hashset to get a reference to the settings. This is VERY fast. So unless we see performance issues, the simplicity of just accessing the setting each time we draw is nice. If there is low hanging fruit to preserve the font change, then thats fine, but I wouldn't bend over backwards.
2. that said, moving the font setting into the guifont is a completely reasonable approach. As mentioned above, guifont is not a standard but rather specific to a given gui. If we feel the need to add an component of the guifont to better select and render fonts, I vote we just go for it.

lucas-miranda: I feel using guifont to do that is the better approach to do that
but as we discussed, isn't guifont options stating that it should use that style everywhere?  (e.g use a font as bold everywhere as if it's a regular style)
at first, I've interpreted in this way, even if there is nothing at the docs saying otherwise (as far as I see)

but if it's just a misinterpretation of my side and that option states that is to use **when** a specific style occurs, I really agree with that, should be the best solution

lucas-miranda: and if we all agree with that, I can make all needed modifications to fit into that

Kethku: I personally don't see an issue with guifont specifying that "when italic text is requested, render this variant". That feels like a top level font specification to me, but I also don't plan on using such a feature so I may be misunderstanding something

last-partizan: Well, struct is already called `FontOptions`, so adding option like `prefers_oblique_for_italic` or `use_italic_as_oblique` feels right think to do.

And we can even make it private, and add `fn get_slant(self, italic: bool) -> Slant` for using in build_clusters.

lucas-miranda: Right, I'll work on things to fit what was discussed

src/renderer/fonts/font_slant.rs
```diff
@@ -0,0 +1,28 @@
+use skia_safe::font_style::Slant;
+
+#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
+pub enum FontSlant {
```
lucas-miranda, line 4: I've created `FontSlant`, instead using skia Slant directly, just to keep uniform with `FontWeight`and how font things are named at neovim, essentially it doesn't brings any benefits aside the name

src/renderer/fonts/font_options.rs
```diff
@@ -59,17 +65,30 @@ impl FontOptions {
     }
 
     pub fn get_weight(&self, bold: bool) -> FontWeight {
-        if self.bold || bold {
+        if bold {
             FontWeight::Bold
```
lucas-miranda, line 69: Here I've decided not to use `self.bold` when getting `FontWeight` (or `FontSlant` below) from `FontOptions`, so the code can be reused at sections where `FontOptions` bold option should not be taken into account (for example, here at [CachingShape](https://github.com/neovide/neovide/pull/869/commits/fbe0c3d8da8ad4bd84b4688631d3393ba2a44bb4#diff-ab22eb44387b5cd97300475e15e02e47e8725088ccbfeebc7fe99cac0a8d9bc2L180-R171))

src/renderer/fonts/font_slant.rs
last-partizan: And i see you did same with FontWeight? or FontWeight is required for something?

I think we should not do this, and just use Weight and Slant from skia directly.

Maybe use `pub use skia_safe::font_style::Weight as FontWeight;`, but in the end we need Skia Weight and Slant, so this is only layer of complexity which could be avoided.

src/renderer/fonts/font_options.rs
```diff
@@ -1,9 +1,11 @@
-use super::font_loader::FontSelection;
+use super::{font_loader::FontSelection, FontSlant, FontWeight};
 
 const DEFAULT_FONT_SIZE: f32 = 14.0;
 
 #[derive(Clone, Debug)]
 pub struct FontOptions {
+    use_italic_as_oblique: bool,
```
last-partizan, line 7: Let's place private properties after public. This way - we first see which is most relevant.

src/renderer/fonts/font_slant.rs
lucas-miranda: I want to avoid using another layer to FontWeight, but, as I've stated at 7208fc4 commit, it doesn't implements Hash, which FontKey requires, and strangely uses `i32` underlying instead `u32`

But I agree with using it to slant as `pub use skia_safe::font_style::Slant as FontSlant;`

src/renderer/fonts/font_options.rs
lucas-miranda: Seems okay to me, I've just tried to replicate what I see around `Renderer`

src/renderer/fonts/font_options.rs
```diff
@@ -40,10 +43,13 @@ impl FontOptions {
                 bold = true;
             } else if part == "i" {
                 italic = true;
+            } else if part == "o" {
```
last-partizan, line 46: i would replace this with `use-italic-as-oblique` or `i-as-o`, becouse simple `o` could be interpretered as oblique.

src/renderer/fonts/font_slant.rs
last-partizan: Hmm, i see. In that case i think it's better to have both structs on neovide side.

src/renderer/fonts/font_options.rs
lucas-miranda: Makes sense to me

src/renderer/fonts/font_options.rs
last-partizan: That's good idea.

src/renderer/fonts/mod.rs
```diff
@@ -1,4 +1,9 @@
 pub mod caching_shaper;
 mod font_loader;
 mod font_options;
+mod font_slant;
+mod font_weight;
 mod swash_font;
+
+pub use font_slant::FontSlant;
+pub use font_weight::FontWeight;
```
last-partizan, line 9: These two modules should not be public, let's keep them private.

last-partizan (commented):
Other that that, everythink looks alright.

I was thinking, do we need weight and slant in the FontKey and in ShapeKey? Actually, Slant and Weight are only needed at `fn load`, but the way our code structured, makes us pass them everywhere.

If FontLoader could have reference or copy of FontOptions, we could go leave almost all code unchanged. Would it work, or have i missed something? What do you think?

This is only "think about it", not "request for changes".

lucas-miranda:
I think you've a fair point, I was very confuse with implementation at first, but `ShapeKey` only need explicitly `FontWeight` and `FontSlant` if we'll support multiple slant and weight at different text sections at the same time, other than that it's pointless to have it there, since it's a global option right now

lucas-miranda:
@last-partizan I don't think we could remove slant and weight from `FontKey` unless we explicitly remove from `FontLoader` cache the loaded italic variant `FontPair` whenever `CachingShape::update_font()` was called and when the new `FontOptions` has a different slant when italic is true

> If FontLoader could have reference or copy of FontOptions, we could go leave almost all code unchanged. Would it work, or have i missed something? What do you think?

I think some parts of `CachingShaper` could be simplified by having a copy or reference to `FontOptions`, but if we do that, we should still provide a fn with custom `FontKey`, since it's used at some sections from `CachingShaper::build_clusters()` to retrieve `FontPair` with custom values to weight and slant, which ignores the ones at `FontOptions`, such as skia fallback and last resort fallback

last-partizan:
@lucas-miranda 
```rs
    pub fn update_font(&mut self, guifont_setting: &str) {
        trace!("Updating font: {}", guifont_setting);

        let options = FontOptions::parse(guifont_setting);
        let font_key = FontKey::from(&options);

        if self.font_loader.get_or_load(&font_key).is_some() {
            trace!("Font updated to: {}", guifont_setting);
            self.options = options;
            self.reset_font_loader();
        } else {
            trace!("Font can't be updated to: {}", guifont_setting);
        }
    }

    fn reset_font_loader(&mut self) {
        let font_size = self.options.size * self.scale_factor;
        trace!("Using font_size: {:.2}px", font_size);

        self.font_loader = FontLoader::new(font_size);
        self.blob_cache.clear();
    }

```

When font is changed (and that means, when our use_italic_as_oblique is changed), it resets font loader and blob cache.

src/renderer/fonts/font_loader.rs
```diff
@@ -119,12 +119,9 @@ impl FontLoader {
...
     fn load(&mut self, font_key: FontKey) -> Option<FontPair> {
-        let font_style = match (font_key.bold, font_key.italic) {
-            (true, true) => FontStyle::bold_italic(),
-            (false, true) => FontStyle::italic(),
-            (true, false) => FontStyle::bold(),
-            (false, false) => FontStyle::normal(),
-        };
+        let weight = font_key.weight.into();
+        let slant = font_key.slant.into();
+        let font_style = FontStyle::new(weight, Width::NORMAL, slant);
```
last-partizan, line 124: > I think some parts of CachingShaper could be simplified by having a copy or reference to FontOptions, but if we do that, we should still provide a fn with custom FontKey, since it's used at some sections from CachingShaper::build_clusters() to retrieve FontPair with custom values to weight and slant, which ignores the ones at FontOptions, such as skia fallback and last resort fallback

FontKey can completely ignore existence of FontOptions. 

and this code will look like

```suggestion
        let font_style = FontStyle::new(self.options.get_weight(font_key.bold), Width::NORMAL, self.options.get_slant(font_key.italic));
```

get_weight and get_slant should be left in it's current state.

Skia Fallback is not a problem, it renders whatever you want (if you ask for oblique, it still renders something like italic).
Last Resort not a problem either, it has `bold: false, italic: false`, and it get's converted to expected normal weight and upright slant.

lucas-miranda:
> When font is changed (and that means, when our use_italic_as_oblique is changed), it resets font loader and blob cache.

Yeah, you're completely right, I'm sorry I didn't see that, I just focused at `FontLoader::get_or_load()` cache hit, but `reset_font_loader()` is been called even when font don't completely changed

src/renderer/fonts/font_loader.rs
lucas-miranda: Understood, I didn't know that

PR #872, opened 2021-08-10
swap feature flag to just be a setting instead

Kethku:
- Refactor

- No

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,6 @@ members = ["neovide-derive"]
 [features]
 default = []
 embed-fonts = []
-debug-renderer = []
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


clason:
> ## Did this PR introduce a breaking change?
> _A breaking change includes anything that breaks backwards compatibility either at compile or run time._

Yes :) This breaks build with
```
error[E0599]: no method named `get_global` found for struct `settings::SETTINGS` in the current scope
```
(which clippy also complained about -- maybe we should listen to it once in a while ;))

clason:
Should be just `SETTINGS.get`, right?

Kethku:
fixing now

Kethku:
Done.

PR #875, opened 2021-08-10
Add missing double quote

mforsb:
- Fix

- No


Kethku:
Woops... Thanks for the catch

PR #886, opened 2021-08-14
fix(logging): Replace println with debug in bridge/events.rs

last-partizan:
This PR moves those annoying messages on startup from stdout to log (with debug level).

```
Ignored client type property: "methods"
Ignored client type property: "attributes"
```

- Fix

- No


Kethku:
Good call. These have been generally bugging me for a while.

PR #889, opened 2021-08-14
fix: Trigger grid resize after font change

last-partizan:
Right now when you change font, window or grid not resized, so window is larger or smaller than picture inside it.

This PR triggers grid resize after font change. Maybe it's better to change window size, but I decided to resize grid (picture), becouse it's simpler in our case.

- Fix

- No

### DIFF
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -230,7 +230,7 @@ impl Renderer {
         canvas.restore();
     }
 
-    pub fn handle_draw_command(&mut self, root_canvas: &mut Canvas, draw_command: DrawCommand) {
+    fn handle_draw_command(&mut self, root_canvas: &mut Canvas, draw_command: DrawCommand) {
         match draw_command {
             DrawCommand::Window {
                 grid_id,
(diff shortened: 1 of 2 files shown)


src/renderer/mod.rs
```diff
@@ -230,7 +230,7 @@ impl Renderer {
         canvas.restore();
     }
 
-    pub fn handle_draw_command(&mut self, root_canvas: &mut Canvas, draw_command: DrawCommand) {
+    fn handle_draw_command(&mut self, root_canvas: &mut Canvas, draw_command: DrawCommand) {
```
last-partizan, line 233: I removed public, it's not needed here.

src/renderer/mod.rs
```diff
@@ -276,16 +276,19 @@ impl Renderer {
...
     pub fn draw_frame(&mut self, root_canvas: &mut Canvas, dt: f32) -> bool {
-        let mut font_changed = false;
-
         let draw_commands: Vec<_> = self
             .batched_draw_command_receiver
             .try_iter() // Iterator of Vec of DrawCommand
             .map(|batch| batch.into_iter()) // Iterator of Iterator of DrawCommand
             .flatten() // Iterator of DrawCommand
             .collect();
+        let mut font_changed = false;
```
last-partizan, line 291: Here i moved var definition closer to first usage.

Kethku:
Looks good to me!

PR #896, opened 2021-08-15
refactor(font): conditional compilation on macos

clason:
Avoids 
```
warning: unused variable: `points_per_inch`
  --> src/renderer/fonts/font_options.rs:94:9
   |
94 |     let points_per_inch = 72.0;
   |         ^^^^^^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_points_per_inch`
   |
   = note: `#[warn(unused_variables)]` on by default
```
when compiling `neovide` on macOS, which is innocuous but has been bugging me for a while.


Kethku:
Sweet. Thanks for the fix :)

PR #897, opened 2021-08-15
ci: allow skipping failing publish step

clason:
Should allow M1 artifacts to be built and published even though `EnricoMi/publish-unit-test-result-action/composite@v1` is broken on M1 due to broken `PyNaCl` dependency.


clason:
Alternative is to just skip the step for M1 since it's known to be broken?

Issue #693, opened 2021-05-27
No Idea How To Run 0.5.0 With WSL


skjbulcher:
The documentation says 

> Neovide supports displaying a full gui window from inside wsl via the --wsl command argument.

However, I have no idea how to make that work. I downloaded both the `neovide` and `neovide.exe` 0.5.0 files from https://github.com/Kethku/neovide/releases. 

- The .exe ran fine in Windows, but doesn't start if I provide a `--wsl` argument via a shortcut or from the command line.
- Running `./neovide --wsl` from WSL (Ubuntu) responds with "nvim: Unknown option argument: "--wsl"`.
- Running `./neovide.exe`  from WSL launches neovide in windows, but it isn't much different from running it from windows directly. All paths are Windows paths, and there is no sense that it is using the "wsl copy of neovim" as the documentation claims.

The instructions leave me clueless. I don't use visualstudio, and I'm not about to install and set up VS to work with WSL just to get an idea about how neovide _might_ work.

I'd like to explore neovide to see if it could work for me, but at this point I don't see how I can?

neovide: v0.5.0
neovim: v0.4.3
This issue has to do with running neovide, not neovim.

EDIT: Clarify that this issue is about neovide and not neovim, since both have a v0.5.0.

Kethku:
I'm happy to help you out. The commands have just changed significantly and I'm guessing the docs haven't caught up.

Some tips:
1. You have to have neovim installed in wsl for it to work. You have to run the neovide exe from the windows side and it will try to start up neovim via the nvim.exe on the wsl side.
2. If that doesn't work, you can also start up neovim in wsl via tcp socket. There are some details here: https://github.com/Kethku/neovide/tree/opengl#remote-tcp-support although this is currently on the opengl branch. I'm going to fix the CI build so you can download a current build of that tonight.
3. If all that fails, you could try running with --log to get a log file written out to disk which might give some further hints.

If any of that works let me know, I'd like to update the docs to make it clearer. If not, I've got some other things to try. Please be patient. Its just me and a few other folks building this and supporting mac, linux, and windows is not trivial especially when using a language as new as rust.

skjbulcher:
Good to know I wasn't missing something!

> You have to have neovim installed in wsl for it to work. You have to run the neovide exe from the windows side and it will try to start up neovim via the nvim.exe on the wsl side.

How does neovide.exe know to start up neovim in WSL? Here's what I'm thinking:

[...]

PR #900, opened 2021-08-16
Find nvim in WSL using `$SHELL -lic` to resolve $PATH

rsynnest:
In Linux the $PATH is traditionally defined in files like `~/.profile`, `~/.bash_profile`. `~/.zprofile` that are loaded on login.
Invoking the user's $SHELL as a login shell will properly load the WSL $PATH.
This is a potential fix for #693

- Fix

- No

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -36,7 +36,7 @@ fn platform_build_nvim_cmd(bin: &str) -> Option<Command> {
         cmd.args(&[
             bin.trim(),
             "-c",
-            "let \\$PATH=system(\"bash -ic 'echo \\$PATH' 2>/dev/null\")",
+            "let \\$PATH=system(\"\\$SHELL -lic 'echo \\$PATH' 2>/dev/null\")",
         ]);
         Some(cmd)
     } else if Path::new(&bin).exists() {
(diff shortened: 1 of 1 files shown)


Kethku:
Very interesting. Would this have negative consequences for folks who don't use bash as their shell?

rsynnest:
No, if someone is using some other login shell like zsh, the worst `bash -lic` would do is fail to load their custom $PATH if it was defined in a zsh-specific place like `~/.zprofile`. The method in the existing code would also miss those cases since it's using bash. Best practice for Linux is to define $PATH in `~/.profile` since it's shell-agnostic, so people really should be defining $PATH there, but in light of that thought, I've updated the PR to call `$SHELL -lic` instead of `bash -lic`. This invokes the user's default shell as a login shell, which should capture 99% of people's custom $PATH.

The only problem I can foresee is if someone is writing to stdout in their `.bashrc` or `.profile`. Something like a `echo "WELCOME TO WSL!"` message would break the nvim lookup, for this PR and the existing code. Users really shouldn't be doing this though, especially if they're trying to use WSL with Windows interoperability.

Kethku:
Thanks so much for this fix. I wouldn't have thought of it :)

Issue #918, opened 2021-08-26
Incorrect window size at start up or after font change


unclechu:
**Describe the bug**

After I just start the Neovide it exceeds available size for its window. If I change window size post-factum (switch i3 tiling layout back and forth) it then automatically resizes to the correct size, to fit the window.

**To Reproduce**
1. Reproducible build of Neovim and Neovide using Nix (pin of *nixpkgs* is from *nixos-unstable* channel, one that is older from stable channel has older Neovide that just fails to start):
 
   ``` sh
   nix-shell --pure -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/5bc8b980b9178ef9a4bb622320cf34e59ea2ea10.tar.gz -p neovim -p neovide --run neovide
   ```

2. Either

   1. Just start the Neovide
   2. Change font family and/or size via `set guifont=...`

   _N.B. I use i3 window manager so it automatically tiles Neovide’s window.
   Maybe without a tiling window manager Neovide just creates a window with the size it wants and it stays the same
   while a tiling window manager forces it to resize and Neovide doesn’t expect that._

3. See that the Neovim window exceeds or less than window size

**Expected behavior**

1. After startup Neovide fits its window side *(when a tiling window manager is used)*, it doesn’t exceed it and doesn’t take less than available space of its window
2. When font is changed Neovide automatically resizes its contents according current window size (if it’s a tiling window manager window’s size stays the same)

**Screenshots**

![2021-08-26 20-06-01](https://user-images.githubusercontent.com/799353/131005617-b2aba50d-2b3e-44e4-bdd3-107c85dc4237.png)

Then if I change the window’s size it fits:

![2021-08-26 20-07-39](https://user-images.githubusercontent.com/799353/131005805-b4ed2163-ad36-41de-8b01-0f53f502e2fe.png)

Before font change:

![2021-08-26 20-10-20](https://user-images.githubusercontent.com/799353/131006138-42631757-1a87-425a-b0d6-6836e881b069.png)

After font change:

![2021-08-26 20-11-02](https://user-images.githubusercontent.com/799353/131006223-7bbabf03-7fe2-459b-b334-84ee77aad4f5.png)

**Desktop (please complete the following information):**
 - OS: NixOS 21.05
 
   ``` console
   $ nixos-version
   21.05.2717.bc1a0f56812 (Okapi)
   ```

 - Neovide Version [e.g. 0.5.0]
 
   ``` console
[...]

PR #921, opened 2021-08-28
fix: Allow resize at startup

last-partizan:
Refs #918

- Fix

- No

### DIFF
--- a/src/window/window_wrapper/mod.rs
+++ b/src/window/window_wrapper/mod.rs
@@ -185,7 +185,9 @@ impl GlutinWindowWrapper {
             self.saved_grid_size = Some(settings.geometry);
             // Font change at startup is ignored, so grid size (and startup screen) could be preserved.
             // But only when --maximize is not used. With maximized window we should redraw grid.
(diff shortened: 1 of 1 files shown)


src/window/window_wrapper/mod.rs
```diff
@@ -185,7 +185,9 @@ impl GlutinWindowWrapper {
             self.saved_grid_size = Some(settings.geometry);
             // Font change at startup is ignored, so grid size (and startup screen) could be preserved.
             // But only when --maximize is not used. With maximized window we should redraw grid.
-            font_changed = false;
+            //
+            // temporary disable this: https://github.com/neovide/neovide/issues/918
+            // font_changed = false;
```
Kethku, line 190: So does this cause the start screen to go away?

last-partizan: Yeah, but look again, we could detect if window is resized, and leave startup screen when it doesn't :)

Kethku:
Beauty. I think this resolves a whole host of issues.

PR #924, opened 2021-08-30
Add install instructions for Fedora

RMPR:
- Other

- No

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,13 @@ REGEX=$(printf 's/{url}/&\#branch=%s/g' '<YOUR-BRANCH-HERE>')
 sed "$REGEX" PKGBUILD
 makepkg -si
 ```
+#### Fedora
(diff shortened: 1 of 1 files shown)


Kethku:
I don't know much about fedora. Is this the default way to install packages in fedora? Who manages the package?

cizordj:
That seems to be a third-party repository, it is like using a PPA for Ubuntu but in this case it is Fedora-based.

RMPR:
Disclaimer I don't have anything to do with the package owner. It seems it's [chrisbouchard](https://github.com/chrisbouchard), the SRPM is publicly available [here](https://github.com/chrisbouchard/neovide-nightly). I proposed this change because afaik this COPR is the only one on Fedora providing neovide.

Kethku:
I don't think I'm going to merge this as I don't know who posted it or really anything about COPR or Fedora. So adding these instructions would make the repository harder to maintain. Thanks for the pr though!

PR #927, opened 2021-09-01
updates rmpv (0.4.4 -> 1.0.0)

bugabinga:
When trying to build neovide via `cargo install --git
https://github.com/neovide/neovide` it failed.
The reason was, that the `--git` option by default ignores the
Cargo.lock file and resolves crates version on-demand.

Probably because the master branch of `nvim-rs` tracked, there was
a version mismatch of rmpv that it and neovide used.

Bumped version of rmpv in neovide to 1.0.0 to match nvim-rs.
Refreshed the Cargo.lock file via `cargo update`.
Also, fixed a warning where a variable need not be mut.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -68,9 +68,9 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
-version = "0.1.50"
+version = "0.1.51"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


clason:
If you `cargo install --git`, you usually need to add the `--locked` flag explicitly.

last-partizan:
But we should update it here.

This updates nvim-rs, which fixes bug with invalid cursor position in command line.

Issue #545, opened 2021-04-12
Neovide hijacks keys that it's not supposed to


tadzik:
I'm not quite sure how to call this behaviour, hence the confusing title – apologies.

I use the "Super" (or "windows") key to interact with me desktop environment and window manager – e.g. super+1 switches to the first workspace, super+p opens the application launcher etc. These get handled by KDE in my case. In case of neovide, and neovide only, the keys themselves get captured and handled by neovide instead: super+p pasting the contents of the pastebuffer (and never reaching KDE), super+1 switching desktops as it should, but also entering "1" into neovim. This makes it unusable for me if I want to use it alongside other applications (which is always).

The expected behaviour would be to not react to these key combinations, just like no other app I've ever seen does :)

Let me know if I can help and/or provide more details.

Kethku:
If you search super https://neovim.io/doc/user/intro.html here, you'll find 
![image](https://user-images.githubusercontent.com/3323631/114456188-6d39cd80-9b91-11eb-88f7-2d61bbe93758.png)

Neovim adds support for binding keys with the D modifier to handle the "windows" key. So we need to have a middle ground where Neovide can handle those bindings while still reacting like you expect. I think its possible that this an sdl2 keyboard handling issue which will go away when we switch to winit, but I'm not sure.

Maybe a good middle ground would be to add a setting to Neovide which configures whether it should respond to D- bindings or not. What do you think?

j4qfrost:
I thought this was an Easter egg feature where the key inputs get randomly switched around.

heimer16:
I'm having a similar issue, although in my case, the key combinations are picked up by both neovide and the window manager (openbox).  As an example, with neovide focused, when I do Super + R to run a program (via rofi) and press enter, neovide is also picking up those bindings and also enters replace mode and replaces the character with a new line.   Does not happen with terminal versions of neovim (at least not in alacritty or kitty)

Kethku:
Does the same issue happen in the opengl branch (which is run on winit)

heimer16:
I tried the opengl branch, and everything seemed to build fine, but running the binary just segfaults.  Maybe I'm missing a dependency or something?  My debugging skills at this point are pretty limit, but if you have any guidance I can give it a shot.  I'm on arch, x11, openbox, with amdgpu drivers.

Kethku:
@heimer16 we have fixed that issue and merged into main. Can you try on main?

adsick:
[...]

PR #931, opened 2021-09-02
Logo/Meta/Super key hang / key ignore on focus change on Linux/Gnome/X11

rti:
- Fix

- No

## Purpose

This ~~fixes~~ describes an issue where Neovide held the logo/super/meta key in pressed state after a window focus change, even though the key was not pressed anymore. Therefor Neovide either ignores all further key presses when `neovide_input_use_logo=false` or registers all keys as `<D-*>` keys (with logo modifier) when `neovide_input_use_logo=true`.

## Environment
I experienced the issue on Arch Linux, Gnome/X11, Neovide 5047637.
Gnome uses the logo key when pressed and released without any further keys to show a dashboard with all windows.
I am using the logo key in combination with other keys to change window arrangements, e.g. logo+f to fullscreen a window.

## Steps to reproduce
- Open Neovide
- Ensure it is focused
- Press and hold the logo key 
  - Neovide will receive a modifier change event and in turn set the logo state [here](https://github.com/neovide/neovide/blob/main/src/window/window_wrapper/keyboard_manager.rs#L64)
  - Neovide will receive another modifier change event and in turn set the logo state to false again
  - Neovide will receive a focus lost event
- Press and release the f key
  - Neovide will resize
 - Release the logo key
   - Neovide will receive a gain focus event
   - Neovide will receive another modifier change event and in turn 🔥  set the logo state to true again 🔥 

This leaves Neovide in a wrong state where it either ignores further key presses when `neovide_input_use_logo=false` ([doc](https://github.com/neovide/neovide/wiki/Configuration#use-logo-key)) or registers all keys as `<D-*>` keys ([with logo modifier](https://github.com/neovide/neovide/commit/726539fbae34b693949d2fa08a6dba786a65a8b9)) when `neovide_input_use_logo=true`.

## ~~The fix~~
~~The code already had a mechanism to ignore non modifier key presses if the current frame also had a focus change event. This behavior was introduced [here](https://github.com/neovide/neovide/commit/fec2711b90b49ad3acaeeccf4c41746e01cf7fc7) and [here](https://github.com/neovide/neovide/commit/0d6cfd54cbe945e1693c9be576ddb26ee96e2d36). This fix uses this very mechanism to also ignore modifier change events under the same conditions.~~

## Related
~~Maybe this also fixes the issue @adsick described [here](https://github.com/neovide/neovide/issues/545#issuecomment-901382568).~~

### DIFF
--- a/src/window/window_wrapper/keyboard_manager.rs
+++ b/src/window/window_wrapper/keyboard_manager.rs
@@ -58,10 +58,13 @@ impl KeyboardManager {
             } => {
                 // Record the modifer states so that we can properly add them to the keybinding
                 // text
-                self.shift = modifiers.shift_key();
-                self.ctrl = modifiers.control_key();
-                self.alt = modifiers.alt_key();
-                self.logo = modifiers.super_key();
+                if !self.ignore_input_this_frame {
+                    self.shift = modifiers.shift_key();
+                    self.ctrl = modifiers.control_key();
+                    self.alt = modifiers.alt_key();
+                    self.logo = modifiers.super_key();
+                }
+
             }
             Event::MainEventsCleared => {
                 // And the window wasn't just focused.
--- a/src/window/window_wrapper/keyboard_manager.rs
+++ b/src/window/window_wrapper/keyboard_manager.rs
@@ -84,7 +84,7 @@ impl KeyboardManager {
                     }
                 }
 
-                // Regardless of whether this was a valid keyboard input or not, rest ignoring and
+                // Regardless of whether this was a valid keyboard input or not, reset ignoring and
                 // whatever event was queued.
                 self.ignore_input_this_frame = false;
                 self.queued_key_events.clear();
--- a/src/window/window_wrapper/keyboard_manager.rs
+++ b/src/window/window_wrapper/keyboard_manager.rs
@@ -58,13 +58,14 @@ impl KeyboardManager {
             } => {
                 // Record the modifer states so that we can properly add them to the keybinding
                 // text
-                if !self.ignore_input_this_frame {
-                    self.shift = modifiers.shift_key();
-                    self.ctrl = modifiers.control_key();
-                    self.alt = modifiers.alt_key();
-                    self.logo = modifiers.super_key();
+                if self.ignore_input_this_frame {
+                    return;
                 }
 
+                self.shift = modifiers.shift_key();
+                self.ctrl = modifiers.control_key();
+                self.alt = modifiers.alt_key();
+                self.logo = modifiers.super_key();
             }
             Event::MainEventsCleared => {
                 // And the window wasn't just focused.


src/window/window_wrapper/keyboard_manager.rs
```diff
@@ -58,10 +58,13 @@ impl KeyboardManager {
             } => {
                 // Record the modifer states so that we can properly add them to the keybinding
                 // text
-                self.shift = modifiers.shift_key();
-                self.ctrl = modifiers.control_key();
-                self.alt = modifiers.alt_key();
-                self.logo = modifiers.super_key();
+                if !self.ignore_input_this_frame {
```
last-partizan, line 61: Better to use `guard clause`,

```
if condition {
  return;
}
```

rti: If you prefer, ok.

Though, I find it a bit dangerous to jump right out of the function. People adding code after `match event {}` later might scratch their heads ;)

rti:
Converted this to a draft as I was just able to restore the initial problem by using a key combination with a modifier key after e.g. logo+f (for fullscreen). Will investigate further and report back. Sorry for the noise.

src/window/window_wrapper/keyboard_manager.rs
last-partizan: Oh, i haven't seen it was inside match event.

I was thinking that was just plain function. In that case we should leave it as it is. (Or maybe split into function, like `self.handle_modifiers_change`.

rti:
I am closing this merge request because this is definitely not a Neovide issue and the proposed change does not fix it.

I was able to reproduce the issue in winit's window example as well, both in [maroider/winit/tree/new-keyboard-linux](https://github.com/maroider/winit/tree/new-keyboard-linux) and [Kethku/winit/tree/new-keyboard-all](https://github.com/Kethku/winit/tree/new-keyboard-all).

I was NOT able to reproduce the problem in XFCE, i3 and Gnome/Wayland. So apparently this is a Gnome/X11 issue.

## Steps to reproduce
 - Gnome 40.4.0 X11 Linux
 - Set the left Super key to be the "Overview shortcut" in Gnome Tweaks App, Keyboard & Mouse
 - Set a key combination using the left Super key (e.g. super-f) to change the window size (e.g. "Fullscreen" or "View Spit on Left") via Gnome Settings, Keyboard, Customize Shortcuts
 - Start neovide or the winit window example 
 - Use your window size change shortcut to resize the window and release both keys, first the letter, then Super
 - The modifier key state during focus gain reported by X will have the Super key set ([code](https://github.com/Kethku/winit/blob/new-keyboard-all/src/platform_impl/linux/x11/event_processor.rs#L822))
 - The app will not receive any key release events, so the Super modifier state will be stuck

## Mitigations
- Move the mouse, because the next motion event will have the correct modifiers set
or
- Set the **right** Super key to be the "Overview shortcut" in Gnome Tweaks App, Keyboard & Mouse. The app will suddenly receive all key events as expected

Issue #300, opened 2020-05-27
under line be only red


findNextStep:
I add setting make every member function and var with underline and the underline only show as red
![image](https://user-images.githubusercontent.com/26270464/82988276-7a0e0b00-a02b-11ea-9de1-8701c7d2f487.png)
I set underline like 
```
highlight LspCxxHlSymClassMethod gui=underline guifg=#e5b124
```
It show in terminal like 
![image](https://user-images.githubusercontent.com/26270464/82988423-b5a8d500-a02b-11ea-9526-4d9df8e3d4e2.png)

j4qfrost:
I'm having some trouble understanding your issue. Is your problem that you want both colors to appear but for different cases or is one color showing when another should be?

findNextStep:
I think the underline color set by `gui=underline`should be same as the font color

j4qfrost:
Have you tried changing other settings? It could be applying the wrong configuration file. Also maybe try changing the color to #ffffff just to check and see that it's not some weird rendering issue.

findNextStep:
it seem that it just happen 

![image](https://user-images.githubusercontent.com/26270464/82993585-6a92c000-a033-11ea-8805-beee7013d96b.png)
by the way i work on mac, compile with master branch

j4qfrost:
I've found the cause of the behavior discrepancy. It looks like the behavior for neovim is to make it so underline would always be the foreground color, but neovide handles this with a "special" color which is set to red. I am a beginner with vim, so I'm not sure how to configure the special value. This looks like a design decision by @Kethku, so I would sit tight for a response from him.

I think it's a fairly nice feature to be able to change the color like this for the underlines to pop, but it's not in line with nvim design. Thanks for pointing it out, @findNextStep

Kethku:
actually this is to spec. Neovim guis have the ability to set the underline color where terminals cannot. The special color is sent by neovim for this exact case.

You can change the underline or undercurl color by setting guisp for a given style

j4qfrost:
Should guisp default to foreground instead of red then?

Kethku:
That's an interesting point and I'm not sure. I'm sorta incline to look up what other guis do and emulate with the exception of neovim-qt which does it wrong

Thanks,
Keith
________________________________

j4qfrost:
I like how it is now. I think this is more a feature than anything else. We could include a link to some vim configuration samples on the README. It would be a good way to explore quirks.

svermeulen:
[...]

PR #940, opened 2021-09-05
fix: Fixed red color for underline

last-partizan:
@Kethku i remember we talked about this, and that time we decided not to fix this, as red default color was provided by neovim.

But, i think that was incorrect decision. We're having access to all required data to make decision about underline color:

- fg color
- special color
- defaults (provided by neovim)

And when selecting special - if current area has explicit guisp - we're using it. If not - we fall back to fg color (becouse this gives better user expirience than using red from neovim, as neovim provides `generic defauls`, not specific for this area).

Closing #300

- Fix

- No

### DIFF
--- a/src/editor/style.rs
+++ b/src/editor/style.rs
@@ -54,7 +54,7 @@ impl Style {
     pub fn special(&self, default_colors: &Colors) -> Color4f {
         self.colors
             .special
-            .unwrap_or_else(|| default_colors.special.unwrap())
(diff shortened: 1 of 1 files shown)


Kethku:
Sounds good to me. Only missing bit is 
image

Side note: :O love the auto clippy messages. I thought that was broken!

last-partizan:
Fixed and merged!

PR #944, opened 2021-09-08
AMD Startup Fix

Kethku:
Adds an srgb setting to fix an issue on some amd machines

- Fix

- No


leon332157:
Yes it works!

jesselawson:
`cargo build --release` on this branch builds successfully without error on both my MBP and Windows 10 Home.

PR #976, opened 2021-09-24
Process at most one key event per frame ("fixes" out-of-order input problem)

### LINKED ISSUE #975: Keystrokes arrive out of order under load
**Describe the bug**
If my environment stresses out nvim/neovide enough, my keystrokes start arriving out of order.

**To Reproduce**
Unfortunately, the way I produce this is pretty specific to my environment right now. I don't know of a simple way to make this happen on any machine.

1. Use Wayland on GNOME 3, 2560x1440 display, 1.5x scaling.
1. Set a pretty small guifont size in init.vim (13 or lower).
2. Launch Neovide with multigrid, maximized.
3. Navigate to a very large project (Neovide itself is too small to trigger this!)
3. Use FZF's `:Rg` command to ripgrep through files.
4. Type some text.

**Expected behavior**
My 7th gen Thinkpad starts to chug from the high resolution and large project. In older versions of Neovide, my text would come in at a delay, but with no problems. In newer versions, my text comes in at a delay and out of order!

**Screenshots**

https://user-images.githubusercontent.com/2793160/134611181-710e4ee9-d905-4216-bfe0-317768c5b2bf.mp4

Actual text typed: "okay here is some text"

 - OS: Arch Linux
 - Neovide main
 - Neovim 0.5.0

**Additional context**
I ran a git bisect and this behavior was introduced in fec2711b90b49ad3acaeeccf4c41746e01cf7fc7. I have some spare time today so I'll try to find a good solution myself, but if anyone knows off the top of their head how to solve it then feel free to beat me to it :wink:

Resonious:
- Fixes #975

Just like the last PR I made, this is kind of a bandaid solution to something I don't fully understand :sweat_smile:.

I did some manual testing and this seems to solve the problem.

At first I was concerned that this would have unintended consequences. For example, if I make 5 key press events in one frame, `MainEventsCleared` only gets called once, and so I would only see 1 keypress, right? But in reality, I imagine that sending a `UiCommand` doesn't result in a redraw until after a roundtrip with nvim. That roundtrip probably triggers a redraw, which would trigger `MainEventsCleared` again, firing off another `UiCommand`. That feedback loop should repeat until `queued_key_events` drains completely. I'm no winit expert, though, so I could be wrong here.

- No

### DIFF
--- a/src/window/window_wrapper/keyboard_manager.rs
+++ b/src/window/window_wrapper/keyboard_manager.rs
@@ -67,24 +67,28 @@ impl KeyboardManager {
                 // And the window wasn't just focused.
                 let settings = SETTINGS.get::<KeyboardSettings>();
 
-                if !self.should_ignore_input(&settings) {
-                    // If we have a keyboard event this frame
-                    for key_event in self.queued_key_events.iter() {
-                        // And a key was pressed
+                if self.should_ignore_input(&settings) {
+                    self.queued_key_events.clear();
+                    self.ignore_input_this_frame = false;
+                } else {
+                    // Process one keyboard event at a time. If we send more than one event per
+                    // frame, then there's a chance that nvim will re-order the events.
+                    loop {
+                        if self.queued_key_events.is_empty() {
+                            break;
+                        }
+
+                        let key_event = self.queued_key_events.remove(0);
                         if key_event.state == ElementState::Pressed {
-                            if let Some(keybinding) = self.maybe_get_keybinding(key_event) {
+                            if let Some(keybinding) = self.maybe_get_keybinding(&key_event) {
                                 self.command_sender
                                     .send(UiCommand::Keyboard(keybinding))
                                     .expect("Could not send keyboard ui command");
+                                break;
                             }
                         }
                     }
                 }
-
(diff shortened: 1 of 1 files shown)


Kethku:
As an update, I have looked into this PR and I do think its a reasonable strategy however before I merge this I want to try playing with how input commands are sent to the neovim executable. Currently all ui commands are processed in parallel. I think a better solution may be to select some ui commands to be processed in series (such as keyboard and mouse inputs) and some which are processed in parallel.

If that doesn't work, then I will attempt something like what you are suggesting. Thanks so much for pushing this forward!

Resonious:
Sounds good. Would make a lot more sense to fix this deeper in the stack :+1:

Kethku:
@Resonious here is my first attempt https://github.com/neovide/neovide/pull/996 could you try building from this branch and let me know if it fixes the input issue you encountered?

Kethku:
I merged the other PR as it won't make things worse. Please let me know if you find current main to be an improvement. If not I'll take a longer look at things and maybe merge this PR

Resonious:
@Kethku Just got around to recompiling now.

#996 seems to work perfectly. Thanks for finding a good solution!

PR #979, opened 2021-09-25
Fix transparency

LoipesMas:
- Fix

- No

# Description
Transparency is broken (I think since c76945? Not sure tho). This PR tries to fix that.
There are still some issues and I could use some help. 

Mainly:
- Transparency not fully updating after changing `g:neovide_transparency` (it refreshes after toggling fullscreen for example).
I tried doing a redraw or something, but unsuccessfully.
- Some backgrounds maybe shouldn't be transparent? Maybe making only `Normal guibg` transparent would look better? Or just make only black color transparent? (that's what my terminal emulator does afaik, and it looks good)
- What exactly should be semi-transparent? You could make window canvas fully transparent and make line background semi-transparent, but then parts inside window but outside nvim canvas are fully transparent. But if you make window canvas semi-transparent then it stacks with line opacity making it darker than it should be.

(I don't have experience with rust or this codebase, so maybe there is something I missed)

### DIFF
--- a/src/renderer/grid_renderer.rs
+++ b/src/renderer/grid_renderer.rs
@@ -100,8 +100,9 @@ impl GridRenderer {
                 .set_color(style.background(&self.default_style.colors).to_color());
         }
 
-        let transparency = {SETTINGS.get::<WindowSettings>().transparency};
-        self.paint.set_alpha_f(transparency);
+        if self.paint.color() == Color::BLACK {
+            self.paint.set_alpha(0);
+        }
         canvas.draw_rect(region, &self.paint);
     }
 
--- a/src/renderer/grid_renderer.rs
+++ b/src/renderer/grid_renderer.rs
@@ -103,7 +103,7 @@ impl GridRenderer {
                 .set_color(style.background(&self.default_style.colors).to_color());
         }
 
-        if self.paint.color() == Color::BLACK {
+        if self.paint.color() == self.get_default_background().with_a(255) {
             self.paint.set_alpha(0);
         }
         canvas.draw_rect(region, &self.paint);
--- a/src/renderer/grid_renderer.rs
+++ b/src/renderer/grid_renderer.rs
@@ -75,7 +75,10 @@ impl GridRenderer {
     }
 
     pub fn get_default_background(&self) -> Color {
-        let transparency = {SETTINGS.get::<WindowSettings>().transparency};
+        let mut transparency = {SETTINGS.get::<WindowSettings>().transparency};
+        if transparency < 1.0 {
+            transparency = 0.0;
+        }
         self.default_style.colors.background.unwrap().to_color().with_a((255.0 * transparency) as u8)
     }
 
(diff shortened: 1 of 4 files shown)


LoipesMas:
I managed to solve these issues.
Default-background-color colored background is fully transparent, `root_canvas` is drawn as background color with transparency as alpha.
This fixes the updating issue because now only `root_canvas` changes and it updates properly. 
Right now `transparency` is actually opacity. It should either be renamed as `opacity` (to fit `floating_opacity`, and it's actual use) or all occurrences replaced with `1-transparency`.

Kethku:
Thanks so much for looking into this! I will review it soon

bmulholland:
I'm giving this branch a try with multigrid enabled and can say that it's a definite improvement over master. There are still some issues, like at some point the text in telescope turns permanently fuzzy. However, on master, sometimes the telescope text doesn't appear at all, so consistent fuzzy text is actually an improvement :)

DimitrisMilonopoulos:
Compiled your branch, great work. It is an improvement. Although plugins like telescope still don't render properly. (Missing font colors or previews)

LoipesMas:
I think those telescope issues are the same as [this](https://github.com/neovide/neovide/issues/1071).
Which means it's something I can't easily solve (at least I don't think so). It's because of the way the telescope creates it's windows.

DimitrisMilonopoulos:
@LoipesMas I have applied the fixes for plenary and telescope works as expected too. I have a weird issue with the mouse when multiple windows are open that are not present on the master. ( I even rebased your branch on the latest upstream and it persists)

Do you experience the same issue?

# Config needed

set mouse=a

# Steps to reproduce

1. Open neovide with multigrid
2. Open more splits try to select something on the second split

bug

src/renderer/grid_renderer.rs
```diff
@@ -103,7 +103,7 @@ impl GridRenderer {
                 .set_color(style.background(&self.default_style.colors).to_color());
         }
 
-        if self.paint.color() == Color::BLACK {
+        if self.paint.color() == self.get_default_background().with_a(255) {
```
DimitrisMilonopoulos, line 111: This causes improper rendering of font shadows. If you look closely on your line numbers you can see the issue, removing it fixes this.

LoipesMas: I don't see it. And this is needed for transparency.

Kethku:
I'm merging this as I think this works as expected. The mouse issue is a separate problem

src/renderer/rendered_window.rs
```diff
@@ -238,7 +242,7 @@ impl RenderedWindow {
         paint.set_color(default_background.with_a(a));
         root_canvas.draw_rect(pixel_region, &paint);
 
-        paint.set_color(Color::from_argb(a, 255, 255, 255));
```
Kethku, line 241: This broke transparent floating windows (windows which render with transparency over other windows with a blurred background)

I'm wondering if you could give some details about this change?

LoipesMas: IIRC this color is used for drawing text, so I set it to fully opaque, because otherwise the text is hard to read.

Issue #946, opened 2021-09-11
Windows artifact is double zipped


Kethku:

Kethku:
Fixed by #993

PR #993, opened 2021-10-02
Fix for Windows artifact doublezip

dineshKumar777:
- fix for #946 

Following are my findings regarding this issue

Many projects facing this same issue. But it seems this the intended behavior. 
On download, a zip is dynamically created with all the file contents that were uploaded, which is explained here in upload-artifact repo [readme](https://github.com/actions/upload-artifact#zipped-artifact-downloads).

Workaround:
- Instead of zipping and uploading, now we directly upload the executable which resolves the issue.

REF : [upload-artifact double-zips a zip](https://github.com/actions/upload-artifact/issues/39)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,15 +48,10 @@ jobs:
         run: |
           cargo build --release
 
-      - name: Archive Windows artifact
-        run: |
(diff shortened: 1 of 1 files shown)


Kethku:
Beauty! Thanks for the fix :)

PR #1009, opened 2021-10-11
[windows] Right click menu setup with no Admin

nganhkhoa:
The old code was using HKEY_CLASSES_ROOT, which requires Admin.
Fixed to use HKEY_CURRENT_USER to work without Admin.

Reference: https://stackoverflow.com/a/29769228/13173256

This emloys two right-click menu (context menu) for files, and background directory (open a directory, right click at an empty place).

It is possible to also add a right-click menu to folder (right click on a folder). Key will be: `Software\\Classes\\Directory\\shell\\Neovide` and . Although I don't use this right-click menu, I would still implement this if it's needed.

PS: I think we should update the readme. A while ago, a person contacted me through email asking for command, which is not documented.

---

If you already register right click menu in previous version of Neovim, open `regedit` and remove these keys:
- Computer\HKEY_CLASSES_ROOT\Directory\Background\shell\Neovide
- Computer\HKEY_CLASSES_ROOT\*\shell\Neovide

### DIFF
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -143,7 +143,8 @@ impl ParallelCommand {
             #[cfg(windows)]
             ParallelCommand::RegisterRightClick => {
                 if unregister_rightclick() {
-                    let msg = "Could not unregister previous menu item. Possibly already registered.";
+                    let msg =
+                        "Could not unregister previous menu item. Possibly already registered.";
                     nvim.err_writeln(msg).await.ok();
                     error!("{}", msg);
                 }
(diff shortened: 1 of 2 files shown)


Kethku:
This looks good from inspection but should be tested. Just requires adding the context menu when run from a non admin install, testing the context menu addition was added, and then removing it and testing that it was removed.

Note: too my knowledge on windows 11 this requires going to the old context menu via the more options button at the bottom of right click.

nganhkhoa:
This is tested on my machine. Now I've updated to Windows 11, it is the expected behaviour that we must traverse through the "Show more options" button.

I have not research into how Windows 11 look up the registry for menu. I've seen Adobe PDF and Winrar updated to this though.

Kethku:
Okeydokey! I will merge then. Thanks so much for the contribution

PR #1015, opened 2021-10-14
Fix special style

LoipesMas:
`Style.special()` was returning foreground instead of default special if `special` was `None`. 
(If that was intended then the test at line 145 should be fixed instead).

### DIFF
--- a/src/editor/style.rs
+++ b/src/editor/style.rs
@@ -144,7 +144,7 @@ mod tests {
         style.colors.special = None;
         assert_eq!(
             style.special(&DEFAULT_COLORS),
-            DEFAULT_COLORS.special.unwrap()
+            style.foreground(&DEFAULT_COLORS),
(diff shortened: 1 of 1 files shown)


Kethku:
Sorry for the delay. There was discussion here: https://github.com/neovide/neovide/pull/940

@LoipesMas do you have some ideas or response to the decisions made there?

LoipesMas:
The reason for this change was that I run `cargo test` and the test failed for default special. 
I didn't know that this behavior was intended. Now that I've read the discussion it makes sense.
So, as I stated, the test should be fixed instead (because now it still expects previous behavior).
I can update this PR with this fix

Kethku:
Wild that the tests aren't being run in the ci build! I'll take a look at that as well

LoipesMas:
Updated.

PR #1016, opened 2021-10-14
Add neovim_bin argument

### LINKED ISSUE #1005: Option to specify path of nvim.exe for neovide to detect.
I have to place nvim\bin into path environment variable just so that neovide can sense nvim.exe.

Add an option to specify nvim path in neovide settings.

LoipesMas:
- Feature

- No

Adds command line argument for specifying neovim bin path.
Closes #1005

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -31,6 +31,7 @@ extern crate derive_new;
 extern crate lazy_static;
 
 use std::sync::mpsc::channel;
+use std::env::args;
 
 use log::trace;
 use tokio::sync::mpsc::unbounded_channel;
@@ -117,7 +118,7 @@ fn main() {
     //   properly or updates to the graphics are pushed to the screen.
 
     //Will exit if -h or -v
-    if let Err(err) = cmd_line::handle_command_line_arguments() {
+    if let Err(err) = cmd_line::handle_command_line_arguments(args().collect()) {
         eprintln!("{}", err);
         return;
     }
@@ -179,12 +180,6 @@ fn main() {
 pub fn init_logger() {
     let settings = SETTINGS.get::<CmdLineSettings>();
 
-    let verbosity = match settings.verbosity {
-        0 => "warn",
-        1 => "info",
-        2 => "debug",
-        _ => "trace",
-    };
     let logger = match settings.log_to_file {
         true => Logger::with_env_or_str("neovide")
             .duplicate_to_stderr(Duplicate::Error)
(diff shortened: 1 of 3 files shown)


Kethku:
This looks good but I would like to test it before merging.

Kethku:
Testing would require:
1. Removing neovim from the path
2. Calling neovide with the bin argument and making sure it works
3. Re-adding neovim to the path
4. Calling neovide with the path to a different neovim install to make sure it works
5. Calling neovide with an invalid neovim path to make sure something useful is messaged

LoipesMas:
```
$ PATH="" neovide
ERROR [neovide::bridge] nvim not found!
$ PATH="" neovide --neovim-bin /usr/bin # this works correctly
$ neovide --neovim-bin /path/to/different/version/nvim # uses the different version (verified by :version)
$ neovide --neovim-bin /incorrect/path
WARN [neovide::bridge] NEOVIM_BIN is invalid falling back to first bin in PATH
# and then launches correctly as if the argument wasn't used
$ PATH="" neovide --neovim-bin /incorrect/path
WARN [neovide::bridge] NEOVIM_BIN is invalid falling back to first bin in PATH
ERROR [neovide::bridge] nvim not found!
```
Also `NEOVIM_BIN` env var overwrites argument. Not sure if this the correct way or should this be the other way around.

Kethku:
This should definitely be the other way around.

src/main.rs
```diff
@@ -194,7 +189,7 @@ pub fn init_logger() {
                 Naming::Timestamps,
                 Cleanup::KeepLogFiles(1),
             ),
-        false => Logger::with_env_or_str(format!("neovide = {}", verbosity)),
+        false => Logger::with_env_or_str(format!("neovide = {}", "trace")),
```
clason, line 192: Why did you remove the `verbosity` option and force the log level to `trace`? People in general don't want the console spammed with debug output when running neovide (not to mention that this has performance implications)!

Kethku: Does this spam to console? The logger is supposed to only write to file when the log argument is given. All other times the log calls are effectively noops

Maybe there is behavior I don't understand on some platforms?

clason: yes, it does...

clason: I think that's configurable; can't find out at first glance what the default is for flexi_logger

clason: https://docs.rs/flexi_logger/latest/flexi_logger/code_examples/index.html#choose-the-log-output-channel:

> By default, logs are written to stderr.

Kethku: Welp. I can fix this :)

clason: There's already a PR up: https://github.com/neovide/neovide/pull/1085

Kethku: I think the verbosity flag is unnecessary if we only log out when --log is present. That way the log always has all the information we need and never effects performance when we don't.

I think the goal originally was to always print errors to the console. Which I may still try to get going

clason: Makes sense; you can do that pretty easily with `flex_logger`.

PR #1027, opened 2021-10-27
add set git clone depth

przemyslawjanpietrzak:
- Other

- No

When you want to build noevide from source you don't have to load hole git history. This should speed up a little cloning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ $ scoop install neovide
 5. Build and install Neovide:
 
     ```sh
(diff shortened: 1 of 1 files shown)


Kethku:
I think this is unnecessary. I realize this improves things but the benefit feels minor compared to more opaque instructions. For example it would feel weird to me to add more than necessary instructions for building and running. One might assume we would have to add the depth parameter which isn't the case.

przemyslawjanpietrzak:
ok

PR #1032, opened 2021-11-03
Added setting to hide mouse cursor when typing

### LINKED ISSUE #935: [Feature Request] Hiding Mouse Cursor While Typing
Many terminal emulators like Alacritty or Kitty have the option to hide the mouse cursor as soon as the user starts typing, would it be possible to implement a similar feature in neovide? Thanks for considering.

jsosulski:

- Feature

- No.

### More

This PR closes #935

*Disclaimer:* I am fairly new to rust and architecturally not sure if this should be handled where it's handled. I am also not aware of the performance overhead checking this on every key/mouse move event.

OT: I think I have never found a way to implement a desired feature so quickly in a new programming language and also like how I barely had to change anything to expose this feature as a setting. Props to the architecture!

### DIFF
--- a/src/window/settings.rs
+++ b/src/window/settings.rs
@@ -8,6 +8,7 @@ pub struct WindowSettings {
     pub fullscreen: bool,
     pub iso_layout: bool,
     pub remember_window_size: bool,
+    pub hide_mouse_when_typing: bool,
 }
 
 impl Default for WindowSettings {
@@ -19,6 +20,7 @@ impl Default for WindowSettings {
             refresh_rate: 60,
             no_idle: SETTINGS.get::<CmdLineSettings>().no_idle,
             remember_window_size: false,
+            hide_mouse_when_typing: false,
         }
     }
 }
--- a/src/window/window_wrapper/keyboard_manager.rs
+++ b/src/window/window_wrapper/keyboard_manager.rs
@@ -1,6 +1,6 @@
-use glutin::{PossiblyCurrent, WindowedContext};
 use glutin::event::{ElementState, Event, KeyEvent, WindowEvent};
 use glutin::keyboard::Key;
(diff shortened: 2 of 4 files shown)


dineshKumar777:
Thanks for this. I am looking forward to this feature for a long time. Tested this in windows machine and works as intended. But it feels little bit slower when typing compared to main branch. (mine is very old machine).

jsosulski:
I have a kind of old machine, but it probably was once powerful (i7-4710MQ) and I did not notice any slowdown. Does it also feel slower when the setting is disabled?
Someone with MACOS should probably also test this first.

The setting could be stored as a flag in the event-handlers which might be faster, but again I do not want to intermingle settings with event_handlers unless the performance degradation is too severe, I guess.

dineshKumar777:
mine is i5 windows machine in which everything is little bit slower :) . so if others don't face any issue we can safely ignore mine. better to wait for core devs to have a look into this.

Kethku:
I like this idea, but rather than calling the winit api to hide the cursor over and over, I'm going to keep track of when it was called and do it once
should resolve the performance issue

Kethku:
I moved the mouse hide handling to the mouse manager and created a variable there which tracks whether it has been hidden so that it doesn't get hidden over and over unnecessarily. I believe this should fix the performance issue and am merging this.

Kethku:
Thanks for the new feature!

jsosulski:
Thank you for the fix and merge!

zhaogangang:
i set g:neovide_hide_mou_when_typing = 1, it work goods just use keyboard, but when i click the mouse, then key the keyboard, the mouse cursor not hide again, i think should hide mouse cursor when i key the keyboard anytime, just like in neovim

os version:ubuntu 20.1
nvim version:0.7.0
neovide version:0.10.1

PR #1042, opened 2021-11-09
Refactor/combinators

purpleP:
- Codestyle
- No breaking changes

Using combinators is more idiomatic IMHO.


Kethku:
Good call again! Thanks

PR #1043, opened 2021-11-09
There’s no need to reallocate characters grid and doing all these copying.

purpleP:
- Refactor
- No breaking changes.

Unless you want to save space when the grid is getting smaller you can avoid allocating new vector and copying data there.

I should have probably added `self.characters.set_len(width * height)` if you will want to use bounds checking from vector instead of doing it yourself in `get_index`.

And if you want to save memory when downsizing you should consider using `drain` 

https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.drain to avoid copying all this cells.

Pick any version I’ll update the PR accordingly If you’re interested.

### DIFF
--- a/src/editor/grid.rs
+++ b/src/editor/grid.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::{iter, sync::Arc};
 
 use super::style::Style;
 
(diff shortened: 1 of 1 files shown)


Kethku:
This seems like a very reasonable change which should have some performance benefits for bigger grids.

I would like to do a cursory test of it before merging though.

Kethku:
Testing should just require building from this branch and moving some windows around to resize them

LoipesMas:
It looks like it breaks things.
Probably because character positions are wrong.
If we have a char at `x: 5, y: 10` with grid size `100x30` then it's at index `5 + 10 * 100 = 1005`. If we resize the grid to `50x30` then it should be at `5 + 10 * 50 = 505` but it's still at `1005`.

Kethku:
Ok. Given that this is an unnecessary change, and to my knowledge it does not create performance problems, I'm going to close this as not merged.

Would totally be happy to revisit this if somebody cares to push it further though

PR #1044, opened 2021-11-09
Remove unnecessary bounds checking

purpleP:
- Refactor
- No breaking changes

There is no need to do bounds checking as it happens in `cell_index`.

### DIFF
--- a/src/editor/grid.rs
+++ b/src/editor/grid.rs
@@ -59,12 +59,13 @@ impl CharacterGrid {
     }
 
     pub fn get_cell(&self, x: u64, y: u64) -> Option<&GridCell> {
-        self.cell_index(x, y).map(|idx| &self.characters[idx])
+        self.cell_index(x, y)
+            .map(|idx| unsafe { self.characters.get_unchecked(idx) })
     }
 
     pub fn get_cell_mut(&mut self, x: u64, y: u64) -> Option<&mut GridCell> {
         self.cell_index(x, y)
-            .map(move |idx| &mut self.characters[idx])
+            .map(move |idx| unsafe { self.characters.get_unchecked_mut(idx) })
     }
 
     pub fn set_all_characters(&mut self, value: GridCell) {


Kethku:
Just trying to understand here. Was this a performance issue? I'd rather avoid an unsafe block if necessary

j4qfrost:
My intuition is that there might be an obscure bug where the window gets resized and a cell gets created that isn't associated with a character.

purpleP:
> Just trying to understand here. Was this a performance issue? I'd rather avoid an unsafe block if necessary

You’re using a vector to store what is essentially 2darray, right?

Then you had to reimplement indexing for it, right?

in rust there’s three methods to get element by index from safest to least safe:

- `get` which returns an option if index is out of bounds
- `index` or `[]` which panics if index is out of bounds
- `get_unchecked` which doesn’t do any bounds checking.

You already doing bounds checking in `cell_index`, there’s no point in doing it again in inside of vec implementation. And if you still think that something might happen during future refactorings then consider that if you will leave the old version it would still panic at runtime.

purpleP:
You know what? Maybe 🤔 you would prefer another solution. Instead of checking an index yourself just delegate that to vector. It’s even less code and no unsafe on your part.

```
fn get_cell(&self, x: usize, y: usize) -> Option<GridCell> {
    self.characters.get(x + y * width)
}
```

LoipesMas:
> ```
> fn get_cell(&self, x: usize, y: usize) -> Option<GridCell> {
>     self.characters.get(x + y * width)
> }
> ```

This one looks the best. Clean and safe. Performance-wise it should be fast enough.

j4qfrost:
> > ```
> > fn get_cell(&self, x: usize, y: usize) -> Option<GridCell> {
> >     self.characters.get(x + y * width)
> > }
> > ```
> 
> This one looks the best. Clean and safe. Performance-wise it should be fast enough.
> 
> 


The issue is that you remove bound checking.

LoipesMas:
> The issue is that you remove bound checking.
Right. `get` does bound checking, but I forgot about checking if `x` and `y` are valid.

j4qfrost:
Yea, just realized this myself. x and y need to be checked individually. Haven't caffeinated yet.

LoipesMas:
x and y are checked in cell_index, so it's back to `[]` vs `get` vs `get_unchecked`.
I'm not sure how much does replacing `[]` with `get_unchecked` impact performance (I would expect it to be close to none, since neovide isn't really that demanding and those few CPU cycles shouldn't make a difference, but feel free to do some benchmarking).
Using `get_unchecked` is *theoretically* faster, but it would be a bit less readable, so I would argue that it's not worth it. 
But again, if it makes a noticeable difference, then it would make sense.

purpleP:
> x and y are checked in cell_index, so it's back to `[]` vs `get` vs `get_unchecked`.
> I'm not sure how much does replacing `[]` with `get_unchecked` impact performance (I would expect it to be close to none, since neovide isn't really that demanding and those few CPU cycles shouldn't make a difference, but feel free to do some benchmarking).
> Using `get_unchecked` is _theoretically_ faster, but it would be a bit less readable, so I would argue that it's not worth it.
> But again, if it makes a noticeable difference, then it would make sense.

I don’t think it’s going to be noticeable, I just don’t want to leave any performance on the table. Now that we’ve Discussed this, I think it doesn’t worth it.

The real solution to this is using proper 2darray from `ndarray` library or doing it by yourself anyway. Which I can do, If anyone is interested.

j4qfrost:
The less code we need to maintain, the better.

Kethku:
I agree with the comments made here. I would definitely be willing to take a PR which replaces our manual 2d array with ndarray. But its low on the priority list atm for me to take on

Issue #1064, opened 2021-12-04
Compiler warnings on linux


jeremycostanzo:
**Describe the bug**
There are currently two compiler warnings on linux on the latest version of main
```
cargo check
    Checking neovide v0.8.0 (/home/.../neovide)
warning: unused import: `super::ui_commands::ParallelCommand`
  --> src/bridge/handler.rs:11:5
   |
11 | use super::ui_commands::ParallelCommand;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: field is never read: `ui_command_sender`
  --> src/bridge/handler.rs:20:5
   |
20 |     ui_command_sender: Arc<Mutex<LoggingTx<UiCommand>>>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: `neovide` (bin "neovide") generated 2 warnings
    Finished dev [unoptimized + debuginfo] target(s) in 1.12s
```

**To Reproduce**
1. Clone the repo on a linux (or macos I believe) machine
2. Run `cargo check`

**Expected behavior**
No compiler warnings

**Desktop (please complete the following information):**
 - OS: debian 11
 - Neovide Version: latest commit: commit 8b1a81ded3a42dc935a5f04a35873d94184d52a3 (origin/main, origin/HEAD, main)

jeremycostanzo:
I resolved one of those in #1065 , but the second one: 
```
warning: field is never read: `ui_command_sender`
  --> src/bridge/handler.rs:20:5
   |
20 |     ui_command_sender: Arc<Mutex<LoggingTx<UiCommand>>>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default
```
would require a lot more refactoring considering that this struct field is used only for windows builds.

PR #1065, opened 2021-12-04
Remove compiler warning on non windows platforms

jeremycostanzo:
Related to: #1064 
ParallelCommand was only used on windows but it was imported for every distro

- Refactor

- No


Kethku:
Thanks for the fix :)

PR #1072, opened 2021-12-05
Properly export PATH on macOS when ran from Finder

pawlowskialex:
Currently, it's impossible to launch Neovide from Finder when using Fish as login shell or when using non-standard environment configs (e.g. system managed by [Nix](https://github.com/NixOS/nix) and [nix-darwin](https://github.com/LnL7/nix-darwin)).

This PR removes hardcoded expectation that `~/.profile` exists or that `$SHELL` is POSIX-compatible.

- Fix

- No

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -236,7 +236,6 @@ fn handle_macos() {
     if env::var_os("TERM").is_none() {
         let shell = env::var("SHELL").unwrap();
         let cmd = "printenv PATH";
-        
         if let Ok(path) = std::process::Command::new(shell)
             .arg("-lic")
             .arg(cmd)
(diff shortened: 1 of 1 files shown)


Kethku:
I don't have a mac to test this. Could you add a comment explaining the new flags?

lethanix:
> I don't have a mac to test this. Could you add a comment explaining the new flags?

I tested it in macOS Monterey (Intel) and it seems to work fine. However, the explanation of the new flags could be really useful before merging.

pawlowskialex:
Sure thing, I can add a couple of commands. This sequence is already used for Windows: https://github.com/neovide/neovide/blob/cbdb28b0946fa96acc9b84ddc3752c648e66a82b/src/bridge/mod.rs#L39
`-l run as login shell`
`-i run as interactive shell`
`-c <text> run text from arguments`

Basically, this simulates opening your preferred terminal emulator. The resulting `$PATH` will be exactly the same as in the terminal.

Kethku:
Awesome. Thanks so much :)

PR #1074, opened 2021-12-06
Update CI to build a universal app for macOS

pawlowskialex:
Github Actions now support macOS 11, this allows building a universal app on Intel VMs.

Instead of the separate `build-m1` workflow, CI will now run `cargo bundle` for x86_64 and aarch64 architectures in a single workflow, and then `lipo` them together.

`cargo bundle` doesn't currently support building universal apps, so this requires a bit of manual hacking on the resulting `.app` bundle.

Self-hosted runners to build for M1 are no longer required.

- Fix

- No

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,7 +54,7 @@ jobs:
           path: ./target/release/neovide.exe
 
   build-mac:
-    runs-on: macos-latest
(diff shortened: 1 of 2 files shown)


Kethku:
Hell yeah! This is awesome. Merging now

Issue #1030, opened 2021-10-31
Center window position on startup


etienne-k:
**Describe the bug**
When neovide is started, the window appears somewhere between the center and the bottom right corner of my screen (see screenshot).
This also means that the window appears smaller than the requested --geometry, because I tend to specify a geometry that takes about 75% screen height and width.
This effect appears in the latest git build (0.7.0.r185.g491973f-1). With the latest stable version, the window was centered at startup. This leads me to believe that this is a regression.

**To Reproduce**
Start neovide using a build of the latest sources.

**Expected behavior**
The window appears centered (or at least, that behavior is optionally configurable).

**Screenshots**
![image](https://user-images.githubusercontent.com/2804556/139603220-3f195a30-c251-4bae-9576-3245ab5f15e5.png)

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Neovide Version: 0.7.0.r185.g491973f-1
 - Neovim Version: 0.5.1 stable release (also tested with 0.5.0 - no difference)

Probably not relevant, but these are the GUI-related options I have set:
```lua
vim.o.guifont = "SauceCodePro Nerd Font Mono:h10" -- <font>:h<size>
vim.g.neovide_cursor_vfx_mode = "pixiedust"
vim.g.neovide_remember_window_size = false
```
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/7449893/neovide_rCURRENT.log)

Kethku:
Window position is now remembered by default fixing this issue.

Issue #887, opened 2021-08-14
Remember window position


clason:
With `let g:neovide_remember_window_size = v:true`, neovide remembers the window size. That is nice! 

It would be even nicer if it also remembered the window _position_ in this case ;)

Kethku:
I wonder if both of these should happen by default

comiluv:
I second this motion 😸

chrismit3s:
+1

PR #1075, opened 2021-12-06
Center window on startup

jeremycostanzo:
Resolves: #1030 
- Feature

- I changed the default geometry from 50 to 30, I am not sure whether it could break things

### DIFF
--- a/src/settings/window_geometry.rs
+++ b/src/settings/window_geometry.rs
@@ -24,7 +24,7 @@ pub fn try_to_load_last_window_size() -> Result<Dimensions, String> {
 }
 pub const DEFAULT_WINDOW_GEOMETRY: Dimensions = Dimensions {
     width: 100,
-    height: 50,
+    height: 30,
 };
 
 pub fn maybe_save_window_size(grid_size: Option<Dimensions>) {
--- a/src/window/window_wrapper/mod.rs
+++ b/src/window/window_wrapper/mod.rs
@@ -343,9 +343,12 @@ pub fn create_window(
 
         if !was_centered {
             let window = window_wrapper.windowed_context.window();
+            let new_size = window.inner_size();
             if let Some(handle) = window.current_monitor() {
-                center_window(handle, window);
-                was_centered = true;
+                if is_already_resized(new_size) {
+                    center_window(handle, window);
+                    was_centered = true;
+                }
             }
         }
 
@@ -360,8 +363,8 @@ fn is_already_resized(size: PhysicalSize<u32>) -> bool {
 fn center_window(handle: winit::monitor::MonitorHandle, window: &Window) {
     let outer_size = window.outer_size();
     let monitor_size = handle.size();
-    let window_x = (monitor_size.width - outer_size.width) / 2;
-    let window_y = (monitor_size.height - outer_size.height) / 2;
+    let window_x = (monitor_size.width as i32 - outer_size.width as i32) / 2;
+    let window_y = (monitor_size.height as i32 - outer_size.height as i32) / 2;
     window.set_outer_position(PhysicalPosition {
         x: window_x,
         y: window_y,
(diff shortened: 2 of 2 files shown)


jeremycostanzo:
I changed the default geometry to make the window larger than it is high.
I did it because with default settings, the window was too high for my 1920/1080 screen on windows, but I can remove the change from this PR so that can discuss it in another issue ^^.

src/window/window_wrapper/mod.rs
```diff
@@ -339,10 +341,32 @@ pub fn create_window(
...
 
 fn is_already_resized(size: PhysicalSize<u32>) -> bool {
     size != PhysicalSize::from((800, 600))
 }
+
+fn center_window(handle: winit::monitor::MonitorHandle, window: &Window) {
+    let outer_size = window.outer_size();
+    let monitor_size = handle.size();
+    let window_x = (monitor_size.width as i32 - outer_size.width as i32) / 2;
+    let window_y = (monitor_size.height as i32 - outer_size.height as i32) / 2;
```
LoipesMas, line 367: Can this be `(monitor_size.width - outer_size.width) as i32 / 2`?
Should look a bit cleaner.

jeremycostanzo: I added the `as i32` afterwards, because both `monitor_size.width` and `outer_size.width` are usize, so the substraction would fail if `outer_size.width` is larger than `monitor_size.width`.

With the `as i32`, it does not fail and the window is centered.

Although I agree that it is not satisfying to have it 4 times in 2 lines, so don't hesitate if you can think of a better implementation :).

LoipesMas: Ah, they're usize. That changes things.
I'm not sure what the expected behaviour should be when `outer_size` is larger than `monitor_size`.
Maybe [saturating_sub](https://doc.rust-lang.org/stable/std/primitive.usize.html#method.saturating_sub) could be used? So then if `outer_size` is larger than `monitor_size` it will "snap" to top-left corner (I think).

LoipesMas: Maximizing it seems reasonable

jeremycostanzo: It should be better now

LoipesMas:
Looks good. Can't test it myself (tiling window manager), so I'll trust you (or if someone else tests this) that it works.

I think we should leave the default geometry out of this PR, since this wasn't related to the issue.

Also, now that I think of it, should it always center the window? Don't some window managers open windows in particular positions (i.e. kinda like tiling, or opening where it was closed (or is it always handled on application side))?
And always centering could mess with it.

Kethku:
Also how does this interact with https://github.com/neovide/neovide/issues/887 which tries to remember the position of the window on startup

last-partizan:
I don't think neovide should center it's window by default (or at all).

Right now window manager takes care of window positions, and doing this job just fine. (In my case - gnome places windows in less cluttered part of the screen, and it's great behaviour).

> With the latest stable version, the window was centered at startup. This leads me to believe that this is a regression.

@Kethku does this looks like regression? i don't remember it working that way.

I think, it's better not to touch initial position and instead implement remembering last position. If it was at center - it will be restored.

But, if this really adds value for some people, it should be added behind a flag, so it would not interfere with default behaviour.

src/settings/window_geometry.rs
```diff
@@ -24,7 +24,7 @@ pub fn try_to_load_last_window_size() -> Result<Dimensions, String> {
 }
 pub const DEFAULT_WINDOW_GEOMETRY: Dimensions = Dimensions {
     width: 100,
-    height: 50,
+    height: 30,
```
last-partizan, line 27: I think this is unrelated, and better be reverted.

You always can create alias for your default size.

Kethku:
@last-partizan I believe this was an unintended side effect of using sdl2. Now that things are on winit, we have more control.

Also a bunch of changes just went in to the way we persist window position and size, so I'm inclined to say we should abandon this PR or lock the changes behind a setting (and or environment variable).

@jeremycostanzo what do you think?

jeremycostanzo:
@Kethku Yes I understand, let's close this then!

PR #1079, opened 2021-12-08
Install cargo-bundle when building for macOS

pawlowskialex:
The previous [PR](https://github.com/neovide/neovide/pull/1074) broke CI because it doesn't install `cargo-bundle` before building.

I have no idea why it worked during my testing, probably caching.

- Fix

- No


LoipesMas:
Thanks

PR #1083, opened 2021-12-10
fix #243 ime input support

### LINKED ISSUE #243: IME Support
Hi there~

Awesome NeoVide is, though some support of non-English inputs may face some degrees of disability. Dynamic represent [here](https://twitter.com/_kjhl/status/1248993062837039105). Besides, thanks Kethku for friendly reply on my previous tweets, such a warm action.

- **Catch Tapping Events** When user is typing in non-English context, Vide catchs the letter and send it to inside vim. In normal situations with modern IDE, problem above will be (probably I guess) solved with `if_not_us_keyboard_current -> act nothing`.
- **Position of Popup Vocab List** As shown in tweets, the popup input methods was fixed at TL, something like `(0, 0)`.
- **The Convert of Chinese** Seemingly strings of pronunciation always goes under a process of converting to square Chinese chars in my terminal-based vim, while Vide does not.

I've noticed that some similar bugs were tagged but I hope this helps, at least the former two (I guess) will be a common faults in **Input Methods Situation**.

vzex:
fix #243

no breaking changes
no breaking build, tested on windows 10.
chineseinput

### DIFF
--- a/src/window/keyboard_manager.rs
+++ b/src/window/keyboard_manager.rs
@@ -1,6 +1,8 @@
 use glutin::event::{ElementState, Event, KeyEvent, WindowEvent};
 use glutin::keyboard::Key;
 
+use glutin::platform::modifier_supplement::KeyEventExtModifierSupplement;
+
 use crate::bridge::{SerialCommand, UiCommand};
 use crate::channel_utils::LoggingTx;
 use crate::settings::SETTINGS;
@@ -121,8 +123,11 @@ impl KeyboardManager {
             Some(self.format_keybinding_string(true, true, key_text))
         } else {
             let key_text = key_event.text;
-
-            if let Some(key_text) = key_text {
+            if let Some(ori_key_text) = key_text {
+                let mut key_text = ori_key_text;
+                if let Some(modify) = key_event.text_with_all_modifiers() {
+                    key_text = modify;
+                }
                 // This is not a control key, so we rely upon winit to determine if
                 // this is a deadkey or not.
                 let keybinding_string = if let Some(escaped_text) = is_special(key_text) {
--- a/src/window/keyboard_manager.rs
+++ b/src/window/keyboard_manager.rs
@@ -18,7 +18,7 @@ pub struct KeyboardManager {
     shift: bool,
     ctrl: bool,
     alt: bool,
-    pre_is_dead: Option<char>,
+    prev_dead_key: Option<char>,
     logo: bool,
     ignore_input_this_frame: bool,
     queued_input_events: Vec<InputEvent>,
@@ -31,7 +31,7 @@ impl KeyboardManager {
             shift: false,
             ctrl: false,
             alt: false,
-            pre_is_dead: None,
+            prev_dead_key: None,
             logo: false,
             ignore_input_this_frame: false,
             queued_input_events: Vec::new(),
@@ -84,14 +84,14 @@ impl KeyboardManager {
 
                 if !self.should_ignore_input(&settings) {
                     // If we have a keyboard event this frame
-                    let mut pre_is_dead = self.pre_is_dead;
+                    let mut prev_dead_key = self.prev_dead_key;
                     for input_event in self.queued_input_events.iter() {
                         match input_event {
                             InputEvent::KeyEvent(key_event) => {
                                 // And a key was pressed
                                 if key_event.state == ElementState::Pressed {
                                     if let Some(keybinding) =
-                                        self.maybe_get_keybinding(key_event, &mut pre_is_dead)
+                                        self.maybe_get_keybinding(key_event, &mut prev_dead_key)
                                     {
                                         self.command_sender
                                             .send(SerialCommand::Keyboard(keybinding).into())
@@ -100,12 +100,12 @@ impl KeyboardManager {
                                 } else if key_event.state == ElementState::Released {
                                     // dead key detect here
                                     if let Dead(dead_key) = key_event.logical_key {
-                                        pre_is_dead = dead_key; //should wait for the next input text_with_all_modifiers, and ignore the next ime.
+                                        prev_dead_key = dead_key; //should wait for the next input text_with_all_modifiers, and ignore the next ime.
                                     }
                                 }
                             }
                             InputEvent::ImeInput(raw_input) => {
-                                if pre_is_dead.is_none() {
+                                if prev_dead_key.is_none() {
                                     self.command_sender
                                         .send(SerialCommand::Keyboard(raw_input.to_string()).into())
                                         .expect("Could not send keyboard ime string");
@@ -113,7 +113,7 @@ impl KeyboardManager {
                             }
                         }
                     }
-                    self.pre_is_dead = pre_is_dead;
+                    self.prev_dead_key = prev_dead_key;
                 }
 
                 // Regardless of whether this was a valid keyboard input or not, rest ignoring and
@@ -132,26 +132,26 @@ impl KeyboardManager {
     fn maybe_get_keybinding(
         &self,
         key_event: &KeyEvent,
-        pre_is_dead: &mut Option<char>,
+        prev_dead_key: &mut Option<char>,
     ) -> Option<String> {
         // Determine if this key event represents a key which won't ever
         // present text.
         if let Some(key_text) = is_control_key(key_event.logical_key) {
-            if pre_is_dead.is_some() {
+            if prev_dead_key.is_some() {
                 //recover dead key to normal character
-                let real_char = String::from(pre_is_dead.unwrap());
-                *pre_is_dead = None;
+                let real_char = String::from(prev_dead_key.unwrap());
+                *prev_dead_key = None;
                 Some(real_char + &self.format_keybinding_string(true, true, key_text))
             } else {
                 Some(self.format_keybinding_string(true, true, key_text))
             }
         } else {
-            let key_text = if pre_is_dead.is_none() {
+            let key_text = if prev_dead_key.is_none() {
                 key_event.text
             } else {
                 key_event.text_with_all_modifiers()
             };
-            *pre_is_dead = None;
+            *prev_dead_key = None;
             if let Some(ori_key_text) = key_text {
                 let mut key_text = ori_key_text;
                 if self.alt {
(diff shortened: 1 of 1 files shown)


vzex:
btw,this code
image
i don't konw the purpose, this will cause the editor received more string than the normal ime input, this is wrong on mac.

src/window/window_wrapper/keyboard_manager.rs
```diff
@@ -8,14 +8,18 @@ use crate::channel_utils::LoggingTx;
 use crate::settings::SETTINGS;
 use crate::window::KeyboardSettings;
 
+enum KeyOrIme {
```
last-partizan, line 11: Maybe this should be named in some generic way, like KeyboardEvent, or InputEvent?

src/window/window_wrapper/keyboard_manager.rs
```diff
@@ -8,14 +8,18 @@ use crate::channel_utils::LoggingTx;
...
+}
 pub struct KeyboardManager {
     command_sender: LoggingTx<UiCommand>,
     shift: bool,
     ctrl: bool,
     alt: bool,
     logo: bool,
     ignore_input_this_frame: bool,
-    queued_key_events: Vec<KeyEvent>,
+    queued_key_or_ime_events: Vec<KeyOrIme>,
```
last-partizan, line 22: And, after renaming enum, this should also be renamed to something generic.

last-partizan (approved):
Looks good to me.

last-partizan:
@clason @j4qfrost @LoipesMas @smolck please tale a look, and if you don't see any problems, i'll merge this.

last-partizan:
@vzex looks like you need to sync changes with latest main branch, to fix those failing checks.

vzex:
> @vzex looks like you need to sync changes with latest main branch, to fix those failing checks.

done

clason (changes requested):
I'm afraid this has broken dead keys (at least) on macOS: typing `´e` now inserts `éé` instead of a single `é` as on main (and similarly for other dead key combinations).

Also, typing `´´` (or other dead keys) makes dead keys "sticky", so any other dead key is inserted literally _twice_, and this continues while typing (other) dead keys.

LoipesMas:
Looks good to me, but I didn't have any issues before. At least I don't see any regression (on Linux).

vzex:
> I'm afraid this has broken dead keys (at least) on macOS: typing `´e` now inserts `éé` instead of a single `é` as on main (and similarly for other dead key combinations).
> 
> Also, typing `´´` (or other dead keys) makes dead keys "sticky", so any other dead key is inserted literally _twice_, and this continues while typing (other) dead keys.

as is mentioned before, https://github.com/neovide/neovide/pull/1083#issuecomment-991083715, this line code, i think it's not a right way, but i don't know what this code mean, so i can't fix this. But i think this is another bug hidden before, and it's  bring up after i fix the ime input.

clason:
Just to make clear: dead keys are working just fine on the main branch; this is a definite regression with this PR.

clason:
Protip: don't paste screenshots but links to the file through the Github API, like so: https://github.com/neovide/neovide/blob/7b6a8f31d2de1fca52a5cb039b97f5c7f92e5cbd/src/window/keyboard_manager.rs#L104-L108

This makes it easier for people to check the context (especially blame).

vzex:
> Protip: don't paste screenshots but links to the file through the Github API, like so:
> 
> https://github.com/neovide/neovide/blob/7b6a8f31d2de1fca52a5cb039b97f5c7f92e5cbd/src/window/keyboard_manager.rs#L104-L108
> 
> This makes it easier for people to check the context (especially blame). Here's the rationale explained: #767 (@smolck)

ok, i'll reproduce this later, I never used dead key before.

vzex:
> I'm afraid this has broken dead keys (at least) on macOS: typing `´e` now inserts `éé` instead of a single `é` as on main (and similarly for other dead key combinations).
> 
> Also, typing `´´` (or other dead keys) makes dead keys "sticky", so any other dead key is inserted literally _twice_, and this continues while typing (other) dead keys.

please check, I've tested on my mbp, it's ok now.

clason:
Sorry, now dead keys work (partly -- typing a dead key and a non-composing key swallows the latter: `´b`  should result in a literal `´b`, but you only get `´`), but Alt-modifiers are broken on non-US keyboards: Alt-8 should (and used to) insert `{`, but now I only get `8`.

vzex:
> 

Alright, I'll have a try, there are so many special cases.

clason:
> Alright, I'll have a try, there are so many special cases.

Yes indeed; I'm afraid there's no alternative to a lot of special cases depending on platform and language. Any attempt to "simplify" things _will_ break one or more use case in my experience. 

(The core of a GUI is input and output -- which otherwise a terminal will handle for you. People tend to focus on the output part, since that's where "the shiny" is, but the input part is just as important and arguably harder; especially if you want to be cross-platform without pulling in a huge framework like QT...)

vzex:
> > Alright, I'll have a try, there are so many special cases.
> 
> Yes indeed; I'm afraid there's no alternative to a lot of special cases depending on platform and language, and supporting IME necessarily means adding another. Any attempt to "simplify" things _will_ break one or more use case in my experience.
> 
> (The core of a GUI is input and output -- which otherwise a terminal will handle for you. People tend to focus on the output part, since that's where "the shiny" is, but the input part is just as important and arguably harder; especially if you want to be cross-platform without pulling in a huge framework like QT...)

I tested on windows and mbp, with chinese input, dead keys(alt+e u, alt+8(with us keyboard and austria keyboard layout)).All works well now(just for now).

vzex:
> > Alright, I'll have a try, there are so many special cases.
> 
> Yes indeed; I'm afraid there's no alternative to a lot of special cases depending on platform and language, and supporting IME necessarily means adding another. Any attempt to "simplify" things _will_ break one or more use case in my experience.
> 
> (The core of a GUI is input and output -- which otherwise a terminal will handle for you. People tend to focus on the output part, since that's where "the shiny" is, but the input part is just as important and arguably harder; especially if you want to be cross-platform without pulling in a huge framework like QT...)

Check again please

clason:
Yes, now modifier keys work again 🚀 

(The `´b` only printing `´` regression is still there, but that's minor if it means adding support for IME.)

vzex:
> Yes, now modifier keys work again 🚀
> 
> (The `´b` only printing `´` regression is still there, but that's minor if it means adding support for IME.)
 
I 've solve the problem now ,but I found a new problem, that is:
if I am in normal mode, when i am typing alt+q, I bind a hot-key to close buffer, but in the insert mode, the alt+q means show a character 'œ'. So I can't directly decide to send '<A-q>' or 'œ' before I know the vim mode. I think after bring in the ime support, the keyboardmgr should know what mode the vim in or just send the keyevent to 

> Yes, now modifier keys work again 🚀
> 
> (The `´b` only printing `´` regression is still there, but that's minor if it means adding support for IME.)

I have added more support for dead keys, and list every type of cases.

clason:
> I have added more support for dead keys, and list every type of cases.6 (Austria keyboard):[= =] =>´´q, this means [= = q] =>´´q

Yes, that works now, as well. (I don't use `alt` keybindings, since terminals don't allow them anyway, sorry...)

vzex:
> if I am in normal mode, when i am typing alt+q, I bind a hot-key to close buffer, but in the insert mode, the alt+q means show a character 'œ'. So I can't directly decide to send '' or 'œ' before I know the vim mode. I think after bring in the ime support, the keyboardmgr should know what mode the vim in or just send the keyevent to
> 
> >

Never mind, I found that Neovide didn't support alt on mac yet, so my pr doesn't break this future

last-partizan (changes requested):

src/window/keyboard_manager.rs
```diff
@@ -8,14 +9,19 @@ use crate::channel_utils::LoggingTx;
...
+enum InputEvent {
+    KeyEvent(KeyEvent),
+    ImeInput(String),
+}
 pub struct KeyboardManager {
     command_sender: LoggingTx<UiCommand>,
     shift: bool,
     ctrl: bool,
     alt: bool,
+    pre_is_dead: Option<char>,
```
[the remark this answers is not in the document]
vzex: agree

vzex:
So is there anyone can approval this?

src/window/keyboard_manager.rs
```diff
@@ -69,22 +84,42 @@ impl KeyboardManager {
...
-                                    .send(SerialCommand::Keyboard(keybinding).into())
-                                    .expect("Could not send keyboard ui command");
+                    let mut pre_is_dead = self.pre_is_dead;
+                    for input_event in self.queued_input_events.iter() {
+                        match input_event {
+                            InputEvent::KeyEvent(key_event) => {
+                                // And a key was pressed
+                                if key_event.state == ElementState::Pressed {
+                                    if let Some(keybinding) =
+                                        self.maybe_get_keybinding(key_event, &mut pre_is_dead)
```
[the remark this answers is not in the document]
last-partizan: @vzex this is unresolved, prev_dead_key should not be passed as mutable here. Or there is a reson why this impossible without passing prev_dead_key?

vzex:
@last-partizan I've uploaded the new changes, it's now  don't need to pass the prev_dead_key

last-partizan (approved):

last-partizan:
@Kethku i think this code is in better shape that it was before, and it's ready for your review.

Kethku:
Awesome. Thanks for the ping. I will take a look

src/window/keyboard_manager.rs
```diff
@@ -97,19 +132,27 @@ impl KeyboardManager {
...
+            }
+        } else {
+            let key_text = if self.prev_dead_key.is_none() {
                 key_event.text
+            } else {
+                key_event.text_with_all_modifiers()
             };
-
-            if let Some(key_text) = key_text {
+            if let Some(ori_key_text) = key_text {
```
Kethku, line 149: Super minor, where does the name `ori_key_text` come from?

Kethku: is it original?

Kethku:
This looks good to me! Thanks for the amazing persistence. This has been an open issue for a while. Once the lint build comes back I will merge

src/window/keyboard_manager.rs
last-partizan: Looks like "original" to me, i'll fix it in main branch :)

PR #1085, opened 2021-12-10
Fix cmdline args

last-partizan:
Add back `--verbosity` and `files_to_open`, removed in https://github.com/neovide/neovide/pull/1016

- Fix

- No

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -124,7 +124,8 @@ pub fn create_nvim_command() -> Command {
     let mut cmd = build_nvim_cmd();
 
     cmd.arg("--embed")
-        .args(SETTINGS.get::<CmdLineSettings>().neovim_args.iter());
+        .args(SETTINGS.get::<CmdLineSettings>().neovim_args.iter())
+        .args(SETTINGS.get::<CmdLineSettings>().files_to_open.iter());
 
     info!("Starting neovim with: {:?}", cmd);
 
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -124,8 +124,7 @@ pub fn create_nvim_command() -> Command {
     let mut cmd = build_nvim_cmd();
 
     cmd.arg("--embed")
-        .args(SETTINGS.get::<CmdLineSettings>().neovim_args.iter())
-        .args(SETTINGS.get::<CmdLineSettings>().files_to_open.iter());
+        .args(SETTINGS.get::<CmdLineSettings>().neovim_args.iter());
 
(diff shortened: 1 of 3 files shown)


LoipesMas:
Shouldn't neovim args be last?

last-partizan:
@LoipesMas yes, i'll fix this.

last-partizan:
@LoipesMas @Kethku 

In the last commit i removed "files_to_open" from struct and using neovim_args to pass files also. I think this is cleaner method.

LoipesMas:
Can't we just have file opening handled by neovim, by passing the filename in neovim args?
`neovide -- foo.txt`

Kethku:
Yeah I also think the files to open arg is superfluous

last-partizan:
@Kethku @LoipesMas nobody expects `--` in editor to open files, every other neovim gui (and gvim) - just open files if you give filenames. That's text editor.

Kethku:
Hmmm. Ok maybe I misunderstand how clap works. The way I read it, to open multiple files as is, one would have to pass them in with a `--files_to_open` flag. Which seems idk worse?

But I'm probably misunderstanding something

last-partizan:
@Kethku yeah, we're talking about positional arguments.

without `files_to_open` we need to use `neovide -- filename.rs`, with it - just `neovide filename.rs` like normal editors do.

I see you removed verbosity in the main branch and defaulted to log only errors. I think i'm okay with that. Do we have some variable to increase log level when needed?

LoipesMas:
> nobody expects `--` in editor to open files, every other neovim gui (and gvim) - just open files if you give filenames. That's text editor.

I don't use any editor like that, so I can't really say what the user expects.

Will it still work with, for example, `neovide --some-flag filename.rs -- --some-neovim-flag`? (I'm not too familiar with clap)

last-partizan:
@LoipesMas yes, this is exactly how it works.

```
./target/debug/neovide --multigrid README.md -- --noplugin
```

LoipesMas:
Nice, then I don't see any issues with this being merged.

Kethku:
Ok awesome. Re log level, when --log is present, it logs everything so the verbosity was only for if you want to print the log to console.

I think I'm good with the plan where the only way to access the non error logs is when --log is present. So I think we're good.

Last thing, I recently added some unit tests to this file. I think the files_to_open flag is a good case for a unit test because I want to make sure we are on the same page for how it should work. I will write a test which captures the functionality after merging this.

Thanks for the quick fix, and sorry for the confusion.

PR #1088, opened 2021-12-11
Fix Full Disk Access for universal macOS app

pawlowskialex:
The universal app bundle building process had some problems that prevented granting Full Disk Access to the app, those being:
- Due to build caching, there was a copy of `Neovide.app` inside the main `Neovide.app`
- Binary after `lipo` contained conflicting signatures, x86_64 segment having no signature, while aarch64 had an ad-hoc signature.

These problems resulted in macOS not being able to remember which binary was granted what permissions, so TCC prompt popped up on every file access. It was impossible to grant Full Disk Access to the app.

This PR removes the redundant copy of the `Neovide.app` and adds force Ad-Hoc codesign of the app bundle after `lipo`.

- Fix

- No

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -123,6 +123,7 @@ jobs:
       - name: Merge into a universal app
         run: |
           mkdir -p target/release/bundle/osx
+          rm -rf target/release/bundle/osx/Neovide.app || true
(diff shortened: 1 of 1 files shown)


shaunsingh:
lgtm

PR #1090, opened 2021-12-13
List Available Fonts when guifont=*

Kethku:
Display documentation for the guifont option, and list available fonts in a scratch buffer when the user sets guifont equal to "*".

guifont_list

- Feature

- No

### DIFF
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -4,7 +4,7 @@ use std::sync::Arc;
 use log::error;
 use log::trace;
 
-use nvim_rs::Neovim;
+use nvim_rs::{Neovim, call_args, rpc::model::IntoVal};
 use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver};
 
 use crate::bridge::TxWrapper;
@@ -15,7 +15,9 @@ use crate::windows_utils::{
 };
 
 // Serial commands are any commands which must complete before the next value is sent. This
-// includes keyboard and mouse input which would cuase problems if sent out of order.
+// includes keyboard and mouse input which would cause problems if sent out of order.
+//
+// When in doubt, use Parallel Commands.
 #[derive(Debug, Clone)]
 pub enum SerialCommand {
     Keyboard(String),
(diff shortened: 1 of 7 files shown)


LoipesMas:
This is definitely a step in right direction.
Perhaps a wild idea, but could we somehow *"inject"* vim script into this buffer which would make pressing a hotkey (i.e. Enter) on one of those lines do `set guifont={text of this line (with correct '\' and all)}`?
(Actual vim script part should be easy, just not sure if injection is doable)

dartacao:
> This is definitely a step in right direction. Perhaps a wild idea, but could we somehow _"inject"_ vim script into this buffer which would make pressing a hotkey (i.e. Enter) on one of those lines do `set guifont={text of this line (with correct '\' and all)}`? (Actual vim script part should be easy, just not sure if injection is doable)

If neovide can set buffer local commands, this should be as simple as
`:nnoremap <buffer> <CR> <cmd>lua vim.opt.guifont=vim.fn.getline('.')<CR>`

Here's a gif of me switching between Iosevka and FiraCode NF with this :)
Animation

LoipesMas:
Doing `nvim.command("nnoremap <buffer> <CR> <cmd>lua vim.opt.guifont=vim.fn.getline('.')<CR>").await.ok();` in `ui_commands.rs:174` seems to work as expected.
Pressing Enter on lines with incorrect font name still changes `guifont`, but the font used for rendering stays the same.
Maybe we could print a warning if neovide can't find current font? Something like `nvim.command("echoerr \"Couldn't find provided font.\"");`

Kethku:
This seems like a reasonable request. The other thing I'm concerned about with this flow is that it doesn't set the font size and is not stored in config. But I think that can just be explained in the text and it will be fine.

LoipesMas:
> The other thing I'm concerned about with this flow is that it doesn't set the font size  

We could do some more verbose lua scripting to get current height/width and just append it to the font name when selecting.

Storing in config is something the user has to do themselves, unless we would store the selected font as fallback for when guifont is not set, but I think that would be messy.

LoipesMas:
```
nvim.command("nnoremap <buffer> <CR> <cmd>lua tmpfnt=vim.fn.getline('.') for x in vim.o.guifont:gmatch(\":.*\") do tmpfnt=tmpfnt..x end vim.opt.guifont=tmpfnt<CR>").await.ok();
```
This should work fine. I'm not familiar with lua, so maybe it can be simpler/shorter.

Kethku:
Done. This seems good to me. Merging.

PR #1093, opened 2021-12-13
Enable default features on winit

### LINKED ISSUE #1092: Neovide on Ubuntu compiles only with explicit window-manager feature in Cargo toml
**Describe the bug**
Currently neovide does not compile on ubuntu.

**To Reproduce**
On Ubuntu on current main: `cargo build --release`

**Expected behavior**
Neovide compiles.

**Screenshots**
```
error: Please select a feature to build for unix: `x11`, `wayland`
  --> /home/jan/.cargo/git/checkouts/winit-58efae9df3f5264a/3b5462e/src/platform_impl/linux/mod.rs:10:1
   |
10 | compile_error!("Please select a feature to build for unix: `x11`, `wayland`");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

 - OS: Ubuntu 20.04
 - Neovide Version: current main 
 - Neovim Version: 0.5.2

As I use x11, I can fix this by specifying `x11` as a feature for winit in the `Cargo.toml`. But I dont get why this error started to occur now.

Is it possible to detect if x11 or wayland are available at compile time and choose that feature? Or add a feature to neovide which is passed to winit so on Ubuntu you could for example `cargo build --release --features x11` instead of modifying the `Cargo.toml` when the window manager changes?

### Possible solutions

1. Build winit with default features, should work for both wayland and x11.
2. Add x11 or wayland as a feature of neovide, which needs to be provided manually and enables the respective feature in winit.

Features specific target features do not seem to work due to https://github.com/rust-lang/cargo/issues/1197

jsosulski:
Currently neovide does not compile on Ubuntu (closes #1092).
If the second solution in #1092 is preferred, i.e. having x11 and wayland as features in neovide, I can also do that.

- Fix

- No

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ rand = "0.7"
 pin-project = "0.4.27"
 futures = "0.3.12"
 glutin = { git = "https://github.com/neovide/glutin", branch = "new-keyboard-all" }
-winit = { git = "https://github.com/neovide/winit", branch = "new-keyboard-all", default-features = false }
+winit = { git = "https://github.com/neovide/winit", branch = "new-keyboard-all" }
 gl = "0.14.0"
 swash = "0.1.4"
 clap="2.33.3"


LoipesMas:
I've recently thought about having separate features for wayland and x11, because right now it's building for both. But then you would have to specify the display server and also have two different builds to publish.

PR #1097, opened 2021-12-17
remove "-complete=expression" with -nargs=0

nganhkhoa:
Newer versions of Vim/Neovim throw error when using `-complete` with `-nargs=0`

https://github.com/neoclide/coc.nvim/issues/3414
https://github.com/neoclide/coc.nvim/pull/3210

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -109,7 +109,7 @@ pub fn build_neovide_command(channel: u64, num_args: u64, command: &str, event: 
     };
     if num_args == 0 {
         return format!(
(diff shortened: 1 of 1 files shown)


last-partizan:
Looks good to me. This is only for windows, and this shouldn't break older versions?

nganhkhoa:
This part of code currently only used in Windows, `NeovideRegisterRightClick` and `NeovideUnregisterRightClick`.

I was suprised to see this error to, a quick search on Google gives me the related issues and PR. I haven't tested on older version of Neovim.

PR #1099, opened 2021-12-19
Add remember_window_position setting

### LINKED ISSUE #887: Remember window position

LoipesMas:
- Fix

- No
## Description
Adds `remember_window_position` setting.
Closes #887 
Works for me, but feel free to also test this yourself.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2589,6 +2589,7 @@ dependencies = [
  "percent-encoding",
  "raw-window-handle",
  "scopeguard",
+ "serde",
  "smithay-client-toolkit",
  "unicode-segmentation",
  "wayland-client",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ dirs = "2"
 rand = "0.7"
 pin-project = "0.4.27"
 futures = "0.3.12"
-glutin = { git = "https://github.com/neovide/glutin", branch = "new-keyboard-all" }
+glutin = { git = "https://github.com/neovide/glutin", branch = "new-keyboard-all", features=["serde"] }
 winit = { git = "https://github.com/neovide/winit", branch = "new-keyboard-all" }
 gl = "0.14.0"
 swash = "0.1.4"
--- a/src/settings/mod.rs
+++ b/src/settings/mod.rs
@@ -9,7 +9,10 @@ use log::trace;
 use nvim_rs::Neovim;
 use parking_lot::RwLock;
 pub use rmpv::Value;
-pub use window_geometry::{maybe_save_window_size, parse_window_geometry, DEFAULT_WINDOW_GEOMETRY};
+pub use window_geometry::{
+    load_last_window_position, maybe_save_window_position, maybe_save_window_size,
+    parse_window_geometry, DEFAULT_WINDOW_GEOMETRY,
+};
 
 use crate::bridge::TxWrapper;
 use crate::error_handling::ResultPanicExplanation;
--- a/src/settings/window_geometry.rs
+++ b/src/settings/window_geometry.rs
@@ -1,6 +1,7 @@
 use crate::settings::SETTINGS;
 use crate::utils::Dimensions;
 use crate::window::WindowSettings;
+use glutin::dpi::PhysicalPosition;
 use std::path::PathBuf;
 
 #[cfg(unix)]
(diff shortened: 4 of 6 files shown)


Shatur:
Maybe enable it by default?

Shatur:
Oh, I see, `remember_window_size` is `false` too. That's weird defaults.

LoipesMas:
Changed them both to `true` by default

fredizzimo:
These changes are in conflict with my pull request https://github.com/neovide/neovide/pull/1111, which were merged at almost the same time. So right now the code tries to save to the wrong path and Neovim crashes on exit. 

I haven't checked completely yet, but it also looks like the code doesn't really work for saving both at the same time, since both save functions generates a completely new json file.

I will check that further and create a pull request that fixes both issues.

LoipesMas:
> These changes are in conflict with my pull request

I also just noticed that. Whoops. That should be an easy fix.

> looks like the code doesn't really work for saving both at the same time

Yep, it looks like it. My bad. I think maybe we should have a more complex settings structure that gets saved, i.e. a dict that has "window_size" and "window_pos" keys, that gets saved in json, so when saving one we would read the json, update only what's needed and write it back. Or maybe you have a better idea?

Also, I think you forgot to change [this line](https://github.com/fredizzimo/neovide/blob/fix_settings/src/settings/window_geometry.rs#L35) so it tries to load from directory and fails so it just always uses default geometry.

fredizzimo:
Yes, I currently looking at how to structure everything the best. 

At the moment it looks like it would make most sense to use the existing Settings structs, so that the window save and load functions would just use those, and don't touch the physical files at all. A dedicated settings module could be added to handle the actual saving and loading from disk. .

For the actual file, I'm not sure if JSON is the best format. It's OK for settings that are automatically read and write like these, but perhaps not for something that is manually edited. And at least for manual editing I would rather have an additional level of hierarchy so it becomes `Window.Position` for example. 

Additionally the manually set settings should probably be stored in a config file rather than data file, and never be touched automatically. There could also be an override system like this `command line` > `config` > `data`. So that a setting on the command line always overwrite anything set in the config files, and if you have a custom position set in your config file it would override the automatically saved data for example.

I don't think I have time to do big changes like these though, at least not right now, so I will just make a quick fix for the bugs, and save the settings as a nested dictionary in the JSON file. That should still allow for things to be refactored and improved later on, and even migrate to other data formats.

And yes, you are right about the I forgot to change the loading part, I will fix that.

LoipesMas:
I'm not sure if I understand you correctly, but I don't think there will be any settings saved in a file that are supposed to be edited manually. I wouldn't even call window position/size a "setting". All settings are handled by neovim, so the user edits their init.vim (or whatever) to change the settings. JSON is only used for "persistent storage" for neovide, which user doesn't even have to know about.
Manually set settings aren't saved, the user has to manually put them in their nvim config, but that's how (n)vim works, so it's fitting.
Unless we're talking about those settings that aren't accessible from neovim, such as multigrid. Those are only relevant at startup and are taken from command line or environment variables. We *could* create a config for those, which would be edited only by the user, but that's a different thing from storing window position/size.

fredizzimo:
Yes, I'm mostly talking about the stuff that's currently passed through the command line and I think there should be an option to read them from a config file in addition to passing them through the command line. Or at least that should be something that's taken into account if bigger changes are made to this system.

I would still call position/size a setting though, especially since the size maps to the geometry command line option. And both of them might actually be something someone would want to configure. But other things, that could potentially be saved in the `neovide-settings` might be purely internal stuff.

So as I see it, all three systems, user configuration, command line and the neovide-settings, can potentially control the same thing, and should therefore share as much as possible in common. 

Anyway, as I said earlier, those changes will remain for the future, since I don't have time to do it now, and it also needs a lot more thought to do properly. So for now, I will use the same json file and just add nested json objects and keys to be able to save multiple things. But if you prefer I can also do it without nesting,  and use prefixes like `window_` like you suggested instead.

LoipesMas:
After thinking about it more, config file is starting to make a lot more sense. I just got used to the way it is now, but a single config file instead of environment variables and what-not seems just better. It's definitely something to think about in the future.

I think nested json objects instead of prefixes are more reasonable and easier to extend, so go with that.

PR #1111, opened 2021-12-31
Fix for non-existing data directory

### LINKED ISSUE #1013: Neovide Windows 11 WSL2 not responding when quit
**Describe the bug**
I'm on Windows 11, I'm using neovide with WSL2. Everything seems to be fine except when I quit, neovide becomes not responding.

**To Reproduce**
1. Open WSL terminal
2. Run `neovide.exe --wsl`
3. Quit neovide either using `:q` or clicking close button
4. Neovide becomes not responding

**Expected behavior**
Neovide should exit easily.

**Screenshots**
https://user-images.githubusercontent.com/47636806/137082110-03095cae-7b9a-4d7d-a239-d380c549a6aa.mov

 - OS: Windows 11 Pro Version 10.0.22000 Build 22000
 - Neovide Version: Build from source with commit https://github.com/neovide/neovide/tree/7f76ad4764197ba75bb9263d25b265d801563ccf
 - Neovim Version:  v0.6.0-dev

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/7335655/neovide_rCURRENT.log)

**Additional context**

fredizzimo:
Neovide was crashing on exit if you had not run Neovim on Windows previously, and therefore no 'AppData\Local\nvim-data\' directory exists. 

While fixing this I noticed that `XDG_DATA_HOME` was not properly respected on Unix platforms and since I had to modify the same functions that deals with that, I fixed it at the same time.

Fix #1013

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -571,6 +571,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "dirs"
+version = "3.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "30baa043103c9d0c2a57cf537cc2f35623889dc0d405e6c3cccfadbc81c71309"
+dependencies = [
+ "dirs-sys",
+]
+
+[[package]]
 name = "dirs-sys"
 version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1325,7 +1334,7 @@ dependencies = [
  "cfg-if 0.1.10",
  "clap",
  "derive-new",
- "dirs",
+ "dirs 2.0.2",
  "euclid",
  "flexi_logger",
  "futures 0.3.17",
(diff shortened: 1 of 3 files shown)


LoipesMas:
Looks good to me.
Regarding respecting `XDG_DATA_HOME`, this change means that if someone has this changed to something else than default, now the settings won't be loaded (only once, since after that it will use the new dir) and the old settings will be left in the old directory? IIRC it's only used for remembering window size, so it's not something major nor a big file. Also I don't think there is a pretty solution to that. And linux users who have `XDG_DATA_HOME` changed and save window size are probably a small percentage of all users, so it shouldn't have much impact.

fredizzimo:
Yes, I had the same thought myself when I made the change. But I thought that someone that had customized the `XDG_DATA_HOME` would have noticed that it goes to the wrong place and reported a bug, so it's probably not a big issue.

LoipesMas:
Makes sense.
Thanks for the fix!

PR #1112, opened 2022-01-01
Fix window settings load and save

fredizzimo:
Some of the load and save paths that were broken in pull request https://github.com/neovide/neovide/pull/1111 and https://github.com/neovide/neovide/pull/1099 were fixed. And the JSON is now structured to properly support saving and loading of both position and size data, and is easily extendible for more fields and objects in the future.

### DIFF
--- a/src/settings/window_geometry.rs
+++ b/src/settings/window_geometry.rs
@@ -4,8 +4,6 @@ use crate::window::WindowSettings;
 use glutin::dpi::PhysicalPosition;
 use serde::{Deserialize, Serialize};
 use std::path::PathBuf;
-#[cfg(unix)]
-use xdg;
 
 const SETTINGS_FILE: &str = "neovide-settings.json";
 
--- a/src/settings/window_geometry.rs
+++ b/src/settings/window_geometry.rs
@@ -2,6 +2,7 @@ use crate::settings::SETTINGS;
 use crate::utils::Dimensions;
 use crate::window::WindowSettings;
 use glutin::dpi::PhysicalPosition;
+use serde::{Deserialize, Serialize};
 use std::path::PathBuf;
 #[cfg(unix)]
 use xdg;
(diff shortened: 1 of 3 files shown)


src/settings/window_geometry.rs
```diff
@@ -48,50 +70,49 @@ pub fn try_to_load_last_window_size() -> Result<Dimensions, String> {
...
 }
 
-pub fn maybe_save_window_size(grid_size: Option<Dimensions>) {
+fn maybe_save_window_size(grid_size: Option<Dimensions>) -> Dimensions {
     let settings = SETTINGS.get::<WindowSettings>();
-    let saved_window_size = if settings.remember_window_size {
+    if settings.remember_window_size {
         grid_size.unwrap_or(DEFAULT_WINDOW_GEOMETRY)
     } else {
         DEFAULT_WINDOW_GEOMETRY
```
fredizzimo, line 88: Note, that I retained the old behaviour here of resetting to the default value. But I think it would make more sense to not modify the saved value at all.

src/settings/window_geometry.rs
```diff
@@ -13,6 +14,23 @@ pub const DEFAULT_WINDOW_GEOMETRY: Dimensions = Dimensions {
     height: 50,
 };
 
+#[derive(Serialize, Deserialize)]
+struct SerializedWindowSettings {
```
LoipesMas, line 18: I'm not sure about the "Serialized" prefix in name, here and in `SerializedSettings`. But I also don't have a better idea.

LoipesMas (changes requested):

fredizzimo:
Sorry for being so uncareful regarding the naming, I was rushing the original pull request, since the build was broken and I didn't have much time to look into it today. But I have now updated it, and I actually removed both `maybe_save` functions, since i could not think of a good name for them, and they did not really need to be separate functions. I also rewrote them a bit, but I'm totally new to Rust, so I'm not sure if I did it in a good way or if it can be simplified further.

I renamed `SerializedSettings` to `PersistentSettings`

LoipesMas (approved):

LoipesMas:
Looks good to me.
Removing those function seems reasonable.

Also, small thing that I only noticed just now: in `window_geometry.rs` we have `use xdg;` which is unused, because later we "explicitly" do `xdg::BaseDirectories::...`. So it can be either removed, or changed to `use xdg::BaseDirectories` and the call below changed to just `BaseDirectories::...`. But personally I'd just remove it.

I don't see any other issues, so fix that and I'll merge this.

LoipesMas:
Thanks!

Issue #1089, opened 2021-12-13
Create a website


Kethku:
This issue is to discuss pros and cons for building a website and what it might contain.

Some pros:
- Can make a slick landing page which sells the key ideas behind neovide
- Simple easy to read place for FAQs and initial setup instructions
- Feels more official

Some cons:
- Extra thing to maintain
- Not clear what it gives over just improving the existing readme
- Requires somebody with experience to design it

I personally don't have much opinion either way other than that I don't feel confident I could design something better than what we already have, so this is sorta up to whoever wants to take it on.

dineshKumar777:
This we can do. It will be most likely a static website so maintaining won't be that hard once project got stabilized.
Also you don't need to do maintain both Readme and website contents. We can parse contents from Github markdown automatically.

cljoly:
> We can parse contents from Github markdown automatically.

In case you don’t already have a tool in mind @dineshKumar777, I use [RISS][] to reuse a Readme on a static website.

*disclaimer*: I wrote RISS

[RISS]: https://cj.rs/readme-in-static-site/

cljoly:
Also, this may be a good theme (even though I haven’t used Zola much): https://juice.huhu.io/

Kethku:
I have used and really like Zola. Good find

cljoly:
I’ve put together a website in a subdirectory [on this branch](https://github.com/cljoly/neovide/tree/site). There may be some rough edges still, but this should give an idea of a possible way forward on this. The way it is set up, update to images in assets and README.md should be propagated automatically.

Let me know if you want me to open a PR to discuss this further.

![image](https://user-images.githubusercontent.com/7347374/147296543-e5202948-c7f3-4b8b-9974-6de00d31755a.png)

![image](https://user-images.githubusercontent.com/7347374/147296570-ced2eb7c-da7d-4a3d-ab4b-f00ceb23c48d.png)

PyGamer0:
That looks amazing

cljoly:
Thanks, I’ll try to finish documenting the code and to make a PR over the weekend.

cljoly:
I’ve made the PR, we can of course further refine the website look and content, but hopefully that’ll be a starting point.

PR #1114, opened 2022-01-02
Starting point for a website

cljoly:
- Create a simple Zola website in a subdirectory
- Automate website content update from the README.md and existing graphical assets
- Document how to build and edit the website
- Deploy automatically to GitHub Pages

- Feature (#1089)

- No


cljoly:
See https://cljoly.github.io/neovide/ for a preview.

**Things that need to be done right before merge:**
1. [ ] Make sure that there is a `TOKEN` as described in https://github.com/marketplace/actions/zola-deploy-to-pages#secrets

cljoly:
If we want to have a custom domain, I have neovide.eu.org (free domain managed by an non-profit that nether expires, so low-maintenance), I’m willing to update DNS records as may be needed (and to transfer it, although I’m unsure about the process).

Kethku:
Good point. Another option is I could grab up neovide.dev (or .com) and pay for it with sponsorship money. That way it could be a top level domain.

What do you think?

Kethku:
Can't say enough how awesome this looks btw :) Thanks so much for pushing this forward

cljoly:
> Good point. Another option is I could grab up neovide.dev (or .com) and pay for it with sponsorship money. That way it could be a top level domain.
> 
> What do you think?

As you want. .eu.org is [in](https://publicsuffix.org/list/public_suffix_list.dat) the known [public suffix list](https://publicsuffix.org/), so it should behave like say `.co.uk`. But maybe a .dev is more appealing.

cljoly:
> Can't say enough how awesome this looks btw :)

I've mostly used the defaults of the theme 😉

Kethku:
Maybe I'm just being weird, but I'm not in the EU (or uk) and I know people use neovide from all over, so I'm sorta inclined to get a domain which is location agnostic. Idk if .dev does that or not, but I figure its not a bad way to go

cljoly:
> Maybe I'm just being weird, but I'm not in the EU (or uk) and I know people use neovide from all over, so I'm sorta inclined to get a domain which is location agnostic. Idk if .dev does that or not, but I figure its not a bad way to go

Sure.

Kethku:
Ok I grabbed it up. Its getting late where I'm at, so I'm gonna drop off for now, but this looks good to me :) I see that there is a requirement to add a token. I'll take a look at what that entails soon unless somebody else gets to it sooner.

cljoly:
Awesome, thanks for the review!

shaunsingh:
Looks awesome!

Small nitpick, code blocks for me have a black background but the text itself has a white background. Occurs on macOS 12 on Safari, not sure if other browsers have the issue
image

cljoly:
> Looks awesome!
> 

Thanks!

> Small nitpick, code blocks for me have a black background but the text itself has a white background. Occurs on macOS 12 on Safari, not sure if other browsers have the issue image

Yeah, I agree it’s not ideal, it happens to me as well, both on Android and a desktop Linux with a Chromimum-based browser and Firefox.

It should probably be reported upstream to the theme, as it seems to be caused by this CSS:
image

cljoly:
Shall we set up the token and merge?

I would think the upstream issue [with code text](https://github.com/neovide/neovide/pull/1114#issuecomment-1008282292) is minor enough to be addressed separately.

Kethku:
Done! I will look into what it will take to target the neovide.dev domain to the new website. Thanks again!

cljoly:
I'll send a follow up PR to make Zola generate for neovide.dev (and create the CNAME where GitHub pages expects it.)

Issue #1096, opened 2021-12-15
Fonts render with significant spacing


dartacao:
**Describe the bug**
Most fonts render oddly in my machine. In particular, FiraCode NF has unusually large gaps between characters (see screenshots).

**To Reproduce**
1. Open Neovide with font as FiraCode\ NF, whatever height, on windows 10. Not sure if this is a reproducible bug on other machines.

Tried reinstalling FiraCode, using other ligature'd fonts like Iosevka. Similar issues with large gaps between chars.

**Expected behavior**
Fonts should render with no gaps between ligatures or characters in general.
For example, Wezterm renders FC NF (same font) as:
![image](https://user-images.githubusercontent.com/2131630/146268325-3b89bb70-7aaf-49e2-8bfe-0418a9030109.png)

**Screenshots**
![image](https://user-images.githubusercontent.com/2131630/146268067-24caf5d2-031e-41c1-bdc2-473758a79353.png)
Note that characters in statusline are also unusually spaced, no matter the font size.

**Desktop (please complete the following information):**
 - OS: Windows 10 
 - Neovide Version 0.8.0
 - Neovim Version 0.6.0

Logs:
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/7722808/neovide_rCURRENT.log)
[neovide_r2021-12-15_21-29-49.log](https://github.com/neovide/neovide/files/7722807/neovide_r2021-12-15_21-29-49.log)

exten:
1. change the font , code in `src/renderer/fonts/font_loader.rs`
like this `static DEFAULT_FONT: &[u8] = include_bytes!("../../../assets/fonts/Fira Code Regular Nerd Font Complete.otf");`
2. rebuilder

dartacao:
> 1. change the font , code in `src/renderer/fonts/font_loader.rs`
> like this `static DEFAULT_FONT: &[u8] = include_bytes!("../../../assets/fonts/Fira Code Regular Nerd Font Complete.otf");`
> 2. rebuilder

I'm sorry but I don't quite understand what I should do to be honest...

dartacao:
> 1. change the font , code in `src/renderer/fonts/font_loader.rs`
>        like this `static DEFAULT_FONT: &[u8] = include_bytes!("../../../assets/fonts/Fira Code Regular Nerd Font Complete.otf");`
> 
>     2. rebuilder

I think I did what you asked successfully. However, the issue seems to persist... Any other suggestions?

[...]

PR #1116, opened 2022-01-02
Fix the font spacing

fredizzimo:
Previously the font width was always rounded down to the nearest integer, which typically caused the fonts to look to squashed. To correct the issue a fudge factor, which scales the fonts so that the width is always an exact integer, was introduced.

Here's a comparison image, the top one is DejaVuSansMono NF:h11 without these changes, the middle one has the changes, the bottom row is the same font rendered with Windows 11 Notepad

image

A potential additional problem that I noticed is that we use the average glyph width for the metrics, while other code bases I have seen uses a wide character like `M` as the base. For example here https://github.com/microsoft/terminal/blob/main/src/renderer/dx/DxFontRenderData.cpp#L719

### DIFF
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -25,19 +25,23 @@ pub struct CachingShaper {
     blob_cache: LruCache<ShapeKey, Vec<TextBlob>>,
     shape_context: ShapeContext,
     scale_factor: f32,
+    fudge_factor: f32,
 }
 
 impl CachingShaper {
     pub fn new(scale_factor: f32) -> CachingShaper {
         let options = FontOptions::default();
         let font_size = options.size * scale_factor;
-        CachingShaper {
+        let mut shaper = CachingShaper {
             options,
             font_loader: FontLoader::new(font_size),
             blob_cache: LruCache::new(10000),
             shape_context: ShapeContext::new(),
             scale_factor,
-        }
+            fudge_factor: 1.0,
+        };
+        shaper.reset_font_loader();
+        shaper
     }
 
     fn current_font_pair(&mut self) -> Arc<FontPair> {
@@ -54,7 +58,7 @@ impl CachingShaper {
     }
 
     pub fn current_size(&self) -> f32 {
-        self.options.size * self.scale_factor
+        self.options.size * self.scale_factor * self.fudge_factor
     }
 
     pub fn update_scale_factor(&mut self, scale_factor: f32) {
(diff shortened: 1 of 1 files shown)


fredizzimo:
Since it's very likely that this will fix https://github.com/neovide/neovide/issues/1096, I updated this pull request and added code for measuring the width of the letter 'M', and use that instead of the average width if available.

fredizzimo:
I just noticed an issue, but I will fix it tomorrow. 

When I set the font to `Consolas:h11`, I get the following
```
TRACE [neovide::renderer::fonts::caching_shaper] Fudge factor: 0.99
TRACE [neovide::renderer::fonts::caching_shaper] Fudged font size: 14.55px
TRACE [neovide::renderer::fonts::caching_shaper] Fudged font width: 8.00px
TRACE [neovide::renderer::grid_renderer] Updated font dimensions: Dimensions { width: 7, height: 18 }
 ```
So it rounds down the 8.0 to 7 for some reason.

**Comment by fredizzimo** [comment]
No I'll fix it now, I had accidentally removed the rounding when converting to int. But just to be safe I'll use the floor+0.5 trick instead.

**Comment by Kethku** [comment]
I will test this now!

**Comment by Kethku** [comment]
Tried it out and it works great for me :) Thanks so much

File: src/renderer/fonts/caching_shaper.rs [hunk 1]
```diff
@@ -79,29 +83,57 @@ impl CachingShaper {
...
+        let mut font_size = self.current_size();
         trace!("Using font_size: {:.2}px", font_size);
+        self.font_loader = FontLoader::new(font_size);
+        let (metrics, font_width) = self.info();
+        trace!(
+            "Font width: {:.2}px {:.2}px",
+            font_width,
+            metrics.average_width
+        );
+        self.fudge_factor = font_width.round() / font_width;
```
last-partizan, line 99: @fredizzimo this `fudge_factor` is actually has broken something. Now we can't use non-exact font sizes, and for some reason, my font (JetBrains Mono, looks good only with some non-integer sizes).

And difference is pretty big, compare this two screens:

With fudging set to 1:

image

```
TRACE [neovide::renderer::fonts::caching_shaper] Fudge factor: 1.00
TRACE [neovide::renderer::fonts::caching_shaper] Fudged font size: 23.53px
TRACE [neovide::renderer::fonts::caching_shaper] Fudged font width: 14.12px
```

With actual fudging:
image

```
TRACE [neovide::renderer::fonts::caching_shaper] Fudge factor: 0.99
TRACE [neovide::renderer::fonts::caching_shaper] Fudged font size: 23.33px
TRACE [neovide::renderer::fonts::caching_shaper] Fudged font width: 14.00px
```

That little difference between 14.00 and 14.12 changes something in font rendering, and at 14.12 it renders properly, with correct height. (JetBrainsMono actually uses that `Increased height for better reading expirience` at ther website, and i can clearly notice when something is wrong).

I think it would be good to add option to disable this fudging. Or maybe i'll look for options to resolve issue with squashed fonts without scaling fonts to nearest integer.

Where is that rounding you talked about in pr description?

> Previously the font width was always rounded down to the nearest integer

Issue #983, opened 2021-09-28
Neovide fails to read linux environment variables when launched with --wsl


HawkinsT:
**Describe the bug**
When launching Neovide from Windows with --wsl it doesn't read in all environment variables from .profile (or .bashrc etc.), only $PATH (so far as I can tell) which leads to inconsistent behaviour. For example, executing commands via `:!` will act differently to executing them from `:term` (which does read in environment variables). In order to get around this I am (for now) having to specifically set the environment variables I require in my init.vim.

**To Reproduce**
1. Set $TEST to some value in .profile
2. In neovide enter `:!echo $TEST`
3. In neovide enter `:term echo $TEST`
4. See difference in output.

**Expected behavior**
Neovide --wsl should behave the same as nvim (when nvim is launched from within a wsl2 shell); that is, it should read in all environment variables since many vim scripts and packages can depend on these.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Neovide Version main compiled by me (reported as 0.8.0 in Windows properties menu)
 - Neovim Version v0.6.0-dev

HawkinsT:
Addendum: It seems most environment variables are actually being set correctly. One I have had issues with though is `$DISPLAY`, which always seems to be null if not set within init.vim.

HawkinsT:
Hmm, I'm using main (compiled two days ago).

HawkinsT:
I'll try to test later, but I think at least part of the issue is Neovide using `$SHELL` blindly in https://github.com/neovide/neovide/blob/main/src/bridge/mod.rs and assuming POSIX compliance.

I've been using bash as my default shell until now with fish called at the end of my .bashrc but had issues with `$DISPLAY` not being set when in Neovide. Instead of calling fish from bash, if I make fish my default shell I've discovered that Neovide completley craps out and can't find anything on `$PATH` (e.g. I get a bunch of errors when launching Neovide and `:checkhealth` finds almost nothing, despite `:echo $PATH` and e.g. `:!which python3`, working correctly and being specified explicitly in config.fish as well as .profile and .bashrc), even though neovim within wsl has no issues. Changing `$SHELL` to `/bin/bash` in either my init.vim or config.fish files fixes this and the `$DISPLAY` issue.

If shell POSIX compliance is going to be assumed by Neovide, then perhaps `$SHELL` should be replaced with just invoking bash directly within mod.rs, or at least first looking for bash and using if found?

fredizzimo:
[...]

Issue #950, opened 2021-09-11
no node in path when running in --wsl mode


Lite5h4dow:
**Describe the bug**
when you run neovide from any terminal (WSL, Powershell, CMD) in WSL mode, there is no node in path despite if you launch nvim from the WSL terminal it is there because .zshrc initialised the path.

**To Reproduce**
1. run `neovide.exe --wsl` from any environment
2. enter `<ESC>:!which node<CR>`
3. See `node not found`
4. Coc will crap itself on launch if you have that installed 

**Expected behavior**
I would expect node to be in path

**Screenshots**
![image](https://user-images.githubusercontent.com/18059545/132961123-13564c9f-4d3b-4b64-82d0-2afedd02aa63.png)

**Sidenote**
if anyone knows a decent alternative to Coc that dosent require node, and has Typescript support (Deno and Node) i would be open to switching.

PyGamer0:
Does it work if you don't use neovide? (calling neovim from wsl)

Lite5h4dow:
@PyGamer0 yes it does.

Lite5h4dow:
nvim in terminal:
![image](https://user-images.githubusercontent.com/18059545/133072303-da8b6582-ca48-48d0-8e06-822d19157d79.png)

Kethku:
I don't know much about linux integration with windows.

Recently a change went in to try to harvest the path from the wsl bash before running to fix issues like this. The relevant line is here:
https://github.com/neovide/neovide/blob/main/src/bridge/mod.rs#L39

Sounds like you're using zsh? I don't know, but is it possible that that config isn't loaded with the wsl command linked above?

fredizzimo:
I created a pull request that should fix this issue https://github.com/neovide/neovide/pull/1118

fredizzimo:
I merged #1118. But see the comments there, there might be unexpected problems, so open a new issue if you find any.

williamJmelton:
I'm also having this issue having just built Neovide from source last week. When launching neovide.exe --wsl I don't get the executables added to my path from my .zshrc. When I launch with --wsl with the WSL distro's default login shell set to bash I don't have the PATH problems, but I do when using zshrc

fredizzimo:
Hm, are you sure if you were building the latest master version? If not I will check later today or tomorrow to see if something has broken, since there were some refactoring to the code after I wrote it, and I haven't updated to that version myself yet.

Hackder:
[...]

PR #1118, opened 2022-01-03
Run WSL Neovim inside a login shell

fredizzimo:
Currently WSL is launched with a hack to launch a temporary interactive shell in order to fix the path inside Neovim, but it leaves all other environment variables unset, and does not correctly load all your login scripts.  See this flowchart from https://shreevatsa.wordpress.com/2008/03/30/zshbash-startup-files-loading-order-bashrc-zshrc-etc/

flowchart

This  pull request changes that, so that Neovim is launched inside an interactive shell, and all settings and scripts are read correctly, at least I haven't seen any problems on my system. 

I also refactored the command line processing to unify the code paths taken on different systems and to avoid duplicating the code.

Note that there's a tiny problem, it does not necessarily work if you spawn another shell from within your bash scripts, like the Genie instructions here for example suggests https://github.com/arkane-systems/genie/wiki/Automatically-start-genie-on-every-shell-session. I had that problem when using Zsh as the default shell, but not when using Bash. I'm not sure if that problem can be solved at all, but I got rid of the problem by switching to https://github.com/nullpo-head/wsl-distrod, which is superior to Genie in many other ways too IMO.

My suggested fix for this is to document this, and instruct the users to fix their shell scripts to not start any subshells, another option is to add some command line parameter to choose between interactive and login shell.

Fixes https://github.com/neovide/neovide/issues/950
Fixes https://github.com/neovide/neovide/issues/983

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -29,8 +29,12 @@ fn set_windows_creation_flags(cmd: &mut Command) {
     cmd.creation_flags(0x0800_0000); // CREATE_NO_WINDOW
 }
 
-#[cfg(windows)]
-fn platform_build_nvim_cmd(bin: &str) -> Option<Command> {
+fn build_nvim_cmd_with_args(bin: &str) -> Command {
+    let mut args = Vec::<String>::new();
+    args.push("--embed".to_string());
+    args.extend(SETTINGS.get::<CmdLineSettings>().neovim_args);
+
+    #[cfg(windows)]
     if SETTINGS.get::<CmdLineSettings>().wsl {
         let mut cmd = Command::new("wsl");
         cmd.args(&[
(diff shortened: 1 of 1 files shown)


Lite5h4dow:
looks good. thanks for this.

fredizzimo:
Have anyone been able to test this yet, does it work for you? Do you think it's a good default behaviour, and do we need to add some flags to revert back to the old behaviour?

Kethku:
This looks good to me imho. My wsl instance is pretty borked atm (I think I accidentally deleted some of the files trying to free up space). But the refactoring to simplify creation of the neovim process is worth it in and of itself. @fredizzimo feel free to merge w/e you feel its ready

fredizzimo:
I will merge this now, since I think that's the best way of getting actual feedback. I have been using it myself for over a month without problems on a couple of different machines. But obviously that's not a big sample. 

If people do get problems, a likely cause is the one that I described above. Some init script that spawns an inner shell. This needs to at least be documented some way, But I think the discussion of exactly how, and what can be done about it can be discussed in some issue. 

Also some sort of error handling needs to be done, since right now it just hangs if the nvim process can be launched, but the communication fails, but that's a general problem and should be handled on all platforms.

HUAHUAI23:
hi I'm starting neovide via neovide --wsl, how do I source .bashrc to set some environment variables when I start neovide?

PR #1119, opened 2022-01-03
Config file

LoipesMas:
# What kind of change does this PR introduce?
Adds configuration file for some settings. It's described in more detail in `website/docs/config-file.md` file.

# Did this PR introduce a breaking change?
No


shaunsingh:
Personally I think such settings should be handled from within neovim (via `vim.g.<option>`), thats just what I think though. Would love to hear others' perspectives as well.

clason:
I've made my thoughts known elsewhere, but briefly I am _strongly_ in favor of a dedicated config file -- I don't see a reason for shoehorning configuration for one program (Neovide) into that of another program (Neovim). I know that was historically the done thing (coming from vanilla Vim, where the GUI was _not_ a separate program). I believe this no longer serves a purpose in the Neovim and Lua world, for basically two reasons:

1. Neovim GUIs are increasingly diverse, and a centralized configuration from Neovim only makes sense if these settings can apply to multiple GUIs (like the old `GuiFont` directive, which however not even all GUIs support). 
2. Neovim itself is moving away from dumping a bucket of options into the global scope and more towards a `vim.foo.config({ ... })` style.

So why not do something that from the beginning is tailored towards Neovide? (Be it TOML, YAML, JSON, or a text file, as long as I can plop it in `.config/neovide` and keep it under version control.)

fredizzimo:
The situation is a bit tricky. Some options like `multigrid`, paths to nvim and so on, absolutely needs to be set before connecting to Neovim. Currently these are set by either command line options or environment variables. Well `multigrid` could theoretically be moved to, if we first starts nvim with a special script to read the setting, but the other ones can still not be moved there.

A settings file in *addition* to those would be ideal IMO, where the settings can be specified in both, but the command line always overrides the stored settings. 

But when it comes to runtime stuff, that you can change at runtime, I think all of that should be inside vim. That allows you to bind keybindings and stuff. Even the current frameless command line option, should be there IMO, which would allow you to toggle frames on and off, on demand.

I think the regular `g:` style options are fine, since they are easy to use inside Neovide, they even support dictionaries for more complex things. But I think it's possible to pass custom scripts, so a lua based interface for those could also be provided. 

For guifont and Neovide, I think that needs to be deperecated in favor of something more structured and extensible, to support all kinds of stuff like custom font weights, separate bold and non-bold fonts, stylistic sets and so on. It's hard to get that all into a single string, and it's non-standard anyway.

clason:
For the sake of completeness: I was only considering one-time setup options and not runtime changes (which I don't see a lot of use for, to be honest); for the latter, I would similarly prefer exposing functionality via a Lua module (similarly to Uivonim) rather than global variables.

LoipesMas:
I agree with pretty much everything fredizzimo said. Some things are impossible (or at least hard) to change at runtime and don't make sense to change at runtime. Config file makes sense for those settings. Rather than specifying command line options or setting environment variables.
But other things should probably stay configured by nvim, for keybindings etc.
But then there are two config files (nvim config and neovide config).
Not sure what's the best solution for all that.
I think trying to parse nvim config before starting nvim (however it would be implemented) is the worst solution though.

clason:
> But then there are two config files (nvim config and neovide config).

Yes, that is exactly what I am suggesting. I don't consider this bad. (Just to make it clear: anything Neovim-specific should of course stay in `init.vim` and friends -- it makes no sense to "take over" or (worse) duplicate Neovim configuration!)

shaunsingh:
> Personally I think such settings should be handled from within neovim (via vim.g.<option>), thats just what I think though. Would love to hear others' perspectives as well.

Note that this comment was supposed to be for any settings set *after* we connect to neovim. I didn't realize this PR was for settings connecting *before* that. In that case, I supporting putting options such as multigrid under a configuration file

src/settings/config.rs
```diff
@@ -0,0 +1,135 @@
...
+fn neovide_config_path() -> PathBuf {
+    let xdg_dirs = xdg::BaseDirectories::with_prefix("neovide").unwrap();
+    xdg_dirs.get_config_home()
+}
+
+#[cfg(windows)]
+fn neovide_config_path() -> PathBuf {
+    let mut data_path = dirs::home_dir().unwrap();
+    // I have no idea where this should be
+    todo!("I have no idea where this should be");
```
MultisampledNight, line 94: Following NeoVim, I'd say `~/AppData/Local/neovide/`.

fredizzimo: It's not fully that simple, it should follow the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), which means that it can be changed by environment variables.

Since it can be a bit tricky we already have a library in use and use it here
https://github.com/neovide/neovide/blob/92bc1725f1733547eb0ae25b740425f03f358c2a/src/settings/window_geometry.rs#L42-L46

Note that in this case, you should use [get_config_home](https://whitequark.github.io/rust-xdg/xdg/struct.BaseDirectories.html#method.get_config_home) instead of `get_data_home`, or perhaps better `place_config_file` which will take care of creating the required directories.

fredizzimo: Ah sorry, we were talking about Windows, there I would put it in `~/AppData/Roaming/neovide`. The difference is that roaming is synchronized between machines if you are on a domain, so if you use multiple machines at work for example, the settings would be synchronized.

Lite5h4dow:
im very in favor of this addition, ive stated in other issues that i agree with the idea of having a seperate config file for neovide.

khjorth:
This would be very useful for those of us who start nvim instances on remote systems and then connect to them with Neovide. Currently, we have to configure Neovide on each remote system, which is rather unusual for a locally running client.

src/settings/config.rs
correabuscar:
```
$ cat ~/.local/share/nvim/neovide-settings.json
{"window":{"Windowed":{"position":{"x":0,"y":0},"size":{"width":100,"height":50}}}}
```


> It's not fully that simple, it should follow the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), which means that it can be changed by environment variables.
> 
> Since it can be a bit tricky we already have a library in use and use it here
> 
> https://github.com/neovide/neovide/blob/92bc1725f1733547eb0ae25b740425f03f358c2a/src/settings/window_geometry.rs#L42-L46
> 
> Note that in this case, you should use [get_config_home](https://whitequark.github.io/rust-xdg/xdg/struct.BaseDirectories.html#method.get_config_home) instead of `get_data_home`, or perhaps better `place_config_file` which will take care of creating the required directories.

LoipesMas:
Hey everyone, it has been... *a while* :sweat_smile: 
But I decided to get back to it and finish this finally.

Let me know if there are any problems. But if it's something bigger and not a major issue (e.g. adding a particular setting which would be complicated to implement), then I'd rather leave it for a separate PR. 
The people have waited long enough.

Cargo.toml
```diff
@@ -22,7 +22,7 @@ cfg-if = "1.0.0"
 clap = { version = "4.0.8", features = ["cargo", "derive", "env"] }
 csscolorparser = "0.6.2"
 derive-new = "0.5.9"
-dirs = "4.0.0"
+dirs = "5.0.0"
```
fredizzimo, line 25: Is there a reason for using `dirs` instead of the library that we already are using `xdg`?

Perhaps Windows support? But in that case I think we can remove the xdg library, and use only `dirs`

LoipesMas: Indeed, `dirs` also has Windows support. We could probably replace our `xdg` usage with `dirs`.
Should I do it in this PR or we'll leave that for a separate one?

shaunsingh: One weird thing with `dirs` is on macOS, `config_dir()` refers to `$HOME/Library/Application Support`, though you probably want it to be under `$HOME/.config`. 

Same thing with caches and local/share, dirs uses `$HOME/Library/Caches` and `$HOME/Library/Application Support` when most people expect it to be `~/.cache` and ~/.local/share`

I think it makes more sense to keep using `xdg` for linux/mac then "hardcode" the path for windows

LoipesMas: I'm not familiar with macOS directory structure, but this makes sense. I changed it to use `xdg` on mac/linux

last-partizan:
@LoipesMas let's finish this. Need any help or feedback?

src/main.rs
```diff
@@ -142,6 +145,14 @@ fn protected_main() {
     #[cfg(target_os = "windows")]
     windows_attach_to_console();
 
+    match Config::load_from_path(&config_path()) {
+        Ok(config) => config.write_to_env(),
+        Err(err) => {
+            eprintln!("{err}");
+            error!("{err}");
```
last-partizan, line 152: Do we need them both?

src/settings/config.rs
```diff
@@ -0,0 +1,89 @@
...
+            env::set_var("NEOVIDE_MAXIMIZED", maximized.to_string());
+        }
+        if let Some(vsync) = self.vsync {
+            env::set_var("NEOVIDE_VSYNC", vsync.to_string());
+        }
+        if let Some(srgb) = self.srgb {
+            env::set_var("NEOVIDE_SRGB", srgb.to_string());
+        }
+        if let Some(no_idle) = self.no_idle {
+            env::set_var("NEOVIDE_NO_IDLE", no_idle.to_string());
```
last-partizan, line 59: we're changed variable name to `NEOVIDE_IDLE` recently, so, behaviour is reversed. We should use `idle: false`  here too.

src/settings/config.rs
```diff
@@ -0,0 +1,89 @@
...
+    pub vsync: Option<bool>,
+    pub srgb: Option<bool>,
+    pub no_idle: Option<bool>,
+    pub neovim_bin: Option<PathBuf>,
+    pub frame: Option<Frame>,
+    pub geometry: Option<Dimensions>,
+}
+
+impl Config {
+    pub fn write_to_env(&self) {
```
last-partizan, line 45: by the way, is there any alternatives to setting variables from config?

last-partizan:
@fredizzimo take a look, maybe you spot some rough spots also.

src/settings/config.rs
```diff
@@ -0,0 +1,89 @@
...
+    let mut path = dirs::config_dir().unwrap();
+    path.push("neovide");
+    path
+}
+
+pub fn config_path() -> PathBuf {
+    let mut config_path = neovide_config_dir();
+    config_path.push(CONFIG_FILE);
+    config_path
+}
```
fredizzimo, line 30: I think these functions should be in the same place as the ones getting the data path for the save window size, so that it's easy to see where everything is saved. Perhaps you could move those functions from `window_size.rs` to here?

src/settings/config.rs
```diff
@@ -0,0 +1,89 @@
...
+#[derive(Debug, Deserialize, Default)]
+pub struct Config {
+    pub multigrid: Option<bool>,
+    pub maximized: Option<bool>,
+    pub vsync: Option<bool>,
+    pub srgb: Option<bool>,
+    pub no_idle: Option<bool>,
+    pub neovim_bin: Option<PathBuf>,
+    pub frame: Option<Frame>,
+    pub geometry: Option<Dimensions>,
```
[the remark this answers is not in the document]
fredizzimo: Yes,  I don't think it should be a setting that you are supposed to edit manually.

src/settings/config.rs
fredizzimo: Actually I did not change `NEOVIDE_IDLE` only `NEOVIDE_SRGB` and `NEOVIDE_VSYNC`. But it would probably make sense to change it the same way.

Hubro:
This would be great for when you want to run Neovide as a client to a remote Neovim instance. Config such as GUI font and `neovide_refresh_rate` needs to be set on the client, not Neovim.

A dedicated config file would be perfect for this.

To that end, I would like to request that `neovide_refresh_rate` and `guifont` be added to the config file, overriding whatever is set in Neovim :smile: That would be awesome.

src/settings/config.rs
LoipesMas: What do you mean by alternatives?

last-partizan: I mean, instead of writing to env, maybe we there is some other way? But it it works, and there is no better way, it's fine.

src/main.rs
LoipesMas: I don't remember my exact reasoning, but now I realize that logger is not set up at this point anyway, so `error!` is useless. I'll delete it then.
Thanks for pointing that out.

src/settings/config.rs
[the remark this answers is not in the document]
LoipesMas: There is a command line argument for that, so I thought it's something that the user might want to change. I guess it should be removed then?
I'll remove it from config anyway.

src/settings/config.rs
LoipesMas: Actually, you did change `NEOVIDE_IDLE` and it's broken now. I created a PR to fix that:  #1889. I'll update the code here as well.

src/settings/config.rs
LoipesMas: I see your point, but looking at both of those things now, I see that they can easily get confused. I guess `PersistentSettings` would need to be renamed. Maybe something like `PersistentState`?

LoipesMas:
> To that end, I would like to humbly request that `neovide_refresh_rate` and `guifont` be added to the config file, overriding whatever is set in Neovim

@Hubro That would be nice. But those settings are handled a bit differently, so it's not that simple. I'd suggest creating a new issue for that (if there isn't one already open).

Hubro:
> > To that end, I would like to humbly request that `neovide_refresh_rate` and `guifont` be added to the config file, overriding whatever is set in Neovim
> 
> @Hubro That would be nice. But those settings are handled a bit differently, so it's not that simple. I'd suggest creating a new issue for that (if there isn't one already open).

Dang :face_with_head_bandage: 

I opened a new issue: https://github.com/neovide/neovide/issues/1891

src/settings/config.rs
last-partizan: Makes sense to me.

And I think it's better to have single serialization format, right now it's json there and toml here. Toml is really preferable for config over json, so it makes sense to store those settings as toml also. But this is suggestion for another refactoring.

And maybe moving those functions here is also better to leave for separate refactoring.

LoipesMas: Yeah, I agree. No need to do all that in this PR.

src/settings/config.rs
```diff
@@ -0,0 +1,95 @@
...
+    fn load_from_path(path: &Path) -> Result<Self, String> {
+        let toml = std::fs::read_to_string(path).map_err(|e| {
+            format!(
+                "Error while trying to open config file {}:\n{}\nContinuing with default config.",
+                path.to_string_lossy(),
+                e
+            )
+        })?;
+        let config = toml::from_str(&toml).map_err(|e| {
+            format!(
```
last-partizan, line 87: Would be nice to display error only in this case, first is perfectly fine for use-case without config.

But that's for another refactoring, maybe i'll fix it after the merge.

Everything else looks fine, merging it.

PR #1123, opened 2022-01-04
Detach stdin, stdout and stderr from terminal

### LINKED ISSUE #1122: Neovide process not fully detaching from terminal
**Describe the bug**
When running neovide (without passing the `--nofork` flag) the spawned process doesn't detach from the terminal.

**To Reproduce**
1. Run neovide from a terminal emulator without the `--nofork` flag
2. Exit the terminal with `exit`
3. See that the terminal emulator window will not close until neovide is closed

**Expected behavior**
I expected the terminal window to close without having to close neovide.

 - OS: Arch Linux
 - Neovide Version: 0.8.0
 - Neovim Version: v0.6.0-dev

**Additional context**
I believe this happens because neovide does not override stdin/stdout/stderr when spawning the new process:

https://github.com/neovide/neovide/blob/d220477b7379335968d972975e562aefe5061383/src/main.rs#L213-L217

By default, stdin/stdout/stderr inherit from the parent process, so they remain attached to the terminal session, preventing it from closing. Redirecting them to `/dev/null` seems to fix the problem:

```rust
assert!(process::Command::new(current_exe)
    .stdin(process::Stdio::null())
    .stdout(process::Stdio::null())
    .stderr(process::Stdio::null())
    .arg("--nofork")
    .args(env::args().skip(1))
    .spawn()
    .is_ok());
```

This was only tested on Linux. Should I make a PR?

bpandreotti:
Previously, stdin/stdout/stderr would remain attached to the terminal, meaning the terminal session could not be closed until neovide was closed. This fixes #1122.

- Fix

- No

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -211,6 +211,9 @@ fn maybe_disown() {
 
     if let Ok(current_exe) = env::current_exe() {
         assert!(process::Command::new(current_exe)
+            .stdin(process::Stdio::null())
+            .stdout(process::Stdio::null())
+            .stderr(process::Stdio::null())
             .arg("--nofork")
             .args(env::args().skip(1))
             .spawn()


Kethku:
Done! Thanks for the change, worked on my machine 👍🏻

Kethku:
(note I manually applied the change in main. This has been merged)

bpandreotti:
Great, thanks for the merge!

PR #1131, opened 2022-01-06
Fix notification event ordering

fredizzimo:
Use a single thread instead of spawning a new task for each notification. This makes sure they are processed in the correct order.

Thank you @Kethku, your comments #1129 made it easy to find the issue. I first checked the `nvim-rs` implementation to see if there were any problems, but that one seems to be OK. Then it was quite easy to see that, there was a problem in Neovide's implementation of the notification handler. A new task was spawned for each notification, which of course does not not guarantee any ordering. So instead I just spawn a new thread at the start, and use a channel to transfer the notification in the correct order to the thread for processing.

Note that I struggled a bit with the actual implementation since I don't really know Rust. In particular I'm not happy with the mutex around the channel. But It seems that nvim-rs requires the sync trait, and I did not see any other way around that. 

For me the bug as I describe it in #1129, was a 100% repeat, and I have not been able to reproduce it anymore with these changes. I have not done very extensive testing if it causes any other problems either, but at least after some use I did not detect any issues or performance problems, but it's probably worth testing some more before merging.

So this fixes https://github.com/neovide/neovide/issues/1129

### DIFF
--- a/src/bridge/handler.rs
+++ b/src/bridge/handler.rs
@@ -1,8 +1,11 @@
 use async_trait::async_trait;
+use futures::lock::Mutex;
 use log::trace;
 use nvim_rs::{Handler, Neovim};
 use rmpv::Value;
-use tokio::task;
+use std::sync::mpsc::{channel, Receiver, Sender};
+use std::sync::Arc;
+use std::thread;
 
 #[cfg(windows)]
 use crate::bridge::ui_commands::{ParallelCommand, UiCommand};
@@ -14,12 +17,54 @@ use crate::{
     settings::SETTINGS,
 };
 
+struct NotificationEvent {
+    event_name: String,
+    arguments: Vec<Value>,
+}
+
 #[derive(Clone)]
-pub struct NeovimHandler {}
+pub struct NeovimHandler {
+    sender: Arc<Mutex<Sender<NotificationEvent>>>,
+}
 
 impl NeovimHandler {
     pub fn new() -> Self {
-        Self {}
+        let (sender, receiver): (Sender<NotificationEvent>, Receiver<NotificationEvent>) =
+            channel();
+        thread::spawn(|| {
+            for event in receiver {
+                match event.event_name.as_ref() {
+                    "redraw" => {
+                        for events in event.arguments {
+                            let parsed_events = parse_redraw_event(events)
+                                .unwrap_or_explained_panic("Could not parse event from neovim");
+
+                            for parsed_event in parsed_events {
+                                EVENT_AGGREGATOR
+                                    .send(EditorCommand::NeovimRedrawEvent(parsed_event));
+                            }
+                        }
+                    }
+                    "setting_changed" => {
+                        SETTINGS.handle_changed_notification(event.arguments);
+                    }
+                    #[cfg(windows)]
+                    "neovide.register_right_click" => {
+                        EVENT_AGGREGATOR
+                            .send(UiCommand::Parallel(ParallelCommand::RegisterRightClick));
+                    }
+                    #[cfg(windows)]
+                    "neovide.unregister_right_click" => {
+                        EVENT_AGGREGATOR
+                            .send(UiCommand::Parallel(ParallelCommand::UnregisterRightClick));
+                    }
+                    _ => {}
+                }
+            }
+        });
+        Self {
+            sender: Arc::new(Mutex::new(sender)),
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/bridge/handler.rs
```diff
@@ -14,12 +17,54 @@ use crate::{
...
+    sender: Arc<Mutex<Sender<NotificationEvent>>>,
+}
 
 impl NeovimHandler {
     pub fn new() -> Self {
-        Self {}
+        let (sender, receiver): (Sender<NotificationEvent>, Receiver<NotificationEvent>) =
+            channel();
+        thread::spawn(|| {
+            for event in receiver {
```
Kethku, line 35: Just out of curiosity, have you tried just sending the events in the handler directly rather than sending over a channel? I am guessing there may be some perf issues if done that way, but its much simpler, so might be worth trying at the very least

Kethku:
I still don't have a consistent repro of this bug, but made a branch that is maybe a tad cleaner here https://github.com/neovide/neovide/tree/alternate_fix_event_ordering could you try on this branch? If it doesn't work we'll just go with your channel based strategy

fredizzimo:
Yes. I will check later today, I guess during the night for you. 

I was initially planning to not use a thread, since the whole io loop already runs in a separate thread. But I decided against that, for a few different reasons, so I never tried it.

1. The original code uses a task so I figured that there would be a good reason for it. And even if it seems to work for me, there could be some special case that requires it.
2. If a thread is not used, we can’t handle notifications and other rpc requests at the same time, but that’s probably not a problem, since the heavyweight processing is already on another thread.
3. A separate thread could be faster, since it allows the stream processing and event processing at the same time. Although it could also be slower due to overhead.

BTW, so far my version has worked nice for me. Neovide even feels slightly faster, but that’s probably a placebo  effect.

Kethku:
I'm relatively sure that nvim-rust has a thread there already because processing is in an async tokio context, but thats just a guess which is why I figured trying without a new thread on our side would be smart.

Your concerns are reasonable, but assume that we arrived at the current state by some kind of design :) I think we should try without the extra thread. At least on my machine it works great!

fredizzimo:
Yes it does, it has two IO threads, one for input and another for output, and then for each RPC request it spawns a new task, allowing the task and the IO thread to to run in parallel. And since the IO thread can proceed to handle the next task, it also allows tasks to run in parallel, or nested tasks as they call it.

The notifications however, run directly in the IO thread, therefore blocking both future tasks and new IO, and possibly even Neovim from progressing when the buffers become full. That was actually the fix for the serialization order bug, to not spawn a new tasks for notifications. 

This is not a big problem as long as the notification processing on the Neovide side is quick enough, and if we never try to run any RPC requests from within that thread directly, since that would deadlock, as it never allows the response to be processed. I think it's fine for the current Neovide code, since all it seems to do is sending events to another thread for actual processing.

fredizzimo:
Your version without the thread fixes the issue for me. So if you prefer that, go ahead and use it instead.

Kethku:
I think I'm just going to run with my branch as theres not THAT much working going on there. But I will definitely remember this conversation in case perf issues do crop up.

KillTheMule:
Great you got that sorted out, nice :+1:  I wanted to make some comments from the perspective of `nvim-rs`, just in case it matters at some point:

1. `nvim-rs` doesn't have threads per se, it's a fully async lib in the rust sense. That means that the runtime the consumer chooses is responsible for using threads (or not), but `nvim-rs` itself doesn't know anything about that. I have tests in place that supposedly ensure it all works out even when explicitely using a single threaded runtime
2. 

> The notifications however, run directly in the IO thread, therefore blocking both future tasks and new IO

Now, this isn't fully true. We do await notifications indeed, but that does not mean everything else is blocked. As commonplace in async, you'll need to make sure your handler does not block too long, so the runtime can take over and handle other things. [Here](https://tokio.rs/blog/2020-04-preemption) is an article about that and mitigation strategies employed by tokio. One main point I want to make though is that if your notification handler itself sends a request to neovim, it will `await` it, and that enables the runtime to switch tasks and have the response actually be handled, no matter the number of threads.

> If a thread is not used, we can’t handle notifications and other rpc requests at the same time

Note that you also shouldn't. If a notification comes in, and then a request, you first want to fully process the notification, and _after that_ process the request, logically speaking (of course, you might have some smart opportunities to let handling overlap, depending on your application, but you'll first need to update your internal state from the notification so you can give the "right" response to the request). What you can do is send out requests while handling the notification, as explained above, no need for an explicit thread here.

I hope I could clear up some misconceptions. I'm harping a bit about this because that actually was the main reason for me to create `nvim-rs` in the first place :) The fix looks just right to me in its final incarnation.

PR #1134, opened 2022-01-08
added option to enable font subpixel antialiasing

### LINKED ISSUE #1133: Add subpixel antialiasing option
Currently you can turn off antialiasing (haven't tested in a while. May not actually work) but you can't change what kind. Turns out on some systems subpixel antialiasing makes the text look significantly worse with gaps and such.

It would be good to give the user the option to pick which they prefer.

jsosulski:
- Feature

- No

## Other

I dont know if the default should be to leave it enabled as it was before or disabled. I copied the behaviour of current main branch, i.e., have it disabled by default.
The option is exposed as `g:neovide_font_subpixel_antialiasing` to make it clear that it does not concern cursor antialiasing.

Closes #1133

### DIFF
--- a/src/renderer/fonts/font_loader.rs
+++ b/src/renderer/fonts/font_loader.rs
@@ -3,7 +3,11 @@ use std::sync::Arc;
 use lru::LruCache;
 use skia_safe::{font::Edging, Data, Font, FontHinting, FontMgr, FontStyle, Typeface};
 
-use crate::renderer::fonts::{font_options::FontOptions, swash_font::SwashFont};
+use crate::{
+    renderer::fonts::{font_options::FontOptions, swash_font::SwashFont},
+    settings::SETTINGS,
+    window::WindowSettings,
+};
 
 static DEFAULT_FONT: &[u8] = include_bytes!("../../../assets/fonts/FiraCode-Regular.ttf");
 static LAST_RESORT_FONT: &[u8] = include_bytes!("../../../assets/fonts/LastResort-Regular.ttf");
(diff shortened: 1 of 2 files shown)


LoipesMas:
Changing antialiasing requires a font update. Without that you can't see any changes, so, unless user changes font after changing this option, they will think that it doesn't work / doesn't make a difference. 
Even putting it in my nvim config (and putting `set guifont=...` after that) doesn't seem to work.

jsosulski:
I see, I just restart neovide whenever I change the config as sourcing it again sometimes messes with my vimrc, I will check out if I can find a way to trigger a font update when this setting changes.

jsosulski:
I just noticed it works on my machine without restarting neovide when I change `$MYVIMRC` and re-source it.

LoipesMas:
For me even restarting neovide doesn't work.

fredizzimo:
It's most likely dependent on the order, If you have `set guifont` after or before `g:neovide_font_subpixel_antialiasing`

LoipesMas:
Tried both, neither works. But manually setting guifont works.

jsosulski:
I am on Ubuntu 20.04 using KDE Plasma.  See (low quality) gif below:

subpixel

Aside from the setting not working, is there a global update trigger that forces neovide to reload everything? E.g. `source $MYVIMRC` seems to update everything.

jsosulski:
As pointed out by @fredizzimo in #1133 I was using a "buggy" version of the JetBrainsMono font. With a different font, enabling / disabling font subpixel antialiasing does not change the rendering on my machine, so disabling this by default makes sense to me.

src/renderer/fonts/font_loader.rs
```diff
@@ -17,7 +21,13 @@ impl FontPair {
     fn new(mut skia_font: Font) -> Option<FontPair> {
         skia_font.set_subpixel(true);
         skia_font.set_hinting(FontHinting::Full);
-        skia_font.set_edging(Edging::AntiAlias);
+
+        let settings = SETTINGS.get::<WindowSettings>();
+        if settings.font_subpixel_antialiasing {
+            skia_font.set_edging(Edging::SubpixelAntiAlias);
+        } else {
+            skia_font.set_edging(Edging::AntiAlias);
```
last-partizan, line 29: Maybe option should expose all three variants:

Alias, AntiAlias and Subpixel?

And should be named "font_edging" instead of current name.

last-partizan:
Closing it then.

Issue #1021, opened 2021-10-18
Support italized cursor text


dineshKumar777:
**Describe the bug**
For italized font, cursor text is showing normal font.

**Expected behavior**
Cursor text should be italized which is actual behavior in terminal

**Screenshots**
![X1YqxdD6F6](https://user-images.githubusercontent.com/21097695/137657982-d324fba6-9465-4300-832a-c675713403b9.png)

**Desktop (please complete the following information):**
 - OS: [e.g. windows 10]
 - Neovide Version [e.g. 0.8.0]
 - Neovim Version [e.g. 0.1.0]

PR #1136, opened 2022-01-08
Fix cursor bold and italic rendering

PfisterFactor:
- Fix
Addresses #1021

Unsure, I changed the Cursor struct's "character" field to store a GridCell instead of a String in order to carry style information.

### DIFF
--- a/src/editor/window.rs
+++ b/src/editor/window.rs
@@ -63,7 +63,7 @@ impl Window {
         });
     }
 
-    pub fn get_cursor_character(
+    pub fn get_cursor_grid_cell(
         &self,
         window_left: u64,
         window_top: u64,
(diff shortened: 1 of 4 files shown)


src/editor/window.rs
```diff
@@ -63,18 +63,22 @@ impl Window {
...
+            _ => (' '.to_string(), None),
         };
 
         let double_width = match self.grid.get_cell(window_left + 1, window_top) {
             Some((character, _)) => character.is_empty(),
             _ => false,
         };
 
-        (character, double_width)
+        (character.0, character.1, double_width)
```
Kethku, line 81: nit: I think the local variable character here should be renamed. Maybe grid_cell?

PfisterFactor: I agree, I pushed a commit fixing it 👍

PfisterFactor: Gotcha, I renamed:
Cursor struct: "character" -> "grid_cell"
get_cursor_character -> get_cursor_grid_cell

Kethku:
This is a reasonable fix. Thanks for the contribution :)

PR #1139, opened 2022-01-09
Add paste by mouse

balroggg:
## Feature
- Add paste from system clipboard by press middle mouse button
- Add `cli-clipboard` dependency to get content of system clipboard

- No

### DIFF
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -41,6 +41,11 @@ pub enum SerialCommand {
         position: (u32, u32),
         modifier_string: String,
     },
+    Paste {
+        data: String,
+        crlf: bool,
+        phase: i64,
+    },
 }
 
 impl SerialCommand {
@@ -101,6 +106,11 @@ impl SerialCommand {
                 )
                 .await
                 .expect("Mouse Drag Failed");
+            }
+            SerialCommand::Paste { data, crlf, phase } => {
+                nvim.paste(data.as_ref(), crlf, phase)
+                    .await
+                    .expect("Paste failed");
             }
         }
     }
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -110,7 +110,7 @@ impl SerialCommand {
             SerialCommand::Paste { data, crlf, phase } => {
                 nvim.paste(data.as_ref(), crlf, phase)
                     .await
-                    .expect("Paste failed");
+                    .unwrap_or(false);
             }
         }
     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a13739d7177fbd22bb0ed28badfff9f372f8bef46c863db4e1c6248f6b223b6e"
 
 [[package]]
+name = "addr2line"
+version = "0.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b9ecd88a8c8378ca913a680cd98f0f13ac67383d35993f86c90a70e3f137816b"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
 name = "adler"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


dineshKumar777:
@balroggg this pr works in windows 10. but for some reason it pastes duplicate content

https://user-images.githubusercontent.com/21097695/148684011-3cf659bd-2f5f-4090-96ef-994ba352f40b.mp4

.

balroggg:
@dineshKumar777 sorry, I can only test on Linux. I think I fix this behaviour by checking release state of mouse press, may be on windows it's different...

dineshKumar777:
@balroggg  it seems content is pasted during both middleclick press and release event.

balroggg:
> @balroggg it seems content is pasted during both middleclick press and release event.

With condition in code(I check release event) it should not.

src/bridge/ui_commands.rs
```diff
@@ -102,6 +107,11 @@ impl SerialCommand {
                 .await
                 .expect("Mouse Drag Failed");
             }
+            SerialCommand::Paste { data, crlf, phase } => {
+                nvim.paste(data.as_ref(), crlf, phase)
+                    .await
+                    .unwrap_or(false);
```
balroggg, line 113: Skip error when paste in `modifiable=off buffer` (read-only files, startup screens and etc). We can check `modifiable` var directly and don't call paste, but it's shorter to just ignore error.

Kethku:
I'm confused by this change. on Main, if I'm in insert mode and have some text in the clipboard and middle click the mouse, text is pasted into the buffer.

This is built in neovim behavior. So I'm sorta hesitant to change it. Am I missing something? Does that not work on non windows platforms?

balroggg:
> 

It works in neovim/vim in terminal, but if you have some gui layer you need pass clipboard to the application.
For example `alacrity` has clipboard mechanism for supported platforms. `Qt` has builtin support of clipboard(https://doc.qt.io/qt-5/qclipboard.html) in gui apps.

Kethku:
Makes sense. Problem is it seems to be working in neovide on my machine before this PR. So seems like theres a problem with the system clipboard integration.

I did some reading and I think your PR would be a great starting point for building integration which would enable WSL system clipboard and remote clipboard transparently. https://neovim.io/doc/user/provider.html this page under "Clipboard integration" outlines a way to specify a custom vimscript lambda for reading and writing to the clipboard. This combined with the cli-clipboard crate and some careful async handling should let us fix this for everyone.

To do it, we'll probably need a couple of new changes:
1. Register a new handler for a read clipboard and set clipboard notifications https://github.com/neovide/neovide/blob/main/src/bridge/handler.rs#L38
2. On startup, set the g:clipboard variable such that it calls rpcnotify and rpcrequest to hit our new handler implementations. An example of this can be found here https://github.com/equalsraf/neovim-qt/pull/479/files

... I think thats it! The neovim-qt PR I've linked is a pretty good example of what it should look like.

I much prefer solving the issue this way as manually pasting content into the buffer seems incorrect to me, and doesn't handle yanking properly. This seems like the more "correct" way to go. But I'm interested in your thoughts.

Kethku:
Side note: given that middle mouse already pastes on my machine, it makes me think something else is causing problems on your local install. @LoipesMas's point that it pastes twice on their machine using your branch supports this idea

Kethku:
Also forgot that, much of this work is already completed here: https://github.com/neovide/neovide/pull/1003/files so I'm sorta inclined to take this instead

balroggg:
@Kethku thanks for the information! I will check these PR's.

Kethku:
Awesome. Please try the provider pr and let me know if it fixes your issue. If so, it just got updated to main and we should be able to merge it very soon

fredizzimo:
Yes, middle mouse paste should work if all of these conditions are satisfied, even without #1003, which is really only needed when you run neovim remotely through a TCP connection without special ssh clipboard tools.

* The mouse support enabled, for examples `mouse=a` to enable it in all modes.  https://neovim.io/doc/user/options.html#'mouse'
* You have the correct clipboard tool executable enabled depending on the system https://neovim.io/doc/user/provider.html#clipboard, especially Windows needs manual setup https://github.com/neovim/neovim/wiki/FAQ#how-to-use-the-windows-clipboard-from-wsl
* You haven't remapped the middle mouse button to something else

Kethku:
Closing in favor of https://github.com/neovide/neovide/pull/1003

PR #1141, opened 2022-01-10
Trim newlines from end of path on macOS

### LINKED ISSUE #1140: Last location listed in path not honored when launched from Finder on macOS
**Describe the bug**
When launching from Finder on macOS, the last location in my `$PATH` is not searched for executables.

**To Reproduce**
1. Append a new location to $PATH
2. Create a script in that directory such as `foo.sh` and then `chmod +x foo.sh`
2. Launch Neovide from Finder and run `echo $PATH` to verify the path is present
3. Run `echo executable("foo.sh")` and it will return `0`.

If `neovide` is launched from the terminal, it correctly returns `1`.

**Expected behavior**
Should return `1`

 - OS: macOS 12.0.1
 - Neovide Version: 0.8.0 (`main`)
 - Neovim Version: 0.6.0

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/7842584/neovide_rCURRENT.log)

**Additional context**
I do notice a difference in what is returned by `:echo $PATH` from terminal and non terminal. When launched from Finder, I'm seeing an extra blank line at the end of the path. I think that it is including a newline in the last path element.

ViViDboarder:
Fix

No

Fixes #1140

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -209,7 +209,10 @@ fn handle_macos() {
             .arg(cmd)
             .output()
         {
-            env::set_var("PATH", std::str::from_utf8(&path.stdout).unwrap());
+            env::set_var(
+                "PATH",
+                std::str::from_utf8(&path.stdout).unwrap().trim_end(),
+            );
         }
     }
 }


shaunsingh:
LGTM, simple change

PR #1146, opened 2022-01-13
Fix floating windows transparency

LoipesMas:
Floating windows were not rendered transparent with `neovide_transparency=1.0`.

### DIFF
--- a/src/renderer/grid_renderer.rs
+++ b/src/renderer/grid_renderer.rs
@@ -106,13 +106,12 @@ impl GridRenderer {
                 .set_color(style.background(&self.default_style.colors).to_color());
         }
 
-        if (SETTINGS.get::<WindowSettings>().transparency - 1.0).abs() > f32::EPSILON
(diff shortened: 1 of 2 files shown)


src/renderer/grid_renderer.rs
```diff
@@ -106,13 +106,12 @@ impl GridRenderer {
...
-        if (SETTINGS.get::<WindowSettings>().transparency - 1.0).abs() > f32::EPSILON
-            && self.paint.color() == self.get_default_background()
-        {
+        // Only make background color transparent
+        if self.paint.color() == self.get_default_background() {
             if is_floating {
                 self.paint
                     .set_alpha((255.0 * SETTINGS.get::<RendererSettings>().floating_opacity) as u8);
-            } else {
+            } else if (SETTINGS.get::<WindowSettings>().transparency - 1.0).abs() > f32::EPSILON {
```
Kethku, line 114: On current main, after merging this, floating windows are not transparent. I think there are still issues here.

PR #1148, opened 2022-01-14
Touch support (scrolling, click-by-tap, visual selection by dragging)

MultisampledNight:
- Feature
  - Touch scrolling
  - Clicking by touch tapping
  - Creating a visual selection through touch dragging (hold down finger for a bit and move)

  - Touch input is actually handled now instead of ignored

This PR introduces general touch support (despite the branch name). From what I understood this has been supported in the past already, but was broken by a move to `sdl2` _and back to `winit`?_

While the new setting values (namely `touch_deadzone` and `touch_drag_timeout`) are undocumented yet, I'd be happy to do that in the `Wiki` tab, but I can't seem to edit there. It should probably belong to the `Input Settings` section though.

### DIFF
--- a/src/window/mouse_manager.rs
+++ b/src/window/mouse_manager.rs
@@ -313,7 +313,7 @@ impl MouseManager {
                 // not updating the position would cause the movement to "escalate" from the
                 // starting point
                 if let Some(last_location) = self.touch_position.insert(finger_id, location) {
-                    let delta = (location.x - last_location.x, location.y - last_location.y);
+                    let delta = (last_location.x - location.x, location.y - last_location.y);
                     self.handle_pixel_scroll(font_size, delta, keyboard_manager);
                 }
             }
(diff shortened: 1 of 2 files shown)


TornaxO7:
It works for me. Tested on the acer spin 5.

MultisampledNight:
Since this is my own PR, I would rather not like to merge it myself. So in case someone with merge permissions stumbles across this, I'd be happy if you'd review a bit and then merge.

fredizzimo (approved):
I didn't see any problems. But I'll leave this open a bit still, in case someone else wants to review. And I have also not tested the code. I don't see how it could break normal mouse behaviour, but there's always the possibility.

Finally I'm not sure if we want to have a feature freeze before the release or not.

Kethku:
I don't have a strong opinion either way. Regardless, I think we should move to more frequent releases after this one so shouldn't be a big deal if we wait

shaunsingh:
The PR likely won't create any new issues, and I imagine its a feature atleast a few people would like. Code LGTM as well, I don't have any issues with merging it before the next release.

Kethku:
I agree.

MultisampledNight:
Nice, I'll document it. Thanks for your time!

PR #1159, opened 2022-01-20
correct repo address as well as bundle id

smartrejames:


Kethku:
Nice catch. Thanks for the fix

PR #1161, opened 2022-01-20
use cargo version string to register neovide channel

nganhkhoa:


Kethku:
Extremely good point. Thanks for the catch

PR #1162, opened 2022-01-20
fix for macos finder issue

Kethku:
@shaunsingh @fredizzimo in that case I vote we merge this.

- Fix

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -53,7 +53,7 @@ async fn start_neovim_runtime() {
     .unwrap_or_explained_panic("Could not locate or start neovim process");
 
     // Check the neovim version to ensure its high enough
-    match nvim.command_output("echo has('nvim-0.4')").await.as_deref() {
+    match dbg!(nvim.command_output("echo has('nvim-0.4')").await.as_deref()) {
         Ok("1") => {} // This is just a guard
         _ => {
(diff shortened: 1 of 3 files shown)


fredizzimo:
Like I mentioned in https://github.com/neovide/neovide/issues/447, I think `platform_which` and `platform_exists` needs to be implemented. It probably works now due to this hack
https://github.com/neovide/neovide/blob/8542333c940b0549c82da1ff89f6b9bd11f73c53/src/main.rs#L201-L220

But if those are implemented, then that should no longer be needed, and it appears as if that hack does not work on all systems anyway.

src/bridge/mod.rs
```diff
@@ -109,6 +109,16 @@ fn build_nvim_cmd_with_args(bin: &str) -> Command {
         cmd.args(&["$SHELL", "-lc", &argstring]);
         return cmd;
     }
+
+    #[cfg(macos)]
```
fredizzimo, line 113: Is this clippy error, something we need to be concerned of?

Kethku: I'll look into it

Kethku:
@shaunsingh any chance you could try the currently pushed version? I don't have a mac so I can't test...

shaunsingh:
I apologize for the delay, I was having a few issues with my GitHub account. Works great, changes lgtm!

PR #1163, opened 2022-01-21
setup remote copy paste for wsl

nganhkhoa:
Update the remote copy paste to wsl sessions.

Updates #1003

PR #1164, opened 2022-01-21
use u64 when register channel

nganhkhoa:
Neovide parses the channel using struct `ClientVersion`, which requires a `u64`. This fixes remote channel used by #1003 and #1163.

Please merge this **before** testing the remote copy/paste.

https://github.com/neovide/neovide/blob/ab0394ae40a466c5e4c7c53fd4ece85fc52f0c24/src/bridge/events.rs#L307-L313

I was careless in #1161, and didn't parse as `u64`. I'm sorry.


Kethku:
Your point is reasonable, so I've created a new PR which does that. It disables features which require  a channel when the channel isn't found.

Thanks for the PR!

PR #1169, opened 2022-01-23
refactor: Move call to SETTINGS.get inside should_ignore_input

last-partizan:
It wasn't required outside, and this way code is cleaner.

- Codestyle

- No

### DIFF
--- a/src/window/keyboard_manager.rs
+++ b/src/window/keyboard_manager.rs
@@ -1,14 +1,13 @@
-use glutin::{
-    event::{ElementState, Event, KeyEvent, WindowEvent},
-    keyboard::{Key, Key::Dead},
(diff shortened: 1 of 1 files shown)


src/window/keyboard_manager.rs
```diff
@@ -1,15 +1,14 @@
...
-    platform::modifier_supplement::KeyEventExtModifierSupplement,
-};
-
 use crate::{
     bridge::{SerialCommand, UiCommand},
     event_aggregator::EVENT_AGGREGATOR,
     settings::SETTINGS,
     window::KeyboardSettings,
 };
+use glutin::{
```
last-partizan, line 7: This part is changed by cargo fmt, so i just leave it here.

lucas-miranda (approved):

PR #1177, opened 2022-01-25
cleanup: Remove super verbose logging from font_shaper

last-partizan:
- Other

- No

This is really not necessary.

### DIFF
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use log::{trace, warn};
+use log::{debug, trace, warn};
 use lru::LruCache;
 use skia_safe::{
     graphics::{font_cache_limit, font_cache_used, set_font_cache_limit},
@@ -69,13 +69,13 @@ impl CachingShaper {
     }
 
     pub fn update_scale_factor(&mut self, scale_factor: f32) {
-        trace!("scale_factor changed: {:.2}", scale_factor);
+        debug!("scale_factor changed: {:.2}", scale_factor);
         self.scale_factor = scale_factor;
         self.reset_font_loader();
     }
 
     pub fn update_font(&mut self, guifont_setting: &str) {
-        trace!("Updating font: {}", guifont_setting);
+        debug!("Updating font: {}", guifont_setting);
 
         let options = FontOptions::parse(guifont_setting);
         let font_key = FontKey {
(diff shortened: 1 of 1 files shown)


fredizzimo:
That logging has been useful in a few instances. But I have been worried that someone leaks some sensitive information, since it sends everything that's on the screen, so it's probably good to disable. But the same information is still there for the grid messages, it's just harder to parse. 

I wonder if the best option would be to actually set the logging level, so that these trace messages are not logged by default, but can still be obtained if really needed in order to track down some bug? But that probably means that we have to go through each log message and properly categorize it.

last-partizan:
I was thinking of replacing all `trace` here with `debug`. if it was useful, i'd better do that.

fredizzimo:
I don't think that would help, since trace is supposed to be the most verbose level. It's just that we always log with the trace level enabled at the moment.  https://docs.rs/flexi_logger/latest/flexi_logger/enum.Level.html

last-partizan:
I'm using something like this to trace font shaping:

```
export RUST_LOG="neovide::renderer::fonts::caching_shaper=trace"
```

And if i change it to debug - i will see only relevant messages, and when you need extra tracing - you will see it with trace level.

Kethku:
This seems like a reasonable solution.

Kethku:
@last-partizan feel free to merge as this preserves the useful data in the log file

fredizzimo:
I still don't understand why trace messages are showing in the log and debug ones not, since trace is a lower log level than debug. Since at least in all the other programming languages and logging libraries that I have used it's the other way around, and the documentation that I linked above also seems to indicate that.

Are we specifically turning on the trace log level or something, or is there something I need to learn about the rust logging system, well I know almost nothing about it, so that's probably it.

last-partizan:
@fredizzimo it works as you expect, with level set to `trace` you would see every other messages too, it is most verbose level. With `debug` you would see everithing except `trace`.

fredizzimo:
Yes, but what I mean is that it should correspond to the statements in the code. When I put `trace` in the code, I expect the message to be only shown when trace logging is enabled. Now when it's changed to `debug` I would expect it to show up when either `debug` or `trace` level logging is enabled, so I'm confused how this can make the logging less verbose.

last-partizan:
The main idea is - not to put everything in `trace`, but set level to appropriately for each message.

My main concern when adding this was: i could not see messages about font size, becouse everything was cluttered bug "Shaping text: ...".

Now, when there is different level of verbosity, i just use debug, and can see only those font related messages.

fredizzimo:
Ah sorry, my bad, I must have been too tired yesterday when I looked at the actual changes, since I thought that you changed the shaping messages to debug, but it was the other messages that you changes. So yes, everything makes perfect sense now.

Issue #1174, opened 2022-01-25
Strange rendering of Lua comments with "JetBrainsMono Nerd Font" when ligatures presented


daniilsvc:
**Describe the bug**
Comments in Lua files are rendered incorrectly in case of using "JetBrainsMono Nerd Font" and presence of some(!) ligatures.

**To Reproduce**

1. Create `test.lua` file with the next content:
```lua
-- I write several comment lines
-- in this file in order to demonstrate
-- effect of weird JetBrainsMono Nerd Font rendering
-- in case SOME ligaure presence.
-- Ligarute: ..
```
2. Set fonts in your config: `vim.cmd [[set guifont=JetBrainsMono\ Nerd\ Font:h12]]`
3. Run command `neovide test.lua`
4. Observe "dancing" chars:
  ![2](https://user-images.githubusercontent.com/14823534/150891805-061be9ea-cd32-48da-ab59-abff3010bf74.png)
5. Compare with file without last `..` ligature:
  ![3](https://user-images.githubusercontent.com/14823534/150893187-46d43382-3138-4a40-ad13-e1eba995fadc.png)

You can try other ligatures to reproduce these "dancing" chars. Arrors like `->` or `=>` are not producing the error. But `...`, `***`, `.=` - lead to incorrect rendering.

**Expected behavior**
No "dancing" chars. Correct rendering.

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04
 - Neovide Version 0.8.0
 - Neovim Version 0.6.1

**Please run `neovide --log` and paste the contents of the `.log` file here:**
`neovide test.lua --log` for ligature case:
[ligature.log](https://github.com/neovide/neovide/files/7929933/ligature.log)

`neovide test.lua --log` for no ligature case:
[noLigature.log](https://github.com/neovide/neovide/files/7929934/noLigature.log)

**Additional context**
I found this bug accidentally becase `..` is used in Lua for string concatenation in one of my configs files. Rendering JetBrainsMono Nerd Font in Neovide has more issues than this one. For example, if I use "dracula" colortheme (dracula/vim) with this fonts and open `pom.xml` in my project I got "dancing" chars again.
You can play around with it and find out that combination Neovide + JetBrainsMono Nerd Font + `some colorscheme` may render some elements (comments, string literals, keywords, etc.) of different file formats incorrectly.

fredizzimo:
[...]

PR #1179, opened 2022-01-25
feat: Allow float font sizes if requested

last-partizan:
To use float font size, define it with dot, example:

monospace:h11 - uses integer size
monospace:h11.0 - uses float font size

---

This adds optional support for float font sizes.

- https://github.com/neovide/neovide/pull/1116#pullrequestreview-862861964

This PR has broken my ideal font size, so i'm adding this as a fix.

It should work seamlessly for anyone, if someone wants integer fonts - with default syntax `h11` gives integer scaling, with `h11.0` fonts will not use fudging.

---

- Feature

- No

### DIFF
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -94,25 +94,28 @@ impl CachingShaper {
     }
 
     fn reset_font_loader(&mut self) {
-        // Calculate the new fudge factor required to scale the font width to the nearest exact pixel
-        // NOTE: This temporarily loads the font without any fudge factor, since the interface
-        // needs a size and we don't know the exact one until it's calculated.
         self.fudge_factor = 1.0;
         let mut font_size = self.current_size();
         debug!("Original font_size: {:.2}px", font_size);
-        self.font_loader = FontLoader::new(font_size);
-        let (metrics, font_width) = self.info();
-        debug!(
-            "Font width: {:.2}px (avg: {:.2}px)",
-            font_width, metrics.average_width
-        );
-        self.fudge_factor = font_width.round() / font_width;
-        debug!("Fudge factor: {:.2}", self.fudge_factor);
-        font_size = self.current_size();
-        debug!("Fudged font size: {:.2}px", font_size);
-        debug!("Fudged font width: {:.2}px", self.info().1);
 
         self.font_loader = FontLoader::new(font_size);
+        let (metrics, font_width) = self.info();
+
+        debug!("Original font_width: {:.2}px", font_width);
+
+        if !self.options.allow_float_size {
+            // Calculate the new fudge factor required to scale the font width to the nearest exact pixel
+            debug!(
+                "Font width: {:.2}px (avg: {:.2}px)",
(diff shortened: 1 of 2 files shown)


Kethku:
This seems like a reasonable fix to me. I'll defer to @fredizzimo though as he introduced the fudge factor (which improved font rendering significantly from what I can tell)

fredizzimo:
Moving my reply form https://github.com/neovide/neovide/pull/1116#pullrequestreview-862861964 here, since I think this is the better place to discuss it

Do you use the latest version of Jetbrains Mono, since the version in included in the Nerd Fonts release is very old and pretty broken, see this https://github.com/neovide/neovide/issues/1174? So that might be why you think it's not correct.

To my eyes, the fudged version looks more pleasant to the eyes, even in your screenshots, and it also matches what I see on their website, except that they use a bigger line spacing, but that's requested separately, and does not come from the actual font size, The non-fudged version is the odd one there. The website sets a font size of 18 px, which is 14 pt, so that's what I set in Neovim
image

Also before my changed, 14.12 px  wide did not render as 14.12px by Neovide, it rounded it down to 14 px, but kept the height, so the font does not have the proportion it's designed to have. 

You are correct that you don't get a 14pt font when you specify 14, you get a font size of 13.86pt, but the proportions and the spacing is correct according to how the font is designed. But that doesn't really matter, as fonts are designed in em units, where 12 pt is one 1em, so unless you use sizes of that you get fractional scaling anyway. And if that doesn't give you fractional scaling, then it will most likely be that when you convert it to actual pixels on the screen..

If you really want the exact pt heights specified, then we need a fractional sized grid.  Skia is supposed to support it and some applications have support for it, for example Firefox, https://www.reddit.com/r/firefox/comments/doco5o/firefox_nightly_is_the_first_major_application_on/. So at least in theory it's possible. However I did not see that happening in my actual Firefox, but it might be some option that you have to enable, or it might be available only on Linux.

BTW, I'm not against this change since it's added an option, but I personally think it's wrong.

last-partizan:
Yeah, i'm using nerd font version.

Normal version does indeed looks better, but still have those differences between fudged and non-fudged version.

Probably that's personal preference, and with this merge we both can have what we want, so i'm merging this :)

last-partizan:
> Also before my changed, 14.12 px wide did not render as 14.12px by Neovide, it rounded it down to 14 px, but kept the height, so the font does not have the proportion it's designed to have.

Btw, where does this rounding occurs? I would like to look into it.

fredizzimo:
It actually works differently now, since I needed it to round exactly for the fudge factor. This is the current code. So this actually means that your changes will make the font too wide in some cases. Also note that 14.12px is quite close to 14, so that's why it still looks reasonable, good even when it's rounded down to be too narrow, but if the requested font width is close to a being halfway between pixels, I don't think that either rounding up or down give a good result.

https://github.com/neovide/neovide/blob/be79bdee60a82ebf3921c61da20014c68cb08267/src/renderer/fonts/caching_shaper.rs#L150-L156

But that code used to round down.

Issue #1176, opened 2022-01-25
Neovide thinks neovim is below version 0.4 only when passing a file name as parameter


LordMZTE:
**Describe the bug**
When starting neovide with `neovide` it works fine.
Starting it with `neovide test.txt` however, will result in an error (see log file).

**To Reproduce**
1. open neovide with `neovide`
2. check the log file

**Expected behavior**
Neovide opens up fine and opens the file

**Desktop (please complete the following information):**
- OS: ArchLinux, kernel 5.16.2
- Neovide Version 0.8.0, commit `gdd74ff4`
- Neovim Version 0.6.1

**Please run `neovide --log` and paste the contents of the `.log` file here:**
<details>
<summary>with file argument</summary>

```log
INFO [neovide::bridge::command] Starting neovim with: Command { std: "/sbin/nvim" "--embed test.txt", kill_on_drop: false }
ERROR [neovide::bridge] Neovide requires nvim version 0.4 or higher. Download the latest version here https://github.com/neovim/neovim/wiki/Installing-Neovim
```

</details>

<details>
<summary>without file argument (TRACE messages removed because of GH length limit)</summary>

```log
INFO [neovide::bridge::command] Starting neovim with: Command { std: "/sbin/nvim" "--embed", kill_on_drop: false }
DEBUG [neovide::bridge::events] Ignored client type property: "version"
DEBUG [neovide::bridge::events] Ignored client type property: "type"
DEBUG [neovide::bridge::events] Ignored client type property: "methods"
DEBUG [neovide::bridge::events] Ignored client type property: "attributes"
INFO [neovide::bridge::setup] Neovide registered to nvim with channel id 1
INFO [neovide::bridge] Neovim process attached
DEBUG [neovide::settings::window_geometry] Loaded window settings: Windowed { position: PhysicalPosition { x: 20, y: 112 }, size: Dimensions { width: 148, height: 99 } }
INFO [neovide::window] window created (scale_factor: 1.0000, font_dimensions: Dimensions { width: 11, height: 22 })
INFO [neovide::running_tracker] Quit with code 0: Quit from neovim
INFO [neovide::running_tracker] Quit neovim processed failed
DEBUG [neovide::settings::window_geometry] Saved Window Settings: {"window":{"Windowed":{"position":{"x":20,"y":112},"size":{"width":148,"height":99}}}}
```

</details>

fredizzimo:
Hm.. 

[...]

Issue #1175, opened 2022-01-25
Does not start on macOS 


caeu:
Neovide does not start on macOS after updating to (today's commit ```dd74ff```). The window pops up and closes almost immediately, logs shows ERROR Neovide requires nvim version 0.4 or higher while I have 0.6.1

Old version works (from a week ago)

**To Reproduce**
Start nevoid from command line or application

 - OS: macOS 12.1
 - Neovide Version 0.8.0
 - NVIM v0.6.1
 
**LOG**
TRACE [neovide] Neovide version: 0.8.0
DEBUG [neovide::settings::window_geometry] Loaded window settings: Windowed { position: PhysicalPosition { x: 1280, y: 1080 }, size: Dimensions { width: 106, height: 41 } }
TRACE [neovide::renderer::fonts::caching_shaper] Using font_size: 28.00px
TRACE [neovide::renderer::fonts::font_loader] Loading font FontKey { bold: false, italic: false, family_name: None }
TRACE [neovide::renderer::fonts::caching_shaper] Font width: 17.23px 17.36px
TRACE [neovide::renderer::fonts::caching_shaper] Fudge factor: 0.99
TRACE [neovide::renderer::fonts::caching_shaper] Fudged font size: 27.63px
TRACE [neovide::renderer::fonts::caching_shaper] Fudged font width: 17.00px
TRACE [neovide::renderer::fonts::font_loader] Loading font FontKey { bold: false, italic: false, family_name: None }
INFO [neovide::window] window created (scale_factor: 2.0000, font_dimensions: Dimensions { width: 17, height: 35 })
TRACE [neovide::redraw_scheduler] Next frame queued
TRACE [neovide::redraw_scheduler] Next frame queued
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text:
INFO [neovide::bridge::command] Starting neovim with: Command { std: "/bin/zsh" "-lc" "\x1b]1337;RemoteHost=ommited@ommited\x07\x1b]1337;CurrentDir=/Users/ommited\x07\x1b]1337;ShellIntegrationVersion=12;shell=zsh\x07/usr/local/bin/nvim --embed", kill_on_drop: false }
ERROR [neovide::bridge] Neovide requires nvim version 0.4 or higher. Download the latest version here https://github.com/neovim/neovim/wiki/Installing-Neovim

fredizzimo:
[...]

PR #1180, opened 2022-01-25
fix: Fix opening files on linux

last-partizan:
I don't know how about other OS where you using SHELL, but on linux you should
pass arguments like this, or this happens:
```
[src/bridge/command.rs:26] cmd = Command {
    std: "/usr/bin/nvim" "--embed README.md",
    kill_on_drop: false,
}
nvim: Unknown option argument: "--embed README.md"
```
- Fix

- No

### DIFF
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -115,7 +115,7 @@ fn build_nvim_cmd_with_args(bin: &str) -> TokioCommand {
         cmd
     } else {
         let mut cmd = TokioCommand::new(bin);
-        cmd.arg(args_str);
+        cmd.args(args);
         cmd
     }
 }


fredizzimo:
The other platforms don't work properly either after https://github.com/neovide/neovide/pull/1162, which introduced this regression.

See https://github.com/neovide/neovide/issues/1176 and https://github.com/neovide/neovide/issues/1175. 

The fix is correct for direct launching, but the the shell command needs a single argument after `-lic`, you can try for example
`/bin/bash -c 'ls ~/.config'`, vs `/bin/bash -c ls ~/.config `

Kethku:
@fredizzimo if I'm understanding you right, I think you are saying this works? Or are the other platforms also broken currently?

fredizzimo:
Sorry for being unclear. On Linux, this fix is correct and should work. But it does not fix the other platforms, and they can’t be fixed the same way, since they need a single argument.

Kethku:
Ok I'm confused then because it seems that wsl works with or without args on my machine with current main

fredizzimo:
Yes, the MacOS problem is probably something else after all. Other than this accidental regression on Linux, I don't see any difference in how the commandline is passed, compared to the version I made earlier, which passes all the rest of the arguments as a single string. And on Mac we do the same, but it fails there..

Kethku:
I commented on the macos change. I think we are getting ANSI escape codes along with the output which is gumming up the works. Maybe theres a way to ask the shell to write unthemed output? Or maybe a crate we can depend on which clears out ANSI codes?

PR #1181, opened 2022-01-26
Make multigrid default and add settings disabling multigrid specific behavior

Kethku:
- Fix

- Yes, multigrid is now default behavior

### DIFF
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -110,7 +110,7 @@ pub fn handle_command_line_arguments(args: Vec<String>) -> Result<(), String> {
         )
         .arg(
             Arg::with_name("no_multi_grid")
-                .long("nomultigrid")
+                .long("no-multigrid")
                 .help("Disable Multigrid"),
         )
         .arg(
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -179,7 +179,8 @@ pub fn handle_command_line_arguments(args: Vec<String>) -> Result<(), String> {
         },
         maximized: matches.is_present("maximized") || std::env::var("NEOVIDE_MAXIMIZED").is_ok(),
         // Multigrid should be enabled by default so set it to false if NEOVIDE_NO_MULTI_GRID is set
-        multi_grid: !(matches.is_present("no_multi_grid") || std::env::var("NEOVIDE_NO_MULTIGRID").is_ok()),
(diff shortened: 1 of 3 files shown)


src/renderer/mod.rs
```diff
@@ -30,7 +30,9 @@ pub use rendered_window::{LineFragment, RenderedWindow, WindowDrawCommand, Windo
 #[derive(SettingGroup, Clone)]
 pub struct RendererSettings {
     position_animation_length: f32,
+    window_animation: bool,
     scroll_animation_length: f32,
+    smooth_scrolling: bool,
```
fredizzimo, line 35: It seems like only the `cmd_line.rs` file deals with multigrid stuff. The rest adds `window_animation` and `smoth_scrolling`, but no corresponding commandline options or environment variables. Is this intentional?

src/cmd_line.rs
```diff
@@ -109,9 +109,9 @@ pub fn handle_command_line_arguments(args: Vec<String>) -> Result<(), String> {
                 .help("Maximize the window"),
         )
         .arg(
-            Arg::with_name("multi_grid")
-                .long("multigrid")
-                .help("Enable Multigrid"),
+            Arg::with_name("no_multi_grid")
+                .long("nomultigrid")
```
last-partizan, line 113:
```suggestion
                .long("no-multigrid")
```

Maybe this is better choice

src/renderer/mod.rs
Kethku: No this makes sense. When multigrid is disabled, the rendering system renders into a single window and doesn't send relevant scrolling events. So no cmd_line argument is needed.

src/cmd_line.rs
Kethku: I debated here, but there were other existing settings which concatted noidle so I went with this. Instead I think I'll add a dash to the other items

Kethku: Done

last-partizan:
Is there something else needs to be done before merging this?

bmulholland:
If this were merged, I'd have to disable it immediately because of https://github.com/neovide/neovide/issues/1503

falcucci:
we still have visual issues related to the `--multigrid` functionality already reported at #1819 #1508 

we need to analise about what is going in the MsgArea behavior when multigrid is enabled.

Kethku:
Closing in favor of improve-render-loop branch

PR #1183, opened 2022-01-27
Fix broken CI

last-partizan:
- Fix

- No

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,19 @@ jobs:
           NEOVIM_BIN: "C:/tools/neovim/Neovim/bin/nvim.exe"
           RUST_BACKTRACE: full
         run: |
-          cargo test -- -- -Z unstable-options --format json | cargo2junit > results.xml
+          # Tag: BROKEN_CI
+          # Refs: https://github.com/neovide/neovide/issues/1182
+          # WARNING: Currently this command does not do what it's expected to do,
+          # for some reason all arguments after -- is ignored.
+          # So results.xml is empty.
+          # There was -- -- before, but that was broken too.
+          cargo test -- -Z unstable-options --format json > results.json
+          # This and `set -eo pipefail` for MacOS/Linux
+          # is added, becouse without it pipe will eat exit status.
+          # Docs are stating that it should work by default, but it does not.
+          # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference
+          # Oh, it still does not fail on Windows. ¯\_(ツ)_/¯
+          cat results.json | cargo2junit > results.xml
 
       - name: Publish Test Results
         uses: EnricoMi/publish-unit-test-result-action/composite@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -98,7 +98,8 @@ jobs:
         env:
           RUST_BACKTRACE: full
         run: |
-          cargo test -- -- -Z unstable-options --format json | cargo2junit > results.xml
+          set -eo pipefail
```
last-partizan, line 101: @Kethku looks like this flag fixes `not faling part`, but i don't know how to do the same for windows.

As for "not working json" part, i have one idea.

I've used my favorite debugging techniques "remove it and see if something changes". And i've removed everything from "main.rs".

And it fixed something.

Problem is: `cargo test -- --help` should pass arguments to libtest, as stated here: https://doc.rust-lang.org/cargo/commands/cargo-test.html

And in neovide, somehow, it does not pass anything until there is code in `main.rs`, so i thing problem is somewhere there.

.github/workflows/build.yml
```diff
@@ -36,7 +36,19 @@ jobs:
           NEOVIM_BIN: "C:/tools/neovim/Neovim/bin/nvim.exe"
           RUST_BACKTRACE: full
         run: |
-          cargo test -- -- -Z unstable-options --format json | cargo2junit > results.xml
+          # Tag: BROKEN_CI
```
last-partizan, line 39: I've added this `Tag: BROKEN_CI`, to mark broken parts and does not repeat comments.

.github/workflows/build.yml
```diff
@@ -167,23 +177,24 @@ jobs:
...
-      - name: Install Dependencies
-        run: |
-          sudo apt-get install -y curl gnupg ca-certificates git gcc-multilib g++-multilib cmake libssl-dev pkg-config libfreetype6-dev libasound2-dev libexpat1-dev libxcb-composite0-dev libbz2-dev freeglut3-dev libxi-dev
-
       - name: Check Formatting
         run: |
           cargo fmt --all -- --check
 
-      - name: Install Neovim
+      - name: Install dependencies and neovim
```
last-partizan, line 184: Here i optimized three steps, `update/install deps/install neovide` into one. It should be faster this way.

src/bridge/mod.rs
```diff
@@ -17,7 +17,7 @@ use crate::{
     settings::*,
 };
 
-use command::create_nvim_command;
+pub use command::create_nvim_command;
```
last-partizan, line 20: This fixes one error in tests, three others i'm leaving for you, @Kethku

last-partizan:
I think, this could be merged as is.

It fixes (meaning: it fails when it should) pipeline for Linux and MacOS. (On Windows i haven't figured out how to make it work)

Other fixes, that is - broken tests, could be fixed later.

last-partizan:
Eureka!

Now i fucking understand it!

Tests was broken at BUILD stage, so they had no opportunity to process arguments after `--`.

Well. It is fixed now ;)

src/editor/window.rs
```diff
@@ -353,9 +347,7 @@ mod tests {
             (114, 64),
             batched_sender.clone(),
         );
-        batched_sender
-            .send_batch()
-            .expect("Could not send batch of commands");
+        batched_sender.send_batch();
```
last-partizan, line 350: @Kethku actually, i fixed everything, and now it works :)

PR #1195, opened 2022-02-03
feat: add support for blend=100

0x00002a:
- Feature

- No

This adds support for `blend=100` when drawing the cursor. In terminal mode if blend=100 is in the active highlight group for `guicursor` then it makes the cursor invisible, my personal use-case for this is hiding the cursor in nerdtree and similar.

Note that I'm not sure if there is more to `blend` than just making the cursor invisible if its 100, `:help blend` says something about blend level in floating windows but I think the cursor is special, in any case I've only implemented the cursor bit.

### DIFF
--- a/src/editor/cursor.rs
+++ b/src/editor/cursor.rs
@@ -80,6 +80,13 @@ impl Cursor {
             .unwrap_or_else(|| default_colors.foreground.unwrap())
     }
 
+    pub fn blend_is_hide(&self) -> bool {
+        self.style
+            .as_ref()
+            .map(|s| s.blend == 100)
+            .unwrap_or(false)
+    }
+
     pub fn change_mode(&mut self, cursor_mode: &CursorMode, styles: &HashMap<u64, Arc<Style>>) {
         let CursorMode {
             shape,
--- a/src/editor/cursor.rs
+++ b/src/editor/cursor.rs
@@ -83,7 +83,7 @@ impl Cursor {
     pub fn alpha(&self) -> u8 {
         return self.style
             .as_ref()
-            .map(|s| 255 * (100 - s.blend))
+            .map(|s| (255 as f32 * ((100 - s.blend) as f32 / (100.0 as f32))) as u8)
             .unwrap_or(255)
     }
 
--- a/src/editor/cursor.rs
+++ b/src/editor/cursor.rs
@@ -80,11 +80,11 @@ impl Cursor {
             .unwrap_or_else(|| default_colors.foreground.unwrap())
     }
 
-    pub fn blend_is_hide(&self) -> bool {
-        self.style
+    pub fn alpha(&self) -> u8 {
+        return self.style
             .as_ref()
-            .map(|s| s.blend == 100)
-            .unwrap_or(false)
+            .map(|s| 255 * (100 - s.blend))
+            .unwrap_or(255)
     }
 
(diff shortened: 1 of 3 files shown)


src/editor/cursor.rs
```diff
@@ -80,6 +80,13 @@ impl Cursor {
             .unwrap_or_else(|| default_colors.foreground.unwrap())
     }
 
+    pub fn blend_is_hide(&self) -> bool {
```
Kethku, line 83: maybe rename this to "should_hide"

Also rather than special casing rendering or not, why not have blend apply transparency?

Kethku: image
Specifically the docs indicate that this may be the intended usecase.

Kethku: Actually this is a solution to the background transparency issue we've been having when trying to make the window transparent. I believe this is a great option for that

0x00002a: So, just to check, do you mean apply an alpha scaled based on blend between 0 and 100 to `foreground` and `background` in `draw`? Also theres a weird issue where the cursor doesn't go invisible when opening nerdtree (which I think is a simple vsplit), even if `guicursor` is set to something with blend=100, it will update upon switching windows or entering command mode though. I'm guessing it isn't doing a redraw when it should but I don't know much about this codebase yet :p, any pointers?

Kethku: Yes an alpha blend seems like the way to go. The cursor is redrawn on the base canvas with each frame, so it should be fine.

Re redraw. Thats odd, one way to force a draw is to take a reference to https://github.com/neovide/neovide/blob/main/src/redraw_scheduler.rs and then call queue_next_frame. This will force the system to draw what it's got at the next available time. I'd probably try queueing a frame when the renderer receives updated styles and or new cursor

0x00002a:
OK I've changed it to use transparency, so now it will scale alpha between 0 and 255 using blend between 0 and 100. I've also fixed the issue with the cursor not updating, the issue was the cursor style needs updating when the guicursor option changes. In order to do this I've added awareness of the current mode to `Editor`.

src/editor/mod.rs
```diff
@@ -79,11 +81,21 @@ impl Editor {
                 RedrawEvent::SetTitle { title } => {
                     EVENT_AGGREGATOR.send(WindowCommand::TitleChanged(title));
                 }
-                RedrawEvent::ModeInfoSet { cursor_modes } => self.mode_list = cursor_modes,
+                RedrawEvent::ModeInfoSet { cursor_modes } => {
+                    self.mode_list = cursor_modes;
+                    if let Some(curr_mode_i) = self.current_mode_index.clone() {
```
Kethku, line 86: nit: please write out current, helps with somebody reading the code for the first time

Kethku:
Looks like there is a formatting error which can be fixed with `cargo fmt --all`. I also commented about a small naming issue, but other than that, looks great to me :) thanks so much for the change

0x00002a:
That should do it, my pleasure, I tried neovide out recently and I'm very much enjoying it

Kethku:
beautiful and probably a new speed record ;)

Will merge after the lint finishes

emmanueltouzery:
Note that this PR caused a regression, I opened a separate bug: https://github.com/neovide/neovide/issues/1207

emmanueltouzery:
Sorry for the confusion: this PR has nothing to do with the bug I opened, it didn't cause any regression, I believe. Disregard my previous message.

PR #1201, opened 2022-02-07
Fix update loop

Kethku:
Addresses low framerate caused by faulty usage of winit event loop.

This PR changes the redraw scheduler to take a reference to the window object. This lets the scheduler queue a redraw when the app is waiting. I also changed the scheduler to keep track of a running buffer of frames.

So now whenever a frame is queued, the render loop switches to poll mode for 100 frames. Once those frames run out, the system swaps to wait mode where it stops the event loop until another frame is queued or some window event occurs. This lets us render as fast as possible when animations are playing, but stop rendering when nothing has changed on the screen in order to save processing power and battery.

- Fix

- Yes

The default configured refresh rate is now set to 0. This falls back to the recorded refresh rate from connected monitors observed on startup. This way the system automatically renders as fast as the fastest monitor by default, but is still configurable if the user desires.

### DIFF
--- a/src/editor/cursor.rs
+++ b/src/editor/cursor.rs
@@ -84,7 +84,7 @@ impl Cursor {
         return self
             .style
             .as_ref()
-            .map(|s| (255 as f32 * ((100 - s.blend) as f32 / (100.0 as f32))) as u8)
+            .map(|s| (255_f32 * ((100 - s.blend) as f32 / (100.0_f32))) as u8)
             .unwrap_or(255);
     }
 
--- a/src/editor/mod.rs
+++ b/src/editor/mod.rs
@@ -83,7 +83,7 @@ impl Editor {
                 }
                 RedrawEvent::ModeInfoSet { cursor_modes } => {
                     self.mode_list = cursor_modes;
-                    if let Some(current_mode_i) = self.current_mode_index.clone() {
+                    if let Some(current_mode_i) = self.current_mode_index {
                         if let Some(current_mode) = self.mode_list.get(current_mode_i as usize) {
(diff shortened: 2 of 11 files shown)


Kethku:
This PR currently freezes from time to time. Its unclear what causes this issue, but this should not be merged until that is addressed.

j4qfrost:
@Kethku how exactly does it freeze? Going to write some automation to try to catch the issue.

Kethku:
I've yet to track it down. I'm currently using this branch on my work mac machine without ever freezing, so its possible that I've fixed the issue. This was also reported in the discord by warlord777

Kethku:
The other blocker for this branch is that it uses WAY too much gpu and cpu. I think I'm going to have to rework it to use a spinsleep in order to hit the right frame times, but I haven't gotten around to it.

Heres some vaguely related things I wrote in the discord:

```
As a side note: I've been testing things out and haven't arrived at a happy place re performance, so I'm going to postpone merging the new pr as its not worse than what is currently committed to main. I've also been testing on the previously mentioned mac on a beefy 5k monitor and have noticed some serious performance problems.

As a related effort, I've been playing around with Zed's graphics system which rumor has it may be opensourced in the not too distant future. So I think my plan for the short run is to wait on making some performance improvements until I can learn what I can from the Zed setup. If things go how I expect they might, I think its very possible that I can take another stab at performance using the custom renderer from Zed which should be much faster, or at the very least, more understandable than the opaque performance issues I've been running into with Skia.

TLDR: I don't think the perf issues should block a release, so the only remaining problems are the ones with shell command setup that may be solved by the #1217 PR
```

PR #1214, opened 2022-02-12
Handle underscores in the font name as spaces

### LINKED ISSUE #1213: Underscores in the `guifont` option get used as-is instead of converted to spaces
**Describe the bug**
According to `:h guifont`, underscores can be used instead of spaces in that option:

>    - A '_' can be used in the place of a space, so you don't need to use
>      backslashes to escape the spaces.

But actually using this currently fails by falling back to the default font.

**To Reproduce**
1. Install the font "Source Code Pro" (or any other font name with spaces in it) on your system
2. Type `:set guifont=Source\ Code\ Pro`
3. See the font changing accordingly
4. Reset the font by typing `:set guifont=`
5. Type `:set guifont=Source_Code_Pro`
6. See the font not changing at all

**Expected behavior**
To change the font to Source Code Pro instead of staying at the previous one.

 - OS: Arch Linux
 - Neovide Version: 0.6.1
 - Neovim Version: Build from main @ 92bc1725f1733547eb0ae25b740425f03f358c2a

Not applicable.

MultisampledNight:
- Fix

- No

This fixes #1213.

### DIFF
--- a/src/renderer/fonts/font_options.rs
+++ b/src/renderer/fonts/font_options.rs
@@ -23,7 +23,7 @@ impl FontOptions {
             let parsed_font_list: Vec<String> = parts
                 .split(',')
                 .filter(|fallback| !fallback.is_empty())
-                .map(|fallback| fallback.to_string())
+                .map(|fallback| fallback.replace('_', " "))
                 .collect();
 
             if !parsed_font_list.is_empty() {


Kethku:
Thanks for the fix! I agree this is probably a good idea, just so that it works how people would expect from looking at guifont docs

PR #1217, opened 2022-02-15
fix: bridge command

### LINKED ISSUE #1175: Does not start on macOS

jaeheonji:
This PR resolves #1175 

- Fix

- No

First of all, I got rid of Interactive (`-i`) options in shell. because, If a banner or similar text is output at the start of the shell, the output buffer of the executed command cannot be retrieved normally.

**Example:**
화면 캡처 2022-02-16 015645

(But, in the case of Windows, the command is executed through `wsl`. It may not be affected by files such as `bashrc`, `zshrc`, `bashenv`, `zshenv`, etc)

And referring to the specification of the `Command` as follows

> * No arguments to the program
> * Inherit the current process’s environment

In macOS, It seems that we do not need to run the command through the shell (with `-c`), but simply execute the run command directly.

This may not be a perfect solution. So, I'd really appreciate it if you could give me feedback.

### DIFF
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -48,18 +48,18 @@ fn build_nvim_cmd() -> TokioCommand {
 }
 
 // Creates a shell command if needed on this platform (wsl or macos)
-fn create_platform_shell_command(command: String) -> Option<StdCommand> {
+fn create_platform_shell_command(command: &str, args: Vec<&str>) -> Option<StdCommand> {
     if cfg!(target_os = "windows") && SETTINGS.get::<CmdLineSettings>().wsl {
         let mut result = StdCommand::new("wsl");
         result.args(&["$SHELL", "-lic"]);
         result.arg(command);
+        result.args(&args);
 
         Some(result)
     } else if cfg!(target_os = "macos") {
-        let shell = env::var("SHELL").unwrap();
-        let mut result = StdCommand::new(shell);
-        result.args(&["-lic"]);
-        result.arg(command);
+        let mut result = StdCommand::new(command);
+        result.args(&args);
+
         Some(result)
     } else {
         None
@@ -67,7 +67,8 @@ fn create_platform_shell_command(command: String) -> Option<StdCommand> {
 }
 
 fn platform_exists(bin: &str) -> bool {
-    if let Some(mut exists_command) = create_platform_shell_command(format!("exists -x {}", bin)) {
+    // if let Some(mut exists_command) = create_platform_shell_command(format!("exists -x {}", bin)) {
+    if let Some(mut exists_command) = create_platform_shell_command("exists", vec!["-x", bin]) {
         if let Ok(output) = exists_command.output() {
             output.status.success()
         } else {
@@ -80,7 +81,8 @@ fn platform_exists(bin: &str) -> bool {
 }
 
 fn platform_which(bin: &str) -> Option<String> {
-    if let Some(mut which_command) = create_platform_shell_command(format!("which {}", bin)) {
+    if let Some(mut which_command) = create_platform_shell_command("which", vec![bin]) {
+        info!("Running which command: {:?}", which_command);
         if let Ok(output) = which_command.output() {
             if output.status.success() {
                 let nvim_path = String::from_utf8(output.stdout).unwrap();
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -51,9 +51,8 @@ fn build_nvim_cmd() -> TokioCommand {
 fn create_platform_shell_command(command: &str, args: Vec<&str>) -> Option<StdCommand> {
     if cfg!(target_os = "windows") && SETTINGS.get::<CmdLineSettings>().wsl {
         let mut result = StdCommand::new("wsl");
-        result.args(&["$SHELL", "-lic"]);
-        result.arg(command);
-        result.args(&args);
+        result.args(&["$SHELL", "-lc"]);
+        result.arg(format!("{} {}", command, args.join(" ")));
 
         Some(result)
     } else if cfg!(target_os = "macos") {
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -4,7 +4,7 @@ use std::{
     process::{Command as StdCommand, Stdio},
 };
 
-use log::{error, info, warn};
+use log::{debug, error, warn};
 use tokio::process::Command as TokioCommand;
 
 use crate::{cmd_line::CmdLineSettings, settings::*};
@@ -12,7 +12,7 @@ use crate::{cmd_line::CmdLineSettings, settings::*};
 pub fn create_nvim_command() -> TokioCommand {
     let mut cmd = build_nvim_cmd();
 
-    info!("Starting neovim with: {:?}", cmd);
+    debug!("Starting neovim with: {:?}", cmd);
 
     #[cfg(not(debug_assertions))]
     cmd.stderr(Stdio::piped());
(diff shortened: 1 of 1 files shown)


src/bridge/command.rs
```diff
@@ -48,26 +48,26 @@ fn build_nvim_cmd() -> TokioCommand {
 }
 
 // Creates a shell command if needed on this platform (wsl or macos)
-fn create_platform_shell_command(command: String) -> Option<StdCommand> {
+fn create_platform_shell_command(command: &str, args: Vec<&str>) -> Option<StdCommand> {
```
MultisampledNight, line 51: Nit: If you change the signature here already, I think it's just fine to take a slice here (`&[&str]`), too.

src/bridge/command.rs
```diff
@@ -80,7 +80,8 @@ fn platform_exists(bin: &str) -> bool {
 }
 
 fn platform_which(bin: &str) -> Option<String> {
-    if let Some(mut which_command) = create_platform_shell_command(format!("which {}", bin)) {
+    if let Some(mut which_command) = create_platform_shell_command("which", vec![bin]) {
+        info!("Running which command: {:?}", which_command);
```
MultisampledNight, line 84: Nit: A `info!` might be a bit too loud than actually needed. I think a `debug!` or even `trace!` might be better, the actual command executed is rarely needed imo.

src/bridge/command.rs
```diff
@@ -48,26 +48,26 @@ fn build_nvim_cmd() -> TokioCommand {
...
+
         Some(result)
     } else {
         None
     }
 }
 
 fn platform_exists(bin: &str) -> bool {
-    if let Some(mut exists_command) = create_platform_shell_command(format!("exists -x {}", bin)) {
+    // if let Some(mut exists_command) = create_platform_shell_command(format!("exists -x {}", bin)) {
```
MultisampledNight, line 69: Even smaller nit: Delete the line rather than commenting it out. Git takes care of old things already, you don't need to preserve them manually.

MultisampledNight (commented):
Some small nits, but looks good to me generally.

src/bridge/command.rs
jaeheonji: Oh! This is my mistake. I forgot to delete 😅  thank you!

bmulholland:
I confirm this works for me on my 2021 Macbook Pro w. M1 Pro

MultisampledNight:
To be honest, then I think nothing keeps this from merging anymore.

shaunsingh:
LGTM and runs fine on my machine

j4qfrost:
Absolutely not a complaint, but this also includes refactor changes in addition to the fix. Thanks for PR. Merging.

clason:
> I confirm this works for me on my 2021 Macbook Pro w. M1 Pro

Did you test the bundle as well (`cargo bundle --release`), or just the standard binary?

j4qfrost:
> > I confirm this works for me on my 2021 Macbook Pro w. M1 Pro
> 
> Did you test the bundle as well (`cargo bundle --release`), or just the standard binary?

Did a problem arise?

clason:
Yes, I can't launch the app from the Finder anymore. (`open neovide.app` still works, though, which implies that there's a problem with the environment passed to Neovide.)

jaeheonji:
@clason Have you ever installed `neovim` through `brew` package manager?

clason:
No.

clason:
I think this assumption is not valid when running from Finder:

> In macOS, It seems that we do not need to run the command through the shell (with -c), but simply execute the run command directly.

(I'm sorry for not testing this before it got merged.)

jaeheonji:
I just assumed it was a problem because neovim installed with brew did not exist in the PATH.
Anyway, I'm not sure what exactly is the problem currently, so I'll try to re-test it.

clason:
The problem is that if you launch an app from Finder and not through a shell, it will not have most of the environment variables set that are necessary for it to find and start Neovim.

clason:
No, that's definitely not it. It's a proper regression.

jaeheonji:
When running an app through Finder, a specific command can be normally executed only when the shell is executed in `Interactive Mode` like `/bin/zsh -i "which nvim"`

However, the problem is that if a specific command is executed interactively, incorrect output is obtained according to the settings of each shell. For example, in the case of zsh, when executing the `neofetch` command or `echo "..."` in the `.zshrc` file.

If we roll back to the previous code, execution through Finder works normally., but the user should never put any code that causes stdout or a command like `echo` in `.zshrc` or `.bashrc`.

I still don't know what is the best way to solve all of this.

jaeheonji:
Okay.. First of all, you need `lc`options to run it in Finder.

However, in this case, if the user installed neovim via `brew`, it will not run because the installation path of `nvim` is not in the `PATH`. So at this point it seems like the best way to specify this like:

```
$ sudo ln -s /opt/homebrew/Celler/neovim/.../bin/nvim /usr/local/bin/nvim
```

clason:
I think that is discouraged by homebrew...

(Personally, I install Neovim by hand, so this won't affect me, but forcing people to do that will lead to a lot of complaints from homebrew users....)

jaeheonji:
```
...
    } else if cfg!(target_os = "macos") {
        let shell = env::var("SHELL").unwrap_or_else(|_| "/bin/sh".to_string());
        let mut result = StdCommand::new(&shell);
        result.env("PATH", "/opt/homebrew/bin");

        result.args(&["-lc"]);
        result.arg(format!("{} {}", command, args.join(" ")));

        Some(result)
}
```

PR #1227, opened 2022-02-19
add width to guifont settings

fritsstegmann:
- Feature

Allows for letter spacing by setting the width in guifont

No

### DIFF
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -176,7 +176,7 @@ impl CachingShaper {
 
     pub fn font_base_dimensions(&mut self) -> (u64, u64) {
         let (metrics, glyph_advance) = self.info();
-        let font_width = (glyph_advance + 0.5).floor() as u64;
+        let font_width = (glyph_advance + (self.options.width - self.options.size + 0.5)).floor() as u64;
 
         let bare_font_height = (metrics.ascent + metrics.descent + metrics.leading).ceil();
         let font_height = bare_font_height as i64 + self.linespace;
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -179,8 +179,7 @@ impl CachingShaper {
 
         let bare_font_height = (metrics.ascent + metrics.descent + metrics.leading).ceil();
         let font_height = bare_font_height as i64 + self.linespace;
-        let font_width =
-            (glyph_advance + (self.options.width - self.options.size + 0.5)).floor() as u64;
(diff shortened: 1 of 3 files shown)


src/renderer/fonts/caching_shaper.rs
```diff
@@ -150,7 +150,8 @@ impl CachingShaper {
     pub fn font_base_dimensions(&mut self) -> (u64, u64) {
         let (metrics, glyph_advance) = self.info();
         let font_height = (metrics.ascent + metrics.descent + metrics.leading).ceil() as u64;
-        let font_width = (glyph_advance + 0.5).floor() as u64;
+        let font_width =
+            (glyph_advance + (self.options.width - self.options.size + 0.5)).floor() as u64;
```
MultisampledNight, line 154: Why minus the font size here? Isn't that size only vertically?

fritsstegmann: so It looks like the glyph_advance is the width derived the font set, I added in the minus so that the width setting is relative to the font size, <fontFamily>:h12:w12 would be equal to the current view but upping the width relative to the fontsize makes the font go wider, less then the font size makes it go tighter. I did this to make VictorMono look a bit better, it is to tight for my liking. 

this was my config:

vim.opt.guifont = "VictorMono Nerd Font:h12:w13.5"

MultisampledNight: Hm, in my opinion the "width" should be in absolute units, and not change when the actual size (which I interpret as height) changes. But fair enough if that works better for you.

MultisampledNight:
Huh, why did you close the PR? The change could be useful for sure!

fritsstegmann:
Ah, sorry, though you did not like it.

fritsstegmann:
marking it as draft because there are one or two things I'd like to do.

src/renderer/fonts/caching_shaper.rs
fritsstegmann: uhm, yeah, but from what I can see the width is partly calculated of the font size, the user won't know what the width that was calculated, so something relavtive to the font size is the best they to go by to set the width. It'll also stop them from having to set both the width and size when they make font updates.

MultisampledNight:
Did anything change in that regard? Else I think this could be just merged.

obmutescences:
any progress about this?

MultisampledNight:
Rebased on top of main, intending to merge this. I do want to change the width behavior though, since I believe an offset would be more comfortable to use. Also need to add it to the docs that were added in meantime.

MultisampledNight:
Woah, quite a time ago. Sorry to have kept you waiting for a year and a half, but thank you still!! Choooooooooo choooooooooooooooo! :steam_locomotive:

PR #1242, opened 2022-03-10
Fix: Neovide.app not working in Finder 

### LINKED ISSUE #1240: Neovide.app wont Open in Finder, but opens if alias in the terminal
Last time I compile before the last merge on bridge patch im not facing this...

well its good that the cli neovide is working now 
my only problem on that instance of neovide is i cannot override CMD + h 
if it is spawn using neovide cli.

Neovide.app the latest one I compile when double click or run 
Just wont start... if done on any GUI file explorer like finder...

I tried to use an alias on it and invoke in the terminal
and It start...

Is there a way both the cli and Neovide.app 
have the same instance?

its kinda side since I cannot find the instance or App being launch by the cli 
to point in macos keyboard to override it...

Looking forward for the next patch 

still thanks for this great project, I used it a lot in Rust :)

jaeheonji:
This PR resolve #1240 

- Fix

- No

The content in #1217 is getting long, so I'm making this PR with a new alternative.
I hadn't thought of this PR (#1072), so currently `Neovide.app` cannot be run through Finder on macOS.

~~However, this still leaves some problems.~~

~~For example, if neovim is installed through `brew`, it will not run (in Finder) because `nvim` does not exist in the `PATH` variable (default installation path is `/opt/homebrew/bin/`).~~

~~But one thing I'm concerned about is that hardcoding the settings of a 3rd party app has maintenance issues later on.
So, if there is a better way, please give feedback.~~

EDIT: Thank you, everyone. I removed direct hardcoding of brew to `PATH` and added a way to solve this problem instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,6 +253,7 @@ snap install neovide
 
 ## Troubleshooting
 - Neovide requires that a font be set in `init.vim` otherwise errors might be encountered. This can be fixed by adding `set guifont=Your\ Font\ Name:h15` in init.vim file. Reference issue [#527](https://github.com/neovide/neovide/issues/527).
+- If you installed `neovim` via Apple Silicon (M1)-based `brew`, you have to add the `brew prefix` to `$PATH` to run `Neovide.app` in GUI. Please see the [homebrew documentation](https://docs.brew.sh/FAQ#my-mac-apps-dont-find-homebrew-utilities). Reference issue [#1242](https://github.com/neovide/neovide/pull/1242)
 
 ### Linux-specific
 - If you recieve errors complaining about DRI3 settings, please reference issue [#44](https://github.com/neovide/neovide/issues/44#issuecomment-578618052).
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -58,6 +58,8 @@ fn create_platform_shell_command(command: &str, args: &[&str]) -> Option<StdComm
     } else if cfg!(target_os = "macos") {
         let shell = env::var("SHELL").unwrap_or_else(|_| "/bin/sh".to_string());
         let mut result = StdCommand::new(&shell);
+        result.env("PATH", "/opt/homebrew/bin");
+
         result.args(&["-lc"]);
         result.arg(format!("{} {}", command, args.join(" ")));
 
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -56,8 +56,10 @@ fn create_platform_shell_command(command: &str, args: &[&str]) -> Option<StdComm
 
         Some(result)
     } else if cfg!(target_os = "macos") {
-        let mut result = StdCommand::new(command);
-        result.args(args);
+        let shell = env::var("SHELL").unwrap_or_else(|_| "/bin/sh".to_string());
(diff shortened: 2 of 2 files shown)


danielnehrig:
For example, if neovim is installed through brew, it will not run (in Finder) because nvim does not exist in the PATH variable (default installation path is /opt/homebrew/bin/).

I don't understand really... if neovim is not part of the Path, that means one can also not run it in the terminal without specifying the full path to the binary. This looks rather like a local dev setup thing problem. I think hardcoding the path like this is not a solution, especially not for a broader audience, as for me the `/opt/homebrew` does not even exist. And the default installation path of my neovim is `/usr/local/bin/nvim` in my case.

jaeheonji:
@danielnehrig Thanks for your feedback

> I think hardcoding the path like this is not a solution

Yes, I agree 100%. As I have already said, this is a very bad idea. And, in your environment this shouldn't be a problem. Because, it's more correct to say that this is a problem in the [brew used by Apple Slicon](https://brew.sh/2021/02/05/homebrew-3.0.0/).

When installing `nvim` through `brew` in Intel X86 environment, it is installed as `/usr/local` path, but in case of M1, it is installed as `/opt/homebrew`. And since this is not a general `$PATH` environment, the `.app` file cannot be executed through Finder.

I tried to assign `--neovim-bin` directly through `.plist` modification, but it still doesn't work. (Of course, I could be wrong about this.)

It would be nice to have a more fundamental solution, but I actually don't know a good solution yet. (This PR is just one example.)
One way to avoid modifying the code is simply to create an nvim symbolic link in the `/usr/loacl/bin` path only for M1 users.

danielnehrig:
what about let the user fill out the path himself in this case add `/opt/homebrew`
how about reading the current loaded shell which is set in /etc/passwd and set via `pwd` or use `/bin/sh` since it's a symlink
and launch neovide with the shell containing your set path

jaeheonji:
Sorry! please make sure I understand correctly.

Is your intention correct to mean that the user registers `neovim` path to the `$PATH` "by himself" in the current default shell, and then loads the shell configuration (with `$PATH`) before `neovide` is run?

henryhchchc:
I found a section in [Homebrew documentation](https://docs.brew.sh/FAQ#my-mac-apps-dont-find-homebrew-utilities) talking about this issue. Hope it helps.

jaeheonji:
@henryhchchc Oh! Thanks 😄 I didn't know that.

owittek:
> I found a section in [Homebrew documentation](https://docs.brew.sh/FAQ#my-mac-apps-dont-find-homebrew-utilities) talking about this issue. Hope it helps.

This seems to work for opening the GUI in finder but it is still not possible to open it from spotlight/alfred. There are currently two fixes for that I know of:

1. Adding the path via .zprofile as suggested here: https://github.com/neovide/neovide/issues/1373#issuecomment-1179268069
2. symlinking nvim to /usr/local/bin ( ln -s /opt/homebrew/bin/nvim /usr/local/bin/nvim ).

But both solutions break the telescope plugin as nvim opens in the bin folder. This leads to telescope not responding to input.

PR #1246, opened 2022-03-16
error: there is no argument named `min_ft` solved

huncut2016:
- Fix (I can't compile on my computer, and this small code change fixed it)

- No

### DIFF
--- a/src/renderer/profiler.rs
+++ b/src/renderer/profiler.rs
@@ -114,19 +114,19 @@ impl Profiler {
 
         // Show min, max, avg
         root_canvas.draw_str(
-            format!("min: {min_ft:.1}ms"),
+            format!("min: {}ms", min_ft),
             (rect.left, rect.bottom),
             &self.font.skia_font,
             &paint,
         );
         root_canvas.draw_str(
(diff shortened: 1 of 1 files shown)


MultisampledNight:
By the way, the change is not needed if you run `rustup update`. Since Rust 1.58, which was released on January 22th, Rust can capture variables in such format strings directly.

j4qfrost (approved):
LGTM

LoipesMas:
As MultisampledNight said, it's a fairly new rust feature, so updating your rust toolchain also "fixes" this. IIRC Neovide doesn't have a specified minimal supported rust version, so we *can* expect users who compile it to use the latest version, but it's a minor readability thing (which already has caused some confusions such as this one), so we can use the older way.
But right now it wouldn't be the same, because you omitted the formatting argument, it should be `{:.1}ms` (in all three lines).
If you fix that, we can merge this to avoid further confusion for such minor thing.

huncut2016:
@LoipesMas I'm done!

LoipesMas:
Thanks!

PR #1247, opened 2022-03-17
Add neovide_confirm_quit option to confirm before quitting

### LINKED ISSUE #874: Prompt for save on exit with window X button
It's easy to lose work with neovide in cases where that work would not be lost in GVim, simply because neovide has no save prompt when exiting with the window's X button

Ideally we would have a save prompt similar to GVim: 

Question_312

As mentioned in the discussion [here](https://github.com/neovide/neovide/discussions/871), switching the exit command from `:qall!` to `:qall` would fix the core issue, though it wouldn't be as smooth a workflow as a save dialogue.

Honestly, I am baffled how any text editor is missing such a basic safeguard against lost work, and more, how I can't find anyone else mentioning this.

viviicat:
Screenshot from 2022-03-16 19-45-05

This adds a new option, `g:neovide_confirm_quit` (default: `0`), which, when set to `1` or `v:true`, will execute `confirm qa` instead of `qa!` when the window is closed.

Mostly fixes #874, by reducing the potential to lose work. It does not pop up a nice dialog box, of course. This is pretty much the way that nvim-qt does things.

Tested on Linux.

Please let me know if there's a more elegant way to query the option than putting it in an inline script.

- Feature

- No, the option is disabled by default.

### DIFF
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -127,7 +127,9 @@ impl ParallelCommand {
     async fn execute(self, nvim: &Neovim<TxWrapper>) {
         match self {
             ParallelCommand::Quit => {
-                nvim.command("qa!").await.ok();
+                nvim.command("if exists('g:neovide_confirm_quit') && g:neovide_confirm_quit == 1 | confirm qa | else | qa! | endif")
+                    .await
+                    .ok();
             }
             ParallelCommand::Resize { width, height } => nvim
                 .ui_try_resize(width.max(10) as i64, height.max(3) as i64)
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -127,7 +127,7 @@ impl ParallelCommand {
     async fn execute(self, nvim: &Neovim<TxWrapper>) {
         match self {
             ParallelCommand::Quit => {
-                nvim.command("if exists('g:neovide_confirm_quit') && g:neovide_confirm_quit == 1 | confirm qa | else | qa! | endif")
+                nvim.command("if get(g:, 'neovide_confirm_quit', 0) == 1 | confirm qa | else | qa! | endif")
                     .await
                     .ok();
             }
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -127,9 +127,11 @@ impl ParallelCommand {
     async fn execute(self, nvim: &Neovim<TxWrapper>) {
         match self {
             ParallelCommand::Quit => {
-                nvim.command("if get(g:, 'neovide_confirm_quit', 0) == 1 | confirm qa | else | qa! | endif")
-                    .await
-                    .ok();
+                nvim.command(
+                    "if get(g:, 'neovide_confirm_quit', 0) == 1 | confirm qa | else | qa! | endif",
+                )
+                .await
+                .ok();
             }
             ParallelCommand::Resize { width, height } => nvim
                 .ui_try_resize(width.max(10) as i64, height.max(3) as i64)


viviicat:
fixing formatting errors...

Shatur:
I would enable it by default.

tobealive:
@viviicat awesome 💪🙏

viviicat:
@Shatur personally I like the idea of it being on by default, but since that would change default behavior I think it would be up to the project owner(s).

ernie:
This does not seem to work on MacOS 12.3, neither via command-q or via clicking the menu option.

dineshKumar777:
> This does not seem to work on MacOS 12.3, neither via command-q or via clicking the menu option.

Just confirming. By default this is disabled. Did you enable it by using g:neovide_confirm_quit=1??

ernie:
> Just confirming. By default this option is disabled. Did you enable it by using g:neovide_confirm_quit=1??

I tried both `g:neovide_confirm_quit=1` and `g:neovide_confirm_quit=v:true` with no luck.

ernie:
> To be safe i'll ask this dump question. Did you build it again from source after the commit containing this functionality was merged?

Not a dumb question at all! I'm a new user (only found Neovide last week), so I've only ever built the app since after it was merged.

tobealive:
Okay nice.

A further step i would try is to enable it for the active session by entering the vim command `:let g:neovide_confirm_quit=1` then check if there is a confirmation request when trying to close a session with modified, unsaved files

viviicat:
Oh, another question: Do you have a modified buffer open? It won't ask to confirm if nothing requires saving.

Also, what happens if you just try to manually run `:confirm qa`?

ernie:
> Oh, another question: Do you have a modified buffer open? It won't ask to confirm if nothing requires saving. (It should prompt with anything as simple as typing gibberish into the default buffer)

Good question, and yes! I even used the `ui_commands.rs` file from your screenshot during my verification efforts. :D

> 
> Also, what happens if you just try to manually run `:confirm qa`?

That works as expected.

viviicat:
I guess the event is being bypassed, then.

dineshKumar777:
@ernie we are not able to reproduce your issue in both windows and mac os. Can you try again with latest neovide.
2020 m1 macbook pro - macOS monterey 12.2.1

if issue still persists with latest version, it would be helpful if you share a screen recording.

https://user-images.githubusercontent.com/21097695/163696796-042abf4b-59ab-40f0-87ed-c7c7021628b9.mp4

tobealive:
Also `ZQ` which is alias for `:q` is not triggering it btw. 

"OS"-window closing should trigger it.

ernie:
> Also `ZQ` which is alias for `:q!` will not triggering it btw.
> 
> "OS"-window closing should trigger it.

Aha! So, Command-Q didn't trigger it. Only clicking the close button in the top left. That explains it, as I never quit Mac apps using the X

I was able to get it "working" by clicking the X.

Neither Command-Q nor selecting neovide -> quit neovide worked.

tobealive:
Yes, this is not saying one must use the X. Command Q should do the thing. I'm not having access to a mac right now. Maybe the fellow thread readers can give some more input.

ernie:
I ran with `—log` to see if an event was seen before quitting. The only time I saw the CloseRequested event was when clicking the X. Upon command-q the key press wasn’t reflected anywhere, the log just ceased.

tobealive:
> @ernie we are not able to reproduce your issue in both windows and mac os. Can you try again with latest neovide. 2020 m1 macbook pro - macOS monterey 12.2.1
> 
> if issue still persists with latest version, it would be helpful if you share a screen recording.
> 
>  2022-04-16_16-57-15.1.mp4

This is off topic but this is a wonderful job on desktop customization in this little preview. Didn't knew it's possible to go on Mac that far. Could provide a link or recommend a resource?

dineshKumar777:
> I ran with `—log` to see if an event was seen before quitting. The only time I saw the CloseRequested event was when clicking the X. Upon command-q the key press wasn’t reflected anywhere, the log just ceased.

@ernie Thanks for the feedback. good thing we are progressing on finding the root cause. 
FYI, in windows I tried with ALT+ F4, and this works as intended. Seems like mac specific issue. Will create a separate bug to track this issue.

> This is off topic but this is a wonderful job on desktop customization in this little preview. Didn't knew it's possible to go on Mac that far. Could provide a link or recommend a resource?

@tobealive I never used Mac device so far :( . This is verified and shared in neovide discord channel by another community member. You can join in if you are interested.

technicalpickles:
Since the last comment, https://github.com/neovide/neovide/pull/1247 was closed by https://github.com/rust-windowing/winit/pull/1583 . It looks like we are using the fork at [neovide/winit on the new-keyboard-all branch](https://github.com/neovide/winit/tree/new-keyboard-all). It doesn't look like the commits in that tree are upstream though.

MultisampledNight:
Yes, and we'll have to wait until we can use upstream's `new-keyboard` branch, because merging in every single upstream change is not really viable.

PR #1257, opened 2022-03-31
Render block cursor as an outline when window is not focused

### LINKED ISSUE #1253: Change cursor state when neovide window is not focused
The cursor state does not change when neovide window is not focused. Check the attached screenshots on how gVim changes the state of the cursor when its window is not focused.

Check the state of the cursor:

When gVim is not focused:
gvim-not-focused

When neovide is not focused:
neovide-not-focused

hallettj:
- Feature

- No

Hi! I'm really enjoying Neovide! This PR changes the way the cursor is rendered when the editor window is unfocused. If the cursor shape is set to block, then when the window loses focus the cursor will be drawn as an outline. This provides a visual cue that can be easier to notice than other cues like highlighted window borders or title bar because the user tends to focus on the cursor. After using Neovide for a bit I have realized that I rely on the cursor rendering cue - I've been typing into the wrong window much more often recently.

There are still some tweaks that could be made. I have not added tests yet, and I think the outline is a little too thick. I would appreciate any feedback, and I will try to make updates accordingly.

Fixes #1253

### DIFF
--- a/src/renderer/cursor_renderer/mod.rs
+++ b/src/renderer/cursor_renderer/mod.rs
@@ -430,13 +430,12 @@ impl CursorRenderer {
         subtract.line_to(self.corners[3].current_position + offsets[3]);
         subtract.close();
 
-        match op(&rectangle, &subtract, skia_safe::PathOp::Difference) {
-            Some(path) => {
-                canvas.draw_path(&path, &paint);
-                path
-            }
-            // TODO: How to handle the failure case?
-            None => self.draw_rectangle(canvas, paint),
-        }
+        // We have two "rectangles"; create an outline path by subtracting the smaller rectangle
+        // from the larger one. This can fail in which case we return a full "rectangle".
+        // TODO: Is this the best way to handle the failure case? Would unwrap be ok?
+        let path = op(&rectangle, &subtract, skia_safe::PathOp::Difference).unwrap_or(rectangle);
+
+        canvas.draw_path(&path, &paint);
+        path
     }
 }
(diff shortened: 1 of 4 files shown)


src/renderer/cursor_renderer/mod.rs
```diff
@@ -286,6 +299,10 @@ impl CursorRenderer {
         )
             .into();
 
+        // The width for the outline of a block cursor should be the same as the width of
+        // a vertical cursor.
+        let outline_width = cursor_dimensions.x * DEFAULT_CELL_PERCENTAGE * 2.0;
```
MultisampledNight, line 304: Would it matter to make this a configurable option? This is more of a selfish request, I like these kinds of outlines to be very thin.

To introduce the new config option, just add it to the `CursorSettings` struct, you can access it here through `settings` which is already in scope then.

hallettj:
I added a configuration option for the outline width. (Thanks for the feedback @MultisampledNight!) I used ems as the unit for the option because I believe that is the most widely-used convention for specifying widths relative to the configured font size. To make that work I also took a stab at inferring the appropriate em size, and I added a property for that to `GridRenderer`. In theory the em size is the same as the font size.

MultisampledNight (approved):
Looks good to me! When this PR will be merged, I'll take care of transferring the documentation comment to the wiki.

njkevlani:
Working as expected, thanks for the work!

komar007:
Thanks for this super-useful change!. I was going to ask for this kind of improvement, because I was constantly trying and failing to work without focus. I am so used to the way most terminal emulators work that I would automatically assume focus when the cursor was filled.

PR #1275, opened 2022-04-16
Open multiple files as Tabs

SergioRibera:
- Feature

- No

## More details
This PR brings with it the ability to open multiple files in multiple tabs, to improve the experience as a text editor

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -86,14 +86,11 @@ async fn start_neovim_runtime() {
 
     start_ui_command_handler(nvim.clone());
 
-    // Verify if the files to open is more than one
-    if settings.target_files.len() > 1 {
-        // Open the rest of the files into new tabs
-        for file in settings.target_files.iter().skip(1) {
-            nvim.command(format!("tabnew {}", file).as_str())
-                .await
-                .unwrap_or_explained_panic("Could not create new tab");
-        }
+    // Open the files into new tabs
+    for file in settings.target_files.iter().skip(1) {
+        nvim.command(format!("tabnew {}", file).as_str())
+            .await
+            .unwrap_or_explained_panic("Could not create new tab");
     }
     SETTINGS.read_initial_values(&nvim).await;
     SETTINGS.setup_changed_listeners(&nvim).await;
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -87,9 +87,9 @@ async fn start_neovim_runtime() {
     start_ui_command_handler(nvim.clone());
 
     // Verify if the files to open is more than one
-    if settings.files.len() > 1 {
+    if settings.target_files.len() > 1 {
         // Open the rest of the files into new tabs
-        for file in settings.files.iter().skip(1) {
+        for file in settings.target_files.iter().skip(1) {
             nvim.command(format!("tabnew {}", file).as_str())
                 .await
                 .unwrap_or_explained_panic("Could not create new tab");
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -85,6 +85,17 @@ async fn start_neovim_runtime() {
     let nvim = Arc::new(nvim);
 
     start_ui_command_handler(nvim.clone());
+    
+    // Verify if the files to open is more than one
+    if settings.files.len() > 1 {
+        // Open the first file into a first window
+        let first_file = settings.files.first().unwrap();
+        nvim.command(format!("e {}", first_file).as_str()).await.unwrap_or_explained_panic("Could not create new tab");
+        // Open the rest of the files into new tabs
+        for file in settings.files.iter().skip(1) {
+            nvim.command(format!("tabnew {}", file).as_str()).await.unwrap_or_explained_panic("Could not create new tab");
+        }
+    }
     SETTINGS.read_initial_values(&nvim).await;
     SETTINGS.setup_changed_listeners(&nvim).await;
 
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -86,12 +86,6 @@ async fn start_neovim_runtime() {
 
     start_ui_command_handler(nvim.clone());
 
-    // Open the files into new tabs
-    for file in settings.target_files.iter().skip(1) {
-        nvim.command(format!("tabnew {}", file).as_str())
-            .await
-            .unwrap_or_explained_panic("Could not create new tab");
-    }
     SETTINGS.read_initial_values(&nvim).await;
     SETTINGS.setup_changed_listeners(&nvim).await;
 
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -85,7 +85,6 @@ async fn start_neovim_runtime() {
     let nvim = Arc::new(nvim);
 
     start_ui_command_handler(nvim.clone());
-
     SETTINGS.read_initial_values(&nvim).await;
     SETTINGS.setup_changed_listeners(&nvim).await;
 
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -85,15 +85,14 @@ async fn start_neovim_runtime() {
     let nvim = Arc::new(nvim);
 
     start_ui_command_handler(nvim.clone());
-    
+
     // Verify if the files to open is more than one
     if settings.files.len() > 1 {
-        // Open the first file into a first window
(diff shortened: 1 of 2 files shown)


src/cmd_line.rs
```diff
@@ -197,6 +201,11 @@ pub fn handle_command_line_arguments(args: Vec<String>) -> Result<(), String> {
             .map(|v| v.to_owned())
             .or_else(|| std::env::var("NEOVIDE_X11_WM_CLASS").ok())
             .unwrap_or_else(|| "neovide".to_owned()),
+        // Command-line arguments with multiple files
```
MultisampledNight, line 204: Rather than throwing the comment all over the place, it could be more useful to rename the variable to something like `command_line_files` or `target_files` instead.

MultisampledNight (commented):
Looks good to me.

src/cmd_line.rs
SergioRibera: Hey, that looks good to me, I'll update with this change now :D

LoipesMas:
Can't this be done just by passing [`-p`](https://neovim.io/doc/user/tabpage.html#tab-page-commands) argument to neovim?  
So you would just do `neovide -- -p filename1 filename2`

src/bridge/mod.rs
```diff
@@ -85,6 +85,16 @@ async fn start_neovim_runtime() {
     let nvim = Arc::new(nvim);
 
     start_ui_command_handler(nvim.clone());
+
+    // Verify if the files to open is more than one
+    if settings.target_files.len() > 1 {
```
LoipesMas, line 90: Pretty sure this check can be omitted, because `.skip` will then just return an empty iterator.

SergioRibera: I just checked and it seems that the -p parameter does the same as my PR :c

SergioRibera: And I think that `.skip` is a good option, thank you very much.

SergioRibera: The if has been removed, and regarding the -p option, I think this PR would be more of a shortcut on the neovide side

LoipesMas:
I don't think we should change the default behavior to be different from neovim's.
*Maybe* we could add `-p` argument to neovide, which passes `-p` to neovim (so you could do just `neovide -p filename1 filename2` instead of `neovide -- -p filename1 filename2`, but this doesn't feel right to me either.  
But I'm open to hearing more opinions (since personally I don't use neovide/neovim like that).

SergioRibera:
> I don't think we should change the default behavior to be different from neovim's. _Maybe_ we could add `-p` argument to neovide, which passes `-p` to neovim (so you could do just `neovide -p filename1 filename2` instead of `neovide -- -p filename1 filename2`, but this doesn't feel right to me either. But I'm open to hearing more opinions (since personally I don't use neovide/neovim like that).

I think you are right, my intention was because calling neovide with multiple files only opens one, maybe even adding the -p parameter to the neovim commands is a better idea than what I already did

LoipesMas:
>calling neovide with multiple files only opens one

When you call neovim (or neovide) with multiple files, it opens them in separate buffers. 

Slightly off-topic: I think some users coming to vim (me included) expect to have currently open files in separate tabs (because it resembles how it is in other editors), while actually buffers are for that. So unless you have a reason to use tabs, using buffers will probably be better.

SergioRibera:
> > calling neovide with multiple files only opens one
> 
> When you call neovim (or neovide) with multiple files, it opens them in separate buffers.
> 
> Slightly off-topic: I think some users coming to vim (me included) expect to have currently open files in separate tabs (because it resembles how it is in other editors), while actually buffers are for that. So unless you have a reason to use tabs, using buffers will probably be better.

Hey what's up, I understand that the command I use here, the `:tabnew` opens them in different buffers as you said, the detail and the nice thing is that it generates those tabs visually which in my opinion is more friendly, but tell me, what other suggestion do you have, I've been seeing the same way to insert in the commands for neovim the parameter `-p` in case the files are more than one, maybe this solution is more aligned to what you say, but I'm attentive

MultisampledNight:
Sorry for being unresponsive to this: Generally I'd say we should not have this functionality, since it might break already existing workflows. But in this case I think it's pretty useful even though replicating functionality, and is not too intrusive as the files are still opened in buffers — they are only visible now.

SergioRibera:
Ok, thinking well, I think it is convenient to keep the buffers management, although I think that showing it in tabs is more friendly for those who just start and want to use neovide as their graphic tool, so I can propose two solutions to the problem:

1.- have a "open in tabs" option, like -t or --tabs to enable showing tabs, by default it would not be shown and only opens in buffers.
2.- Default open in tabs unless you pass an option like --no-tabs and only use buffers.

Personally I prefer the second option because it prioritizes ease of use for beginners and more advanced users who prefer buffers can afford to add aliases or the parameter

MultisampledNight:
I prefer the latter one, too. We really don't need to cause even more confusion for someone who knows NeoVim.

LoipesMas:
It bothered me that neovim doesn't have an option to have `-p` be default behavior (because I'd expect it to). But I found a solution:
```vim
" Open all buffers as tabs on enter
au VimEnter * exec "tab all"
```
This works the same as `-p`, when put into `init.vim`.

But beginners' ease of use is a fair point, even though I'm against changing the default behavior.
So now I'm perhaps more in favor of the second option (more advanced users could do something like this:
```vim
set tabpagemax=1
au VimEnter * exec "set tabpagemax=50"
```
see [`'tabpagemax'`](https://neovim.io/doc/user/options.html#'tabpagemax')).

Either way one group gets an inconvenience. (Also I don't know how big either of them is.)

SergioRibera:
Ok, then I update the PR with the second option, which I think affects a group of people more able to disable the option :D

SergioRibera:
> ‎Me molestó que neovim no tenga la opción de tener un comportamiento predeterminado (porque esperaría que lo hiciera). Pero encontré una solución:‎`-p`
> 
> ```viml
> " Open all buffers as tabs on enter
> au VimEnter * exec "tab all"
> ```
> 
> This works the same as `-p`, when put into `init.vim`.
> 
> ‎Pero la facilidad de uso de los principiantes es un punto justo, a pesar de que estoy en contra de cambiar el comportamiento predeterminado. ‎‎ Así que ahora estoy quizás más a favor de la segunda opción (los usuarios más avanzados podrían hacer algo como esto:‎
> 
> ```viml
> set tabpagemax=1
> au VimEnter * exec "set tabpagemax=50"
> ```
> 
> see [`'tabpagemax'`](https://neovim.io/doc/user/options.html#'tabpagemax')).
> 
> ‎De cualquier manera, un grupo recibe un inconveniente. (Tampoco sé qué tan grande es ninguno de ellos).‎

Or maybe in the execution of the embedded lua code check the --tabs flag and include  that solution that @LoipesMas found

SergioRibera:
Ok, I did the integration of this automand and I tell you that nvim does not recognize the filetype of the buffers and throws some lags when I move to another tab.
Should I upload the change? or should we keep this "tabnew"?

SergioRibera:
Hey, I've done some tests and I've come up with this, I think it's one of the best there is, I'm attentive to your opinions :D

MultisampledNight:
Ah, right, forgot about that: I'll add the `--no-tabs` option to the wiki once this is merged, of course.

src/bridge/mod.rs
```diff
@@ -85,6 +85,7 @@ async fn start_neovim_runtime() {
     let nvim = Arc::new(nvim);
 
     start_ui_command_handler(nvim.clone());
+
```
LoipesMas, line 88: I assume this is an accidental leftover, right?

LoipesMas (changes requested):
Save for the leftover newline, all looks good to me.

src/bridge/mod.rs
LoipesMas: (It's nothing breaking, but let's keep the commits/changes tidy)

SergioRibera:
@MultisampledNight Hey thanks for the suggestion, and yes, in this occasion I decided to play it because the change was not very big, but I will take it into account for another occasion, thank you very much :D

src/bridge/mod.rs
SergioRibera: Of course, no problem

MultisampledNight:
Thank you!

MultisampledNight:
Added documentation to the wiki. Thanks again!

PR #1303, opened 2022-05-06
Add configurable blur amount

### LINKED ISSUE #1265: Make floating blur amount configurable

I've recently started using the zen-mode.nvim plugin, which (when enabled) places the current buffer as a floating window covering the entire screen.

Unfortunately, since Neovide makes the background of floating windows transparent, I can still see everything, although it's slightly blurred:

image

This would probably look awesome if I could just increase the blur amount so the background contents were less distracting. At this blur amount, I can basically still read the background text.

I could disable transparency altogether, but that means I would lose out on some of the eye candy that I use Neovide for.

**Additional context**

Would this involve just adding another config option (i.e. `g:neovide_floating_blur_amount = x`) and passing it in here?

https://github.com/neovide/neovide/blob/main/src/renderer/rendered_window.rs#L258

Would there be any reason to expose both blur values? (I assume they mean blur amount in the X and Y axes.) Would that enable some cool effects?

MultisampledNight:
Fixes #1265.

## Open questions (feel free to participate!)

- ~~Should there be different configuration values for the X and Y amount of blur values? (see https://github.com/neovide/neovide/issues/1265#issuecomment-1115091016)~~

- Feature

- No

### DIFF
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -35,6 +35,7 @@ pub struct RendererSettings {
     scroll_animation_length: f32,
     floating_opacity: f32,
     floating_blur: bool,
+    floating_blur_amount: f32,
     debug_renderer: bool,
     profiler: bool,
 }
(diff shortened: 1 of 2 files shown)


MultisampledNight:
Actually, I think split X and Y look pretty cool (though distracting). So I added that.

MrHeer (approved):
LGTM

PR #1306, opened 2022-05-08
don't error on messagepack values encoding invalid strings

cgadski:


src/bridge/events.rs
```diff
@@ -380,7 +380,10 @@ fn parse_map(map_value: Value) -> Result<Vec<(Value, Value)>> {
 }
 
 fn parse_string(string_value: Value) -> Result<String> {
-    string_value.try_into().map_err(ParseError::String)
+    match string_value {
+        Value::String(s) => Ok(s.into_str().unwrap_or_else(|| String::from("?"))),
```
MultisampledNight, line 384: Looks good to me. Maybe using `\u{FFFD}` would be more appropiate though, as `?` is a pretty common character.

cgadski: Sure, sounds good. I was thinking about somehow indicating the length of `s.into_bytes` in the placeholder, but maybe that's not important.

MultisampledNight:
Awesome. Thank you for looking into this!

Issue #1358, opened 2022-06-18
Modifiers (e.g. Alt) are sometimes ignored when using programmable keyboard


quadruple-output:
**Describe the bug**
I am using a programmable keyboard on a Mac. Sometimes when I enter, for example, the "{" character, neovide outputs "8" (Alt-8 is how you enter "{" on a mac). This only seems to occur when I have neovide maximized on my external 4K monitor. It happens at seemingly random times, but often enough to really drive me crazy.

Spoiler: I think I found a fix for the issue and plan to submit a PR.

**To Reproduce**
1. Setup a programmable keyboard, such that a single key press corresponds to pressing "Alt-8". In almost all apps, pressing this key should produce "{" in any kind of text input control.
2. Startup neovide and maximize it on a high resolution monitor (it may be possible to achieve the same effect by putting the CPU under load, but I did not test this at all).
3. Enter Insert Mode and keep hitting your programmed key. Use different time intervals, for example hit the key fast a few times, then wait 2 seconds, continue, wait longer or shorter, continue …
4. When I follow these steps, I see something like this: 8{{88{{8{{8{{{{8{{{ (in a random order). This is an extreme but real example. Sometimes it just works as expected, especially when I keep hitting the key in short intervals.

**Expected behavior**
I would expect to see a series of "{" characters without any "8" in-between.

**Screenshots**
<I think this is not needed>

**Desktop (please complete the following information):**
 - macOS 12.4
 - Neovide Version 0.9.0
 - Neovim Version 0.7.0

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

I had to add a few log statements in `KeyboardManager::handle_event()` in order to see something. This is the condensed result. The first block reflects the "Ok" situation where hitting the key results in "{", while the second block reflects the error. Watch out for the `ModifiersChanged` event which is processed out-of-order with respect to the queued input events. (The time stamps show how quickly the events follow each other. 

```
[2022-06-17 20:01:15.096005 +02:00] DEBUG [neovide::window::keyboard_manager] src/window/keyboard_manager.rs:59: handle_event(WindowEvent { window_id: WindowId(Id(5198875936)), event: KeyboardInput { device_id: DeviceId(DeviceId), event: KeyEvent { physical_key: AltLeft, logical_key: Alt, text: None, location: Left, state: Pressed, repeat: false, platform_specific: KeyEventExtra { text_with_all_modifiers: None, key_without_modifiers: Super } }, is_synthetic: false } })
[...]

PR #1359, opened 2022-06-18
Push Event `ModifiersChanged` to `queued_input_events`

quadruple-output:
The input event `ModifiersChanged` has to be processed in the correct
order together with `KeyEvent` and `ImeInput`. Otherwise, glitches can
occur, resulting in modifiers being ignored.

Fixes neovide/neovide#1358

No

### DIFF
--- a/src/window/keyboard_manager.rs
+++ b/src/window/keyboard_manager.rs
@@ -6,13 +6,15 @@ use crate::{
 };
 use glutin::{
     event::{ElementState, Event, KeyEvent, WindowEvent},
-    keyboard::{Key, Key::Dead},
+    keyboard::{Key, Key::Dead, ModifiersState},
     platform::modifier_supplement::KeyEventExtModifierSupplement,
 };
 
+#[derive(Debug)]
 enum InputEvent {
     KeyEvent(KeyEvent),
     ImeInput(String),
+    ModifiersChanged(ModifiersState),
 }
 pub struct KeyboardManager {
     shift: bool,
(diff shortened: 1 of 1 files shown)


MultisampledNight:
I can't test the specific case in the linked issue, as I don't have a programmable keyboard, but if it improves the situation for you, great.

Though this change seems to break input after re-focus on Wayland.

Also, please consider changing the branch name from `main` _in future_, this allows checking out only your PR branch through `git fetch origin pull/1359/head:whatever-branch-name` and not every single other branch that still exist (by adding an entire new remote).

quadruple-output:
Thanks a lot for your feedback, @MultisampledNight .

> Though this change seems to break input after re-focus on Wayland.

I have a linux laptop running wayland (though I rarely use it), and I would like to test this. I just need to be clear on how to reproduce the broken input. (Playing around a bit, changing focus by keyboard and mouse, I could not recognize any issue. Running arch linux, all packages up-to-date.)
1) What does "break input" mean exactly?
2) Does it matter _how_ I achieve the re-focus? Should I enable focus-follows-mouse, or rather change focus with the keyboard?

Sorry for re-using the `main` branch. I will be happy to open a _new_ PR if that makes sense (in case I manage to fix the "broken input" issue).

MultisampledNight:
Sorry, I'll be more specific: I'm using Sway 1.7 stable. As long as the window focused the first time, input works perfectly. But once Neovide loses focus in any way, may it just be due to workspace switching, no input at all seems to get through to Neovim. Not even `:` works, but closing the window does.

Sway uses a focus-follow-mouse model, for me it's reproduceable every time. I don't know if that's reproduceable with other window managers.

The branch name is fine, no worries. Just for the next time.

last-partizan:
@quadruple-output try to test branch `fix-stuck-shift`, does that fixes issues for you?

quadruple-output:
> @quadruple-output try to test branch `fix-stuck-shift`, does that fixes issues for you?

Unfortunately not. I have to say it generally got better, meaning that the issue occurs much less frequently. I can type `}` ~100 times or more now, before the issue occurs once. I see this improvement also for the main branch, but I have no clue where the difference could come from. Could be due to improvements in neovide or just a MacOS update.

fredizzimo:
This is no longer relevant with #1899. No events are queued anymore.

PR #1363, opened 2022-06-21
fix(bridge): switch to copypasta clipboard provider

barklan:
Switch to [alacritty/copypasta](https://github.com/alacritty/copypasta). This adds Wayland support without XWayland requirement. Also seems to have less dependencies than abroad.

Tested manually using X11 backend as clipboard has no tests - so I might be missing something.

Fix

No

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -26,7 +26,7 @@ version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
 dependencies = [
- "getrandom 0.2.6",
+ "getrandom 0.2.7",
  "once_cell",
  "version_check",
 ]
@@ -66,25 +66,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
 dependencies = [
  "winapi",
-]
-
-[[package]]
-name = "arboard"
-version = "2.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dc120354d1b5ec6d7aaf4876b602def75595937b5e15d356eb554ab5177e08bb"
-dependencies = [
- "clipboard-win",
- "core-graphics 0.22.3",
- "image 0.23.14",
- "log",
- "objc",
- "objc-foundation",
- "objc_id",
- "parking_lot 0.12.1",
- "thiserror",
- "winapi",
- "x11rb",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


MultisampledNight:
Sounds great, and from my testing it works great, too! I'll leave this a bit open in case any objections come in.

MultisampledNight:
And there it goes, thank you!

MultisampledNight:
Oh well, could you test with the commit before this PR? I suppose you found this PR by git bisecting, but worth asking anyways. I'd revert it then.

khjorth:
Thanks @MultisampledNight. I found the commit in the history and played around with the code to investigate the issue with setting the clipboard, finding that I couldn't even set the clipboard to a literal string.

I'll try to revert the commit and see whether that fixes the issue. I'll also see if I can find a fix, such that we can retain proper Wayland support.

khjorth:
It is indeed the present change that have broken the clipboard for me.

I'm not so sure that we should revert it though. Copypasta uses [x11-clipboard](https://github.com/quininer/x11-clipboard) under the hood. I've tried different versions of this library, and copy is broken for me in every release since and including 0.6.0. This version span includes an implementation based on xcb 1.1 and a reimplementation using x11rb. I find it strange that both of these should be broken with no one reporting the issue, so I suspect something in my setup doesn't match the prerequisites for x11-clipboard: maybe one of the dependencies is missing or is too old. I can't figure out which, though.

I'm on xserver-xorg-core version 2:21.1.4-1 on a proprietary Debian derivative. Other packages:

* xcb 2.4-8
* xcb-proto 1.15.2-1

khjorth:
The examples and tests from both *copypasta* and *x11-clipboard* work for me. With a clipboard manager active, I also get correct behavior in Neovide if I add a clipboard load immediately after setting the clipboard:

```rust
clipboard_ctx.set_contents(lines)?;

clipboard_ctx.get_contents()?;
Ok(())
```

I haven't tried this without a clipboard manager active, and I suspect it might still fail in that case.

My hunch is that the root cause has something to do with the fact the code in question is called by an async function (`handle_notify()` in bridge/handler.rs) that returns immediately afterwards. I have this hunch because similar behavior would be expected under X11 from a program that set the clipboard and then exited immediately afterwards. I don't actually know Rust, let alone its async functionality, so I could use some help deciding whether this hypothesis has any merit.

MultisampledNight:
It's quite possible you're on to something there, though the function being async probably doesn't have much of an effect here. The process still persists even after the `async` is done, though the thread might not (tokio, Neovide's async runtime, is allowed to manage threads at will). Maybe copypasta's X11 code has some logic to commit only on `get_contents` or the like.

khjorth:
I'm working on a PR that runs the set clipboard command through the event aggregator. This seems to solve the problem.

khjorth:
Here's [the draft](1537).

PR #1366, opened 2022-06-22
Alternative Alt + character handling for MacOS

### LINKED ISSUE #1046: Alt send Esc in MacOs
Keymaps with the Alt key in MacOs doesn't work. What other apps that handle terminal does is configure it to send esc. [Alacritty](https://github.com/alacritty/alacritty/blob/1df7dc5171abfe1eab3e95be964f61c5876198f1/alacritty/src/input.rs#L809) and iTerm does that and it works fine with neovim.

ioreshnikov:
By default on MacOS pressing a key combination containing an Alt key produces a special character:

    <Alt+s> -> ß
    <Alt+v> -> √
    ...

This is really unfortunate if you want to use those key combinations as shortcuts and usually this case has to be separately handled by the app.

This commit introduces a new configuration flag `neovide_macos_alt_is_meta`. When the flag is set to true then all the
<Alt+key> combinations are treated as shortcuts with no extra character modifications, i.e. pressing `Alt` and `s` indeed emits a `<M-s>` shortcut.

Fixes #1046 

No, as far as I can tell

### DIFF
--- a/src/window/keyboard_manager.rs
+++ b/src/window/keyboard_manager.rs
@@ -147,7 +147,7 @@ impl KeyboardManager {
             if let Some(original_key_text) = key_text {
                 let mut key_text = original_key_text;
                 if self.alt {
-                    if let Some(modify) = key_event.text_with_all_modifiers() {
+                    if let Some(modify) = key_event_text(key_event) {
                         key_text = modify;
                     }
                 }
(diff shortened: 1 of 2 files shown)


makiri1993:
Hello guys! Any timeline when this fix will be rolled out? 😁

MultisampledNight:
As soon as someone tests this out and reports that everything's working fine/there are potential issues.

augustocdias:
Remember to add the docs about the new option

loqusion:
It's not working for me. Instead of executing keybinds, it just inserts the characters e.g. ø. I rebuilt from `main` and added `let g:neovide_macos_alt_is_meta = v:true` to my `ginit.vim`.

tpict:
It seems this option doesn't work correctly in Neovim terminal buffers, e.g. I have my shell configured to open the current prompt in `$EDITOR` when `<M-e>` is pressed. But under Neovide, it seemingly behaves as a dead key, and then the next key pressed inserts `´`.

ychin:
I don't think it's related to terminal buffers. The meta key handling in Neovide just seems to pass dead keys along instead of intercepting it when `g:neovide_input_macos_alt_is_meta=1`, so mapping `<M-e>` (dead key for ´) and `<M-i>` (dead key for ˆ) both don't seem to work for me. I think Neovide may need to detect that situation and have special handling for them.

FWIW I'm working on an overhaul for MacVim's key input system for the next release, and I think a good extension is to provide more controls. If you let the user specify the meta key behavior for left vs right Option key, this allows people to use say the left Option key as Vim meta key for shortcuts, and right Option key as normal (which kinds of act like AltGr in other platforms as it allows you to enter special characters). VimR does that, along with other terminals like iTerm. As for dead keys like Option-E, I'm debating making an option to configure the behavior, but just for simplicity you could just treat it as <M-e> when `g: neovide_input_macos_alt_is_meta` is set.

PR #1371, opened 2022-06-27
fixing some text

DRuppFv:
- Fix

- No

I was reading it without pretension, but the fact that some texts ends with ".", and others doesn't, and that some texts begans with a capital letter, and others doesn't, made me upset, haha.


MultisampledNight:
Thank you!

Issue #1375, opened 2022-06-28
Tests failing on Windows


MultisampledNight:
The CI tests on Windows are failing. It's likely not to be a code change, since previous commits [which tested just fine before](https://github.com/neovide/neovide/actions/runs/2491442429/attempts/1) are [suddently failing](https://github.com/neovide/neovide/actions/runs/2491442429/attempts/2). The CI logs [are pretty silent](https://github.com/neovide/neovide/runs/7063205182?check_suite_focus=true#step:8:395), probably because [every output on stdout is being eaten by that unpublished `results.json`](https://github.com/neovide/neovide/blob/main/.github/workflows/build.yml#L43).

Since I don't have a Windows machine and skia-safe doesn't like being cross-compiled, could anyone look into that? I believe it's "just" a toolchain mistake, but it could be anywhere, from cargo2junit over nightly Rust to a mistake on our side.

MultisampledNight:
@scottwillmoore, you mentioned you'd be happy to help with the build issues. Could you run `cargo test` in this repo on your machine and see what it yields?

dineshKumar777:
@MultisampledNight  i am not able to reproduce this issue in my local. all tests are passing in my windows machine.

dineshKumar777:
I tried with both stable and nightly
rustup default nightly-x86_64-pc-windows-msvc
rustup default stable-x86_64-pc-windows-msvc

![image](https://user-images.githubusercontent.com/21097695/176462370-efe03916-779d-4708-bb5a-54ccc0a37c1f.png)

scottwillmoore:
I can confirm that I was able to successfully run `cargo test -- -Z unstable-options --format json > results.json` on both stable and nightly on my machine for the latest commit and `7a3bf52` (the failed build).

**I also just noticed that @dineshKumar777 found the issue in #1377. Does this mean this issue can be closed?

MultisampledNight:
Ah yeah, sorry. Forgot about still being open. Thanks anyways!

PR #1377, opened 2022-06-30
GA cargo test fail fix for windows

dineshKumar777:
- Fix
Tests failing on Windows #1375

- No

this occurred due to choco installing nvim in different location and not adding env path correctly.
Because of that, this test (test_read_initial_values ) failed.
since we are using std::process:exit(1), program exits in middle when it's not able to find nvim exec. 
Thats why when this particular test failed no useful error msg or info shown. maybe future refactoring!!

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,20 +20,20 @@ jobs:
       - name: Install Utilities
         run: |
           cargo install cargo2junit
-
       - uses: Swatinem/rust-cache@v1
 
       - name: Check Formatting
         run: |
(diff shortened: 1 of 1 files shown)


MultisampledNight (approved):
Love you. Thanks! Then I think nothing's anymore in the way of a release.

MultisampledNight:
...nice, now the macOS CI fails.

dineshKumar777:
yes noticed. but none of these changes are related to mac workflow.. dont know why it fails now. @MultisampledNight

PR #1380, opened 2022-07-01
Fixup Arch Linux instructions

alerque:
Arch Linux has official packages now, these new instructions reflect that.

Also there was duplicated/conflicting instructions for building from the AUR. I really think it's better to scrap all that and let the Arch Wiki help people build packages if they don't know how to already, upstream project README's aren't a good home for detailed instructions on how to use a distro.


MultisampledNight:
Thank you! Seeing that you are the maintainer for the Arch package, you may want to change the source URL to actually point here (the neovide organization).

alerque:
Thanks. I started with the earlier AUR build and didn't notice the URL had changed since redirects were working. I just updated it.

PR #1385, opened 2022-07-04
Use grid_scroll event for smooth scrolling

### LINKED ISSUE #1334: Smooth scrolling is jumpy sometimes
**Describe the bug**
When scrolling with smooth scrolling enabled, the animation is jumpy for some lines.

**To Reproduce**
Open a long file and scroll the view port up or down slowly with "j/k".

**Expected behavior**
Smooth scrolling should be smooth all the time.

**Screenshots**

https://user-images.githubusercontent.com/552045/170488444-f168a5ce-c06e-4a39-a16d-398883c384d4.mp4

 - OS: Debian Testing
 - Neovide 0.9.0
 - Neovim v0.8.0-dev+52-g89df042a8

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/8779024/neovide_rCURRENT.log)

**Additional context**
Not sure if it is a duplicate #794. But the issue appears also while scrolling up here. See the attached file.

MDeiml:
This removes usage of the `win_viewport` event, which was previously used for smooth scrolling. It replaces its usage with the `grid_scroll` event which is better fit to trigger smooth scrolling behavior, as
1. `win_viewport` contains positional information in terms of buffer lines, but these do not always relate 1:1 to drawn lines e.g. when using folds
2. `win_viewport` is triggered in cases where there is no actual scrolling happening e.g. when jumping around in files while doing substitution. This sometimes causes jumpy behavior.

- Fixes #1334.

- No

### DIFF
--- a/src/bridge/events.rs
+++ b/src/bridge/events.rs
@@ -223,14 +223,6 @@ pub enum RedrawEvent {
         scrolled: bool,
         separator_character: String,
     },
-    WindowViewport {
-        grid: u64,
-        top_line: f64,
-        bottom_line: f64,
-        current_line: f64,
-        current_column: f64,
-        line_count: Option<f64>,
-    },
     CommandLineShow {
         content: StyledContent,
         position: u64,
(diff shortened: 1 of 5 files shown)


0x00002a:
Can confirm this fixes the issues I was encountering in #1334, thanks!

MDeiml:
Yes it seems I have to work on this some more.

The first problem arises, because there could be scroll events that do not scroll the entire grid, just portions of the grid. (Which is the case with multigrid disabled). The animation should only be triggered if a scroll event concerns the entire grid.

The second problem came up because it seems I messed up the `snapshots` attribute of `RenderedWindow`.

MDeiml:
I think these problems would be easy to solve, but I'm not sure anymore if this is the right approach. As someone pointed out there actually is a (blog post)[http://02credits.com/blog/day96-neovide-smooth-scrolling/#scroll-event] on why neovide doesn't use `grid_scroll`. See the discussion in the linked issue.

MDeiml:
I'm gonna close this as I don't believe anymore that this approach is a solution.

PR #1386, opened 2022-07-05
ci: remove unncessary dirs in linux artifact

kskarthik:
## Problem
The linux archive ships with un-necessary dir's, 
```
kar@earth:~$ tar -tf Downloads/neovide.tar.gz
./target/release/neovide
```
## Solution
This patch removes them & archives just the binary

- Fix

- No

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -198,7 +198,8 @@ jobs:
 
       - name: Archive Linux artifact
         run: |
-          tar czvf ./target/release/neovide.tar.gz ./target/release/neovide
+          cd ./target/release/
+          tar czvf neovide.tar.gz neovide
 
       - uses: actions/upload-artifact@v1
         with:


kskarthik:
@Kethku please review & merge this patch :)

MultisampledNight:
Looks good to me, thank you! I'm wondering whether the `cd` has any forward effect outside of the step though.

MultisampledNight:
Okay, it's not, so your full path specification works just fine. Thanks again!

kskarthik:
@MultisampledNight happy to know that worked! :tada:

kskarthik:
@MultisampledNight currenly, only linux x86_64 binaries are built? Then it would be nice to rename archive with `neovide-linux-$VERSION-$ARCH.tar.gz`

MultisampledNight:
Hm, 32-bit computers have barely any presence anymore. The Windows archive doesn't really need the architecture either. Is there any specific reason you need the architecture in the filename? Or is it for pure convenience?

kskarthik:
@MultisampledNight just for convenience sake & a visual clue

PR #1413, opened 2022-07-21
Add note about python2

zdog234:
idk if this is actually worth merging, but I had to figure out how to install python2 to get rust-skia to build (macos >= 12.3 doesn't include python2 any more)

docs

- No

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,13 +171,14 @@ brew install --cask neovide
 1. Install the latest version of Rust. Using homebrew: `brew install rustup-init`
 2. Configure rust by running `rustup-init`
 3. Install CMake. Using homebrew: `brew install cmake`
-4. `git clone https://github.com/neovide/neovide`
-5. `cd neovide`
-6. `cargo build --release`
-7. Copy `./target/release/neovide` to a known location and enjoy.
-8. `cargo install cargo-bundle`
-9. `cargo bundle --release`
-10. Copy `./target/release/bundle/osx/neovide.app` to `~/Applications` and enjoy.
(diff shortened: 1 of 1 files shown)


MultisampledNight:
First, thank you! Then, that's... good to know I guess, though I on Linux don't have python2 installed? I guess that's a macOS specific thing then.

MultisampledNight:
~~The only mention of Python 2 in rust-skia is [here and surrounding](https://github.com/rust-skia/rust-skia/blob/master/skia-bindings/build_support/skia/config.rs), though that seems to be used only for checking if that's a Python version to _skip_. This seems- oh, wait~~

`skia-safe` seems to have switched to Python 3 in https://github.com/rust-skia/rust-skia/pull/595. This PR was merged on May 23, while the latest release was on May 22, how unfortunate. I think we could ask for a new minor release, given that they're already on `0.52.0` on their master branch, while `0.51.0` would be "enough", and I'll do that now.

MultisampledNight:
I just updated `rust-skia` to `0.52.0`, which shouldn't require Python 2 anymore (but uses Python 3 instead). Can you confirm that on your end, @zdog234?

MultisampledNight:
I assume that's a "yes", then. As Python 2 is not required anymore this should be obsolete, but thank you for noticing and explaining the circumstances!

zdog234:
oh shoot, sorry. 

Yep, this did fix things, although I had to figure out a workaround for a **separate** `rust-skia` -related issue that I posted in a [discussion thread](https://github.com/neovide/neovide/discussions/1370#discussioncomment-3530498)

PR #1415, opened 2022-07-22
macos: open filepath which contains spaces correctly

### LINKED ISSUE #1399: Cannot open a file when filepath contains space on Mac
**Describe the bug**
On Mac, execute such command `neovide "/FILE/PATH WITH SPACE/filename"`, only open an empty file.

**To Reproduce**
1. Open a terminal
2. execute `neovide "/FILE/PATH WITH SPACE/filename"`
3. See neovide open an empty buffer

**Expected behavior**
Neovide should open  the selected file

**Screenshots**

 - OS: macOS 12.4
 - Neovide Version: 0.8.0
 - Neovim Version: 0.7.0

The log file is too long to post here.

**Additional context**
nvim can open such "/FILE/PATH WITH SPACE/filename" file

dineshKumar777:
- Fix
#1399

- No

Currently tested against mac and windows os.

for some reason enclosing singlequotes around filepath doesnt worked as expected as mentioned by aJuvan. so went with escaping space with backslash.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1367,6 +1367,7 @@ dependencies = [
  "rmpv",
  "serde",
  "serde_json",
+ "shlex",
  "skia-safe",
  "swash",
  "tokio",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ rand = "0.8.5"
 rmpv = "1.0.0"
 serde = { version = "1.0.136", features = ["derive"] }
 serde_json = "1.0.79"
+shlex = "1.1.0"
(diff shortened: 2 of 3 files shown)


MultisampledNight:
Thank you! If we're trying to handle shell stuff like this at all, we can just directly use `shlex` which we already indirectly depend on, I'll add this to this PR. This also brings the nice benefit of allowing other special chars such as `"`.

MultisampledNight:
(the build & deploy of the website failed, I guess due to the non-inherited GitHub token, but that's irrelevant anyways as no website changes were made)

PR #1440, opened 2022-07-31
Enable confirm quit by default

dineshKumar777:
- Refactor

- No

This is the default behaviour in nvim-qt. So enabled it by default.
#1247 - original PR with flag.

### DIFF
--- a/website/docs/configuration.md
+++ b/website/docs/configuration.md
@@ -95,6 +95,19 @@ Setting `g:neovide_fullscreen` to a boolean value will set whether the app shoul
 screen. This uses the so called "windowed fullscreen" mode that is sometimes used in games which
 want quick window switching.
 
+
+#### Confirm Quit
+
+```vim
+let g:neovide_confirm_quit=v:false
+" or
+let g:neovide_confirm_quit=0
+```
+
+When buffer having unsaved changes, setting `g:neovide_confirm_quit` to a boolean value will confirm quit
+before closing neovide window. Enabled by default.
+
+
 #### Remember Previous Window Size
 
 ```vim
--- a/website/docs/configuration.md
+++ b/website/docs/configuration.md
@@ -95,7 +95,6 @@ Setting `g:neovide_fullscreen` to a boolean value will set whether the app shoul
 screen. This uses the so called "windowed fullscreen" mode that is sometimes used in games which
 want quick window switching.
 
-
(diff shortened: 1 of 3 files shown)


MultisampledNight:
Thank you! While I understand the motivation, I think it's a bit overkill to disallow non-confirmed exit. Even browsers and large IDEs allow ignoring the confirm popup if they really want to.

So I think it'd be preferable to just change the default of the variable to 1 instead of `0`.

dineshKumar777:
@MultisampledNight 
Most IDE and Editor(vscode, visual studio, jetbrains IDE) which i used doesn't support non-confirmed exit. can you give any other IDE which supports this!!
Sublime is the only editor which i know supports non-confirmed exit but that will preserve my changes on next launch.

but here we are losing data when we accidentally close the window. also this is the default behaviour with nvim-qt too. All i am looking for is better defaults.

MultisampledNight:
PyCharm community allows doing so, under `Appearance & Behavior > System Settings` and then `Confirm before exiting the IDE`. I didn't mean to disable the behavior at all, just to keep it configurable in case a user (such as me) doesn't want that, I probably misworded my first comment. But at the moment the PR is enforcing the behavior, ignoring the value of `g:neovide_confirm_quit`.

dineshKumar777:
Now its configurable.
Confirm_quit is enabled by default for any unsaved changes.
to disable, we can disable by :let g:neovide_confirm_quit=0 or v:false

website/docs/configuration.md
```diff
@@ -77,6 +77,19 @@ Setting `g:neovide_fullscreen` to a boolean value will set whether the app shoul
...
+#### Confirm Quit
+
+```vim
+let g:neovide_confirm_quit=v:false
+" or
+let g:neovide_confirm_quit=0
+```
+
+When buffer having unsaved changes, setting `g:neovide_confirm_quit` to a boolean value will confirm quit
+before closing neovide window. Enabled by default.
```
LoipesMas, line 90:
```suggestion
If set to `true`, quitting while having unsaved changes will require confirmation.
Enabled by default.
```

I think this wording is clearer.

dineshKumar777: Done

MultisampledNight:
(rebased to resolve the merge conflict)

PR #1441, opened 2022-08-01
fix(cargo.toml): rm debug symbols in release build

kskarthik:
Helps further slimming of the binary. 

Info:
1. https://doc.rust-lang.org/cargo/reference/profiles.html#strip
2. https://github.com/johnthagen/min-sized-rust#optimize-for-size

- Feature

- No

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,6 +69,7 @@ version = "0.52.0"
 lto = true
 incremental = true
 strip = true
+opt-level = "z"  # Optimize for size.
 
 [package.metadata.bundle]
 name = "Neovide"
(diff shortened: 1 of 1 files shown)


last-partizan (approved):
Looks good to me

LoipesMas:
I think we'd rather optimize for speed, not size. 
And from my testing it looks like the difference is negligible anyway:
image
That's current profile, with `strip = true` added and with `strip = true` and `opt-level = "z"` added, respectively.
It's a 4mb difference at most, so I'm not sure if potential performance loss is worth it.
Maybe if we did a benchmark and there were no performance loss.

kskarthik:
@LoipesMas so `strip=true` should be good to go, if performace is the choice

LoipesMas:
Yeah, just stripping should be fine.

Issue #1310, opened 2022-05-14
Neovide constantly using cpu at idle


0x0013:
**Describe the bug**
I'm seeing constant CPU usage of ~3% when neovide is running. This happens both with my normal config, as well as with `neovide -- --clean`, and persists with multigrid both enabled and disabled. Both when neovide is in focus or not in focus, or on a different (not visible) window manager workspace.

**To Reproduce**
1. Open neovide.
2. Observe cpu usage while neovide is idle.

**Expected behavior**
Neovide keeps idle while not actively used.

**Desktop (please complete the following information):**
 - OS: Arch Linux with Sway (also observed on different computer with i3)
 - Neovide Version 0.8.0 (AUR neovide) and 0.9.0 (AUR neovide-git)
 - Neovim Version 0.7.0

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

Log keeps growing with the following lines endlessly repeating:
```
TRACE [neovide::window] Settings geometry PhysicalSize { width: 69, height: 45 }
TRACE [neovide::window] Inner size: PhysicalSize { width: 3212, height: 1020 }
TRACE [neovide::window] Inner size scaled: PhysicalSize { width: 1606, height: 510 }
TRACE [neovide::window] Settings geometry PhysicalSize { width: 69, height: 45 }
TRACE [neovide::window] Inner size: PhysicalSize { width: 3212, height: 1020 }
TRACE [neovide::window] Inner size scaled: PhysicalSize { width: 1606, height: 510 }
TRACE [neovide::window] Settings geometry PhysicalSize { width: 69, height: 45 }
TRACE [neovide::window] Inner size: PhysicalSize { width: 3212, height: 1020 }
TRACE [neovide::window] Inner size scaled: PhysicalSize { width: 1606, height: 510 }
```
Full log attached. 
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/8692349/neovide_rCURRENT.log)

**Additional context**

Similar issue #440 but that seems to be a different issue, as that was reported only happening with multigrid enabled, while my issue persists with both options.

This does not impact performance for me, but the constant usage can impact battery life. While the idle usage is not high, I think it is too high for idle usage, and is probably not expected behavior, seeing as the usage is higher than a web browser at idle.

CPU usage increases when I am actively using Neovide (which is expected). I can also see GPU usage spike while I am using Neovide, so GPU is working.

[...]

PR #1454, opened 2022-08-04
Detect when the window is not in focus and lower the frame rate

DerFetzer:
- Feature

- No

This is to lower the load when the window is not in use.
See #1310.

### DIFF
--- a/website/docs/configuration.md
+++ b/website/docs/configuration.md
@@ -27,6 +27,17 @@ let g:neovide_refresh_rate=140
 Setting `g:neovide_refresh_rate` to a positive integer will set the refresh rate of the app. This is
 limited by the refresh rate of your physical hardware, but can be lowered to increase battery life.
 
+#### Idle Refresh Rate
+
+```vim
+let g:neovide_refresh_rate_idle=10
+```
+
+Setting `g:neovide_refresh_rate_idle` to a positive integer will set the refresh rate of the app when
+it is not in focus.
+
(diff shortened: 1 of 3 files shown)


website/docs/configuration.md
```diff
@@ -27,6 +27,17 @@ let g:neovide_refresh_rate=140
 Setting `g:neovide_refresh_rate` to a positive integer will set the refresh rate of the app. This is
 limited by the refresh rate of your physical hardware, but can be lowered to increase battery life.
 
+#### Idle Refresh Rate
+
+```vim
+let g:neovide_refresh_rate_idle=10
```
MultisampledNight, line 33: The configuration docs document the default values, but here it's listed at `10` while the `Default` implementation of `WindowSettings` lists `5`.

MultisampledNight (changes requested):
Sorry for the long time with no reaction, but thank you! One small inconsistency though, then we can roll this up.

website/docs/configuration.md
DerFetzer: Thank you for having a look!
I thought a non default value was a deliberate choice there since the default value for `refresh_rate` is 60 and not 140 as in the lines above.
I'll fix both.

MultisampledNight: Oh, sorry from our side then!

DerFetzer: No reason to apologize for!

Issue #1263, opened 2022-04-07
I can't use the cmd+v to paste text on my mac; 


CG-man:
**Describe the bug**
I can't use the cmd+v to paste text on my mac; 

**To Reproduce**
1. use cmd+c to copy something
2. switch to neovide
3. cmd+v to paste the text 

**Expected behavior**
paste the text to buffer/other

**Screenshots**

**Desktop (please complete the following information):**
 - OS: 12.3
 - Neovide Version 0.8.0
 - Neovim Version 0.7.0

**Please run `neovide --log` and paste the contents of the `.log` file here:**

**Additional context**

MultisampledNight:
Neovide just forwards all key commands to NeoVim, which indeed does not have <kbd>Cmd</kbd> + <kbd>V</kbd> (or `<M-v>` in Vim) bound to anything. You can use the `clipboard` option and set it to `unnamedplus` if you want to use <kbd>p</kbd> and <kbd>y</kbd> as replacements for it. Or rebind `<M-v>` to `"+p` and `<M-c>` to `"+y`.

Either way, closing as that's not an issue with Neovide itself.

terhechte:
@CG-man 
Actually, using `<D-v>` or `<D-c>` does nothing on macOS because [this PR](https://github.com/neovide/neovide/pull/857) disables the command key. Took me way too long to figure this out as it isn't really documented.

To enable Cmd + V / Cmd + P on macOS, I have these bindings:

```
" Allow copy paste in neovim
let g:neovide_input_use_logo = 1
map <D-v> "+p<CR>
map! <D-v> <C-R>+
tmap <D-v> <C-R>+
vmap <D-c> "+y<CR> 
```

terhechte:
@MultisampledNight So it is a bit of a Neovim issue because Neovim introduces a non-standard setting that disables the command key by default. Personally, I think this setting should be default-enabled on macOS. Would be willing to crate a PR for that.

MultisampledNight:
@terhechte (you probably meant this is a Neovide issue) That's valid, feel free to try it out! I think it should be as simple as manually implementing the `Default` trait for `KeyboardSettings`.

amitkot:
Based on @terhechte , I used this in `init.lua`:

```lua
-- Allow clipboard copy paste in neovim
vim.g.neovide_input_use_logo = 1
vim.api.nvim_set_keymap('', '<D-v>', '+p<CR>', { noremap = true, silent = true})
[...]

PR #1455, opened 2022-08-05
Changed default for use_logo to true for MacOS

jonasdiemer:
Changes the default for use_logo for MacOS to avoid confusion with new users. 

See https://github.com/neovide/neovide/issues/1263


MultisampledNight:
Thank you! Took the liberty of making the check simpler.

PR #1458, opened 2022-08-05
fix: Change how window resize detection works

### LINKED ISSUE #1307: Neovide closes immediately after starting in arch linux
**Describe the bug**
I installed neovide-git from AUR and on opening, it appears for a second and then closes without any warnings or errors 

**To Reproduce**
1. Install neovide-git from AUR
2. launch neovide

**Expected behavior**
Neovide opens without any issue

**Screenshots**
output

 - OS: Arch Linux 5.17.5
 - Neovide Version 0.9.0
 - Neovim Version 0.7.0

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/8649344/neovide_rCURRENT.log)

**Additional context**
**neovide** package from AUR runs as expected

### LINKED ISSUE #1384: Neovide crashes instantly
Recently I updated Neovide from version 0.8.0 to 0.9.0, but when I launched it, it instantly crashed. I tried to use --nofork argument to see what is the issue and I got this panic error:

```
thread 'main' panicked at 'attempt to divide by zero', src/window/mod.rs:207:20
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I managed to fix it by building it from source and replacing scale_factor as 1 at line 207 in src/window/mod.rs

### LINKED ISSUE #1450: `--geometry` flag has no effect on Linux
**Describe the bug**
Passing `--geometry=MxN` to neovide doesn't change the window size on startup.

**To Reproduce**
Run `neovide --geometry=MxN --log`.

**Expected behavior**
Window size should be modified appropriately by `--geometry`.

**Screenshots**
image

 - OS: Arch Linux
 - Neovide Version 0.9.0
 - Neovim Version 0.7.2

Unlike #1384 and #1307 (which look potentially related), my scale factor is 2.0, which becomes 2 in the current code.

I tried to paste the contents, but GitHub got mad about it being more than 64k characters, so have an attachment instead.

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/9255549/neovide_rCURRENT.log)

**Additional context**
I have `let g:neovide_remember_window_size = v:true` in my `init.vim`, but this also seems to have no effect: the window will always be this size when opening.

I'm pretty sure this is being caused by some combinations of things unique to my system. I'm willing to do some local debugging if needed. I've already bisected this to b9739a9c3e2711115c59118eaf47772b75e81258 (the commit associated with #1267), but I don't understand how this could possibly cause the behavior I'm seeing right now. If anybody has ideas about what more to look into to figure out the root cause of this bug, I'd be happy to spend some time digging this weekend.

chipbuster:
- Fix

- No

Changes how the detection of window resizing works. Previously relied
on hardcoded 800 x 600 value, which was not always correct on MacOS, or
rescaling by the scaling factor, which was not always correct on Linux.

The physical size of the window is now stored right after the window is
initialized, and this value is used whenever we check for window
resizing.

I have tested this with `WINIT_X11_SCALING_FACTOR` set to 0.8, 1.0, 
1.5, 2.0, and 2.1, and observed seemingly-correct behavior on Linux,
(appropriate response to `--geometry` flag, no panics), so I believe
that this closes a number of scaling bugs. I also have a Macbook on
which I was able to reproduce #1216, and this PR does not cause a
regression of behavior on that system.

Closes #1450
Closes #1384 
Closes #1307 
Does not regress #1216 

One potential perf issue is the presence of `self.windowed_context.window().inner_size()`
in `draw_frame()`. I'm hoping that the optimizer will be able to inline
and eliminate this call, but if we're worried about it, I can change
the implementation.

### DIFF
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -65,6 +65,7 @@ pub struct GlutinWindowWrapper {
     fullscreen: bool,
     saved_inner_size: PhysicalSize<u32>,
     saved_grid_size: Option<Dimensions>,
+    size_at_startup: PhysicalSize<u32>,
     window_command_receiver: UnboundedReceiver<WindowCommand>,
 }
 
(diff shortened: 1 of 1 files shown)


MultisampledNight:
Awesome, thank you!

Issue #1298, opened 2022-05-01
Naturally/transparent colored title bar


shaunsingh:
**Is your feature request related to a problem? Please describe.**
Quite a few other GUI editors have a feature called "natural title bar" in which the title bar of the application matches the background of the user's current theme. 

For example, this is how neovide currently looks: 
<img width="715" alt="image" src="https://user-images.githubusercontent.com/71196912/166150619-ef5b8490-a7e2-49b0-9ceb-45b92f85b58d.png">

An example in alacrity: 
![](https://user-images.githubusercontent.com/845857/39506095-726794b0-4d8b-11e8-9014-5d8693abe84c.png)

Or iterm: 
![](https://user-images.githubusercontent.com/4488655/28492796-2c6a02e6-6f0a-11e7-917e-aec4cd3fc4a3.png)

**Describe the solution you'd like**
An option or flag to change the titlebar behavior would be nice

**Describe alternatives you've considered**
N/A

**Additional context**
I'm a tad bit busy this weekend, but I'll work on an implementation this week unless someone wants to pick it up.

TimothyKrell:
I want this so badly, but don't know enough rust to make it happen.

MultisampledNight:
Tbh, it's not impossible to learn Rust. I can't estimate the size of this task though, and I can't test/develop on macOS either. With [the Book](https://doc.rust-lang.org/stable/book/) read and a bit of personal experience made, you can tackle pretty much everything. And if you don't know where to look or are stuck somewhere, just ask.

In this case, it's winit's job for providing us a window to draw on. You could maybe use `raw-window-handle` and dabble around with Cocoa to paint the window title background, the color would be then extracted out of `Normal`'s `ctermbg`/`guibg`. `Normal` is a... normal highlight group.

nickychow:
Based on the code, this is already implemented. You can find here https://github.com/neovide/neovide/blob/main/src/window/mod.rs#L313-L327

You should be able to use it by the cli option `neovide --frame transparent`, but I haven't figured out how to change the padding.

shaunsingh:
I didn't even realize. Thanks!

hhofner:
@nickychow wow thank you! should small things like this be documented somewhere ? like somewhere in the wiki perhaps?

TimothyKrell:
Is there some way to add padding to the top to keep the window buttons to keep from overlapping the content? 
[...]

PR #1459, opened 2022-08-05
Support background color on MacOS

Matts966:
- Feature
- Documentation

- No

Support colorized title bar by `g:neovide_background_color` for macOS, discussed in #1298 

BackgroundColor


Matts966:
https://user-images.githubusercontent.com/28551465/183221643-2a74287b-9c1f-4a54-bd3b-8233c9452daf.mov

src/window/mod.rs
```diff
@@ -18,6 +18,7 @@ use tokio::sync::mpsc::UnboundedReceiver;
 
 #[cfg(target_os = "macos")]
 use glutin::platform::macos::WindowBuilderExtMacOS;
+use crate::settings::draw_background;
```
MultisampledNight, line 21: If you intended to have this import only active if the target OS is `macos`, note that `#[...]` attributes only apply to the next item.

website/docs/configuration.md
```diff
@@ -38,6 +38,25 @@ let g:neovide_transparency=0.8
...
+let g:neovide_background_color = '#0f1117'.printf('%x', float2nr(255 * g:neovide_transparency_point))
+
+" Add keybinds to change transparency
+function! ChangeTransparency(delta)
+  let g:neovide_transparency_point = g:neovide_transparency_point + a:delta
+  let g:neovide_background_color = '#0f1117'.printf('%x', float2nr(255 * g:neovide_transparency_point))
+endfunction
+noremap <expr><D-]> ChangeTransparency(0.01)
+noremap <expr><D-[> ChangeTransparency(-0.01)
+```
```
MultisampledNight, line 56: While it's sure nice to have a whole ready-to-use machine for changing the transparency, this page is more or less only intended to show the name of those config values, what they do and their defaults. The mechanism allowing to change this through a keybinding might have a better home in `website/docs/faq.md`- while we should probably call that thing more like a "tips & tricks" section instead.

src/settings/draw_background.rs
```diff
@@ -0,0 +1,31 @@
...
+use cocoa::appkit::{NSWindow, NSColor};
+use cocoa::base::{id, nil};
+use objc::{
+    rc::autoreleasepool,
+    runtime::{YES},
+};
+use csscolorparser::Color;
+
+#[cfg(target_os = "macos")]
+pub fn draw_background(window: &WindowedContext<PossiblyCurrent>) {
```
MultisampledNight, line 15: For compartization's sake, I'm not sure if this window logic belongs into `src/settings`. It feels more like it should go into `src/window` instead.

MultisampledNight (commented):
Thank you! Three small things I'd like to have addressed before driving this thing home.

src/window/mod.rs
Matts966: I used code below to import the function only on macOS.
Thank you!

```rust
#[cfg(target_os = "macos")]
mod draw_background;
```

website/docs/configuration.md
Matts966: I separated them. Thank you!

src/settings/draw_background.rs
Matts966: I moved it. Thank you !

Matts966:
@MultisampledNight 
Thank you for your nice review :)
I fixed your points and confirmed `cargo fmt --check` exit with code 0.

MultisampledNight:
Alright then, I think this should be ready to go. Thank you again!

One small thing for the future in case you contribute again somewhere (also applies not only to this repo): Consider making a branch on your fork for every PR you make, and don't directly commit to your fork's main. This avoids some tricky local git scenarios where technically both origin and your remote are checked out, but both branches are named `main`.

Matts966:
@MultisampledNight 
Thank you for your kind fix 🥹

zhaozg:
I got error on MacOS X86_64

```
[I] ✘101 ➜ PYTHON=python3 cargo build --release
   Compiling neovide v0.9.0 (/Users/zhaozg/work/extra/vide)
error[E0433]: failed to resolve: use of undeclared crate or module `cocoa`
 --> src/window/draw_background.rs:4:5
  |
4 | use cocoa::appkit::{NSColor, NSWindow};
  |     ^^^^^ use of undeclared crate or module `cocoa`

error[E0433]: failed to resolve: use of undeclared crate or module `cocoa`
 --> src/window/draw_background.rs:5:5
  |
5 | use cocoa::base::{id, nil};
  |     ^^^^^ use of undeclared crate or module `cocoa`

error[E0433]: failed to resolve: use of undeclared crate or module `objc`
 --> src/window/draw_background.rs:8:5
  |
8 | use objc::{rc::autoreleasepool, runtime::YES};
  |     ^^^^ use of undeclared crate or module `objc`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `neovide` due to 3 previous errors

```

MultisampledNight:
Oh welp, I'll try to take care of it.

MultisampledNight:
I think I got it. Could you try again?

zhaozg:
> I think I got it. Could you try again?

It's works, thank you.

PR #1469, opened 2022-08-08
Handle different underlining styles supported by neovim

### LINKED ISSUE #764: Enable customization of the undercurl
**Describe the bug**

Undercurl looks like a dotted line:

изображение

**To Reproduce**
1. `hi SpellBad gui=undercurl guisp=Red`
2. `hi SpellBad`

**Expected behavior**
изображение

**Current behavior**
изображение

 - OS: Linux
 - Neovide Version: 82b745c92bb6dab1402aef70b9c8e07a02979a41 (latest commit at the time of writing)
 - Neovim Version: bdf3df4027a5389a46b4c9a75b50aff5c0ecb351 (latest commit at the time of writing)

[neovide_rCURRENT.log](https://github.com/Kethku/neovide/files/6721384/neovide_rCURRENT.log)

calops:
- Feature
- Fixes #764 

- No

## Content
This overhauls the drawing of underlines and actually implements the
undercurl that was simulated with underdashes before.

Implemented in this PR:
* underline
underline

* underdouble
underdouble

* underdashed
underdashed

* underdotted
underdotted

* undercurl
undercurl

## Points to note
* Previously the underline thickness was proportional to the stroke width. This led to weird aliasing artifacts when drawing lines (especially dashes and dots) and isn't in line with how most other terminal emulators handle this. With this PR, all the lines are pixel-perfect and 1px thick. If desired, it'll be easy to add a configuration option to make it possible to configure the line thickness, but I don't recommend making it proportional to the text itself.
* The undercurl looks (in my opinion) nice, but I don't really know what I'm doing here. I'm using quadratic interpolation from skia by generating points under the text every time. I haven't noticed any hangups, but I'm sure there's a better optimized way to do this if it becomes an issue.
* I'm not sure if there should be an order of precedence between the different line styles (cf the TODO left in the code for now). I'm inclined to say we leave it like that, as I don't think multiple underline styles can happen at the same time in neovim (don't quote me on this though).
* Maybe the gap between the dots for `underdotted` should be widened? I'm worried it'll look like a thin solid line on high-dpi screens.

### DIFF
--- a/src/bridge/events.rs
+++ b/src/bridge/events.rs
@@ -4,7 +4,7 @@ use std::{
     fmt::{self, Debug},
 };
 
-use log::{debug};
+use log::debug;
 use rmpv::Value;
 use skia_safe::Color4f;
 
@@ -496,11 +496,21 @@ fn parse_style(style_map: Value) -> Result<Style> {
                 ("blend", Value::Integer(blend)) => style.blend = blend.as_u64().unwrap() as u8,
 
                 // TODO: handle correct precedence?
-                ("underline", Value::Boolean(true)) => style.underline = Some(UnderlineStyle::Underline),
-                ("undercurl", Value::Boolean(true)) => style.underline = Some(UnderlineStyle::UnderCurl),
-                ("underdotted", Value::Boolean(true)) => style.underline = Some(UnderlineStyle::UnderDot),
-                ("underdashed", Value::Boolean(true)) => style.underline = Some(UnderlineStyle::UnderDash),
-                ("underdouble", Value::Boolean(true)) => style.underline = Some(UnderlineStyle::UnderDouble),
+                ("underline", Value::Boolean(true)) => {
+                    style.underline = Some(UnderlineStyle::Underline)
+                }
+                ("undercurl", Value::Boolean(true)) => {
+                    style.underline = Some(UnderlineStyle::UnderCurl)
+                }
+                ("underdotted", Value::Boolean(true)) => {
+                    style.underline = Some(UnderlineStyle::UnderDot)
+                }
+                ("underdashed", Value::Boolean(true)) => {
+                    style.underline = Some(UnderlineStyle::UnderDash)
+                }
+                ("underdouble", Value::Boolean(true)) => {
+                    style.underline = Some(UnderlineStyle::UnderDouble)
+                }
 
                 _ => debug!("Ignored style attribute: {}", name),
             }
--- a/src/bridge/events.rs
+++ b/src/bridge/events.rs
@@ -502,13 +502,13 @@ fn parse_style(style_map: Value) -> Result<Style> {
                 ("undercurl", Value::Boolean(true)) => {
                     style.underline = Some(UnderlineStyle::UnderCurl)
                 }
-                ("underdotted", Value::Boolean(true)) => {
+                ("underdotted" | "underdot", Value::Boolean(true)) => {
                     style.underline = Some(UnderlineStyle::UnderDot)
                 }
-                ("underdashed", Value::Boolean(true)) => {
+                ("underdashed" | "underdash", Value::Boolean(true)) => {
                     style.underline = Some(UnderlineStyle::UnderDash)
                 }
-                ("underdouble", Value::Boolean(true)) => {
+                ("underdouble" | "underlineline", Value::Boolean(true)) => {
                     style.underline = Some(UnderlineStyle::UnderDouble)
                 }
 
--- a/src/bridge/events.rs
+++ b/src/bridge/events.rs
@@ -4,11 +4,11 @@ use std::{
     fmt::{self, Debug},
 };
 
-use log::debug;
+use log::{debug};
 use rmpv::Value;
 use skia_safe::Color4f;
 
-use crate::editor::{Colors, CursorMode, CursorShape, Style};
+use crate::editor::{Colors, CursorMode, CursorShape, Style, UnderlineStyle};
 
 #[derive(Clone, Debug)]
 pub enum ParseError {
@@ -493,9 +493,15 @@ fn parse_style(style_map: Value) -> Result<Style> {
                 ("strikethrough", Value::Boolean(strikethrough)) => {
                     style.strikethrough = strikethrough
                 }
-                ("underline", Value::Boolean(underline)) => style.underline = underline,
-                ("undercurl", Value::Boolean(undercurl)) => style.undercurl = undercurl,
                 ("blend", Value::Integer(blend)) => style.blend = blend.as_u64().unwrap() as u8,
+
+                // TODO: handle correct precedence?
+                ("underline", Value::Boolean(true)) => style.underline = Some(UnderlineStyle::Underline),
+                ("undercurl", Value::Boolean(true)) => style.underline = Some(UnderlineStyle::UnderCurl),
+                ("underdotted", Value::Boolean(true)) => style.underline = Some(UnderlineStyle::UnderDot),
+                ("underdashed", Value::Boolean(true)) => style.underline = Some(UnderlineStyle::UnderDash),
+                ("underdouble", Value::Boolean(true)) => style.underline = Some(UnderlineStyle::UnderDouble),
+
                 _ => debug!("Ignored style attribute: {}", name),
             }
         } else {
--- a/src/bridge/events.rs
+++ b/src/bridge/events.rs
@@ -495,7 +495,6 @@ fn parse_style(style_map: Value) -> Result<Style> {
                 }
                 ("blend", Value::Integer(blend)) => style.blend = blend.as_u64().unwrap() as u8,
 
-                // TODO: handle correct precedence?
                 ("underline", Value::Boolean(true)) => {
                     style.underline = Some(UnderlineStyle::Underline)
                 }
--- a/src/editor/window.rs
+++ b/src/editor/window.rs
@@ -309,7 +309,7 @@ impl Window {
         self.send_command(WindowDrawCommand::Clear);
         // Draw the lines from the bottom up so that underlines don't get overwritten by the line
         // below.
-        for row in self.grid.height..0 {
+        for row in (0..self.grid.height).rev() {
             self.redraw_line(row);
         }
     }
--- a/src/editor/window.rs
+++ b/src/editor/window.rs
@@ -225,7 +225,17 @@ impl Window {
                 );
             }
 
+            // Due to the limitations of the current rendering strategy, some underlines get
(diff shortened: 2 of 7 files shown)


src/bridge/events.rs
```diff
@@ -493,9 +493,25 @@ fn parse_style(style_map: Value) -> Result<Style> {
...
+                ("undercurl", Value::Boolean(true)) => {
+                    style.underline = Some(UnderlineStyle::UnderCurl)
+                }
+                ("underdotted", Value::Boolean(true)) => {
+                    style.underline = Some(UnderlineStyle::UnderDot)
+                }
+                ("underdashed", Value::Boolean(true)) => {
+                    style.underline = Some(UnderlineStyle::UnderDash)
+                }
+                ("underdouble", Value::Boolean(true)) => {
```
MultisampledNight, line 511: `:h attr-list` calls those `underdot`, `underdash` and `underlineline`, do you have any place where the `ed` suffix comes from? I also like `double` more than `lineline`, but I don't want to break compatibility with the rest of the ecosystem due to poor naming.

```suggestion
                ("underdot", Value::Boolean(true)) => {
                    style.underline = Some(UnderlineStyle::UnderDot)
                }
                ("underdash", Value::Boolean(true)) => {
                    style.underline = Some(UnderlineStyle::UnderDash)
                }
                ("underlineline", Value::Boolean(true)) => {
```

src/renderer/grid_renderer.rs
```diff
@@ -209,4 +193,58 @@ impl GridRenderer {
...
+                path.move_to(p1);
+                let mut i = p1.0;
+                let mut sin = -2.;
+                while i <= p2.0 {
+                    sin *= -1.;
+                    i += 4.;
+                    path.quad_to((i - 2., p1.1 + sin), (i, p1.1));
+                }
+                canvas.draw_path(&path, &underline_paint);
+            }
```
MultisampledNight, line 239: Your concerns about HiDPI and the curl appearing too thin are rightfully:

Screenshot of Neovide showing the undercurl and underline modifiers in comparison to normally sized characters on a HiDPI setup

I think this could be patched up by making the line slightly thicker, doing less iterations of the loop and doing larger steps each time, though this is indeed tricky. To be honest, I don't think we'll be coming to a really aesthetic solution there.

MultisampledNight (commented):
Thank you! This is _great._

By the way, how did you even manage to set the "invalid" modifier names? `:hi! Normal gui=underdouble` results into neovim yelling at me with "invalid value".

src/bridge/events.rs
calops: As mentioned in my other comment, the latest neovim versions seem to have changed the names so that they become aligned with vim's names. I will amend this to support both names so that people aren't forced to upgrade.

src/renderer/grid_renderer.rs
calops: Yeah I was afraid of this.

To be fair: no terminal I've tested handles it any better than what we're seeing here. So, maybe we could just release this like this and then improve upon it in some way?

A quick win would be to make the line thickness a configuration option, as well as maybe the pattern length (which can be used to scale the dashes and dots as well). Trying to automatically render this the best way seems like a doomed idea, or at least much harder than the rest.

MultisampledNight:
Yeah, this is fair. In this case you should be able to just match both patterns using the `|` operator.

src/bridge/events.rs
```diff
@@ -493,9 +493,25 @@ fn parse_style(style_map: Value) -> Result<Style> {
                 ("strikethrough", Value::Boolean(strikethrough)) => {
                     style.strikethrough = strikethrough
                 }
-                ("underline", Value::Boolean(underline)) => style.underline = underline,
-                ("undercurl", Value::Boolean(undercurl)) => style.undercurl = undercurl,
                 ("blend", Value::Integer(blend)) => style.blend = blend.as_u64().unwrap() as u8,
+
+                // TODO: handle correct precedence?
```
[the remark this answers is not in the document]
calops: That sounds good to me, I removed the comment.

src/renderer/grid_renderer.rs
MultisampledNight: Allowing configuration seems like the most sensible option. There's lots of cases where relying on the scale factor the system gives us is wrong.

calops: Another idea would be to actually scale with the font height, but round the numbers so that they stay pixel perfect. Everything should still be 1px thick at low DPIs, but it should scale nicely after that, without any aliasing shenanigans. I'll see if I can come up with a quick example of this.

MultisampledNight: Just a note, how currently everything looks like with HiDPI

All underline styles tried out and applied on a HiDPI display. Notably the distance is too small every time.

calops: Alright, so, I worked on this a little and I'm a bit stuck.

On one hand, automatic scaling looks pretty nice: uvNaekl
 
On the other hand... this can't actually work consistently and this seems to be a limitation of the whole rendering strategy of neovide.

The issue is that everything is drawn asynchronously through a message queue, which works under the premise everything that'll get drawn (within a window) is on its own independent region of the screen, meaning there's no conflict anywhere.

But to draw thick underlines, we need them to be drawn between two lines. They can't be confined to the line the text belongs to, or they will simply strike through it at higher thicknesses (especially the undercurl). But by removing them from the line's canvas so they don't get clipped, they get randomly overwritten by the line below, leaving only two pixels to work with.

What's needed would be to redraw the underlines alone every time the line of text below them gets drawn. It seems pretty hard to do with how the code is currently set up, unless there's an easy solution I'm missing?

MultisampledNight: I think you're right, `draw_foreground` only ever gets called with text fragments, and not the whole current text buffer. This is yet another issue which could be handled through keeping state about the _whole_ text visible at all times, and also redrawing the line below when drawing underlines. But as of now, Neovide doesn't have such a facility.

For now, the line below "sometimes" overlapping the curl is as far as you can get without taking quite a few hoops in terms of re-writing code architecture, I fear.

calops: It's a little better when I:
* Draw lines from the bottom up instead of the top down when drawing everything
* Redraw the line above after any line that's individually drawn
But there are still plenty of glitches here and there. I'm hopeful they can be ironed out, but it'll all just be hacks for now.

MultisampledNight: Alright. Then feel free to push your changes, I'll create an issue about possible ways to approach the larger needed change after merging this PR.

calops: Yeah that's sensible.

For now, the automatic scaling works well so long as line thickness doesn't get above 1. Above that, underlines get clipped here and there. Perhaps it would make sense to put the automatic scaling behind a configuration option for now? I'll push those changes up for review.

src/editor/window.rs
```diff
@@ -297,7 +307,9 @@ impl Window {
 
     pub fn redraw(&self) {
         self.send_command(WindowDrawCommand::Clear);
-        for row in 0..self.grid.height {
+        // Draw the lines from the bottom up so that underlines don't get overwritten by the line
+        // below.
+        for row in self.grid.height..0 {
```
MultisampledNight, line 312: Those two aren't quite the same. Note that `..` ranges are start inclusive, end exclusive. Also, ranges are a bit weirdly designed in Rust imo so this likely doesn't do what you'd expect, [I put a short playground example together for that](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b771ce3e17b0215124fdbee93a978fc6). TLDR is, `start..end` is an empty iterator if `start >= end`.

src/renderer/grid_renderer.rs
```diff
@@ -209,4 +193,79 @@ impl GridRenderer {
...
+                    .set_style(skia_safe::paint::Style::Stroke);
+                let mut path = Path::default();
+                path.move_to(p1);
+                let mut i = p1.0;
+                let mut sin = -2. * stroke_width;
+                while i <= p2.0 {
+                    sin *= -1.;
+                    i += 4. * stroke_width;
+                    path.quad_to((i - 2. * stroke_width, p1.1 + sin), (i, p1.1));
+                }
```
MultisampledNight, line 250: To be honest, this could maybe get more precise. Instead of doing the last "half" curve, we could maybe try to clip it or stop beforehand and use `p2.0` as the final X coordinate of the last target point, plus some arcsin hackery on the Y coordinate. But that might or might not look ugly.

MultisampledNight (commented):
Text rendering has one issue with lines not being redrawn when window size changes, this can be reproduced easily by changing the window size in floating window mode:

Neovide only draws the top half of the text

Else, one

```vim
:let g:neovide_underline_automatic_rescaling=v:true
:hi! Keyword gui=undercurl
```

later, I can confirm that clipping of the undercurl does happen, but I don't think it's too bad for now:

Casual zshrc with every keyword being undercurled, some undercurls are clipped to the below line

Nice work regardless.

MultisampledNight:
(also, small sidenote since formatting seems to be quite your nemesis: check out `:h g:rustfmt_autosave`, which will format the file automatically upon writing in neovim)

src/renderer/grid_renderer.rs
calops: Absolutely, I had that planned and then completely forgot to do it. Clipping is easy enough, but it may look awkward on lines where multiple consecutive fragments have different undercurls.

An easy fix would be to simply have the width of the wave be the same as the font width. That way it'll repeat seamlessly and we don't need to clip it anymore.

src/editor/window.rs
calops: Huh, so it is. I wonder why everything still renders correctly then?

src/renderer/grid_renderer.rs
calops: There's the result:
olpD34x

website/docs/configuration.md
```diff
@@ -114,6 +114,19 @@ let g:neovide_profiler = v:false
 Setting this to `v:true` enables the profiler, which shows a frametime graph in the upper left
 corner.
 
+#### Underline automatic scaling
+
+```vim
+let g:neovide_underline_automatic_scaling = v:true
```
MultisampledNight, line 120: Note that the config docs should document the default, and not one possible configuration.

MultisampledNight (commented):
Looks great! One last thing I'd like to have addressed before merging, though I'll just patch it myself quickly.

MultisampledNight:
Alright. Truly, thank you! I'll merge this now.

PR #1483, opened 2022-08-15
Fix #1475

### LINKED ISSUE #1475: Ctrl+Shift keybinds not working properly <C-S-x>
**Describe the bug**
Ctrl+Shift keybindings don't register correctly.

**To Reproduce**
Steps to reproduce the behavior (minimal example):
1. Run clean neovide: `neovide -- --clean`
2. Create a normal Ctrl+Shift+key binding: `nnoremap <c-s-x> <cmd>echo 'Ctrl+Shift+X'<cr>`

3. Press `Ctrl+Shift+X`
4. Nothing prints in command line
**Note:** x is just an example key, it doesn't work with any key

**Expected behavior**
Command line displays Ctrl+Shift+X, as it does when executed within the terminal

 - OS: Arch Linux x86_64
 - Neovide Version 0.9.0
 - Neovim Version  0.8.0-dev-nightly-12-g977790746

**Additional context**
I have tried with <s-c-x> instead of <c-s-x> and it doesn't work either, but with <c-x> and <s-x> separately it works perfectly:
```vim
nnoremap <c-s-x> <cmd>echo 'Ctrl+Shift+X'<cr>
nnoremap <s-c-x> <cmd>echo 'Ctrl+Shift+X'<cr>
nnoremap <c-x> <cmd>echo 'Ctrl+X'<cr>
nnoremap <s-x> <cmd>echo 'Shift+X'<cr>
```

`neovide_rCURRENT.log` Logfile is too long to paste it here
`neovide_r2022-08-11_21-10.log`: TRACE [neovide] Neovide version: 0.9.0

Daniel-Esteban:
Before, CTRL+SHIFT keystrokes registered as like so:

CTRL+SHIFT+p = \<C-P> instead of \<C-S-p> or \<C-S-P>
CTRL+SHIFT+x = \<C-X> instead of \<C-S-x> or \<C-S-X>
...

So different keymaps couldn't be assigned to CTRL+x and CTRL+SHIFT+x
because nvim assumes it is the same mapping.

Now, CTRL+SHIFT+x is interpreted as \<C-S-X>, so different mappings can
be assigned to CTRL+x and CTRL+SHIFT+x.

- Fix

- Maybe, I don't know

### DIFF
--- a/src/window/keyboard_manager.rs
+++ b/src/window/keyboard_manager.rs
@@ -170,7 +170,7 @@ impl KeyboardManager {
         let special = special || self.ctrl || use_alt(self.alt) || self.logo;
 
         let open = or_empty(special, "<");
-        let modifiers = self.format_modifier_string(use_shift);
+        let modifiers = self.format_modifier_string(special || use_shift);
         let close = or_empty(special, ">");
 
         open.to_owned() + &modifiers + text + close


MultisampledNight:
Thank you!

Even though this is technically a breaking change, since someone could've wanted to map <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>...</kbd> and mapped it to `<C-...>` instead, but technically it just opens up a new usecase. So I wouldn't consider this a breaking change but a change worth documenting anyways.

PR #1485, opened 2022-08-15
Fix glitchy smooth scrolling (#1334)

outfoxxed:
Fixes broken smooth scrolling with multigrid, as seen in #1334.

Comparison video: (Before, After)
https://user-images.githubusercontent.com/83010835/184727581-bacd386a-8752-4c9c-9cca-54a935fc36e6.mp4

The cause of the issue is that a window's top line is changed before it is redrawn when the window's viewport updates.
If it has not been redrawn before neovide renders it, it would draw in the wrong position.

This PR keeps track of the top line, and renders the window at the old top line until the canvas has been redrawn.

- Fix

- No

### DIFF
--- a/src/renderer/rendered_window.rs
+++ b/src/renderer/rendered_window.rs
@@ -124,6 +124,7 @@ impl LocatedSurface {
 pub struct RenderedWindow {
     snapshots: VecDeque<LocatedSnapshot>,
     pub current_surface: LocatedSurface,
+    predraw_top_line: Option<u64>,
 
     pub id: u64,
     pub hidden: bool,
@@ -162,6 +163,7 @@ impl RenderedWindow {
         RenderedWindow {
             snapshots: VecDeque::new(),
             current_surface,
+            predraw_top_line: None,
             id,
             hidden: false,
             floating_order: None,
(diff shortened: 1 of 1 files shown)


outfoxxed:
^ squashed commits

MultisampledNight:
Thank you! Frankly I couldn't seem to reproduce the original issue, but this PR somehow feels... smoother in general. I guess I've been so used to minor glitches in the scrolling that I don't notice them anymore.

Anyways, squashing was not necessary. We automatically squash upon merge, and squashing makes commit-based reviewing impossible. For such a small change it's acceptable though.

msooseth:
This is somehow back in `neovide 0.15.0` that I am using as part of Arch.  It's super-new:

https://gitlab.archlinux.org/archlinux/packaging/packages/neovide/-/commits/main

It landed 3rd of April. I have been seeing this the past few days. I'm on Sway (so under Wayland), not running neovide under XWayland, but natively running it.

I am seeing exactly the behaviour as the video in the original issue, glitch on scroll when I'm repeatedly pushing the "down" or "up" arrow and hence moving the screen, a momentary flicker, exactly like here: https://github.com/neovide/neovide/issues/1334

PR #1489, opened 2022-08-16
Fix regression in #1485

outfoxxed:
This pr fixes a regression in #1485 which breaks window viewports (sorry).
It is most noticeable when creating or moving windows.

Regression:
https://user-images.githubusercontent.com/83010835/184974106-07ca0de0-249f-45e9-a05a-1a2cbfd56332.mp4

This pr fixes it by only using the old draw position if the window is scrolling
https://user-images.githubusercontent.com/83010835/184974417-6ddadbea-3567-4e02-8ce5-8883776f073b.mp4

As far as I can tell this shouldn't cause any additional regressions, and the smooth scrolling fix from #1485 still works.

- Fix

- No

### DIFF
--- a/src/renderer/rendered_window.rs
+++ b/src/renderer/rendered_window.rs
@@ -121,10 +121,16 @@ impl LocatedSurface {
     }
 }
 
+enum ScrollPredraw {
+    None,
+    Scrolling,
+    Ready(u64),
+}
+
 pub struct RenderedWindow {
     snapshots: VecDeque<LocatedSnapshot>,
     pub current_surface: LocatedSurface,
-    predraw_top_line: Option<u64>,
+    scroll_predraw: ScrollPredraw,
 
     pub id: u64,
     pub hidden: bool,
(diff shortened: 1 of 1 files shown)


outfoxxed:
Have not noticed any new regressions since yesterday (I cargo installed my fork systemwide)

outfoxxed:
^ Also changed names to make it easier to understand at a glance

src/renderer/rendered_window.rs
```diff
@@ -121,10 +121,15 @@ impl LocatedSurface {
     }
 }
 
+enum PositionOverride {
+    None,
+    Override { top_line: u64, current_scroll: f32 },
+}
+
```
LoipesMas, line 128: I think using `Option` would be more idiomatic and have this be just the struct or maybe even only use a tuple (since variable name should give enough context). You could then do `.unwrap_or((self.current_surface.top_line, self.current_scroll))` (or the struct version). And `is_some`/`is_none`.

outfoxxed: Yeah, that makes sense.
I'll change it when I can if you dont want to.

MultisampledNight: I'll just do that.

MultisampledNight:
Awesome. Thank you!

PR #1490, opened 2022-08-17
Keep the same framerate until a redraw when unfocusing the window.

### LINKED ISSUE #1484: Cursor setting to outline only upon unfocus takes a split second
**Describe the bug**
Upon un-focusing Neovide, the cursor switches to showing an unfilled rectangle with only an outline only after a sizable delay passed.

**To Reproduce**
1. Startup Neovide
2. Unfocus Neovide
3. See the cursor switching to a non-filled rectangle only after a slight delay

**Expected behavior**
To have the cursor turn into a rectangle with outline only immediately after unfocus.

 - OS: Irrelevant, though Arch Linux
 - Neovide Version 0.10.1
 - Neovim Version 0.7.2

N/A

**Additional context**
Since the delay matches 0.2 seconds and a bit less from my feeling, this is likely a bug introduced by #1454. Instead of lowering the framerate right the frame Neovide is unfocused, we might need to wait an additional frame for un-focus events to actually take effect.

I believe this is an easy issue to tackle. We might need to take additional state around the area the PR in question modified in order to not only detect when Neovide was unfocused, but also what the current frame number is to even record when Neovide was unfocused.

EDIT: On a second thought, this might be overkill. A reordering of check/update could be enough.

Pangoraw:
This PR replaces the logic to get the framerate between event updates to use the higher framerate until a first render has been done when unfocusing. This makes the first render following an unfocus event use the faster framerate, fixing #1484.

The reordering solution suggested in #1484 does not seem to be enough since multiple events can be scheduled before the frame is redrawn in which case `focused` will be false and the slower framerate will be used.

Fixes #1484.

- Fix

- No

### DIFF
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -413,7 +413,12 @@ pub fn create_window() {
 
     let mut previous_frame_start = Instant::now();
 
-    let mut focused = true;
+    enum FocusedState {
+        Focused,
+        UnfocusedNotDrawn,
+        Unfocused,
+    }
+    let mut focused = FocusedState::Focused;
 
     event_loop.run(move |e, _window_target, control_flow| {
         // Window focus changed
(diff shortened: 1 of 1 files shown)


MultisampledNight:
Looks awesome, thank you!

PR #1511, opened 2022-08-27
Make mouse more responsive

ad-chaos:
- Fix (I think?)

Previously every *tick* of the mouse wheel didn't result in a scroll, now it does.

- No (at least not intentionally)

### DIFF
--- a/src/window/mouse_manager.rs
+++ b/src/window/mouse_manager.rs
@@ -68,6 +68,25 @@ struct TouchTrace {
     left_deadzone_once: bool,
 }
 
+struct ScrolledLines {
+    x: f32,
+    y: f32,
+}
+
+impl ScrolledLines {
+    fn new() -> ScrolledLines {
+        ScrolledLines { x: 0., y: 0. }
+    }
+
+    fn reset(&mut self) {
+        self.x = 0.;
+        self.y = 0.;
+    }
+
+    fn line_scrolled(&self, font_width: f32, font_height: f32) -> bool {
+        self.x.abs().floor() > font_width || self.y.abs().floor() > font_height
+    }
+}
 pub struct MouseManager {
     dragging: Option<String>,
     drag_position: PhysicalPosition<u32>,
@@ -78,7 +97,7 @@ pub struct MouseManager {
 
     // the tuple allows to keep track of different fingers per device
     touch_position: HashMap<(DeviceId, u64), TouchTrace>,
-
+    pixels_scrolled: ScrolledLines,
     window_details_under_mouse: Option<WindowDrawDetails>,
 
     mouse_hidden: bool,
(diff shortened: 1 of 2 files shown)


src/window/mouse_manager.rs
```diff
@@ -263,16 +256,12 @@ impl MouseManager {
                 modifier_string: keyboard_manager.format_modifier_string(true),
             }
             .into();
-            for _ in 0..(new_y - previous_y).abs() {
+            for _ in 0..(y as i32).abs()+1 {
```
MultisampledNight, line 259: And I think the culprit for that is here. Consider the case of a very small movement under `1.0`, which is almost every scroll movement on my system. Previously, those changes accumulated in `self.scroll_position` until they were large enough to cause a difference of `1.0`. But now, every single one of these is reflected and basically `1.0` (because `0.whatever as i32` is `0` and `+ 1`), which means every single frame I'm scrolling has an effect.

I'm not too sure how to correctly handle this though. Maybe we could restore the previous behavior and multiply `y` by a configurable value to accommodate every possible system, since it probably worked fine on yours.

MultisampledNight (commented):
Thank you!

Fix-wise, this PR makes the view scroll _insanely_ fast at every small scroll, both on a touchscreen and with a touchpad. What's a slow 25px movement with the finger on a touchscreen results into a 120 line change.

Format-wise, consider running `cargo fmt` before committing or configure your editor of choice for doing so (tip: check `:h rustfmt-autosave`).

src/window/mouse_manager.rs
ad-chaos: Most mouse wheels are *ticked* meaning they are not smooth, so I think that it is fine to call the `nvim_input_mouse()` api function, every tick for mouse input. I think we should have different handling of touch, touchpad and mouse instead.

For touch and touchpad, I think what we could do is to calculate `pixels per line` and when those many pixels are scrolled, we can sent a single `nvim_input_mouse()` api call. But, this is not without caveats!
Under `:h 'mousescroll'` you'll see that a single api call results in different amount of lines scrolled.
The default being `ver:3,hor:6` meaning it'll scroll 3 vertical lines or 6 horizontal lines depending on the parameters passed.

Possible solutions:
1) We scroll by 3 lines every `pixels per line` scroll
2) We scroll by 3 lines every 3*`pixels per line` scroll
3) Enforce a default of `mousescroll=ver:1,hor:1` on neovide users

ad-chaos:
> (tip: check :h rustfmt-autosave).

Wow! Never knew the rust filetype already had such features, nice!

src/window/mouse_manager.rs
MultisampledNight: (sorry for the delay, lost track on this one) To be honest, I think option 3 sounds like the best one, even if I usually advocate for not touching user-configurable options at all if possible. I can't figure out a case where scrolling 3 lines at once is more desirable than a flexibly configurable sensitivity config option.

ad-chaos:
I think this should be it:
1) Added a default mouse scroll of one line (vertical) and one character (horizontal) (This is configurable)
2) Changed pixel scrolling to match what  I said [here](https://github.com/neovide/neovide/pull/1511#discussion_r957536694)

ad-chaos:
#2057

PR #1525, opened 2022-09-05
Fix unwanted resizing at startup when the window is remembered to be maximized

s-cerevisiae:

- Fix

- No

Previously the window would be resized at startup even if it is remembered to be maximized.

This PR fixes that by adding a new field in `GlutinWindowWrapper` to record whether the window is really maximized at startup instead of only reading from `CmdLineSettings`.


MultisampledNight:
Thank you!

NielsKersic:
I'm still facing this issue after installing Neovide from source (MacOS Ventura 13.0).
All other window sizes are remembered correctly, just not maximized.

PR #1529, opened 2022-09-06
Solved C-M not working in Linux

daniprado:
- Fix

- No

## Comment
This is the [solution proposed](https://github.com/neovide/neovide/issues/994#issuecomment-1063500913) by @Polish-Civil to [#994](https://github.com/neovide/neovide/issues/994).


MultisampledNight:
Thank you!

By the way, for future PRs here and also anywhere else: Please create a new branch for all and every change you make. It makes local testing setups less messy.

Issue #1542, opened 2022-09-16
Some Digraphs Recently Broken


jamesb6626:
**Describe the bug**
Since 346dc05, digraphs starting with the characters `<` and `|` are broken.
Instead of allowing digraph entry, it inserts the text `<S-lt>` or `<S-Bar>` respectively.
Note: `>`, and digraphs with these as the second character are unaffected.

**To Reproduce**
1. `$ neovide -- -u NONE`
2. `i<C-K><lt>` or `i<C-K><Bar>` behaves as described above.

**Expected behavior**
Some affected digraphs are `||`, `<<`, `<=`. For anyone testing this.

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04.5
 - Neovide Version 346dc05, d453921 (main)
 - Neovim Version 0.7.2

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/9586434/neovide_rCURRENT.log)
[neovide_r2022-09-16_19-23-15.log](https://github.com/neovide/neovide/files/9586438/neovide_r2022-09-16_19-23-15.log)

**Additional context**
Neovide built from source: `$ cargo install --path .`
Bug is not present in dc35636.

s-cerevisiae:
This is probably the same problem as #1539. I'm looking for a fix.

s-cerevisiae:
It's a bit late now, but the fix is merged. Would you like to test it?

jamesb6626:
Sorry for the delay, seems like its fixed. Also fixed an issue I was having with hop.nvim, similar to #1539.
Thanks. I'll close the issue.

Issue #1539, opened 2022-09-15
Motions from lightspeed.nvim don't work for certain char arguments


hallettj:
**Describe the bug**
This bug comes up specifically when using [lightspeed.nvim](https://github.com/ggandor/lightspeed.nvim). I'm reporting it here because the problem occurs in Neovide, but does not occur when running nvim in a terminal.

Lightspeed overrides some movement commands like `f` and `t`. When attempting to use those custom movements to move to certain characters such as `<` instead of the expected behavior I see this error message:

    not found: <80><fc>^B<

I have seen this problem with the characters `<` and `|`. The problem does not occur with `>` or with alphanumeric characters.

Using git bisect I found that this issue was introduced by #1483.

I tried reading the Lightspeed source to come up with a more minimal reproduction, but the Fennel code is a bit much for me to decipher right now. The function that handles 1-character search is [ft.go](https://github.com/ggandor/lightspeed.nvim/blob/main/fnl/lightspeed.fnl#L577).

**To Reproduce**
1. Install [lightspeed.nvim](https://github.com/ggandor/lightspeed.nvim)
2. Open a file with this content: `const x = 1 < 3`
3. With the cursor at the beginning of that line in normal mode, type `yt<`

**Expected behavior**
There should be no error message. The default register should contain the text `const x = 1 `.

**Actual behavior**
This error message is reported upon typing `<`:

    not found: <80><fc>^B<

**Desktop (please complete the following information):**
 - OS: Debian Testing (bookworm)
 - Neovide Versions 346dc05 and later commits in main
 - Neovim Versions NVIM v0.7.2, NVIM v0.8.0-dev-1064-g48e060f64, and NVIM v0.8.0-dev-1092-g89b9eab63 (latest nightly as of Sept 15 2022)

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
https://gist.github.com/hallettj/d5ed3158d1ee9a89a34968502b4cd4ae

hallettj:
I've noticed another issue with the `<` key. When viewing git status with [fugitive](https://github.com/tpope/vim-fugitive) there are bindings for `>` to expand the diff for a file, and `<` collapses the diff. After #1483 pressing `<` in a fugitive buffer puts the editor into operator pending mode instead of collapsing the diff. (But `>` still works as expected.) It's possible there is an interaction with another one of my plugins here. If I get more information I may open another issue.

s-cerevisiae:
[...]

PR #1547, opened 2022-09-18
Fix key string formatting for special keys

s-cerevisiae:
- Fix

- No

---
Previously (introduced in #1483) some of the special keys (namely `<` and `|`) are mistakenly formatted with a `S-` prefix, which has caused problem in some plugins and Neovim itself (see #1539 and #1542 ).

This PR fixes that by
 - separating the formatting of normal keys (which should always be prefixed with `S-` if other modifiers are pressed, aka `use_shift = true`) and special keys (in which some should be but some others should not), and
 - introducing a new return value for `is_special` function indicating whether `S-` prefix should be used for this special key.

---

Also during this I've found several other problems in the `keyboard_manager` module. I will open another issue for these.

### DIFF
--- a/src/window/keyboard_manager.rs
+++ b/src/window/keyboard_manager.rs
@@ -134,9 +134,9 @@ impl KeyboardManager {
             if self.prev_dead_key.is_some() {
                 //recover dead key to normal character
                 let real_char = String::from(self.prev_dead_key.unwrap());
-                Some(real_char + &self.format_keybinding_string(true, true, key_text))
+                Some(real_char + &self.format_special_key(true, key_text))
             } else {
-                Some(self.format_keybinding_string(true, true, key_text))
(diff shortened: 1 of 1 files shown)


s-cerevisiae:
I assumed that non-special keys should always be prefixed with `C-S-` if both modifier keys are pressed, which is **not** true. 
`^` and `_` should not.

And there seems to be much more edge cases that I have expected.

For this case would it be better to append `S-` to alphabetic keys only? I'm in need of a thorough documentation on `nvim_input`.

MultisampledNight:
Yeah, that sounds fair enough. May break some layouts where Shift on an alphabetic character doesn't automatically mean the capital version of it, but I don't know any off the top of my head.

MultisampledNight:
Looks good to me, 39! Sorry that it took so long, lost track of this.

Issue #1562, opened 2022-09-28
Enable "Open With" on MacOS


n8henrie:
**Is your feature request related to a problem? Please describe.**

Neovide doesn't appear in the `Open With` options when you right-click a file in Finder on MacOS. If you go through the trouble of manually navigating to it, you can then `Open With` neovide, but it opens to a blank buffer instead of the file.

**Describe the solution you'd like**

1. I would like neovide to appear in the list of apps that I can `Open With` for common text-based filetypes (`.txt`, `.md`, `.xml`, `.ics`, etc. etc.).
2. When I choose to `Open With` neovide, I would like it to open the files in question.

It looks like # 2 is blocked: https://github.com/neovide/neovide/issues/644

I would be happy to implement # 1 as a PR if it would be considered for merging; I have a working POC locally. However it won't do a whole lot of good until # 1 works.

I would likely take logic for MacOS packaging out of the GitHub Actions and put it into a bash script that could be called by GA or run locally to create a universal binary, which I think would be better than leaving business logic in the GitHub Actions script. I could probably implement the logic in rust (or python) instead, but bash seems like the simplest solution, and since it would only be run for packaging I think it is a reasonable approach.

**Additional context**

Related issues:
- https://github.com/neovide/neovide/issues/1326
- https://github.com/neovide/neovide/issues/1259
- https://github.com/neovide/neovide/issues/644

Info.plist files for comparison:
- `/System/Applications/TextEdit.app/Contents/Info.plist`
- [goneovim](https://github.com/akiyosi/goneovim/blob/c87cba639be176b6a4379e4fc0d4c9a8cb1760ff/cmd/goneovim/darwin/Contents/Info.plist)
- [macvim](https://github.com/macvim-dev/macvim/blob/951e80d7a470e2e45af7acf132e235229ac9c077/src/MacVim/Info.plist)

chrisgrieser:
I would also like to request that feature. 
(especially since the CLI is not working for me, I cannot open any file with Neovide, except for clunky `:e` stuff.... 🙈 )

jacopogrecodalceo:
Same here. I found a solution with an AppleScript and Automator, giving the cmd 'open -a neovide my_file.py', but without, if I double click on a file in Finder Neovide opens a new file. I'm on MacBook Pro M1 2021 - Monterey.

kabouzeid:
This is a huge blocker for me too

MultisampledNight:
Closing as duplicate in favor of #1259.

PR #1563, opened 2022-09-28
Add script to show neovide as an option in "Open With" menus

n8henrie:
- Feature

- No

Addresses https://github.com/neovide/neovide/issues/1562

### DIFF
--- a/scripts/addtypes.sh
+++ b/scripts/addtypes.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+
+set -Eeuf -o pipefail
+shopt -s inherit_errexit
+set -x
+
+run() {
+  local cmd=$1
+  /usr/libexec/PlistBuddy -c "${cmd}" target/release/bundle/osx/Neovide.app/Contents/Info.plist
+}
+
+main() {
+  plist=target/release/bundle/osx/Neovide.app/Contents/Info.plist
+
+  if [[ -e "${plist}.bak" ]]; then
+    # If the backup exists, overwrite the work copy during development to start fresh
+    cp "${plist}"{.bak,}
+  else
+    # If the backup doesn't exist, create it
+    cp "${plist}"{,.bak}
+  fi
+
+  run 'Add :CFBundleDocumentTypes array'
+
+  run 'Add :CFBundleDocumentTypes:0:CFBundleTypeExtensions array'
+  run 'Add :CFBundleDocumentTypes:0:CFBundleTypeExtensions:0 string txt'
+
+  run 'Add :CFBundleDocumentTypes:0:CFBundleTypeMIMETypes array'
+  run 'Add :CFBundleDocumentTypes:0:CFBundleTypeMIMETypes:0 string text/plain'
+
+  run 'Add :CFBundleDocumentTypes:0:CFBundleTypeName string "Plain Text File"'
+  run 'Add :CFBundleDocumentTypes:0:CFBundleTypeRole string Editor'
+
+  run 'Add :CFBundleDocumentTypes:1:LSItemContentTypes array'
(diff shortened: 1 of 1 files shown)


n8henrie:
NB: This is proof-of-concept at this point, pending discussion / review.

If it seems like a good idea and this implementation is reasonable, my plan would be to:

1. Move the logic from https://github.com/neovide/neovide/blob/0b6e20c1994c761814eb6a8ad33dea90ade97573/.github/workflows/build.yml#L128 to e.g. `scripts/build_macos.sh`, which will make it easier for users and devs to build their own copy locally instead of relying on GA
2. Add this logic into that script to modify the `Info.plist` (prior to the `codesign` step)

Could also use some feedback regarding the filetypes that this would make the most sense for. The related issue links to the MacVim `Info.plist` for comparison, though its licensing might preclude following its example too closely.

mgax:
I guess the natural place to set file types is through [cargo-bundle](https://github.com/burtonageo/cargo-bundle), which is used to generate the app bundle, but it doesn't seem to have support.

MacVim has a long list of filetypes in its [Info.plist](https://github.com/macvim-dev/macvim/blob/master/src/MacVim/Info.plist). I'm not sure if we can copy it outright; it's licensed with the [Vim License](https://choosealicense.com/licenses/vim/).

If we do use `PlistBuddy` to modify `Info.plist`, which TBH sounds like the best option, I suggest we try to [copy](https://choosealicense.com/licenses/vim/) the relevant content from a static `.plist` file in the repository into the app bundle plist.

shaunsingh:
Sorry for the delay, script looks good to me so far

> If it seems like a good idea and this implementation is reasonable, my plan would be to:
> Move the logic from
> [neovide/.github/workflows/build.yml](https://github.com/neovide/neovide/blob/0b6e20c1994c761814eb6a8ad33dea90ade97573/.github/workflows/build.yml#L128)
> Line 128 in [0b6e20c](https://github.com/neovide/neovide/commit/0b6e20c1994c761814eb6a8ad33dea90ade97573)
>        - name: Merge into a universal app 
> to e.g. scripts/build_macos.sh, which will make it easier for users and devs to build their own copy locally instead of relying on GA
> Add this logic into that script to modify the Info.plist (prior to the codesign step)

I'm completely fine with this, if we're straying away from `cargo-bundle` it makes sense to have a way for the user to build a bundle properly as well. You can also take advantage of the `build.rs` file if you'd like. Just make sure to update the documentation with the new steps for building a binary bundle on macOS

> I guess the natural place to set file types is through [cargo-bundle](https://github.com/burtonageo/cargo-bundle)

While I'd personally prefer getting support for modifying Info.plist merged into cargo-bundle as well, it's not getting support anytime soon (https://github.com/burtonageo/cargo-bundle/issues/33). Using PlistBuddy sounds like a good solution for now.

 Regarding the Vim License, I believe as long as we include a copy of the vim license as a comment at the top of the plist file, it should be fine. I'll ping the macvim author (@ychin) for his opinion on it as well.

ychin:
I don't see a problem with the Vim license but then I'm not a copyright lawyer. I believe it should be compatible. Does Neovide not include anything Vim related to begin with? I guess you don't bundle Neovim (which is partially Vim licensed).

Which lines are you thinking of copying from MacVim btw? There are actually some issues with the Info.plist not properly setting up the UTIs (see https://github.com/macvim-dev/macvim/issues/916 and https://github.com/macvim-dev/macvim/issues/1169) and I'm ashamed to admit that I never looked into it and properly fixing them. Just make sure you understand those issues before blindly copying.

n8henrie:
Thanks for responses and feedback all around. It looks like [my issue for this](https://github.com/neovide/neovide/issues/1562) was closed with a note of going a different direction, so I'll close this PR as well.

PR #1583, opened 2022-10-03
Clean up cmdline handling

MultisampledNight:
Uses clap 4.0 in order to make adding new arguments way, way cleaner. Would be very happy if someone with a specialized usage of Neovide's CLI could actually test this, the testsuite ran through but that doesn't mean anything.

- Refactor

- No (practically, theoretically arguments passed to NeoVim have been reordered a bit)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -96,7 +96,7 @@ dependencies = [
  "bitflags",
  "cexpr",
  "clang-sys",
- "clap",
+ "clap 3.2.22",
  "env_logger",
  "lazy_static",
  "lazycell",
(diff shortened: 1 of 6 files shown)


jdrouhard:
I haven't dug too deeply yet, but this commit broke the ability for Neovide to remember the window size. I bisected it to this commit and it looks like the `parse_window_geometry()` function is no longer called if the `--geometry` command line option isn't passed. This means that if you don't pass `--geometry`, you just get the default, but if you do, then it loads the previous values but doesn't use them inside `parse_window_geometry()`.

Issue #1301, opened 2022-05-05
Scale/zoom, configurable at runtime


Hubro:
**Is your feature request related to a problem? Please describe.**

Setting the font size in Neovide is very clunky. It requires the `guifont` setting to be changed, which requires a very specific format including the font family name, the font size and any font weights and/or styles.

There are a couple of use cases in which I really miss the ability to just *increase the scaling* a little bit:

- **Screen sharing**

  I often screen share code as part of presentations at work. Very often I need to zoom in a bit for everybody to be able to see the code. For Vim in a terminal, I can just press <kbd>Ctrl</kbd><kbd>+</kbd>, but in Neovide I'm stuck at a specific font size. A perfect solution would be to have a variable like `g:neovide_scale` that defaults to `1`. I could bind <kbd>Ctrl</kbd><kbd>+</kbd> and <kbd>Ctrl</kbd><kbd>-</kbd> to increase and decrease this number by something like `0.05`.

- **Zen mode**

  Zen mode plugins like https://github.com/folke/zen-mode.nvim have hooks that run when enabling/disabling Zen mode. How cool wouldn't it be to be able to add `vim.g.neovide_scale = 1.5` to this hook, making everything larger while in Zen mode?

**Describe the solution you'd like**

Add a config value like `g:neovide_scale`. Changes to that value changes the scaling of the Neovide GUI at runtime.

Depending on how this is implemented, since Neovide is hardware accelerated, it might even be possible to animate the changing scale, which would be an extra nice touch.

**Describe alternatives you've considered**

Changing the font size. This is clunky. It would require parsing the current GUI font, modifying the font size value and formatting the result back into a font specification. Not very user friendly.

0x0013:
As someone who also struggled a bit with this initially, I also think it would be a good enhancement to have a more user-friendly option.

For anyone else struggling with this, I eventually solved it by defining a few global variables, and lua functions to change size, and map keybindings to them to increase, decrease or reset size to default.

Below is code snippet (neovim 0.7.0)
```
vim.g.gui_font_default_size = 12
vim.g.gui_font_size = vim.g.gui_font_default_size
vim.g.gui_font_face = "Fira Code Retina"

RefreshGuiFont = function()
  vim.opt.guifont = string.format("%s:h%s",vim.g.gui_font_face, vim.g.gui_font_size)
end

ResizeGuiFont = function(delta)
  vim.g.gui_font_size = vim.g.gui_font_size + delta
  RefreshGuiFont()
[...]

PR #1589, opened 2022-10-06
Introduce scale_factor option

BHatGuy:
- Feature

- No

I tried to implement the scaling proposed in #1301. I decided to use an additional factor to the os_scale_factor.

Example for using the new scale_factor with key bindings:
```
let g:neovide_scale_factor=1.0
function! ChangeScaleFactor(delta)
    let g:neovide_scale_factor = g:neovide_scale_factor * a:delta
endfunction
nnoremap <expr><C-=> ChangeScaleFactor(1.25)
nnoremap <expr><C--> ChangeScaleFactor(1/1.25)
```

### DIFF
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -81,10 +81,14 @@ pub struct Renderer {
 
     pub batched_draw_command_receiver: UnboundedReceiver<Vec<DrawCommand>>,
     profiler: profiler::Profiler,
+    os_scale_factor: f64,
+    user_scale_factor: f64,
 }
 
 impl Renderer {
-    pub fn new(scale_factor: f64) -> Self {
+    pub fn new(os_scale_factor: f64) -> Self {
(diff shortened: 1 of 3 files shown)


MultisampledNight:
This is _great_, thank you! I'll add the snippet to the FAQ.

MultisampledNight:
(by the way, you might want to check out `cargo fmt` and `g:rustfmt_autosave`)

TRPB:
Thanks for this @BHatGuy. I've noticed on 0.12.2 at least that I have to press another key after changing the scale factor, it's not instant. Is there any way we can trigger an instant change? It doesn't seem to matter what is key is pressed, even a mouse click works.

I wonder if there's some kind of `refresh()` function we can call to force it each execution?

fredizzimo:
It should be instant in the latest 0.13.3 release, and some releases before that.

Issue #1325, opened 2022-05-23
Font rendering issues


oredaze:
**Describe the bug**

EDIT: After playing with neovide for a while I have stopped using it because it has weird font rendering in general.

In the screenshot below you can see neovide (top) vs terminal nvim (bottom). First the capital letters can't seem to render with a consistent height. Second, see the period, comma and semicolons? They are all squashed and small. It's hard to see them sometimes. Other such small glyphs have similar problems. There are other things that are not supposed to look like they do (for example "t"), but that is not as important.
It is important to note that one of the the issues is produced due to this font (capital height), not others. It is a modified by me ubuntu mono (with fontforge). Obviously I haven't done a perfect job at generating it or something, HOWEVER, neovide is the only program that has issues rendering it. Things that don't have issue include: Alacritty, urxvt, firefox, dmenu, polybar, openbox.
I really want to use that font, and I really like neovide..., but this is hard to work with. Also other fonts are also not cleanly rendered like they should.

**Expected behavior**
Font looks like the bottom example.

**Screenshots**
![screenshot](https://user-images.githubusercontent.com/51076873/169873016-5e39c004-c9e5-4fda-be9c-c113b1c3d411.png)

**Desktop (please complete the following information):**
 - OS: Slackware (happens on my other one - gentoo as well)
 - Neovide Version 0.8.0 binary from github
 - Neovim Version 0.7.0

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/8756956/neovide_rCURRENT.log)

If anyone wants me to send them the font to examine it and give me advice on how to generate it properly (to work with neovide) I would appreciate it.

M1cha:
neovides rendering also seems to be too bright or bold:
alacritty:
![image](https://user-images.githubusercontent.com/878458/173221930-b1c8e186-c1d6-4999-b103-222161458c0c.png)

neovide:
![image](https://user-images.githubusercontent.com/878458/173221909-e395c0d1-e35f-40d4-acac-6efa782fcb74.png)

oredaze:
Your case is even better than usual. I get many other problems even when using a different font.
- 1 pixed added horizontal spacing between every character
- certain characters are blurry as compared to the rest of them or as compared to same character in every other program
- stretched font overall, but only with certain fonts (?)
[...]

PR #1603, opened 2022-10-17
Font edging and hinting moved in settings

Nawy:
This PR adds support to Font Edging and Font Hinting settings from vim configuration files.

Neovide renders with too bright and too fat fonts on macOS and probably other systems. These settings bring flexibility to adjust Neovide's font loading without rebuilding.

Problem:
https://github.com/neovide/neovide/issues/1325

- Feature
- Documentation

- No

### DIFF
--- a/src/renderer/fonts/font_loader.rs
+++ b/src/renderer/fonts/font_loader.rs
@@ -3,7 +3,10 @@ use std::sync::Arc;
 use log::{error, trace};
 use lru::LruCache;
 use nvim_rs::Value;
-use skia_safe::{font::Edging, Data, Font, FontHinting, FontMgr, FontStyle, Typeface};
+use skia_safe::{
+    font::Edging as SkiaEdging, Data, Font, FontHinting as SkiaHinting, FontMgr, FontStyle,
+    Typeface,
+};
 
 use crate::{
     renderer::fonts::swash_font::SwashFont,
@@ -25,15 +28,8 @@ impl FontPair {
         skia_font.set_subpixel(true);
 
         let settings = SETTINGS.get::<WindowSettings>();
-        let mut font_hinting = FontHinting::Full;
-        font_hinting.parse_from_value(Value::from(settings.font_hinting));
-        skia_font.set_hinting(font_hinting);
-
-        let mut font_edging: Option<Edging> = Some(Edging::AntiAlias);
-        font_edging.parse_from_value(Value::from(settings.font_edging));
-        if let Some(edging) = font_edging {
-            skia_font.set_edging(edging);
-        }
+        skia_font.set_hinting(settings.font_hinting.0);
+        skia_font.set_edging(settings.font_edging.0);
 
         let typeface = skia_font.typeface().unwrap();
         let (font_data, index) = typeface.to_font_data().unwrap();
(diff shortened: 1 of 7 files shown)


src/renderer/mod.rs
```diff
@@ -39,7 +39,7 @@ pub struct RendererSettings {
     floating_blur_amount_y: f32,
     debug_renderer: bool,
     profiler: bool,
-    underline_automatic_scaling: bool,
+    underline_automatic_scaling: bool
```
MultisampledNight, line 42: You might want to consider running `cargo fmt`, which will format the whole repo with the Rust community's defaults (which include a trailing comma). Or you set `g:rustfmt_autosave` to `v:true` directly.

src/window/settings.rs
```diff
@@ -35,6 +37,8 @@ impl Default for WindowSettings {
             touch_drag_timeout: 0.17,
             background_color: "".to_string(),
             confirm_quit: true,
+            font_edging: "AntiAlias".to_string(),
+            font_hinting: "Full".to_string()
```
MultisampledNight, line 41: All other settings are in lowercase, except these two. I'd prefer consistency here and by that, also adjust the docs, but your choice.

src/renderer/fonts/font_loader.rs
```diff
@@ -158,3 +165,23 @@ fn font_style(bold: bool, italic: bool) -> FontStyle {
...
+
+fn font_edging(edging: &str) -> Option<Edging> {
+    let edging = edging.to_lowercase();
+    match edging.as_str() {
+        "alias" => Some(Edging::Alias),
+        "antialias" => Some(Edging::AntiAlias),
+        "subpixelantialias" => Some(Edging::SubpixelAntiAlias),
+        _ => None
+    }
+}
```
MultisampledNight, line 187: Maybe implementing `ParseFromValue` like it's done in the cursor renderer with `VfxMode` would make this less messy and also allow including the enum _directly_ in the struct, without a potentially broken invariant.

MultisampledNight (commented):
Cool, thank you! Some small things I want to be looked at before throwing this in.

MultisampledNight:
Take yourself as much time as you need or would like to have. :heart:

src/renderer/mod.rs
Nawy: Fixed formatting with fmt.  Actually, I'm running it. Don't have fmt_autosave in my nvim config. Thanks for the remark.

src/window/settings.rs
Nawy: Done, all of them are lowercase now.

src/renderer/fonts/font_loader.rs
Nawy: Updated with `ParseFromValue`, I don't think that it looks better. But good to have consistency.

Nawy:
@MultisampledNight It's done. Actually, I recently discovered Neovim, and fall in love. Switched to using it for my daily work. I'm lacking a couple of things of course (including this). But going to help with that.

MultisampledNight:
That's nice to hear, GL HF!

The `ParseFromValue` thing was _almost_ what I meant, I'll just quickly patch it myself (by putting the enum itself in the settings without `String` in between).

Nawy:
@MultisampledNight I got it! Let me update it.

MultisampledNight:
No need to duplicate work, I'm almost done.

MultisampledNight:
...distractions, experiments and build times on this machine

I just found out that a `:set guifont=...` is required after setting one of those options, probably due to the font pair loading mechanisms. Maybe we can somehow re-trigger a reload of the current font pair if one of those options changes, I'll investigate that more before merging.

Nawy:
@MultisampledNight Looks cool! I'm a little bit worried about font aliasing because you removed the option-style. So it means edging is all the time there. I'm not sure what's the default value for this parameter. 

 What do you mean by "required"? I run this build on windows/macOS with no `:set guifont` in my config. I see some double initialization of these parameters, but probably because of loading multiple fonts.

MultisampledNight:
About "required": I wasn't precise enough, try setting `:let g:neovide_font_edging="alias"`. The change doesn't seem like it'd be applied until you run `:set guifont=whatever`.

MultisampledNight:
...actually, now I'm wondering whether it would be better to integrate the newly introduced config variables into `guifont` directly.

MultisampledNight:
To be honest, I think option 2 is preferrable (while removing the `edging-` part for the (anti)aliasing). The `guifont` option is kind of _meant_ to be stuffed with all font configuration, looking at the current fallback mechanism and the like. Thus, you can split it in your config on multiple lines anyways if you don't feel like reading/editing one large line.

MultisampledNight:
Feel free to do so (and also argument for one or another should you prefer option 1), but I can do that should that be out of your current motivation fuel reach.

Nawy:
I prefer option 1 because the font edging and hinting sound like a type of renderer. The same I would say about size, bold, and italic settings, they sound like renderer types. 
But if we prefer option 2, how would you detect FontEdging and FontHinting? I see not some literals (h, b, i) are used. So we have to use at least a similar mechanism.
I'm going to try to implement it. Give me a shot. Just would be nice to deal with the format.

MultisampledNight:
Option 1 has the problem of these variables being seemingly unrelated to the guifont option, and we suddently have 2 places for configuring font settings.

If you want to go for option 2, this scheme sounds good to me:

- `hinting-full`, `hinting-normal`, `hinting-slight` and `hinting-none` are tokens for the appropiate hinting strategy
- `alias`, `antialias`, `subpixel`/`subpixelantialias` are tokens for the appropiate antialiasing strategy

If you want to include "edging" for the aliasing either way, fair enough. But I'd say option 2 uses existing machinery and makes it easy for users to set one, comprehensive guifont config, while option 1 adds partly split machinery where one part relies on the other one to be applied.

Nawy:
@MultisampledNight Okay, I integrated edging and hinting into `guifont`. It works at runtime. Also, a few tests have been added to cover `FontOptions`.

MultisampledNight (commented):
Almost perfect! Small things left, but otherwise I think we can get this going. Format seemed a bit odd at first due to the "special-seeming" `#` prefix, but I realized that's only for avoiding ambiguity with `hXX`.

MultisampledNight:
Fwiw, the build issued a few warnings about a few unused imports as well:

```
warning: unused import: `error`
 --> src/renderer/fonts/font_loader.rs:3:11
  |
3 | use log::{error, trace};
  |           ^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused import: `nvim_rs::Value`
 --> src/renderer/fonts/font_loader.rs:5:5
  |
5 | use nvim_rs::Value;
  |     ^^^^^^^^^^^^^^

warning: `neovide` (bin "neovide") generated 2 warnings
```

MultisampledNight:
(for a few minutes I was confused on whether the changes had actually any effect when using them, until I realized I'm running a HiDPI display so the changes were barely visible, heh)

website/docs/configuration.md
```diff
@@ -42,39 +42,20 @@ as such it's also documented in `:h guifont`. But to sum it up and also add Neov
     - `hX` — Sets the font size to `X` points, while `X` can be any (even floating-point) number.
     - `b` — Sets the font **bold**.
     - `i` — Sets the font _italic_.
+    - `#e-X` — Sets edge pixels draw opaque or with partial transparency, while `X` is a type of edging:
```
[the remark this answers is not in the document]
Nawy: True - updated

src/renderer/fonts/font_loader.rs
```diff
@@ -51,11 +46,36 @@ impl PartialEq for FontPair {
...
+        edging: FontEdging,
+    ) -> Self {
+        Self {
+            bold,
+            italic,
+            family_name,
+            hinting,
+            edging,
+        }
+    }
```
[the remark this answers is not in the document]
Nawy: Updated. I kinda agree with your point. Just wanted to help with these comments. But if it's just a data struct - fine.

Nawy:
image

Nawy:
Fixed the line size.

Nawy:
Awesome! Now it's ready to merge. Thanks for your support @MultisampledNight

MultisampledNight:
Chooooooo Chooooooo! Thank you for the patch!

oredaze:
I have tried every combination with every font setting and it does not seem to remove the bold-like font rendering.

PR #1612, opened 2022-10-22
Appimage build

ananthb:
- Other: Appimage builds for Linux

- No

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,6 +9,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  VERSION: 0.10.1
 
 jobs:
   build-windows:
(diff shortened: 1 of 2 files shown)


MultisampledNight:
Sorry for being silent on this one for so long!

Having Neovide available in another packaging format for end-users to choose from would be awesome, but I'm not familiar with the process of packaging AppImages _at all_. So I don't really know if linuxdeploy is a commonly used tool or not, and if using an alpha release of that is normal either. However [the official appimage documentation](https://docs.appimage.org/packaging-guide/from-source/native-binaries.html) mentions it, so I guess it's pretty standard.

Other than that, this seems pretty good to merge, I still want to take a peek around the linuxdeploy source code though.

ananthb:
[AppDir](https://docs.appimage.org/reference/appdir.html)  is a format that describes a structured directory that contains apps. Appimages are built out of AppDirs.

The manual packaging guide walks you through how an appdir is typically structured: https://docs.appimage.org/packaging-guide/manual.html

linuxdeploy does make this much simpler by not needing to care about the AppDir too much.

MultisampledNight:
Looks good to me, thank you!

PR #1616, opened 2022-10-28
Fix default window geometry to use the last saved value (broken in #1583)

jdrouhard:
#1583 caused the last saved window size to no longer be read or used at startup. The `parse_window_geometry()` function is now only used by the parser for the `Dimensions` struct, which is only used if `--geometry` is passed. If that's the case, though, it'll simply use the parsed command line option instead of using the last saved size.

This PR refactors the `parse_window_geometry()` function into two separate pieces--
1. A new function `last_window_geometry()` has been added that attempts to use the last saved window geometry but falls back to the default. This function is used as the default value expression for the `geometry` command line argument/setting.
2. `parse_window_geometry()` now _only_ parses a command line geometry value (like `100x50`). The diff for this looks worse than it actually is. It's now simply the previous map function that was being applied to the input if present, unchanged.

This also adds a `Default` trait implementation for `Dimensions` which returns the default window geometry.

- Fix
- Refactor

- No

### DIFF
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -24,7 +24,7 @@ pub struct CmdLineSettings {
     pub neovim_args: Vec<String>,
 
     /// The geometry of the window
-    #[arg(long, default_value_t = DEFAULT_WINDOW_GEOMETRY)]
+    #[arg(long, default_value_t = last_window_geometry())]
     pub geometry: Dimensions,
 
     /// If to enable logging to a file in the current directory
@@ -41,7 +41,7 @@ pub struct CmdLineSettings {
 
     /// Which window decorations to use (do note that the window might not be resizable
     /// if this is "none")
-    #[arg(long, env = "NEOVIDE_FRAME", default_value_t = Frame::default())]
+    #[arg(long, env = "NEOVIDE_FRAME", default_value_t) ]
     pub frame: Frame,
 
     /// Maximize the window on startup (not equivalent to fullscreen)
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -41,7 +41,7 @@ pub struct CmdLineSettings {
 
     /// Which window decorations to use (do note that the window might not be resizable
     /// if this is "none")
-    #[arg(long, env = "NEOVIDE_FRAME", default_value_t) ]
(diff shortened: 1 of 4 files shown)


src/cmd_line.rs
```diff
@@ -41,7 +41,7 @@ pub struct CmdLineSettings {
 
     /// Which window decorations to use (do note that the window might not be resizable
     /// if this is "none")
-    #[arg(long, env = "NEOVIDE_FRAME", default_value_t = Frame::default())]
+    #[arg(long, env = "NEOVIDE_FRAME", default_value_t) ]
```
MultisampledNight, line 44: Was this intended? It's not really related to the window geometry. Though still, fine!

MultisampledNight (commented):
Haven't tested yet, but code looks good to me (except one partly unrelated thing)! Thank you! Feel free to ping me the few days should I not get back on this.

src/cmd_line.rs
jdrouhard: Unrelated, but I noticed it was unnecessary when I was fixing the default for geometry. I can take this out if you prefer it in a separate commit.

MultisampledNight: It's fine. I'm just not sure what clap does when you don't give a value, also the space (` `) after after the parenthesis annoys my inner eye for no reason.

jdrouhard: I’ll fix the space. `default_value_t` with no expression uses the Default trait for the type so you don’t have to explicitly say `= Type::default()`

MultisampledNight: Yup, clap just seems to use the `Default` impl if not provided. I'll fix the formatting for you really quick.

MultisampledNight: ...oh whew, my GH tab didn't update so I didn't see your message in before, sorry.

jdrouhard: No worries!

jdrouhard:
Ping @MultisampledNight. Checks look good, now!

MultisampledNight:
pong /s

And it works _great_! Tested it under i3 with `for_window [class="neovide"] floating enable. Thank you! I'll merge this now.

PR #1633, opened 2022-11-08
Only escape commandline arguments on macos targets

Jesse-Bakker:
- Fix

- No

Escaping commands breaks passed commands containing multiple words, which are often used like `nvim -c 'echo "test"'`. Command arguments are passed to TokioCommand, which does not need escaping.

### DIFF
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -51,7 +51,7 @@ fn build_nvim_cmd() -> TokioCommand {
 fn create_platform_shell_command(command: &str, args: &[&str]) -> Option<StdCommand> {
     if cfg!(target_os = "windows") && SETTINGS.get::<CmdLineSettings>().wsl {
         let mut result = StdCommand::new("wsl");
-        result.args(&["$SHELL", "-lc"]);
+        result.args(["$SHELL", "-lc"]);
         result.arg(format!("{} {}", command, args.join(" ")));
 
         Some(result)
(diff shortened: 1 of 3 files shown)


MultisampledNight:
Thank you! If arguments passed to `TokioCommand` don't need to be escaped though, then I'm confused to what has been the cause of #1399 — which this change would likely regress to.

Jesse-Bakker:
I'll look into it. I *think* I tried file names with spaces, but I'm not sure

Jesse-Bakker:
Ah, on macOS, the command is synthesized by passing it as a string to the shell, breaking unescaped file paths, whereas on linux (which I use), the neovim command is created directly using `TokioCommand`, where shlex breaks multi-word commands. I moved the shell escaping to where it's obvious why it's needed and made the dependency only compile on macos.

Shell escaping used to only be used on macos, but this was inadvertently changed when the command line handling was cleaned up to use clap derive.

src/bridge/command.rs
```diff
@@ -102,23 +102,34 @@ fn platform_which(bin: &str) -> Option<String> {
     }
 }
 
-fn build_nvim_cmd_with_args(bin: &str) -> TokioCommand {
-    let mut args = vec!["--embed".to_string()];
-    args.extend(SETTINGS.get::<CmdLineSettings>().neovim_args);
-    let args_str = args.join(" ");
+#[cfg(target_os = "macos")]
+fn build_nvim_comd_with_args_platform(bin: &str, args: &[&str]) -> TokioCommand {
```
MultisampledNight, line 106: The name for this function and its companion is slightly too long. Maybe `nvim_cmd_impl` would be better?

MultisampledNight (commented):
Apart from that, I think we're good to go.

MultisampledNight:
Awesome!

MultisampledNight:
Lol, the CI failed to build the macOS `#[cfg]`s, guess I'll revert this for now.

PR #1651, opened 2022-11-22
Only escape commandline arguments on macOS take 2

Jesse-Bakker:
Fixed the signature of the macos version

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,6 @@ rand = "0.8.5"
 rmpv = "1.0.0"
 serde = { version = "1.0.136", features = ["derive"] }
 serde_json = "1.0.79"
-shlex = "1.1.0"
 swash = "0.1.4"
 tokio = { version = "1.17.0", features = ["full"] }
 tokio-util = { version = "0.7.1", features = ["compat"] }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -68,6 +67,7 @@ version = "0.52.0"
 [target.'cfg(target_os = "macos")'.dependencies]
 cocoa = "0.24.0"
 objc = "0.2.7"
+shlex = "1.1.0"
```
williamspatrick, line 70: nit: appears to be a DOS-style new line here?

Locally I see this.
```
+shlex = "1.1.0"^M
```

williamspatrick (approved):
I tested this commit and it fixes #1659

Cargo.toml
Jesse-Bakker: Weird, because I'm on linux and I don't see it

Jesse-Bakker: I see it now. The whole file seems to have DOS line endings when I open it with `set ffs=unix` :frowning:, but the diff only shows an added one on that line. I have removed it from that line.

MultisampledNight:
Thank you!

Issue #1634, opened 2022-11-10
Poor performance after update to 0.10.3


fecet:
The scroll become very slow after update to 0.10.3, I also tried 0.10.2, the speed looks the same.

0.10.1 works well before.

MultisampledNight:
Thanks for reporting! I haven't been able to experience that myself, but my system is also pretty chonk-y.

I believe the performance hit came from the recent smooth scrolling fixes, could you both try [the commit before that fix](https://github.com/neovide/neovide/actions/runs/2897737596) and [the commit after that fix](https://github.com/neovide/neovide/actions/runs/2897828197)?

fecet:
I modify PKGBUILD from https://aur.archlinux.org/neovide-git.git, both commits work fine.

MultisampledNight:
So the hit didn't come from there. Since you seem to have access to the repo locally and can build from source, would it be possible for you to `git bisect` from 0.10.1 to HEAD to find out which commit introduced the hit?

fecet:
OK I will check it later

fecet:
Here is what I found
```
❯ git bisect bad
d253648e8a97f1e01c3fef0c80a88651f344d880 is the first bad commit
commit d253648e8a97f1e01c3fef0c80a88651f344d880
Author: multisn8 <contact@multisamplednight.com>
Date:   Sat Oct 8 13:37:12 2022 +0200

    Make vsync configurable (#1523)

    Tries to request vsync first, and retries window creation without if it fails.

 src/cmd_line.rs                        |  4 ++++
 src/window/mod.rs                      | 25 +++++++++++++++++++++----
 website/docs/command-line-reference.md |  8 ++++++++
 3 files changed, 33 insertions(+), 4 deletions(-)
```

MultisampledNight:
Thank you! This is very interesting. I believe Neovide's sub-par frame pacing interferes with the vsync requested on the window.

For now, you can work around it by passing `--novsync` to Neovide, I'll investigate if it's worth removing it later on. Maybe after #1312 enabling vsync has real effect.

fecet:
I note #1312 mention something about frame rate, so I investigate it further, I have set 
```
let g:neovide_refresh_rate = 138
```
in init, if I set it to 60, neovide works fine even without `novsync`. If I do not change, 
anything, i.e., refresh_rate=138 and start neovide as usual, the profiler fixed at 60fps,
with `novsync`, the profiler is 138fps. 

[...]

Issue #1312, opened 2022-05-14
Extremely poor performance (depending on the window size)


akim-13:
**Bug:**
The bigger the size of the window is (after a certain point), the more unbearable the performance becomes. Full screen is practically unusable.

**To Reproduce:**
1. Open neovide
2. Make it full screen (or make the window >~80% of the screen size)
3. Move the cursor

**Expected behavior:**
Stable performance at 60 fps as it is for smaller window sizes. 

**Demonstration:**
https://user-images.githubusercontent.com/77539913/168424950-0a9273a9-32bd-45ba-8e49-54e2d618206f.mp4

**Desktop:**
 - OS: Arch linux 5.17.5-arch1-1
 - Specs: Intel i7-10750H, NVIDIA GeForce GTX 1650 Ti Mobile, **3840x2400** *(I think this issue might have something to do with 4k resolution)*
 - Neovide Version 0.9.0
 - Neovim Version 0.7.0

**Log:**
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/8692768/neovide_rCURRENT.log)

**Additional context**
[Neovide (and Neovim) config](https://github.com/akim-13/neovim/blob/master/lua/rc/neovide.lua)

akim-13:
My suspicion about this issue being related to a 4k resolution has been confirmed by [this reply](https://github.com/neovide/neovide/issues/1299#issuecomment-1127603519).

vabenil:
I think I have this same issue. When I run neovide with --multigrid in fullscreen I usually get 60 FPS but when scrolling the FPS can go all the way down to the the 20s, if I reduce the size of the window as you mentioned in your post I get full 60 FPS all the time. When running other neovim guis, such as [nvui](https://github.com/rohit-px2/nvui) I get more than OK performance.

Setup:
- OS: Arch Linux 5.15.54-1-lts
- CPU: Intel(R) Core(TM) i5-5200U CPU @ 2.20GHz
- GPU: Intel Corporation HD Graphics 5500 (integrated graphics)

kirillrogovoy:
I have the same issue on a maxed-out 2019 MacBook Pro 13".

MacOS 12.3.1

hatemets:
Same issue with a 27" QHD screen running Arch Linux. Running the window with the size comprising 1/3 of the screen runs smoothly, anything above that starts to become laggy.

ldelossa:
Is this only present with "--multigrid" ?

[...]

PR #1652, opened 2022-11-22
Add documentation about #1634

fecet:
- Documentation

- No

I investigate it further and found that's may because X11 don't support mixed refresh rate monitors well, and that's the reason in my case, I demonstrate it in [`troubeshotting.md`](https://github.com/fecet/neovide/blob/main/website/docs/troubleshooting.md).

This should work for [#1634](https://github.com/neovide/neovide/issues/1634), and maybe for [#1312](https://github.com/neovide/neovide/issues/1312)

### DIFF
--- a/website/docs/troubleshooting.md
+++ b/website/docs/troubleshooting.md
@@ -24,7 +24,7 @@
 - If you recieve errors complaining about DRI3 settings, please reference issue
   [#44](https://github.com/neovide/neovide/issues/44#issuecomment-578618052).
 
-- If your scroling is very stutter
+- If your scrolling is very stutter
 
     - Add flag `--novsync` before startup as a quickfix.
     - Check if the value of `g:neovide_refresh_rate` and the refresh rate of your monitor are matched.
--- a/website/docs/troubleshooting.md
+++ b/website/docs/troubleshooting.md
@@ -26,12 +26,16 @@
 
 - If your scroling is very stutter
 
-    - Add flag "--novsync" before startup as a quickfix.
+    - Add flag `--novsync` before startup as a quickfix.
     - Check if the value of `g:neovide_refresh_rate` and the refresh rate of your monitor are matched.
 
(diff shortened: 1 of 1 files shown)


website/docs/troubleshooting.md
```diff
@@ -23,3 +23,15 @@
...
   [#44](https://github.com/neovide/neovide/issues/44#issuecomment-578618052).
+
+- If your scroling is very stutter
+
+    - Add flag "--novsync" before startup as a quickfix.
+    - Check if the value of `g:neovide_refresh_rate` and the refresh rate of your monitor are matched.
+
+    If your `g:neovide_refresh_rate` is correct, then check if you are using dual monitors with mixed refresh rate, say `144` and `60`, if so, that's because [PSA: X11 does support mixed refresh rate monitors!](https://www.reddit.com/r/linux/comments/yaatyo/psa_x11_does_support_mixed_refresh_rate_monitors/) and X11 force it to the lower one, you can set
+    ```
+    export __GL_SYNC_DISPLAY_DEVICE=DP-0
```
MultisampledNight, line 34: I think `__GL_SYNC_DISPLAY_DEVICE` is an [implementation detail of the Nvidia driver](https://download.nvidia.com/solaris/470.63.01/README/openglenvvariables.html), and `DP-0` suggests that your monitor is connected over DisplayPort, so will likely not work on every setup. Maybe this could be made clearer by mentioning that

- this is an Nvidia-specific environment variable
- the user should check the output of `xrandr` if they're on an X11 system

website/docs/troubleshooting.md
```diff
@@ -23,3 +23,15 @@
 
 - If you recieve errors complaining about DRI3 settings, please reference issue
   [#44](https://github.com/neovide/neovide/issues/44#issuecomment-578618052).
+
+- If your scroling is very stutter
+
+    - Add flag "--novsync" before startup as a quickfix.
```
MultisampledNight, line 29: Lately I've seen some users use ligatures in the browser, maybe having `--novsync` in a codeblock as well isn't too bad.

MultisampledNight (commented):
Otherwise that's awesome, thank you!

fecet:
I do some modification, how about this

shaunsingh:
Quick note that `scrolling` is misspelled, otherwise lgtm

PR #1660, opened 2022-12-01
Allow use of fonts containing `_` in the name

GuiAmPm:
The documentation encourages the use of `\` to escape spaces, and there's no way to escape `_` (underline), in the current main branch, it's not possible to load fonts containing it in the name. 

- Fix

- No

### DIFF
--- a/src/renderer/fonts/font_options.rs
+++ b/src/renderer/fonts/font_options.rs
@@ -27,7 +27,7 @@ impl FontOptions {
             let parsed_font_list: Vec<String> = parts
                 .split(',')
                 .filter(|fallback| !fallback.is_empty())
-                .map(|fallback| parse_font_name(fallback))
+                .map(parse_font_name)
                 .collect();
 
             if !parsed_font_list.is_empty() {
--- a/src/renderer/fonts/font_options.rs
+++ b/src/renderer/fonts/font_options.rs
@@ -1,3 +1,5 @@
+use itertools::Itertools;
+
 const DEFAULT_FONT_SIZE: f32 = 14.0;
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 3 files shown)


GuiAmPm:
Hey!
I missed that part of the documentation, sorry!

I will review my code and do a new PR sometime this weekend.

MultisampledNight:
No worries, feel free to do it anytime you want (or also not at all)!

GuiAmPm:
@MultisampledNight 
I'm glad I can help with this amazing tool.
----------

Now this supports backslash to escape characters.
Since nvim already escapes the string before we are able to process, we need to have a escaped backslash in the config file:
`set guifont=BigBlue\\_TerminalPlus\ Nerd_Font:h10`

for example, the string above is parsed correctly to
`BigBlue_TerminalPlus Nerd Font`

wdyt?

src/renderer/fonts/font_options.rs
```diff
@@ -277,4 +300,43 @@ mod tests {
             FontHinting::Slight,
         );
     }
+
+    #[test]
+    fn test_parse_font_name_with_escapes() {
+        let without_escapes_or_specials_chars = parse_font_name("Fira Code Mono");
+        let without_escapes = parse_font_name("Fira_Code_Mono");
+        let with_escapes = parse_font_name("Fira\\_Code\\_Mono");
```
MultisampledNight, line 308: For readability, you might want to use raw-strings:

```suggestion
        let with_escapes = parse_font_name(r"Fira\_Code\_Mono");
```

src/renderer/fonts/font_options.rs
```diff
@@ -95,6 +95,29 @@ impl PartialEq for FontOptions {
...
+            '_' => Some(' '),
+            _ => Some(current_char),
+        };
+
+        if let Some(result_char) = result_char {
+            parsed_font_name.push(result_char);
+        } else {
+            break;
+        }
+    }
```
MultisampledNight, line 116: This loop is written in a slightly confusing way, imo. Instead of pushing char for char to the string, I'd guess declaring `itertools` as an explicit dependency (which is already in our dep tree anyways, see `cargo tree --invert itertools`) and using `Itertools::batching` would be worth a thought:

```suggestion
    let mut parsed_font_name = char_iter
        .chars()
        .batching(|iter| match iter.next()? {
            '\\' => iter.next(),
            '_' => Some(' '),
            ch => Some(ch),
        })
        .collect();
```

GuiAmPm:
Thanks for the input!
The suggested fixes are applied.

PR #1671, opened 2022-12-12
[Feature] Support Skia Shaders for Cool Effects?

NateAGeek:
This is not meant to be merge, but just a starting spot for the shader feature.

- Feature

- The FPS tanks bc it is compiling a shader on every frame I assume...

### DIFF
--- a/src/renderer/rendered_window.rs
+++ b/src/renderer/rendered_window.rs
@@ -5,7 +5,7 @@ use skia_safe::{
     gpu::SurfaceOrigin,
     image_filters::blur,
     BlendMode, Budgeted, Canvas, Color, Image, ImageInfo, Paint, Point, Rect, SamplingOptions,
-    Surface, SurfaceProps, SurfacePropsFlags,
+    Surface, SurfaceProps, SurfacePropsFlags, FilterMode, runtime_effect::{ChildPtr, Uniform, uniform}, Data,
 };
 
 use crate::{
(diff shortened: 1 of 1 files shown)


Dandelight:
Very nice, really cool. But do you have a way to turn this effect off? Or adjust this effect. As a passer-by, I'm looking forward to seeing it in the main branch, but I think we should make it configurable before merging.

image

A-Lamia:
It's a shader so you would need to modify the shader file.

NateAGeek:
Yeah, so reviewing on the discord community. I need compile the shader on start, check if platform supports it, add in a configurable system to update uniforms, and also expose a setting to load the shader from a config var. Just need to find the time to hack something together :D

Dandelight:
> Yeah, so reviewing on the discord community. I need compile the shader on start, check if platform supports it, add in a configurable system to update uniforms, and also expose a setting to load the shader from a config var. Just need to find the time to hack something together :D

Lots of work to be done! Feel free to ask the community for help! Many thanks to your innovative idea.

Kethku:
Unfortunately I dont think this is going to get merged. We have two attempts at swapping to wgpu for our rendering and dropping skia altogether. One way or another, our rendering system will change significantly once those merge, so this PR won't work for that.

I'm still working on my attempt at switching to wgpu which is making steady progress (details here: https://kaylees.dev/trio/maple/day16-paths/). If you would like I can try to ping you once that work has landed to get another version of your effects integrated

PR #1681, opened 2022-12-17
Save backtrace to file on panic

Chloe-Woahie:
- Feature

- No

I added a panic hook that appends the backtrace to a log file in the current directory so that there is some clue as to what happened when neovide crashes. 

It works by attaching a panic hook at the beginning of `main()`. The original main function has been moved to `protected_main()`. I moved it here for now as I was not sure how to integrate the hook into `main()`'s documentation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,8 @@ unicode-segmentation = "1.9.0"
 which = "4.2.5"
 winit = { git = "https://github.com/neovide/winit", branch = "new-keyboard-all" }
 xdg = "2.4.1"
+chrono = "0.4.23"
+backtrace = "0.3.67"
 
 [dev-dependencies]
 mockall = "0.11.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,9 @@ embed-fonts = []
 [dependencies]
 copypasta = "0.8.1"
 async-trait = "0.1.53"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -47,6 +47,8 @@ unicode-segmentation = "1.9.0"
 which = "4.2.5"
 winit = { git = "https://github.com/neovide/winit", branch = "new-keyboard-all" }
 xdg = "2.4.1"
+chrono = "0.4.23"
+backtrace = "0.3.67"
```
MultisampledNight, line 51: Minor nit: It'd be nice to integrate these new two into the alphabetical order of the deps :)

MultisampledNight (commented):
Awesome, thank you! :tada:

I forgot to tell you this, and it doesn't seem like it's stated anywhere visible, so here I go: Should you contribute here or somewhere else, please consider creating another branch than `main` for your changes, and PR from that branch to this `main`. This makes local testing setups less messy, for a lengthier explanation see https://github.com/neovide/neovide/pull/1468#issuecomment-1215667467.

MultisampledNight:
Looks good to me! I'd leave it bit-rot a day or two, then get it in.

MultisampledNight:
Choooooo chooooooooooo! Thank you!

PR #1706, opened 2022-12-30
Remember windows size via pixel

### LINKED ISSUE #1636: Window size is increasing with every start of neovide
**Describe the bug**
When I start neovide.exe, close the window an restart it again, the window size increases with every new start.
I do not change the window size manually. You can see the problem in the log beneath:
At the top of the top the is size: Dimensions "{ width: 190, height: 44 } }" and at the bottom "size":{"width":232,"height":48}}}}".
After another start it is width: 232, height: 48 at the top and "width":283,"height":52 at the bottom.

**To Reproduce**
Set let g:neovide_remember_window_size = v:true in init.vim (If it is set to false window size will be set to 100 x 50 when quitting neovide)
Start and close the program (no matter if you close by closing the window, using :quit or shortcut ZZ) several times.

**Expected behavior**
The size of the window should stay the same at every start.

 - OS: Windows 10
 
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/9997854/neovide_rCURRENT.log)
- Neovide Version 0.10.3
 - Neovim Version 0.9.0

File too big, see attachment

chomosuke:
- Fix
  - closes #1636.
- Feature
  - added `--size` cmd option.

- No


A-Lamia:
Initial opens very small, but after that works as intended.

chomosuke:
> On first start up window opens in minimal possible size, after that everything works as intended.

This is expected, it's because it's reading the last saved character size and interpreting that as pixel size.

A-Lamia:
Then everything is perfect been using it for a day with no issues :)

fredizzimo:
Can you make sure that everything still works properly when the size has been stored as pixels, and then going to an older version of Neovide? There's this report https://github.com/neovide/neovide/issues/1788, and I'm fairly sure it's caused by this MR.

Using a different name for the setting would probably be sufficient.

chomosuke:
@fredizzimo done and tested :)

chomosuke:
Is there anything that prevents this PR from being merged? It's been sitting here for a while now.

A-Lamia:
as some one who has been using this patch for over 2 months on windows at least it is very solid.

chomosuke:
> as some one who has been using this patch for over 2 months on windows at least it is very solid.

Same, I've been using this on Linux and haven't encountered a single problem so far.

src/settings/window_size.rs
```diff
@@ -1,5 +1,6 @@
 use std::path::PathBuf;
 
+use glutin::dpi::PhysicalSize;
```
fredizzimo, line 3: This sould be `winit::dpi::PyshicalSize`, since already changed all the other references. `winit::dpi::PhysicalPosition` is also used just below this.

fredizzimo (approved):
Except for the tiny comment about a use statement, everything looks good. 

I have also been annoyed with the resizing myself, so I would personally really appreciate if this got merged.

src/settings/window_size.rs
chomosuke: Fixed :)

danielb2:
@MultisampledNight 
@last-partizan hi is there any reason this isn't being reviewed and merged?

last-partizan:
I would really prefer to have original `--geometry` unchanged, and store only window size in pixels. Is that possible?

last-partizan:
Oh, i missed that. I was thinking it was replaced. Looks good to me, than.

@Kethku what do you think, should we merge this?

Kethku:
> @MultisampledNight @last-partizan hi is there any reason this isn't being reviewed and merged?

https://neovide.dev/maintainer-cookbook.html#how-to-keep-your-sanity multisampled wrote this out and I think captures how us maintainers have needed to work on the project. We all have lives and jobs outside of it, and if we tried to track and respond to everything, we would burn out. At least thats how I personally see it.

I personally haven't reviewed this yet because it wasn't clear to me if this is the better solution over storing a grid size. I think the growing window problem likely comes from a disconnect between the stored size which is maybe an internal window size without the frame, and the set size which maybe includes the frame. Or visa versa. Storing a pixel size might be worse if the window was stored on display with a given scaling factor and then restored in a window with a different factor? Idk.

None of above should be a blocker, it was just some thoughts I had that weren't fully investigated. That lack of clarity on the "correct" solution stopped me personally from checking it off or reviewing.

I don't have a strong feeling either way though and this likely improves things so maybe its worth merging. @chomosuke do you have some thoughts on what I've written above? If you're confident I'm over worried Im happy to merge. I appreciate the persistence <3

chomosuke:
No worries @Kethku, I totally understand the maintaining a project like this is hard work and I thank you for everything you've already done for neovide!

For the root cause of growing window I believe that it is caused by font size changing right after nvim starts, there's some further info on a [comment](https://github.com/neovide/neovide/issues/1636#issuecomment-1366382214) I made a while ago and some evidence to support my beliefs. Overall storing the size in pixel has fixed the issue for me :).

I'm not sure about changing scaling factor though, I haven't tested that but regardless I think this PR is still an improvement that fixes the changing font issue and I do think it is worth merging.

Btw, if you think having more maintainer for this project will help, I am happy to be another maintainer. I have a few improvements to neovide in my mind and once I've made them I'd be more well versed in the code base and reviewing PR etc.

Kethku:
Sold! You're probably right about the actual core issue as I had forgotten about the timing issue between reading the config and first window render. Thanks for the attention to detail. Maybe come hang out in the discord? That's where most of the maintainers are atm

YoungHaKim7:
- You can specify the window size when you run it.

```bash
neovide . --size 2930x1300

```

Issue #1376, opened 2022-06-30
os.execute() of lua will crash


MontiL:
**Describe the bug**
in command mode:
: lua os.execute("ls") <-- will crash while executing any command

**To Reproduce**
1. Go to 'command mode'
2. : lua os.execute("ls") or any os commands
3. crash

**Expected behavior**
execute without crash

**Desktop (please complete the following information):**
 - OS: macOS 12.4 M1
 - Neovide Version: 0.8.0
 - Neovim Version: : 0.7.0

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:
[neovide_r2022-06-30_10-57-11.log](https://github.com/neovide/neovide/files/9016097/neovide_r2022-06-30_10-57-11.log)
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/9016098/neovide_rCURRENT.log)

MultisampledNight:
Can reproduce. Shuffling a bit around in `rr` I found this hidden error:

```rs
nvim_rs::error::DecodeError::InvalidMessage(nvim_rs::error::InvalidMessage::NotAnArray(rmpv::Value::Integer(rmpv::Integer {
          n: rmpv::IntPriv::PosInt(97)
        })))
```

The question is... why would Neovide expect an array...

MultisampledNight:
https://docs.rs/nvim-rs/0.4.0/nvim_rs/error/enum.InvalidMessage.html

> This should be very basically non-existent, since it would indicate a bug in neovim.

Hm. Wondering whether we should report this upstream.

starptr:
Took me a bit to figure out why neovide was crashing; maybe this bug should be documented in the [Troubleshooting page](https://neovide.dev/troubleshooting.html) :)

shaunsingh:
I think we should report this upstream rather than asking users to work around it. If its deemed a bug with nvim-rs/neovide instead of neovim itself we can certainly add it to the documentation and work on fixing it here

starptr:
I agree, but for the time being I think as a user of neovide it would have been nice to know about this behavior instead of having to bisect it again. I also don't know enough about neovim to confirm that the bug is upstream; I'm quoting here (it looks like there is a possibility that this is an nvim_rs issue instead? I am not sure).

So maybe the correct order of actions is
1. File a bug in neovim
2. Document this behavior in the neovide docs, and **link to the ticket** in the documentation

I will try to file the bug in neovim in a few weeks if no one else has filed it by then

[...]

PR #1720, opened 2023-01-04
Document the `os.execute()` invalid message upstream bug

starptr:
- Documentation

- No

This PR documents #1376, an upstream neovim bug that's been known for almost half an year.


MultisampledNight:
As newly mentioned in #1376, if it works with neovim-qt I'm not exactly sure if this is an upstream bug anymore. Thank you anyway! I'll just keep this open until we have any kind of hint.

last-partizan:
Looks like it's fixed already.

PR #1755, opened 2023-02-02
Renovate workflows

tobealive:
- Fix
- Feature
- Refactor
- CI

I gave an overhaul to the workflows. `build.yml` mainly.

It eliminates diagnostic warnings and its functionalities were slightly extended.
I grouped related parts and merged redundant sections using a matrix approach, allowing for testing with different Rust versions on each system. 
For now, with the update the tests and build run on stable + nightly, while deployment will use stable. 

If you prefer to have the current way updated
or if the other changes are helpful, please let me know. 

---

Edit: There is probably no way to trigger it for me here with this PR. Tests with act and on the fork are looking good though. https://github.com/tobealive/neovide/actions/runs/4079282571

- No


tobealive:
There came more improvements that could be made to my mind, so I changed it to a draft. I'll take a moment this night and implement then it should be ready.

tobealive:
I think they came out neat now.
https://github.com/tobealive/neovide/actions/runs/4089502504

Updates with the last commit:
- Better structure workflows into jobs that logically build on one another
- Improve OS conditions to better utilizes the matrix
- Use officials actions cargo cache due to better results while testing
- Remove unnecessary setups of stable toolchains for lint- and release
  frozen actions, as GitHub hosted runners have clippy and rustfmt pre-installed

Formatting changes:
- Generally format workflow files to conform to the dominant formatting in the
current workflows
- Consistent use of font-case (in app-related actions, website was mostly left as it is):  
  - Title case for workflow names
  - Sentence case for step names (uniform with GitHub's defaults in actions)
- Rename `clippy` to `lint-app` as it was not only running clippy
- Rename `lint` to `lint-website`

.github/workflows/lint-website.yml
```diff
@@ -0,0 +1,25 @@
+name: Lint Website
+
+on:
+  push:
+    paths:
+      - ".github/workflows/lint.yml"
+      - "website/**/*.md"
+  pull_request:
+    paths:
+      - ".github/workflows/lint.yml"
```
MultisampledNight, line 10: Leftover from the original file, should probably be

```suggestion
      - ".github/workflows/lint-website.yml"
      - "website/**/*.md"
  pull_request:
    paths:
      - ".github/workflows/lint-website.yml"
```

instead?

MultisampledNight (changes requested):
Literally awesome, thank you! Small nit about a copy-paste leftover, but otherwise LGTM!

.github/workflows/lint-website.yml
tobealive: Yes, good catch!

tobealive:
I'll also enable a stricter clippy for build linting then as it's warnings are resolved now.

MultisampledNight:
Hmm, I don't think so. For unexperienced Rust users this could sound like their code suddenly has _hard_ errors when pushed, while they are actually just warnings and the PR content itself should be done first (before addressing clippy's complaints).

tobealive:
SGTM.

If you also prefer this for the builds then lets revert the last commit. 
Else it might be ready 🙂.

MultisampledNight (approved):
Thank you! Choooo chooooo! :steam_locomotive:

MultisampledNight:
I'll take care of the CI failures on main, they're unrelated to this PR.

PR #1758, opened 2023-02-04
Preserve PATH when launching from terminal on MacOS

akateiva:
**Problem**: When neovide is launched from a terminal on MacOS, the $PATH of the launching environment is not preserved due to use of the `-l /--login` option for shell commands.

This causes problems for me, because when launching neovide from CLI with an activated conda python environment, I would like the PATH be preserved from my terminal session so that the pyright language server starts with the desired python interpreter.

**Solution:** This solution is based on j4qfrost's work from https://github.com/neovide/neovide/pull/309, but it seems like the functionality was lost somewhere along the way :D 

So in this PR, I propose using the `--login` option for launching shells if and only if we are launching not from a terminal.


MultisampledNight:
Looks good to me, thank you!

PR #1759, opened 2023-02-04
Resolve clippy warnings

tobealive:
- Codestyle
- Refactor

Fixes all the fluff generated by Clippy. Since there are so few warnings, imho it is best to just fix them, regardless of whether they are all important to neovide.

This will allow to extend the workflows to get more expressive reporting of clippy warnings and thus better detect shortcomings and improve overall code quality.

- No


MultisampledNight:
Thank you!

tobealive:
Thank you! I should have also compared with the CI warnings and make sure my clippy is up to date. 
It was only the warnings that was my local clippy(v0.1.65 at that time) was complaining about.

PR #1762, opened 2023-02-06
Refactor assignment of multiple default font options

tobealive:
- Refactor

I left it in two commits in case you prefer to pick the refactor as an destructured assignment in the former commit. 

- No


MultisampledNight:
Thank you!

Issue #481, opened 2021-02-28
Customizing line height


JoosepAlviste:
Hey!

I was wondering if there's a way to increase the line height (add more space between text lines). It looks quite cramped at the moment and it's usually possible to customize the line height in other apps.

I looked through the docs and `:h guifont`, but didn't find any line-height settings there. Is there a way to customize it in neovide?

If not, would it be difficult to add this?

p00f:
Is `linespace` what you're looking for?
https://neovim.io/doc/user/options.html#'linespace'
> 'linespace' 'lsp'    number    (default 0)
>             global
>             {only in the GUI}
>     Number of pixel lines inserted between characters.  Useful if the font
>     uses the full character cell height, making lines touch each other.
>     When non-zero there is room for underlining.
>     With some fonts there can be too much room between lines (to have
>     space for ascents and descents).  Then it makes sense to set
>     'linespace' to a negative value.  This may cause display problems
>     though!

Kethku:
I'm pretty sure this won't work with neovide today. We can probably make it work though!

rathod-sahaab:
This is easy I just modified font_height(in cacher I believe) and it worked fine. But, symbols don't look good.

I took a look at the kitty terminal's source and it renders symbols(nerd font) itself *i.e.* does not depends on the selected font to provide symbols. We have to do something similar to make it look like that.

Neovide with my hack:
![Screenshot from 2021-04-13 18-40-07](https://user-images.githubusercontent.com/31929374/114558335-32669100-9c88-11eb-879a-e95a3fe8454e.png)

Kitty:
![Screenshot from 2021-04-13 18-37-12](https://user-images.githubusercontent.com/31929374/114558342-34305480-9c88-11eb-8f5d-f28b3ee3b553.png)

It makes symbols of modified cell_height but keeps the font unmodified.

Decisions to take:
1. Should NeoVide render symbols by itself and not depend on the font to provide them?
2. Should we provide an option to use fractional option i.e. % of font height as spacing or utilize neovim's linespace attr.


My Opinion:
1. Yes
2. `g:neovide line_height=...%` if this is present use it otherwise use `linespace=...px`

Kethku:
I read this and will respond soon. I greatly appreciate you digging into this, its been on my list for a while

Kethku:
[...]

PR #1764, opened 2023-02-08
Add linespace support

tobealive:
- Feature

For me personally this tiny setting goes a long way regarding code readability.

Currently `0` is used by default, and the way I used neovide mostly until now was built from source with a static amount of extra space added. This PR now allows to use neovims linespacing setting for this. 

https://user-images.githubusercontent.com/34311583/217552260-7e29077a-ff0c-47ff-aa33-88cd0142d354.mov

Depending on the font configured and the value set for `linespace`, there will be spaces between line characters (as mentioned in #481). However, with a high value for `linespace` these spaces are what a user would also experience when using other clients that support the `linespace` setting, such as goneovim or nvui. Although this is a related problem, it should still be addressed separately.

- No


MultisampledNight:
...GH won't let me link #481 to have it close automatically when this is merged. I'll just mention it in the squash commit description then.

src/renderer/fonts/caching_shaper.rs
```diff
@@ -32,6 +32,7 @@ pub struct CachingShaper {
     shape_context: ShapeContext,
     scale_factor: f32,
     fudge_factor: f32,
+    linespace: u64,
```
MultisampledNight, line 35: According to `:h linespace`, it can also be negative, should likely be a `i64` instead:

> 	With some fonts there can be too much room between lines (to have space for ascents and descents).  Then it makes sense to set 'linespace' to a negative value.  This may cause display problems though!

```suggestion
    linespace: i64,
```

src/renderer/fonts/caching_shaper.rs
```diff
@@ -97,6 +99,26 @@ impl CachingShaper {
         }
     }
 
+    pub fn update_linespace(&mut self, linespace: &u64) {
+        debug!("Updating linespace: {}", linespace);
+
+        let font_key = FontKey {
+            italic: false,
+            bold: false,
+            family_name: self.options.primary_font(),
```
MultisampledNight, line 108: Hm... this'd mean that `linespace` only applies to the first font, not to any fallback fonts. But given the following scenario:

1. `set guifont=Hack,Roboto_Mono`
2. `set linespace=4`
3. `set linespace=Roboto_Mono,Hack`

then `Hack` would have the lineheight of `4`, but `Roboto_Mono` the default of `0`. Is that intended? (I did not test this, only a mindgame. Possible that I missed something. I should go test this, but I need to go in 7 minutes xD)

src/renderer/fonts/caching_shaper.rs
```diff
@@ -153,7 +175,8 @@ impl CachingShaper {
 
     pub fn font_base_dimensions(&mut self) -> (u64, u64) {
         let (metrics, glyph_advance) = self.info();
-        let font_height = (metrics.ascent + metrics.descent + metrics.leading).ceil() as u64;
+        let font_height =
+            (metrics.ascent + metrics.descent + metrics.leading).ceil() as u64 + self.linespace;
```
MultisampledNight, line 179: pong to include the i64 change here later on

MultisampledNight (changes requested):
Awesome! Minor nits.

src/renderer/fonts/caching_shaper.rs
tobealive: Good points. And this should definitely be tested! 

No worries. It can sit here as long as needed.

src/renderer/fonts/caching_shaper.rs
```diff
@@ -175,11 +175,26 @@ impl CachingShaper {
 
     pub fn font_base_dimensions(&mut self) -> (u64, u64) {
         let (metrics, glyph_advance) = self.info();
-        let font_height =
-            (metrics.ascent + metrics.descent + metrics.leading).ceil() as u64 + self.linespace;
         let font_width = (glyph_advance + 0.5).floor() as u64;
+        let mut font_height = (metrics.ascent + metrics.descent + metrics.leading).ceil() as i64;
 
-        (font_width, font_height)
+        let font_height_threshold = 10;
```
tobealive, line 181: We should have a threshold in case the `linespace` value is smaller than the actual font-height.

I set it to `10`. On my screen this results in what you see on the screenshot below when setting something like `linespace=-100`:
Screenshot_20230208_185525

I'm very open to other suggestions on how to solve this.

src/renderer/fonts/caching_shaper.rs
MultisampledNight: False alarm, can't seem to actually reproduce this.

src/renderer/fonts/caching_shaper.rs
MultisampledNight: Hm, could you explain your reasoning behind the threshold? I don't really see the correlation between `linespace` being smaller than the font height and the threshold.

aleksey-rowan:
Apologies since it's completely off topic, but @tobealive, what plugins do you use for the dashboard and command line popup window? Thanks!

src/renderer/fonts/caching_shaper.rs
MultisampledNight: I think instead of applying an opaque default, it might be more clear to ignore the `linespace` setting in that case and throw an error, telling that the `linespace` exceeds the font height. Probably something similar if a font hasn't been found would be a good option, too.

MultisampledNight:
I'll quickly rebase ontop of main and force push to have the clippy changes integrated.

MultisampledNight:
Rebase done, don't forget to pull before making any changes.

MultisampledNight (approved):

tobealive:
Thanks for all the work @MultisampledNight :muscle: 

LVGTM!

tobealive:
@aleksey-rowan 

- Dashboard is `alpha.nvim` 
- For the cmd line I've configured `wilder.nvim`. It serves very well.

I'll leave you a mention in case you are interested in getting the exact config. Wanted to set more stuff to public by next week or so.

MultisampledNight:
chooooooo chooooo :steam_locomotive:

xbladesub:
Do vertical lines display correctly with this PR? (Not dashed?)

MultisampledNight:
@xbladesub No, they aren't displayed correctly.

aleksey-rowan:
> I'll leave you a mention in case you are interested in getting the exact config. Wanted to set more stuff to public by next week or so.

@tobealive Absolutely, I would love to see the exact config.

Also, thanks a lot of the PR--line spacing was something that I missed dearly.

tobealive:
@aleksey-rowan 

sorry for taking a few days, it was a busy week.

this is the full wilder config I'm using to achieve that command palette with as-you-type completion.
https://github.com/tenxsoydev/nxvim/blob/main/lua/nxvim/plugins/wilder.lua

Though it's not the cleanest part of my config. But better that than nothing, I think.

tskj:
> @xbladesub No, they aren't displayed correctly.

I really love this linespacing feature, thank you all for working hard to implement it!

Sorry for replying to a year old PR, I was just wondering if there is an issue to track this problem of vertical dashed lines? Or more specifically I care mostly because it breaks my statusline separators, so I have to make the unfortunate choice between a cool powerline and higher linespacing (which I've just discovered I like very much).

nomasprime:
Is this available in config.toml? Can't seem to get it to work.

PR #1766, opened 2023-02-10
[Feature] Support mouse hover (emit mouse move event)

ryo33:
Calls [nvim_input_mouse()](https://neovim.io/doc/user/api.html#nvim_input_mouse()) with `button = "move"`.
Currently, it always emits the events, but IMO it should not emit the events if ["mousemoveevent" option](https://neovim.io/doc/user/options.html#'mousemoveevent') is not enabled. How should we achieve this behavior?

- [x] Emit the event
- [ ] Do not emit the event when "mousemoveevent" option is disabled
- [ ] Rerender if mouse moved even if not dragged (ignore `self.dragging.is_some()` if "mousemoveevent" option is enabled)

- Feature

- Yes, emits new event

### DIFF
--- a/src/window/mouse_manager.rs
+++ b/src/window/mouse_manager.rs
@@ -181,9 +181,12 @@ impl MouseManager {
                     position: self.drag_position.into(),
                     modifier_string: keyboard_manager.format_modifier_string(true),
                 }));
-            } else if has_moved {
+            } else {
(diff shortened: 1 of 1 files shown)


Kethku:
I'm going to merge this as the other remaining todos dont seem too important.

Issue #1211, opened 2022-02-12
How to change the font size if you're using init.lua (lua environment) and not .vimrc file


nishu-murmu:
I wanted to use Neovide but am not using standard .vimrc file.
my source file is a lua file.
Can we get a feature to edit fonts in lua environment.

[Neovide](https://github.com/nishu-murmu/dotfiles/blob/main/assets/neovide%20(1).png)

MultisampledNight:
The option Neovide uses for this is builtin in NeoVim, it's called `guifont`. So that has nothing to do with Neovide itself. According to `:h lua-vim-opt`, you could just use

```lua
vim.opt.guifont = { "Source Code Pro", "h12" }
```

where `Source Code Pro` sets the font family and `h12` sets the font size. See `:h 'guifont'` for more configuration options.

nishu-murmu:
I know about `guifont`
I have used it in my config in `settings.lua` file
**Here's the settings.lua file**
```
vim.cmd([[
let g:mapleader = " "

set nocompatible
syntax enable
filetype plugin indent on
set encoding=utf-8
set fileencoding=utf-8
set showcmd
set cmdheight=1
set scrolloff=10
set splitbelow
set splitright
set conceallevel=0
set title
set tabstop=2
set shiftwidth=2
set autoindent
set smarttab
set expandtab
set showtabline=2
set noshowmode
set nobackup
set nowritebackup
set updatetime=300
set timeoutlen=500
set clipboard=unnamedplus
set laststatus=0
set number
set ruler
set cursorline
set hidden
set nowrap
set backspace=start,eol,indent
set mouse=a
set lazyredraw
set wildignore+=*/node_modules/*
set termguicolors
set guifont=Hack:h12
set background=dark 
colorscheme onedark

au BufNewFile,BufRead *.es6 setf javascript
au BufNewFile,BufRead *.tsx setf typescriptreact
]])

```
Still the font size doesn't change

Then I read for fonts to work properly and not breaking, you need to use `guifont` in the init file so I also added it in `init.lua` file but still it isn't working

**Here's the init.lua file**
```
require("user.settings")
vim.opt.guifont = { "Source Code Pro", "h12" }
require("user.mappings")
require("user.plugins")
require("user.lsp-config")
require("user.lsp-saga")
require("user.auto-pairs")
require("user.Comment")
require("user.lualine")
require("user.treesitter")
require("user.telescope")
require("user.gitsigns")
require("user.nvim-tree")
[...]

PR #1772, opened 2023-02-18
Update configuration.md doc to reflect reality

kika:
https://github.com/neovide/neovide/issues/1211#issuecomment-1037048651 

the colon is required for the parser to pick up the option

- Documentation

- No

### DIFF
--- a/website/docs/configuration.md
+++ b/website/docs/configuration.md
@@ -41,7 +41,7 @@ set guifont=Source\ Code\ Pro:h14
 Lua:
 
 ```lua
-vim.opt.guifont = { "Source Code Pro:h14" } -- text below applies for VimScript
+vim.o.guifont = "Source Code Pro:h14" -- text below applies for VimScript
 ```
 
 Controls the font used by Neovide. Only setting which is actually controlled through an option, and
--- a/website/docs/configuration.md
+++ b/website/docs/configuration.md
@@ -41,7 +41,7 @@ set guifont=Source\ Code\ Pro:h14
 Lua:
 
 ```lua
-vim.opt.guifont = { "Source Code Pro", "h14" } -- text below applies for VimScript
+vim.opt.guifont = { "Source Code Pro", ":h14" } -- text below applies for VimScript
 ```
 
 Controls the font used by Neovide. Only setting which is actually controlled through an option, and
--- a/website/docs/configuration.md
+++ b/website/docs/configuration.md
@@ -41,7 +41,7 @@ set guifont=Source\ Code\ Pro:h14
 Lua:
 
 ```lua
-vim.opt.guifont = { "Source Code Pro", ":h14" } -- text below applies for VimScript
+vim.opt.guifont = { "Source Code Pro:h14" } -- text below applies for VimScript
 ```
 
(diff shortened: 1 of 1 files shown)


ad-chaos:
I think a better recommendation would be to suggest doing
```lua
vim.o.guifont = "Source Code Pro:h14"
```

last-partizan:
@ad-chaos i like your suggestion.

On the other hand, i think format `{ "Source Code Pro", "h14" }` should be perfectly valid, without colon.

And we should fix our code to accept this.

ad-chaos:
Hmm, in that case you should take that to neovim since they control `vim.opt` behaviour

last-partizan:
```rust
impl FontOptions {
    pub fn parse(guifont_setting: &str) -> FontOptions {
        let mut font_options = FontOptions::default();

        let mut parts = guifont_setting.split(':').filter(|part| !part.is_empty());
```

That's our code, we just expect `:` to be there, but we can make that optional.

ad-chaos:
If you run something like
```vim
lua vim.opt.guifont = {"Source Code Pro", "h14"}
se guifont?
```
You'll see that the way that table is parsed into `Source Code Pro,h14` that is `Source Code Pro` and `h14` are considered two different fonts. And the documentation also says that a `:` is used for options and an `_` can be a space. That's why we do all that parsing the way we do it.

We _could_ add our own bit of logic to check when it is an option and when it is a font, but what if someone wants to use a font named `h14`? We search for a font, if we don't find it then use it as an option. That's sounds a little complex and tricky though.

kika:
Help me understand, please: `vim.opt.guifont` is a list of fonts and the editor tries them in sequence until it finds the font that works and then stops. The rest of the list is not used. Is that right? What happened in my case and what lead me into frustration is that I understood that this list is comprised of either one or two components - font name and optional list of options separated by `:`. I didn't know that vim supports a list despite being a vi/vim user since 1989 :-) My .vimrc always had `guifont` option of exactly one font and I didn't even read the doc on this option once. 
Now, if that is true, why when I specified the "no-antialias" option, it was applied to the font? What happened when I put the following into the `init.lua`: `vim.opt.guifont = { "Fixedsys Excelsior", "h16:#e-alias:#h-none" }` is that `#e-alias` was applied to the font, but `h16` was not (because of the missing colon). And I was running in circles trying to figure out what's going on, until I found a discussion in the issues (mentioned in the PR) and put the colon there. 
According to the logic above, `:#e-alias` should apply to the non-existent font `h16`, right?

ad-chaos:
So @MultisampledNight cleared that up on discord. Fonts are a comma separated list of fallbacks followed by a colon with options that apply to all the fallbacks.

kika:
@ad-chaos so the behavior is "even if `guifont` is a list, it's treated as a single string, with all fonts from all list elements joined together with a comma and all options from all list elements joined together with a colon"? Is that correct?

If yes, super confusing I'd say. Why list then.

ad-chaos:
well, its not a list, its a string all along. `vim.opt` is extra magic that can accept lua tables and convert it to the appropriate format.

kika:
PR updated

MultisampledNight:
Alright, thank you!

PR #1777, opened 2023-02-22
Fix Lua instructions in Background Colour

clvnkhr:
The previous version says to use `vim.g.neovide_transparency_point` which gives an error. To match with the VimScript version, this should be `vim.g.transparency` instead (which does work).

- Documentation

- No


MultisampledNight:
Thank you!

PR #1778, opened 2023-02-22
Refine workflows

tobealive:
- Codestyle
- Refactor

- No

This PR contains refinements and enhancements that can be made to my recent workflow-related commits. It's things I noted down while doing DevOps work on similar GitHub actions afterwards.

result: https://github.com/tobealive/neovide/actions/runs/4275873794

### Changes

- Use cache prior to clippy, as this step is sped up as well if there is a cache hit (~4min vs 30s).
- Call reusable linting workflow instead of re-authoring the linting steps in the build action. This was already intended in the original renovation, which is why `workflow_call` was added as part of the `on` block.
- Update the artifact upload condition -> Check if there is a second artifact; instead of assuming there is none on macOS and others have one.
- Update test result publishing
  - Publishing on Mac was unbearably slow due to the Python setup (should take seconds, but took ~5 min). Adding a Python setup action would improve this, but I went with abstracting the test result publishing into a workflow. This is even faster.
  - It also groups the platform results and fixes adding test results as PR comments for forks. Until now, it only worked for internal but not for PRs from forks. [e.g. PR](https://github.com/tobealive/neovide/pull/3)
- Require a successful build test to start the snap workflow
  

<br>
(Since the `Upload Test Event` and `Test Results` jobs are workflow names and will be used in the job overview on the left side of the build workflow, naming is not fully uniform. I didn't want to be too german about it and didn't update the names for the whole workflow. Please leave a word if it's preferred. I'll quickly add this small commit.)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,16 +20,7 @@ defaults:
 
 jobs:
   lint:
-    runs-on: ubuntu-latest
-
-    steps:
-      - uses: actions/checkout@v3
-
-      - name: Run rustfmt
-        run: cargo fmt --all --check
-
-      - name: Run clippy
-        run: cargo clippy -- -D warnings
+    uses: ./.github/workflows/lint-app.yml
 
   test:
     needs: lint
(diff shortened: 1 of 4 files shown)


MultisampledNight:
Awesome, looks good to me!

PR #1787, opened 2023-03-05
Add profiling support through Tracy Profiler

fredizzimo:
This add profiling support through the [Tracy Profiler](https://github.com/wolfpld/tracy)

To use the profiler you can use the `profiling` or `gpu_profiling` feature, for example

`cargo run --features gpu_profiling`, just remember to launch Tracy first and wait for the connection. It's also helpful to add `--noidle` to better stress the system, and in some cases to disable vsync with `--novsync`

I have also added a few profiling events, that covers most of the basic frame loop. 

Below is a sample, with my screen refresh rate set to 60, matching the default 60 FPS that Neovide tries to maintain. On the surface it looks good, all frames are close to 16.66 ms. But if you look closely it's constantly missing the vsync (over 17ms), followed by a shorter frame to catch up. This will be fixed later, I just have to do some final cleanup for that pull request. Also note that there's almost no load on my system, when it's showing a static screen, you have to zoom in to see something else than `swap_buffers`, which is just a wait.

Screenshot_20230305_020932

This pull request also adds a `--profiling` build profile, which is basically release with debug symbols. This helps a bit with the sampling profiler part of tracy, but I haven't really used that. I did use the profiler of Visual Studio though, which works properly if you use this build profile.

To run that you can do for example
`cargo run --profile profiling --features profiling -- --noidle`

NOTE: I decided to write my own integration, rather than the one provided by the `tracy-client` library, that one copies some strings, allocates and has extra overhead. I also wanted everything to get completely optimized out when the profiling is disabled.

Regarding the enabling of the  GPU profiling, for OpenGL, and this integration there's not much overhead, but I also have a Direct3D implementation, and for that there's some extra overhead, mostly because I can't modify Skia, so that's why it's good to be able to disable it, when you don't need the GPU part.

The Direct3D implementation will come in a later pull request, where I enable Direct3D support for Windows, that was required to get butter smooth frame rates there. The OpenGL support for Windows, especially in Windowed mode is sadly lacking, and it would just drop frames, for now apparent reason randomly, sometimes for almost 100ms.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,8 @@ members = ["neovide-derive"]
 [features]
 default = []
 embed-fonts = []
+profiling = ["dep:tracy-client-sys"]
+gpu_profiling = ["profiling"]
 
 [dependencies]
 async-trait = "0.1.53"
@@ -45,6 +47,7 @@ swash = "0.1.8"
 time = "0.3.9"
 tokio = { version = "1.25.0", features = ["full"] }
 tokio-util = { version = "0.7.4", features = ["compat"] }
+tracy-client-sys = { version = "0.19.0", optional = true }
 unicode-segmentation = "1.9.0"
 which = "4.2.5"
 winit = { git = "https://github.com/neovide/winit", branch = "new-keyboard-all" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,6 +82,11 @@ lto = true
 incremental = true
 strip = true
 
+[profile.profiling]
+inherits = "release"
+strip = false
+debug = true
+
 [package.metadata.bundle]
 name = "Neovide"
 identifier = "com.neovide.neovide"
--- a/src/profiling/profiling_enabled.rs
+++ b/src/profiling/profiling_enabled.rs
@@ -1,7 +1,6 @@
 use std::{
     cell::RefCell,
     ffi::CString,
-    mem,
     ptr::null,
     sync::atomic::{AtomicU8, Ordering},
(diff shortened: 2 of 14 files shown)


fredizzimo:
I made some clippy fixes for when the profiling is enabled.

src/profiling/profiling_enabled.rs
```diff
@@ -146,12 +145,29 @@ static CONTEXT_ID: AtomicU8 = AtomicU8::new(0);
...
 
 impl GpuCtx {
+    fn new() -> Self {
+        let len = 64 * 1024;
+        let mut query = Vec::with_capacity(len);
+        let remaining = query.spare_capacity_mut();
+        unsafe {
+            GenQueries(remaining.len() as i32, remaining.as_mut_ptr() as *mut u32);
+            query.set_len(len);
+        }
```
fredizzimo, line 161: I fixed a stack overflow that sometimes happens (depending on which optimizations the compiler does) with the statically sized array. It was first created on the stack and then moved to the struct.

Cargo.toml
```diff
@@ -49,6 +51,7 @@ unicode-segmentation = "1.9.0"
 which = "4.2.5"
 winit = { git = "https://github.com/neovide/winit", branch = "new-keyboard-all" }
 xdg = "2.4.1"
+tracy-client-sys = { version = "0.19.0", optional = true }
```
last-partizan, line 54: btw, dependencies is already sorted alphabetically 😉

Issue #1236, opened 2022-03-02
Neovide hanging in xorg, seems resolved by using neovide-x11-nt branch in glutin, winit


manfredlotz:
It happens that neovide hangs in xorg when switching to neovide in Gnome xorg with Super-<num> or in Cinnamon using some other keyboard shortcut. This is not happening in Wayland.

Now I rebuilt neovide by changing `Cargo.toml` like this:

```
glutin = { git = "https://github.com/neovide/glutin", branch = "neovide-x11-nt", features=["serde"] }
winit = { git = "https://github.com/neovide/winit", branch = "neovide-x11-nt" } 
```
and the hang situation is gone.

I thought it might be good to report back here.

shekohex:
Hey any updates on this? any why it is not fixed already?

fredizzimo:
Does this MR fix it https://github.com/neovide/neovide/pull/1789? 

Unfortunately Neovide relies on the new winit keyboard API, which has been work in progress for a long time. So we have not been able to merge in new winit versions. Now there's finally some light in the tunnel, that MR uses a a newer work in progress versions of the keyboard API based on a much newer version of winit. If it works well enough we will probably merge it rather soon, but the branch is not well enough tested yet.

In any case the new keyboard API is planned to be released with the next winit version, with a soft deadline of end of July beginning of August. So at least at that point it should be stable enough for us to switch over to it.

shekohex:
Sounds good, hopefully by that time it would be fixed.
For now, it hangs when I use Super key (while Neovide is focused) to switch to other applications or workspace and then go back to Neovide, it hangs and I have to just switch focus to other application and focus back to Neovide to get it to work again; Which, to be honest, is annoying and not convenient 😔

fredizzimo:
https://github.com/neovide/neovide/pull/1789 has been merged to master. Could you try with that?

Issue #907, opened 2021-08-20
Neovide crashes immediately when openning. | Windows 10


0xbalazstoth:
**Describe the bug**
Basically when I open neovide.exe, immediately crashes.

**Expected behavior**
To work and not crash.

**Screenshots**
![image](https://user-images.githubusercontent.com/12668764/130283486-cb04c40e-1104-403b-9244-f0b84fbfbcc5.png)

@rust_panic error
```asm
; id = {0x00008394}, range = [0x00000001406bab80-0x00000001406bac71), name="rust_panic"
; Source location: \rustc\a178d0322ce20e33eac124758e837cbd80a6f633\library\std\src\panicking.rs:667
7FF7D70FAB80: 48 81 EC C8 00 00 00       subq   $0xc8, %rsp
7FF7D70FAB87: 48 89 4C 24 28             movq   %rcx, 0x28(%rsp)
7FF7D70FAB8C: 48 89 54 24 30             movq   %rdx, 0x30(%rsp)
7FF7D70FAB91: 48 8D 4C 24 28             leaq   0x28(%rsp), %rcx
7FF7D70FAB96: E8 35 AD 00 00             callq  0x7ff7d71058d0  ; void panic_unwind::__rust_start_panic() at lib.rs:103
7FF7D70FAB9B: 89 44 24 24                movl   %eax, 0x24(%rsp)
7FF7D70FAB9F: 48 8D 44 24 24             leaq   0x24(%rsp), %rax
7FF7D70FABA4: 48 89 44 24 58             movq   %rax, 0x58(%rsp)
7FF7D70FABA9: 48 8D 05 50 F2 01 00       leaq   0x1f250(%rip), %rax  ; void core::fmt::num::imp::{{impl}}::fmt() at num.rs:279
7FF7D70FABB0: 48 89 44 24 60             movq   %rax, 0x60(%rsp)
7FF7D70FABB5: 48 8D 05 5C 26 CA 00       leaq   0xca265c(%rip), %rax
7FF7D70FABBC: 48 89 84 24 98 00 00 00    movq   %rax, 0x98(%rsp)
7FF7D70FABC4: 48 C7 84 24 A0 00 00 00 >  movq   $0x1, 0xa0(%rsp)
7FF7D70FABD0: 48 C7 84 24 A8 00 00 00 >  movq   $0x0, 0xa8(%rsp)
7FF7D70FABDC: 48 8D 44 24 58             leaq   0x58(%rsp), %rax
[...]

PR #1789, opened 2023-03-05
Update to use new upstream unified keyboard branch

Kethku:
- Fix

This may break ime support. Not sure, this needs to be tested.

### DIFF
--- a/src/renderer/opengl.rs
+++ b/src/renderer/opengl.rs
@@ -1,40 +1,94 @@
+use std::ffi::{c_void, CStr};
+use std::num::NonZeroU32;
+
 use crate::cmd_line::CmdLineSettings;
 
-use glutin::{ContextBuilder, GlProfile, PossiblyCurrent, WindowedContext};
+use glutin::{
+    config::{Config, ConfigTemplateBuilder},
+    context::{ContextAttributesBuilder, GlProfile, PossiblyCurrentContext},
+    display::GetGlDisplay,
+    prelude::*,
+    surface::{Surface, SurfaceAttributesBuilder, WindowSurface},
+};
+use glutin_winit::DisplayBuilder;
+use raw_window_handle::HasRawWindowHandle;
+use winit::{
+    event_loop::EventLoop,
+    window::{Window, WindowBuilder},
+};
 
-use winit::{event_loop::EventLoop, window::WindowBuilder};
+pub struct Context {
+    surface: Surface<WindowSurface>,
+    context: PossiblyCurrentContext,
+    window: Window,
+    config: Config,
+}
 
-pub type Context = WindowedContext<glutin::PossiblyCurrent>;
+impl Context {
+    pub fn window(&self) -> &Window {
+        &self.window
+    }
+    pub fn resize(&self, width: NonZeroU32, height: NonZeroU32) {
+        GlSurface::resize(&self.surface, &self.context, width, height)
+    }
+    pub fn swap_buffers(&self) -> glutin::error::Result<()> {
+        GlSurface::swap_buffers(&self.surface, &self.context)
+    }
+    pub fn get_proc_address(&self, addr: &CStr) -> *const c_void {
+        GlDisplay::get_proc_address(&self.surface.display(), addr)
+    }
+    pub fn get_config(&self) -> &Config {
+        &self.config
+    }
+}
+
+fn gen_config(mut config_iterator: Box<dyn Iterator<Item = Config> + '_>) -> Config {
+    config_iterator.next().unwrap()
+}
 
 pub fn build_context<TE>(
     cmd_line_settings: &CmdLineSettings,
     winit_window_builder: WindowBuilder,
     event_loop: &EventLoop<TE>,
-) -> WindowedContext<PossiblyCurrent> {
-    let builder = ContextBuilder::new()
-        .with_pixel_format(24, 8)
-        .with_stencil_buffer(8)
-        .with_gl_profile(GlProfile::Core)
-        .with_srgb(cmd_line_settings.srgb)
-        .with_vsync(cmd_line_settings.vsync);
+) -> Context {
+    let template_builder = ConfigTemplateBuilder::new()
+        .with_stencil_size(8)
+        .with_transparency(true);
+    let (window, config) = DisplayBuilder::new()
+        .with_window_builder(Some(winit_window_builder))
+        .build(event_loop, template_builder, gen_config)
+        .unwrap();
+    let window = window.expect("Could not create Window");
 
-    let ctx = match builder
-        .clone()
-        .build_windowed(winit_window_builder.clone(), event_loop)
-    {
-        Ok(ctx) => ctx,
-        Err(err) => {
-            // haven't found any sane way to actually match on the pattern rabbithole CreationError
-            // provides, so here goes nothing
-            if err.to_string().contains("vsync") {
-                builder
-                    .with_vsync(false)
-                    .build_windowed(winit_window_builder, event_loop)
-                    .unwrap()
-            } else {
-                panic!("{}", err);
-            }
-        }
-    };
-    unsafe { ctx.make_current().unwrap() }
+    let gl_display = config.display();
+    let raw_window_handle = window.raw_window_handle();
+
+    let size = window.inner_size();
+
+    let surface_attributes = SurfaceAttributesBuilder::<WindowSurface>::new()
+        .with_srgb(Some(cmd_line_settings.srgb))
+        .build(
+            raw_window_handle,
+            NonZeroU32::new(size.width.max(1)).unwrap(),
+            NonZeroU32::new(size.height.max(1)).unwrap(),
+        );
+    let surface =
+        unsafe { gl_display.create_window_surface(&config, &surface_attributes) }.unwrap();
+    let context_attributes = ContextAttributesBuilder::new()
+        .with_profile(GlProfile::Core)
+        .build(Some(raw_window_handle));
+    let context = unsafe {
+        gl_display
+            .create_context(&config, &context_attributes)
+            .unwrap()
+    }
+    .make_current(&surface)
+    .unwrap();
+
+    Context {
+        surface,
+        context,
+        window,
+        config,
+    }
 }
--- a/src/renderer/opengl.rs
+++ b/src/renderer/opengl.rs
@@ -3,6 +3,7 @@ use std::num::NonZeroU32;
 
 use crate::cmd_line::CmdLineSettings;
 
+use gl::MAX_RENDERBUFFER_SIZE;
 use glutin::{
     config::{Config, ConfigTemplateBuilder},
     context::{ContextAttributesBuilder, GlProfile, PossiblyCurrentContext},
@@ -12,6 +13,7 @@ use glutin::{
 };
 use glutin_winit::DisplayBuilder;
 use raw_window_handle::HasRawWindowHandle;
+use winit::dpi::PhysicalSize;
 use winit::{
     event_loop::EventLoop,
     window::{Window, WindowBuilder},
@@ -22,6 +24,13 @@ pub struct Context {
     context: PossiblyCurrentContext,
     window: Window,
     config: Config,
+}
+
+pub fn clamp_render_buffer_size(size: PhysicalSize<u32>) -> PhysicalSize<u32> {
+    PhysicalSize::new(
+        size.width.clamp(1, MAX_RENDERBUFFER_SIZE),
+        size.height.clamp(1, MAX_RENDERBUFFER_SIZE),
+    )
 }
 
 impl Context {
@@ -39,6 +48,10 @@ impl Context {
     }
     pub fn get_config(&self) -> &Config {
         &self.config
+    }
+
+    pub fn get_render_target_size(&self) -> PhysicalSize<u32> {
+        clamp_render_buffer_size(self.window.inner_size())
     }
 }
 
@@ -63,14 +76,14 @@ pub fn build_context<TE>(
     let gl_display = config.display();
     let raw_window_handle = window.raw_window_handle();
 
-    let size = window.inner_size();
+    let size = clamp_render_buffer_size(window.inner_size());
 
     let surface_attributes = SurfaceAttributesBuilder::<WindowSurface>::new()
         .with_srgb(Some(cmd_line_settings.srgb))
         .build(
             raw_window_handle,
-            NonZeroU32::new(size.width.max(1)).unwrap(),
-            NonZeroU32::new(size.height.max(1)).unwrap(),
+            NonZeroU32::new(size.width).unwrap(),
+            NonZeroU32::new(size.height).unwrap(),
         );
     let surface =
         unsafe { gl_display.create_window_surface(&config, &surface_attributes) }.unwrap();
--- a/src/renderer/opengl.rs
+++ b/src/renderer/opengl.rs
@@ -4,6 +4,7 @@ use std::num::NonZeroU32;
 use crate::cmd_line::CmdLineSettings;
 
 use gl::MAX_RENDERBUFFER_SIZE;
+use glutin::surface::SwapInterval;
 use glutin::{
     config::{Config, ConfigTemplateBuilder},
     context::{ContextAttributesBuilder, GlProfile, PossiblyCurrentContext},
@@ -87,6 +88,7 @@ pub fn build_context<TE>(
         );
     let surface = unsafe { gl_display.create_window_surface(&config, &surface_attributes) }
         .expect("Failed to create Windows Surface");
+
     let context_attributes = ContextAttributesBuilder::new()
         .with_profile(GlProfile::Core)
         .build(Some(raw_window_handle));
@@ -94,6 +96,13 @@ pub fn build_context<TE>(
         .expect("Failed to create OpenGL context")
         .make_current(&surface)
         .unwrap();
+
+    // NOTE: We don't care if these fails, the driver can override the SwapInterval in any case, so it needs to work in all cases
+    let _ = if cmd_line_settings.vsync {
+        surface.set_swap_interval(&context, SwapInterval::Wait(NonZeroU32::new(1).unwrap()))
+    } else {
+        surface.set_swap_interval(&context, SwapInterval::DontWait)
+    };
 
     Context {
         surface,
--- a/src/renderer/opengl.rs
+++ b/src/renderer/opengl.rs
@@ -70,7 +70,7 @@ pub fn build_context<TE>(
     let (window, config) = DisplayBuilder::new()
         .with_window_builder(Some(winit_window_builder))
         .build(event_loop, template_builder, gen_config)
-        .unwrap();
+        .expect("Failed to create Window");
     let window = window.expect("Could not create Window");
 
     let gl_display = config.display();
@@ -85,18 +85,15 @@ pub fn build_context<TE>(
             NonZeroU32::new(size.width).unwrap(),
             NonZeroU32::new(size.height).unwrap(),
         );
-    let surface =
-        unsafe { gl_display.create_window_surface(&config, &surface_attributes) }.unwrap();
+    let surface = unsafe { gl_display.create_window_surface(&config, &surface_attributes) }
+        .expect("Failed to create Windows Surface");
     let context_attributes = ContextAttributesBuilder::new()
         .with_profile(GlProfile::Core)
         .build(Some(raw_window_handle));
-    let context = unsafe {
-        gl_display
-            .create_context(&config, &context_attributes)
-            .unwrap()
-    }
-    .make_current(&surface)
-    .unwrap();
+    let context = unsafe { gl_display.create_context(&config, &context_attributes) }
+        .expect("Failed to create OpenGL context")
+        .make_current(&surface)
+        .unwrap();
 
     Context {
         surface,
--- a/src/window/renderer.rs
+++ b/src/window/renderer.rs
@@ -16,13 +16,9 @@ fn create_surface(
     fb_info: FramebufferInfo,
 ) -> Surface {
     let pixel_format = windowed_context.get_config();
-    let size = windowed_context.window().inner_size();
-    let size = (
-        size.width.try_into().expect("Could not convert width"),
-        size.height.try_into().expect("Could not convert height"),
-    );
+    let size = windowed_context.get_render_target_size();
     let backend_render_target = BackendRenderTarget::new_gl(
-        size,
+        size.into(),
         Some(pixel_format.num_samples() as usize),
         pixel_format
             .stencil_size()
@@ -31,8 +27,8 @@ fn create_surface(
         fb_info,
     );
     windowed_context.resize(
-        NonZeroU32::new(size.0 as u32).unwrap(),
-        NonZeroU32::new(size.1 as u32).unwrap(),
+        NonZeroU32::new(size.width).unwrap(),
+        NonZeroU32::new(size.height).unwrap(),
     );
     Surface::from_backend_render_target(
         gr_context,
--- a/src/window/renderer.rs
+++ b/src/window/renderer.rs
@@ -1,8 +1,10 @@
-use std::convert::TryInto;
+use std::num::NonZeroU32;
+use std::{convert::TryInto, ffi::CString};
 
 use crate::redraw_scheduler::REDRAW_SCHEDULER;
 use crate::renderer::WindowedContext;
 use gl::types::*;
+use glutin::prelude::GlConfig;
 use skia_safe::{
     gpu::{gl::FramebufferInfo, BackendRenderTarget, DirectContext, SurfaceOrigin},
     Canvas, ColorType, Surface,
@@ -13,7 +15,7 @@ fn create_surface(
     gr_context: &mut DirectContext,
     fb_info: FramebufferInfo,
 ) -> Surface {
-    let pixel_format = windowed_context.get_pixel_format();
+    let pixel_format = windowed_context.get_config();
     let size = windowed_context.window().inner_size();
     let size = (
         size.width.try_into().expect("Could not convert width"),
@@ -21,16 +23,17 @@ fn create_surface(
     );
     let backend_render_target = BackendRenderTarget::new_gl(
         size,
+        Some(pixel_format.num_samples() as usize),
         pixel_format
-            .multisampling
-            .map(|s| s.try_into().expect("Could not convert multisampling")),
-        pixel_format
-            .stencil_bits
+            .stencil_size()
             .try_into()
             .expect("Could not convert stencil"),
         fb_info,
     );
-    windowed_context.resize(size.into());
+    windowed_context.resize(
+        NonZeroU32::new(size.0 as u32).unwrap(),
+        NonZeroU32::new(size.1 as u32).unwrap(),
+    );
     Surface::from_backend_render_target(
         gr_context,
         &backend_render_target,
@@ -50,13 +53,13 @@ pub struct SkiaRenderer {
 
 impl SkiaRenderer {
     pub fn new(windowed_context: &WindowedContext) -> SkiaRenderer {
-        gl::load_with(|s| windowed_context.get_proc_address(s));
+        gl::load_with(|s| windowed_context.get_proc_address(CString::new(s).unwrap().as_c_str()));
 
         let interface = skia_safe::gpu::gl::Interface::new_load_with(|name| {
             if name == "eglGetCurrentDisplay" {
                 return std::ptr::null();
             }
-            windowed_context.get_proc_address(name)
+            windowed_context.get_proc_address(CString::new(name).unwrap().as_c_str())
         })
         .expect("Could not create interface");
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,6 +57,30 @@ dependencies = [
 ]
 
 [[package]]
+name = "android-activity"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7c77a0045eda8b888c76ea473c2b0515ba6f471d318f8927c5c72240937035a6"
+dependencies = [
+ "android-properties",
+ "bitflags",
+ "cc",
+ "jni-sys",
+ "libc",
+ "log",
+ "ndk 0.7.0",
+ "ndk-context",
+ "ndk-sys 0.4.1+23.1.7779620",
+ "num_enum",
+]
+
+[[package]]
+name = "android-properties"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fc7eb209b1518d6bb87b283c20095f5228ecda460da70b44f0802523dea6da04"
+
+[[package]]
 name = "android_glue"
 version = "0.2.3"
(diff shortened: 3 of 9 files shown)


Kethku:
Note: the ime support seems way better in the new upstream branch, so at some point after this merges, we should pipe the cursor positions to the ime system. Should resolve some long standing bugs.

We also get precommit texts now for ime, so it should also be possible to render in progress texts next under the cursor to fully round out our ime support.

shaunsingh:
image

Small fix to the our glutin fork to build properly on macOS, we can/should upstream it though

shaunsingh:
`Cargo.toml` also still puts `winit` at the old `new-keyboard-all` branch, not sure if that should be updated to v3

fredizzimo:
I think we need to prioritize getting this fixed and merged. `wgpu` will need a more recent winit. And so will my changes to the update loop, so that I can read the display refresh rate.

As far as I see, it's failing because of glutin, we should probably update that to the latest as well. We will need some changes to the context creation, but the `glutin-winit` makes it look almost the same as before.

fredizzimo:
@Kethku, @shaunsingh. I updated to the latest Glutin and cherry-picked the relevant changes from here https://github.com/neovide/neovide/pull/1821, this fixes the build issues, and additionally fixes the Wayland configuration as descirbed in that PR.

Now, since there hasn't been a response to https://github.com/rust-windowing/winit/commit/639e1d40b13a38d9545e452d9f7e7ae5b77e4740#r108100537, the question remains, should we fix that locally in our fork, or try to reach out to the winit guys once more.

I would really like to see this moving forward, not only for the reasons I described in my earlier message, but also because switching to a newer winit should close a lot of the bugs that we have currently.

src/window/renderer.rs
```diff
@@ -13,24 +15,25 @@ fn create_surface(
...
-            .stencil_bits
+            .stencil_size()
             .try_into()
             .expect("Could not convert stencil"),
         fb_info,
     );
-    windowed_context.resize(size.into());
+    windowed_context.resize(
+        NonZeroU32::new(size.0 as u32).unwrap(),
+        NonZeroU32::new(size.1 as u32).unwrap(),
```
fredizzimo, line 35: I think we should force a minimum size here, rather than crashing with unwrap errors. I will fix that myself.

Actually I will also add a maximum size of 8k or something, if someone has a bigger window than that, it's better to get a blurry one, than to crash because of too big buffers. I think we have some reports for that as well. I will add the same limit to the creation as well.

fredizzimo (changes requested):

shaunsingh:
> Now, since there hasn't been a response to https://github.com/rust-windowing/winit/commit/639e1d40b13a38d9545e452d9f7e7ae5b77e4740#r108100537, the question remains, should we fix that locally in our fork, or try to reach out to the winit guys once more.

Feel free to add it into the fork, I feel like it's quite an urgent issue, and can take quite some time to get patches merged upstream

src/renderer/opengl.rs
```diff
@@ -1,40 +1,94 @@
...
+
+    let size = window.inner_size();
+
+    let surface_attributes = SurfaceAttributesBuilder::<WindowSurface>::new()
+        .with_srgb(Some(cmd_line_settings.srgb))
+        .build(
+            raw_window_handle,
+            NonZeroU32::new(size.width.max(1)).unwrap(),
+            NonZeroU32::new(size.height.max(1)).unwrap(),
+        );
```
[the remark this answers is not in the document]
fredizzimo: Only the vsync option was missing, so I added that.

src/window/renderer.rs
fredizzimo: I added validation for all the buffer sizes.

fredizzimo:
The keyboard reperat fix should now be included

fredizzimo (approved):

last-partizan (approved):
I've been using this branch for a few days, and it works great.

Anything else needs to be done before merging?

fredizzimo:
Nothing from my part, except that perhaps it would make sense to lock the commit id of the new-keyboard-v3 branch in `Cargo.toml` instead of relying on just the lockfile to point to the right commit. That we can better control when to update to newer versions.

MultisampledNight:
Sidenote that building this branch fails with `cargo install` or after `cargo update` is ran, with errors in winit and this warning:

```
warning: Patch `winit v0.28.1 (https://github.com/neovide/winit?branch=new-keyboard-v3#4d8d82fd)` was not used in the crate graph.
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
```

I've pushed a commit locking `winit` to `=0.28.1` which fixes this. Seemingly the implicit `^` specifier was not enough and upgraded to `0.28.6` unconditionally (and also fixes a small formatting issue), since this **was** `cargo tree -i winit` after `cargo update` is ran:

```
winit v0.28.1 (https://github.com/neovide/winit?branch=new-keyboard-v3#4d8d82fd)
├── glutin-winit v0.3.0
│   └── neovide v0.10.4 (/home/multisn8/info/lurk/neovide)
└── neovide v0.10.4 (/home/multisn8/info/lurk/neovide)
```

fredizzimo:
We can't merge this. There seems to be a deadlock in the new keyboard handling of winit on Windows. I'm not sure what's wrong, and I'm don't think I have time to debug it today. But we did have a deadlock earlier too that we fixed and notified the upstream about. They did not take our fix as I wrote it  and instead they made another fix by their own, so I wonder if they messed it up. But it could of course be some other deadlock as well.

The deadlock is 100% for me, when I use neovide on a windows remote desktop after I minimize the desktop and return.

fredizzimo:
It's indeed the same bug, or at least very similar, the code calls `PeekMessage` inside the keyboard event handling which deadlocks due to re-entrancy. 

It's either a new regression in the keyboard branch after our previous merge or their fix does not solve the problem https://github.com/ArturKovacs/winit/pull/1

fredizzimo:
It looks like we might be able to switch over to point to the official branch instead. The deadlock appears to be fixed https://github.com/rust-windowing/winit/pull/2662 and the Wayland keyboard repeat might also be fixed. 

I have some minor changes that moves this over to that instead of our fork, along with a few minior fixes due to very slight interface changes. Should I update this PR with those changes, our should we use a separate one for that?

I don't think it makes much sense to pull in the changes to our fork and have Neovide point to that. If we encountered bugs that needs to be fixed, we should send PR upstream instead. And if we lock the commit hash along with the branch it should be just as stable as having our own fork.

last-partizan:
I agree with you, it's better to use official branch.

MultisampledNight:
Tested this in combination with https://github.com/neovide/neovide/pull/1870 for the past 25 minutes on NixOS with Sway (Wayland) with an Intel iGPU, and I cannot observe any change in behavior so far, nice!

fredizzimo:
@last-partizan, could you try that with the official Winit release and on of their examples?  The switch to the upstream version also changes the main Winit version from 0.28.1 to 0.28.6, so it's possible that there's been some regressions.

I also noticed now when checking the diffs that my change converts `cargo.toml` to Linux line endings. I was not supposed to do that, but I think it's a good change, I don't know why we store it with windows line endings, and my mergetool does not detect the line endings automatically so it's been very annoying to deal with that file.

MultisampledNight:
Actually I just noticed there is _one_ thing: If [`g:neovide_hide_mouse_when_typing`](https://neovide.dev/configuration.html#hiding-the-mouse-when-typing) is set to `v:true`, moving the mouse inside the Neovide window does not make it visible again — it needs to be moved out of bounds first. I'm not sure if this should be fixed on our end, since https://github.com/neovide/neovide/blob/4b3c234a49d7db3edbc6ba821964805525f7a02a/src/window/mouse_manager.rs#L443 _should_ trigger the mouse to be visible again, or if this is in winit.

(The file in question has not been changed here or in #1870.)

fredizzimo:
@MultisampledNight, there's an example in winit that you could try `cargo run --example cursor_grab`. Pressing `h` should hide the cursor and `shift-h` show it.

MultisampledNight:
@fredizzimo The winit examples don't display under Wayland, since they don't render. I'll create a wgpu test app and try it with that sometime today.

MultisampledNight:
> Actually I just noticed there is _one_ thing: If [`g:neovide_hide_mouse_when_typing`](https://neovide.dev/configuration.html#hiding-the-mouse-when-typing) is set to `v:true`, moving the mouse inside the Neovide window does not make it visible again — it needs to be moved out of bounds first. I'm not sure if this should be fixed on our end, since
> 
> https://github.com/neovide/neovide/blob/4b3c234a49d7db3edbc6ba821964805525f7a02a/src/window/mouse_manager.rs#L443
> _should_ trigger the mouse to be visible again, or if this is due to winit.
> 
> (The file in question has not been changed here or in #1870.)

Fwiw this was fixed in rust-windowing/winit#2821, but that landed on winit's `master` branch, so we won't see it until https://github.com/rust-windowing/winit/pull/2662 is merged and we depend directly on winit's default branch/a new release after that instead.

fredizzimo:
@last-partizan, I checked Wayland with `WINIT_UNIX_BACKEND=x11` 

Using KDE with Nvidia 3070 it does not crash for me, but flickers badly when resizing. Native Wayland works fine though. 

Did you have any chance to debug where it crashes yet? It would be nice to support xwayland, but if native wayland works, I don't think there's a need for it.

last-partizan:
I does not crashes, but freezes, so i kill it with ctrl-c.

Is there any way to see on what line of code it freezed?

Here's how it looks like 
[Screencast from 2023-05-27 11-04-44.webm](https://github.com/neovide/neovide/assets/301015/6c938c54-15e4-4630-a718-dc2b8f8bd65b)

Here's last debug lines:
```
TRACE [neovide::channel_utils] neovide::window::WindowCommand SetMouseEnabled(true)
TRACE [neovide::editor] Image flushed
TRACE [neovide::channel_utils] alloc::vec::Vec<neovide::renderer::DrawCommand> [ModeChanged(Normal), UpdateCursor(Cursor { grid_position: (6, 1), parent_window_id: 1, shape: Block, cell_percentage: Some(0.0), blinkwait: Some(0), blinkon: Some(0), blinkoff: Some(0), style: None, enabled: true, double_width: false, grid_cell: (" ", None) })]
TRACE [neovide::redraw_scheduler] Next frame queued
```

And when i try to strace it, it looks like this:
```
strace: Process 9602 attached
futex(0x55a54339b228, FUTEX_WAIT_BITSET, 2, NULL, FUTEX_BITSET_MATCH_ANY) = ?
+++ killed by SIGKILL +++
```

Any ideas how to proceed?

Good news: I cannot reproduce it on my laptop, so hopefully not everyone is affected.

fredizzimo:
@last-partizan 

Using a debugger is the easiest, just break into the debugger when the hang happens, and check which mutex/futex the main loop is blocking on. Neovide has many threads so you might need to expand the callstacks one by one to find out. If you need more help, you can post the full callstack here.

I'm usually using codelldb extension for Visual Studio code for debugging, since that works great for rust https://github.com/vadimcn/codelldb with the following launch configuration, the one we have in the repository is outdated

 ```json
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Launch",
            "type": "lldb",
            "request": "launch",
            "args": ["--multigrid"],
            "cargo": {
                "args": ["build"]
            }
        },
    ]
}
```
That enables both debugging and building from within VSCode. I still write the code in Neovim/Neovide though, so VSCode is just for debugging. 

But of course you can also use lldb directly, or use it through Neovim DAP, but I just think that the Visual Studio debugging experience is better.

**Comment by MultisampledNight** [comment]
Fwiw another short and hakcy possibility would be to `kill -SEGV $(pgrep neovide)` and using `systemctl debug`, then in the launched gdb instance `backtrace` to display the backtrace it was hanging on.

**Comment by last-partizan** [comment]
I was able to start debugging session with lldb and codelldb, but i haven't figured out why it hangs.

<details><summary>Here's backtrace from lldb</summary>

```lldb
(lldb) bt
* thread #1, name = 'neovide', stop reason = signal SIGSTOP
  * frame #0: 0x00007f62f24072ed libc.so.6`syscall + 29
    frame #1: 0x00007f62e0ec71de radeonsi_dri.so`___lldb_unnamed_symbol2790 + 142
    frame #2: 0x00007f62e0f269fb radeonsi_dri.so`___lldb_unnamed_symbol4325 + 203
    frame #3: 0x00007f62e0eba210 radeonsi_dri.so`___lldb_unnamed_symbol2634 + 80
    frame #4: 0x00007f62e309c73f libGLX_mesa.so.0`___lldb_unnamed_symbol2575 + 79
    frame #5: 0x00007f62e30885bb libGLX_mesa.so.0`___lldb_unnamed_symbol2294 + 91
    frame #6: 0x00005596969770fd neovide`glutin_glx_sys::glx::Glx::SwapBuffers::h4b5b29833b4e6c72(self=0x00005596986730a0, dpy=0x0000559699af8da0, drawable=12582915) at glx_bindings.rs:561:120
    frame #7: 0x00005596968a5bcb neovide`_$LT$glutin..api..glx..surface..Surface$LT$T$GT$$u20$as$u20$glutin..surface..GlSurface$LT$T$GT$$GT$::swap_buffers::_$u7b$$u7b$closure$u7d$$u7d$::heccc6a8afa9764cd at surface.rs:215:13
    frame #8: 0x00005596968c45d9 neovide`glutin::api::glx::last_glx_error::hb9aa29a8c9e7cfff(display=(__0 = 0x0000559699af8da0), callback={closure_env#0}<glutin::surface::WindowSurface> @ 0x00007ffd24723858) at mod.rs:193:18
    frame #9: 0x00005596968a5b5f neovide`_$LT$glutin..api..glx..surface..Surface$LT$T$GT$$u20$as$u20$glutin..surface..GlSurface$LT$T$GT$$GT$::swap_buffers::h42505733f8afb379(self=0x00007ffd24731240, _context=0x00007ffd24731268) at surface.rs:214:9
    frame #10: 0x00005596969e1859 neovide`_$LT$glutin..surface..Surface$LT$T$GT$$u20$as$u20$glutin..surface..GlSurface$LT$T$GT$$GT$::swap_buffers::h7b1d5dd0cf6ed834(self=0x00007ffd24731238, context=0x00007ffd24731260) at surface.rs:314:17
    frame #11: 0x0000559696a147ea neovide`neovide::renderer::opengl::Context::swap_buffers::h2824e74af5de5207(self=0x00007ffd24731158) at opengl.rs:45:9
    frame #12: 0x000055969693dfe5 neovide`neovide::window::WinitWindowWrapper::draw_frame::hbaf693d67cfc7a2e(self=0x00007ffd24731158, dt=0.0227467846) at mod.rs:229:13
    frame #13: 0x00005596969b5be8 neovide`neovide::window::create_window::_$u7b$$u7b$closure$u7d$$u7d$::h72b767cade56deb1(e=Event<()> @ 0x00007ffd24723ef8, _window_target=0x0000559699bc97c0, control_flow=0x00007ffd2472c520) at mod.rs:518:13
    frame #14: 0x0000559696a9e042 neovide`winit::platform_impl::platform::x11::EventLoop$LT$T$GT$::drain_events::_$u7b$$u7b$closure$u7d$$u7d$::_$u7b$$u7b$closure$u7d$$u7d$::hbbabc9ebd002abe5(event=Event<()> @ 0x00007ffd24724158, window_target=0x0000559699bc97c0, control_flow=0x00007ffd2472c520) at mod.rs:553:29
    frame #15: 0x00005596969b4a85 neovide`winit::platform_impl::platform::sticky_exit_callback::h2d5178cbc5bbab45(evt=Event<()> @ 0x00007ffd24724238, target=0x0000559699bc97c0, control_flow=0x00007ffd2472c520, callback=0x00007ffd247242d0) at mod.rs:924:9
    frame #16: 0x0000559696a9dee3 neovide`winit::platform_impl::platform::x11::EventLoop$LT$T$GT$::drain_events::_$u7b$$u7b$closure$u7d$$u7d$::h4a908025594df04b(event=<unavailable>) at mod.rs:544:17
    frame #17: 0x000055969692ee18 neovide`winit::platform_impl::platform::x11::event_processor::EventProcessor$LT$T$GT$::process_event::hf48cc20b24926f68(self=0x00007ffd2472c8a0, xev=0x00007ffd2472baa0, callback={closure_env#0}<(), neovide::window::create_window::{closure_env#1}> @ 0x00007ffd2472bc20) at event_processor.rs:450:25
    frame #18: 0x0000559696a9de15 neovide`winit::platform_impl::platform::x11::EventLoop$LT$T$GT$::drain_events::hb988dedc72d2c6d4(self=0x00007ffd2472c838, callback=0x00007ffd24731148, control_flow=0x00007ffd2472c520) at mod.rs:543:13
    frame #19: 0x0000559696a9cdff neovide`winit::platform_impl::platform::x11::EventLoop$LT$T$GT$::run_return::single_iteration::h594fc8265cd9b880(this=0x00007ffd2472c838, control_flow=0x00007ffd2472c520, cause=0x00007ffd2472c548, callback=0x00007ffd24731148) at mod.rs:374:13
    frame #20: 0x0000559696a9db1b neovide`winit::platform_impl::platform::x11::EventLoop$LT$T$GT$::run_return::h44f7d6741b033d5a(self=0x00007ffd2472c838, callback={closure_env#1} @ 0x00007ffd24731148) at mod.rs:514:27
    frame #21: 0x0000559696a9e16d neovide`winit::platform_impl::platform::x11::EventLoop$LT$T$GT$::run::h292483a29ba25803(self=<unavailable>, callback=<unavailable>) at mod.rs:529:25
    frame #22: 0x00005596969b4912 neovide`winit::platform_impl::platform::EventLoop$LT$T$GT$::run::h443392cdf7033a9a(self=EventLoop<()> @ 0x00007ffd2472cf08, callback={closure_env#1} @ 0x00007ffd24731148) at mod.rs:832:56
    frame #23: 0x0000559696909bd9 neovide`winit::event_loop::EventLoop$LT$T$GT$::run::h45bbc2de0c89b18b(self=<unavailable>, event_handler={closure_env#1} @ 0x00007ffd24731148) at event_loop.rs:305:9
    frame #24: 0x0000559696940105 neovide`neovide::window::create_window::hd550cbd955f3d46e at mod.rs:474:5
    frame #25: 0x00005596968914ed neovide`neovide::protected_main::hfbb130847588a2ed at main.rs:168:5
    frame #26: 0x00005596968913b8 neovide`neovide::main::hfbbcab7472396a4c at main.rs:73:5
    frame #27: 0x0000559696a5d58b neovide`core::ops::function::FnOnce::call_once::ha6e66dcff1d39b27((null)=(neovide`neovide::main::hfbbcab7472396a4c at main.rs:63), (null)=<unavailable>) at function.rs:250:5
    frame #28: 0x000055969698369e neovide`std::sys_common::backtrace::__rust_begin_short_backtrace::ha13f6e4d92f57739(f=(neovide`neovide::main::hfbbcab7472396a4c at main.rs:63)) at backtrace.rs:134:18
    frame #29: 0x0000559696a10291 neovide`std::rt::lang_start::_$u7b$$u7b$closure$u7d$$u7d$::hc350b52dc7d41547 at rt.rs:166:18
    frame #30: 0x00005596978d992c neovide`std::rt::lang_start_internal::h2947494479068ade [inlined] core::ops::function::impls::_$LT$impl$u20$core..ops..function..FnOnce$LT$A$GT$$u20$for$u20$$RF$F$GT$::call_once::h814f0ad687ab2f92(self=&(dyn core::ops::function::Fn<(), Output=i32> + core::marker::Sync + core::panic::unwind_safe::RefUnwindSafe) @ 0x0000557dbe49c150, args=<unavailable>) at function.rs:287:13
    frame #31: 0x00005596978d9929 neovide`std::rt::lang_start_internal::h2947494479068ade at panicking.rs:487:40
    frame #32: 0x00005596978d9929 neovide`std::rt::lang_start_internal::h2947494479068ade at panicking.rs:451:19
    frame #33: 0x00005596978d9929 neovide`std::rt::lang_start_internal::h2947494479068ade [inlined] std::panic::catch_unwind::he1669253985d4352(f=&(dyn core::ops::function::Fn<(), Output=i32> + core::marker::Sync + core::panic::unwind_safe::RefUnwindSafe) @ 0x0000557dbe49c2f0) at panic.rs:140:14
    frame #34: 0x00005596978d9929 neovide`std::rt::lang_start_internal::h2947494479068ade [inlined] std::rt::lang_start_internal::_$u7b$$u7b$closure$u7d$$u7d$::hd778e2099ecb758e at rt.rs:148:48
    frame #35: 0x00005596978d9929 neovide`std::rt::lang_start_internal::h2947494479068ade at panicking.rs:487:40
    frame #36: 0x00005596978d9929 neovide`std::rt::lang_start_internal::h2947494479068ade at panicking.rs:451:19
    frame #37: 0x00005596978d9929 neovide`std::rt::lang_start_internal::h2947494479068ade [inlined] std::panic::catch_unwind::hb9e5b8ae73c93b8f(f={closure_env#2} @ 0x0000557dbe49c2f0) at panic.rs:140:14
    frame #38: 0x00005596978d9929 neovide`std::rt::lang_start_internal::h2947494479068ade(main=&(dyn core::ops::function::Fn<(), Output=i32> + core::marker::Sync + core::panic::unwind_safe::RefUnwindSafe) @ 0x0000557dbe49c2f0, argc=<unavailable>, argv=<unavailable>, sigpipe=<unavailable>) at rt.rs:148:20
    frame #39: 0x0000559696a1026a neovide`std::rt::lang_start::h5c51dd532b36ed19(main=(neovide`neovide::main::hfbbcab7472396a4c at main.rs:63), argc=1, argv=0x00007ffd24731f88, sigpipe='\0') at rt.rs:165:17
    frame #40: 0x0000559696892b7e neovide`main + 30
    frame #41: 0x00007f62f2327850 libc.so.6`___lldb_unnamed_symbol3141 + 128
    frame #42: 0x00007f62f232790a libc.so.6`__libc_start_main + 138
    frame #43: 0x000055969687bd75 neovide`_start + 37
```
</details>

And screenshots from vscode:
image


Looking at 
```
    frame #1: 0x00007f62e0ec71de radeonsi_dri.so`___lldb_unnamed_symbol2790 + 142
```

 it could be anything driver-related.

We have workaround, so feel free to ignore this issue.

fredizzimo:
@last-partizan,

I found a few of tangential issues https://github.com/rust-windowing/glutin/issues/1231, https://gitlab.freedesktop.org/mesa/mesa/-/issues/4932 and https://gitlab.freedesktop.org/mesa/mesa/-/issues/4180, but none of those matches exactly.

But it does seems like setting a swap interval (vsync) is a bad idea on Wayland. I bet `--novsync` also "fixes" the issue at the loss of some smoothness. My smooth rendering thread PR for example will switch to a software timer in that case.

Fortunately there's a better way to it in Wayland using https://wayland-book.com/surfaces-in-depth/frame-callbacks.html. Especially if `--novsync` works for you I will implement that in my smooth rendering PR. It will also fix the issue that it does not support a `g:neovide_refresh_rate` lower than the actual compositor refresh rate. So for people not  manually increasing `g:neovide_refresh_rate` from 60 to 120 for example, it will also be smoother by default.

last-partizan:
`--novsync` has no effect on this freeze.

fredizzimo:
@last-partizan, 

Hm, if `--novsync` doesn't have any effect, I wonder if it really blocks on `swap_buffers`? 

If you break and there's no hang, it will very likely break there, since most of the rendering loop will be waiting for vsync.

Could you try again, and ensure that neovide is really hung when you break it in the debugger? Maybe using `--novsync` this time.


BTW, meanwhile I have almost implemented vsync using frame callbacks, but now it seems like I have a problem with the event dispatching that I don't really know how to solve at the moment.

fredizzimo:
Unfortunately it looks like using frame callbacks is a dead end. Well it works, but it won't allow us to draw every second frame for example. In order to do that `wl_surface.commit()` has to be called, and that's done deep inside winit, actually I think in `stck-adwaita`. So I don't think it's a good idea to make any extra calls, and we probably also need to actually draw something when doing that, since the buffers will be swapped.

Anyway, upon further research, this is exactly what `SwapBuffers` does internally anyway, but our own implementation would allow a timeout, and therefore not block forever. So in theory it could fix the hang, but if disabling `vsync` doesn't do it, then it won't work anyway.

But on the more positive note, the upsream branch has been merged into winit master https://github.com/rust-windowing/winit/pull/2662 has been merged now while writing this message. I will update this PR to follow that commit instead.

fredizzimo:
We are now finally tracking winit/master! 

There were some quite major refactoring in https://github.com/rust-windowing/winit/pull/2817, since the the last keyboard update, which mostly affects the modifiers. But based on my tests the modifiers appears to work at least on Wayland, which I'm currently on, so no regressions there. But more testing is needed on other platforms before merging.

The biggest concern is probably the Mac modifiers with it's special handling of Option/Alt.

last-partizan:
I tried running debugger with `--novsync`, and it hangs just like before.

image

fredizzimo:
@last-partizan, have you tried the https://github.com/neovide/neovide/pull/1870 branch, it has a separate thread for rendering. I don't think it fixes anything, but worth a shot. 

Anyway, I don't think it's important to fix the XWayland support, since native Wayland works just fine, at most it will cause hangs for users, that used it as a workaround before.

In order to debug it further, we would need symbols for the the glx implementation. It's possible to debug further, but unless we really have to I rather spend my time on something more important.

last-partizan:
@fredizzimo yes, that branch fixes freezing.

fredizzimo:
So far I haven't found anything, either on Windows or Wayland that would prevent us from merging this.

Well, maybe the issue with needing `--nosrgb` at least on some systems. BTW, wezterm also has the same or a very similar issue https://github.com/wez/wezterm/issues/1701, so it's possible that there's something with winit/glutin.

fredizzimo:
I found the problems with SRGB, but I don't have time to properly test and fix it today

This needs to use srgb when the framebuffer is SRGB

https://github.com/neovide/neovide/blob/f623704ffa17d0ba6388d9db01994ac448238175/src/window/renderer.rs#L68-L70


And so does this
https://github.com/neovide/neovide/blob/f623704ffa17d0ba6388d9db01994ac448238175/src/window/renderer.rs#L32-L40

Those lines have not been changed by this PR, and I'm confused how this has actually ever worked, and why it works on other platforms.  Some more investigation is needed, otherwise the fix would be trivial.

fredizzimo:
I have been reading up and experimenting a bit more with srgb stuff. And my conclusion is that at least on Wayland with Nvidia it does not work as it should.

Creating a backbuffer with `srgb` support should by itself not enable `srgb`, it just makes it possible to do so. From this
https://www.khronos.org/opengl/wiki/Image_Format
> [When images with this format are used as a render target](https://www.khronos.org/opengl/wiki/Framebuffer_Colorspace), OpenGL will automatically convert the output colours from linear to the sRGB colorspace if, and only if, GL_FRAMEBUFFER_SRGB is enabled. The alpha will be written as given. When writing multiple outputs, only outputs written to sRGB image formats will undergo such conversion.

`GL_FRAMEBUFFER_SRGB` is automatically enabled by Skia when we tell it to, but  we are not. We request linear formats, so there should be no conversion. To properly request srgb we should set `ColorType::SRGBA8888` and then either convert all our colours to srgb or pass `Some(ColorSpace::new_srgb_linear()),` That gives us the correct colours on the other platforms, but on Wayland with Nvidia we need to either pass `None` or `Some(ColorSpace::new_srgb())`. I think the automatic conversion from linear to srgb is not working, so the output ends up being too dark.

Anyway, since we are not really using srgb, I suggest that we disable it by default, except on Windows, where we need it to workaround this bug https://github.com/neovide/neovide/issues/907.

I also found out that the environment variable `NEOVIDE_NO_SRGB` also works wrong, `NEOVIDE_NO_SRGB=0` means `--nosrgb` while `NEOVIDE_NO_SRGB=1` means srgb enabled. It would be nice if the environment variable was named `NEOVIDE_SRGB` and that we had two command line options `--nosrgb` and `--srgb`.

If someone has time to do those modifications, feel free to do so. It's getting late here, so no more changes by me tonight.

fredizzimo:
I created a separate PR for the srgb issues https://github.com/neovide/neovide/pull/1885, so we should preferably merge that one first.

last-partizan:
@fredizzimo i want to convert `Cargo.toml` to unix line endings in the separate branch, and merge it first. But it will introduce conflicts in this branch. Is that okay? Or i can wait until this merged.

fredizzimo:
@last-partizan, go ahead. I think I accidentally already converted it here.

Actually I would prefer if you also sorted it alphabetically, so that it’s easier to find exising dependencies. I think that I already did that in one of my other pull requests. That also slightly reduces the risk of merge conflicts in the future, since you don’t always add new dependencies  to the end.

fredizzimo:
I have rebased this (and https://github.com/neovide/neovide/pull/1787, https://github.com/neovide/neovide/pull/1790, https://github.com/neovide/neovide/pull/1870) on top of the latest master.

last-partizan:
@fredizzimo  Is there any blocking issues?

I have used #1870 branch (based on this) for a week now, and haven't noticed any keyboard-related issues. Linux/x11 and wayland works fine.

@Kethku Is there anything you want to check before merging this?

fredizzimo:
@last-partizan. I don't know of any blockers either, unless we want to wait for the planned beta https://github.com/rust-windowing/winit/issues/2686. But I think it makes sense to merge now, and update to the beta as soon as it is out.

shaunsingh:
LGTM, tested on macOS

For some reason, trying to open neovide from finder on macOS no longer works, but if I do `open -a /Applications/Neovide.app` and run it from the terminal ... it opens just fine. Not sure if that's a result of these changes or not

last-partizan:
Let's merge this.

I was about mo merge it myself, but i think @fredizzimo can do better job at writing commit message.

@fredizzimo can you do it? And please, use "Squash and merge", and [conventianl commits](https://www.conventionalcommits.org/en/v1.0.0/) syntax when writing commit message.

last-partizan:
I merged main into this, and resolved conflicts in the "Cargo.lock", by resetting it to "main" branch version and then running `cargo r`.

fredizzimo:
Later today took a couple of days, but now it's merged :) Let's see what all this fixes and breaks.

fredizzimo:
I quickly went through all our open issues to see what could potentially be fixed, but I probably missed a lot of issues, so it would be good to do another pass at some point.

It looks like the update itself might not have fixed that many keyboard issues, we still need to do some work especially around the modifier support.

manfredlotz:
Once I had opened #1236.

My system: EndeavourOS using Xorg

Having build neovide from commit _4b3c234a49d7db3edbc6ba821964805525f7a02a_ (May 14) and neovide doesn't react when having started neovide and switched to other app and switching back to neovide. Only moving the mouse makes it work.

Now haven git pulled f412399676c83ea670cd22745b850dad9ee210a0 (June 14) and rebuilt neovide. 

Problem seems to be fixed as when switching to other apps and switching back to neovide then neovide is responsive without any intermediate mouse movement.

So, from my side things seems to be fixed! Thanks a lot for your work.

Issue #1782, opened 2023-02-28
Text display bug using opacity


Valwex:
**Describe the bug**
Text display bug using opacity.

**To Reproduce**
1. Add an opacity in the config file
2. Resize the window

**Expected behavior**
Do not have the old text displayed in the background.

**Screenshots**
<img width="853" alt="Capture_decran_2023-02-28_a_21 20 26" src="https://user-images.githubusercontent.com/62778739/221977870-6ffebde3-a3dd-4de9-a2c6-888d775aca26.png">

**Desktop (please complete the following information):**
 - OS: [macOS 13.3]
 - Neovide Version [0.10.3]
 - Neovim Version [0.8.3]

xbladesub:
same

Issue #1771, opened 2023-02-18
Smooth scrolls around when changing buffers


DasGandlaf:
**Describe the bug**
When f.e. cursor is on line 500, then switch to a buffer where the cursor is on line f.e. 10, it makes a scrolling animation.

**To Reproduce**
1. Go to line f.e. 500
2. Switch to a buffer where cursor is on line f.e. 10

**Expected behavior**
When switching buffers, it should disregard the line position on the previous buffer, and should not smooth scroll to the line number on the new buffer.

**Desktop**
 - OS: Windows 21H1
 - Neovide Version 0.10.3
 - Neovim Version 0.8.2

goduck777:
I still saw a scroll animation when switching buffer using the new git version.

opalmay:
```lua
vim.api.nvim_create_autocmd("BufLeave", {
	callback = function()
		vim.g.neovide_scroll_animation_length = 0
		vim.g.neovide_cursor_animation_length = 0
	end,
})
vim.api.nvim_create_autocmd("BufEnter", {
	callback = function()
		vim.fn.timer_start(70, function()
			vim.g.neovide_scroll_animation_length = 0.3
			vim.g.neovide_cursor_animation_length = 0.08
		end)
	end,
})
```
This works as a workaround

fredizzimo:
I'm re-opening this, but we need to open an upstream Neovim bug report for that. I will do that a bit later, since the proper bug reports that needs a bit more information than this

There are two problems.
1. It does not send a `clear` command when switching buffer, instead it just re-sends the lines. This causes our scrollback buffer to remain valid from Neovide's perspective, even if it really isn't.
2. It sends a `win_viewport` containing a scroll delta, which causes the actual scroll

Issue #1638, opened 2022-11-15
Smooth scrolling jumps wildly when scrolling across folds


Hubro:
**Describe the bug**

When smooth scrolling is enabled, scrolling across folded lines causes the window to jump wildly. This has the opposite effect of what smooth scrolling seeks to do. Instead of making it easy to see the start- and end positions of a scroll, this makes it impossible by basically shaking the window, making it even less clear than if smooth scrolling were disabled.

This issue basically makes using folds a terrible experience. In files with lots of folds, such as Python files with dozens of classes and functions, folds become completely unusable. See the videos below for examples.

**To Reproduce**

1. Enable smooth scrolling
2. Enable folds and collapse the folds
3. Scroll a fold into or out of the current window
4. The window jumps wildly whenever a closed fold is encountered

**Expected behavior**

The smooth scrolling should work just like normal.

If this is currently impossible because of Neovim API limitations, I think Neovide should detect that the scrolled content includes folded regions and disable animations for the current scroll operation.

**Screenshots**

https://user-images.githubusercontent.com/597206/201907715-a774c63f-221e-4ef3-91e1-07c8c1347a45.mp4

https://user-images.githubusercontent.com/597206/201909958-2b160049-3d69-4a37-b4bd-c0d223c2a34f.mp4

**Desktop (please complete the following information):**
 - OS: Arch Linux (Linux 6.0.8-arch1-1)
 - Neovide Version 0.10.1
 - Neovim Version 0.8.0

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

(This is trivial to reproduce so I don't think this is necessary)

**Additional context**

\-

Issue #1630, opened 2022-11-07
Part of buffer is cleared instantly instead of remaining until out of view


Hubro:
**Describe the bug**

When scrolling up or down, the portion of the buffer that is scrolled out of view is instantly cleared, i.e. blanked. This has been happening to me for a long, long time (over a year?) but I haven't realized what the issue is because of my short animation time, I just noticed a very annoying flash any time I scrolled. Increasing the animation time makes it clear what the problem is (see the video below.)

For a nice scroll effect, this part of the buffer should remain visible until it leaves the viewport.

Looking at the [smooth scrolling demo in the GIF on the features page](https://neovide.dev/features.html), this doesn't appear to happen. Does this only occur to me?

**To Reproduce**
1. Enable multigrid
2. Scroll up or down using Ctrl+f or Ctrl+b

**Expected behavior**

A nice scroll animation.

**Screenshots**

https://user-images.githubusercontent.com/597206/200279267-4b78b925-53a9-4d14-a330-957ed6f82c18.mp4

**Desktop (please complete the following information):**
 - OS: Linux (Linux 6.0.7-arch1-1)
 - Neovide Version 0.10.1
 - Neovim Version 0.8.0

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

[neovide_rCURRENT.log.gz](https://github.com/neovide/neovide/files/9950234/neovide_rCURRENT.log.gz)
[neovide_r2022-11-07_10-53-45.log.gz](https://github.com/neovide/neovide/files/9950235/neovide_r2022-11-07_10-53-45.log.gz)

(These are some huge files, I don't think you want me to paste the contents... :smile:)

**Additional context**

\-

fredizzimo:
This will be greatly improved when I finish this https://github.com/fredizzimo/neovide/commits/fsundvik/improve-performance. At least cttrl-f and ctrl-b should work perfectly. But there will still be problems when scrolling completely outside of the previous view. For that we probably need to snap to the last line and animate only the scrolling of the final line.

Furher improvements probably need new support from Neovim itself.

Awpteamoose:
Not sure if related, but seems to be related:
If open fzf's `:Files` (or anything that pops up a quickfix) - a portion of the buffer would just get blanked out. Unless I'm at the very top of the buffer, the blanked out lines will be redrawn, otherwise I have to either switch buffer and go back or scroll or smth like that:

[...]

PR #1790, opened 2023-03-05
Replace scrolling snapshots with a buffer of individual lines

fredizzimo:
This completely changes how the smooth scrolling and multi-grid are handled internally. Instead of storing several snapshot textures that are scrolled, each line is stored as a Skia Picture, which is a lightweight container for rendering that particular line. Each internal Neovide Window contains a ringbuffer of such pictures, big enough to contain double the window height, which means that lines outside the screen are available for scrolling.

Neovide listens to the `win_viewport` event and uses the `scroll_delta` field to control how far it actually scrolls.

The actual scrolling algorithm also had to be changed, since everything works  a bit smoother, and the simple easing based scrolling didn't look good at all, constantly speeding up and slowing down. This is even more noticeable with my rendering improvements, which will come later. I tried to to roughly match the `scroll_animation_length`, but it's not possible to get it fully the same so it might have to be adjusted a bit if the user found something to their liking. This is possibly something that could be pulled out from this pull request into it's own PR.

Finally the window transparency rendering has been fixed, and I think it now works as it should. This is also something that could be extracted to it's own PR. But I keep it here, at least initially so users can test everything together a bit. Note that the `floating_opacity` setting has been completely removed, I don't really know what that is supposed to represent and how it should work together with everything.

The smooth scrolling itself is also a bit faster in terms of framerate, at least on my machine, but that was not the actual goal, the big performance improvements will be in another PR. Note, that we are actually doing more actual rendering work, than before, so it's slightly surprising.

This PR fixes a lot of issue, at least these ones I think (but I have not verified everything, I need help with that, and also with finding other fixed bugs)
https://github.com/neovide/neovide/issues/1782
https://github.com/neovide/neovide/issues/1771
https://github.com/neovide/neovide/issues/1638
https://github.com/neovide/neovide/issues/1630
https://github.com/neovide/neovide/issues/1626
https://github.com/neovide/neovide/issues/1559
https://github.com/neovide/neovide/issues/1198
https://github.com/neovide/neovide/issues/794
https://github.com/neovide/neovide/issues/1478 (the background and foreground rendering is separated now, so that should fix it)

Not sure about these, 
https://github.com/neovide/neovide/issues/1550 (All windows should anchor to the actual grid, not the scrolled position, and I'm not sure if this is fixed or not)
https://github.com/neovide/neovide/issues/1446

But with these big changes, I also expect there to be a lot of new bugs, so let's test this thoroughly before merging. I would also appreciated if someone did the required documentation changes, especially related to the removal of the `floating_opacity` setting.

Edit: I forgot to mention that this also affect the normal single grid rendering. In the end both are using the same functionality inside Neovide


src/editor/mod.rs
```diff
@@ -174,6 +203,7 @@ impl Editor {
                     rows,
                     columns,
                 } => {
+                    tracy_zone!("EdtitorScroll");
```
Kethku, line 206: Minor typo

Kethku:
I'm worried about the usage of the scroll event for semantic scrolls. The docs for it say that https://github.com/neovim/neovim/blob/master/runtime/doc/ui.txt#L394 it is semantically unrelated to the ui concept of scrolling a buffer. This shows up when you scroll larger distances than the rendered text. The screen will jump rather than scrolling. I remember also finding some other examples where the scroll events would be used for things other than scrolling which is why I didn't go this route originally.

That said, there are clearly problems with the current scrolling. So maybe this is better until a more official approach can be done using neovim apis  or w/e

Kethku:
For what its worth, there appears to be some forward progress for giving us what we need to do more correct smooth scrolling here: https://github.com/neovim/neovim/pull/19270. Its been open for a while though...

Kethku:
The user MDeiml summarized the issue well here https://github.com/neovide/neovide/issues/1334#issuecomment-1174322614

fredizzimo:
It's true that this does not deal with scrolling more than one screen at a time, but in practice I don't think it matters. I personally prefer a quick jump to the actual destination over scrolling empty lines like the old implementation did. 

It's not really possible to implement bigger jumps than that without having access to the whole buffer text, including it's highlights, all line wraps and folds, which as far as I know is not available even inside Neovim itself.

I don't believe anything relying on `win_viewport` can be used to implement reliable scrolling, it has `top_line` and `bottom_line`, but those are physical buffer lines, not the virtual lines that are displayed on the screen. In the worst case both can be the same, with a long wrapped line. So it can't really deal with either wrapped or folded lines, and trying reverse engineer anything from that is a recipe for disaster. Even the proposal here https://github.com/neovim/neovim/issues/19227, would most likely not be possible, I don't think neovim does any line wrapping outside the visible screen, so it would not be able to tell how many virtual lines you scrolled if you jump a completely different part of the document. And even if it was possible, we are still missing all the lines between, in order to actually show something on the screen, so to do a completely correct implementation we need something different entirely.

Using the `scroll` event deals with wraps and folds just fine, and also anything else I have been able to throw on it. But you are right, and I have read the documentation too that it does not guarantee anything. In practice, I haven't found any issues, and I have also read the source code of Neovim trying to find anything that break. In the multigrid mode it looks like it's only used for scrolling, and in regular mode, it's also used for scrolling inside splits, but I detect and filter those by the `is_pure_updown` checks, that is when the whole window region is scrolled. I also think that we are in a good position to get the current behaviour standardized, perhaps even add a flag to tell if it's a scroll or not.

So far I have personally not found anything that feels wrong about this implementation,

A-Lamia:
just compiled with the window size fix and i have a blank screen with nothing, will now attempt to recompile without it and will update this message.

A-Lamia:
not sure if this would help.
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/10894299/neovide_rCURRENT.log)

cshuaimin:
Tested with my config, smooth scrolling is much better than before, thank you very much for working on this!

Found two issues though
1. `set winbar=%f` prevents smooth scrolling, scrolling is back to line wise
2. in `:term` insert mode I can't see what's typed and command output, which only show when I exit to term normal mode.

fredizzimo:
@A-Lamia, Unfortunately I don't see anything wrong in the log, but could you test without the transparency changes, that is up to commit 00a240d43e728cd577a7db90a52ef3b1b8a82c04. I tested this myself on  both Windows and Linux, with multi-grid and without, and for me all combinations work,.

@cshuaimin, 
1. It looks like winbar isn't it's own multigrid window in Neovim for some reason,. I think we will have to fill a bug to the Neovim team for that. I might be able to work around it if we can identify windows with Winbars, and know how many lines they are, but it would be very messy and hacky.
2. `:term` indeed behaves strangely, I use `vim-floatterm` locally so I haven't noticed the issues. I will check that out at some point

A-Lamia:
@fredizzimo thought video might be easier to see, top corner you can also see the branch :)
https://user-images.githubusercontent.com/6450181/223069924-a5faa5b2-1186-434e-8410-005cdc5805e7.mp4

https://user-images.githubusercontent.com/6450181/223071033-97e85f0b-d57c-42a0-bcb9-4010758d8461.mp4

fredizzimo:
@A-Lamia, I have no idea what the problem could be currently. But you said that it works with `--clean`, so it's probably some setting or plugin causing it, and then it would help to narrow it down.

It could also be that I'm doing something with Skia that is not really standard, and it fails on certain graphics cards.

cshuaimin:
> in :term insert mode I can't see what's typed and command output, which only show when I exit to term normal mode.

Just found this is caused by `set cmdheight=0`.

A-Lamia:
@fredizzimo yeah i have been trying to think what plugin it could be, for the most part everything is lazy loaded the onlything that is actually "loads" on start is my statusbar / tabbar and internal logic like mappings, i don't use any dashboards or anything like that nvim just opens to a blank buffer.

A-Lamia:
@fredizzimo found it!! back before multigrid was a mess i used to use it with `    neovide_scroll_animation_length = 0` to disable animation because it was super jank it seem when this is set to 0 everything like that video breaks.

A-Lamia:
@fredizzimo something strange with transparency some parts that shouldn't be transparent are transparent.

image

image


>https://user-images.githubusercontent.com/6450181/223092633-26bb561c-3a33-4d0d-b259-bd0bf4945503.mp4

A-Lamia:
Animation on scroll also seems to be pretty stuttery and has some line artifacts when used with visual select, anyway ill stop bothering you now thanks for all the hard work! If i was even remotely as smart as you would help out more, but all i can do is test stuff :)

fredizzimo:
@A-Lamia, no keep reporting any issues you find. I'm not sure when I will have time, but I will come back to this and start fixing.

* The `scroll_animation_length=0`, that should be easy to fix
* The transparency issues. Did you switch back to the latest of this branch? Do you have `neovide_transparency` set. This setting now applies globally to all backgrounds of all windows, including floating ones and the statusline, so that you always can see your desktop behind. This transparency is multiplied by what neovide internally sets, `winblend`, `pumblend`, highligt groups and so on It's hardfor me to tell if it works like that in your images or not. 

  I'm also not sure if that's the best design, so any suggestions here are welcome. Personally I think you should ignore `neovide_transparency`, and tune your color scheme instead. Complete control through that should now be possible.
* The visual select artifacts. Yes I see the same, I guess there's some rounding issues somewhere, I'll try to figure that out.
* The stuttering. It's mostly caused by the way we implement our frame loop, but also a big part is caused by Windows having very poor windowed mode opengl support, on Linux it's much smoother even without further rendering changes. Anyway, this all is going to be fixed with my rendering loop changes and by using Direct3D on Windows

A-Lamia:
@fredizzimo I'll keep reporting then :)

On the topic of transparency i have taken some images with it set to 0 and a red background everything that is not red will be visible.
image
 
image


as for normal transparency i have tried everything and can't get it to work for the window or the blurred popup can see i have my nvim open in my terminal with transparency working fine.
image

as for how transparency should work some interesting ideas could be:
1.	 a default style that works as most terminals work
2.  a global window transparency property with a property that could accept a table of highlight groups that would override what is effected by the global property.

I'm not sure if 2 would solve a particular issue i have with current neovide and other terminals where some highlight groups can not be set to `#000000` or `NONE` and just end up turning white I'm not sure if that is a multiplier issue or some other bug but that could me nice for maybe only having selected things be transparent with an adjustable global.

fredizzimo:
From the first screenshots with the red background, to me it looks like it's working as I planned it to do. With full transparency, only foreground elements, like text will be visible.

For the issue with not getting the built-in transparency to work. For `pumblend` the values should be integers between 0 and 100, 0.5 is invalid . And I don't think you can set `winblend` like that, since it's a per window property that you can set.

For telescope you can add `winblend` to the configuration, look at the `telescope.defaults.winblend` in the help.

For the ideas
1. I'm not sure how that would work in terms of nvim. At least from my initial testing with alacritty the terminal just turns the default background transparent and nothing else. When I launch nvim with my color scheme everything is opaque. 
  
   Turning only the default background transparent could be implemented, but then any elements using accent colors for example would be totally opaque, which is probably not what you want.
4. That's a much better idea, but I think it should be a lua plugin that modifies the actual highlight groups and the winblend of the windows. These changes should support that. `neovide_transparency` can be left at 1 for that

A-Lamia:
@fredizzimo Hmmm, yeah transparency is always complicated, like the way I'm actually used to transparency with the background is what im trying to get but i see that is now actually not possible,

fredizzimo:
It looks like the built-in blend factor only applies to floating windows
> blend={integer}					*highlight-blend*
	Override the blend level for a highlight group within the popupmenu
	or floating windows. Only takes effect if 'pumblend' or 'winblend'
	is set for the menu or window. See the help at the respective option.

And that `NONE` should mean transparent
> 	There are a few special names:
		NONE		no color (transparent)

So, I think the only sensible thing to do is to let `NONE` take the the color of `g:neovide_background_color`,  which supports an alpha value.  And then let `g:neovide_transparency` control the transparency of all other background colors.  And looking at our existing documentiation, that seems to roughly match what's documented https://neovide.dev/configuration.html, except that `g_neovide_background_color` now applies to all platforms for the content, but only where the background is set to `NONE`

This would already get some nice support for some teams that support it like [everforest](https://github.com/sainnhe/everforest/blob/master/doc/everforest.txt#L254)

A-Lamia:
@fredizzimo I'm fine with this, will `g:neovide_background_color` also be able to be updated in real time, and is it possible to also have NONE work on foreground ?

fredizzimo:
I will wait a bit for more input, maybe from others before I make those changes (not that I have time to do it right now anyway). 

But yes, it will update in real-time. 

And I think it can apply to the forground too, without much extra effort. I'm sure it can be done, but if it is too much effort, and requires too many changes it has to be outside of this pull request (which really should not include any transparency changes, but as explained in the first post it's easier to test it this way)

A-Lamia:
also idk if it's just me but pop ups seem to not be centered.

A-Lamia:
Seems like the background for some popups are a little strange
image

you can see the consistency difference with the blur.
image

fredizzimo:
@A-Lamia. I'm not sure that I understand your problem correctly. But I don't think that's an issue with how the transparency is rendered.

It looks like your color scheme uses two different blend factors for the `lazy.nvim` window.  One for the "button" backgrounds, and another for the rest of the window.  The first one looks to be the same as your telsecope window. Another possibility is that the windows use different background colors, which could also affect the transparency.

A-Lamia:
@fredizzimo telescope like popups are the issue it happens with all the colorschemes. the background colors are also the same. it seems the example i gave is not good one moment.

A-Lamia:
@fredizzimo 
No blur you can see the same color
image
image

Blur enabled with a value of `10`.
image
image

The blur is much less, but from reading what you said there might be something to it, it could be the reason why this effects more the none theme is that there are 2 highlight groups one that is a box and then the fills within the boarder and thats why the edges and in-between are normal possibly. 🤔

A-Lamia:
@fredizzimo haha you where right <3
image

fredizzimo:
Yes, I did not implement `NONE` for floating windows either. But I will fix that, so in the end `NONE` will be completely transparent, and take the color of the parent, or in the case of the root, the color of `neovide_background_color `

A-Lamia:
@fredizzimo I've been using your patch all day and honestly it's pretty solid, no crashes only none implementation problem on this topic (not accounting for performance stuff that will be fixed in the future) is the precision issue when scrolling you see it in some cases with blur or solid colored lines.

src/renderer/rendered_window.rs
```diff
@@ -225,26 +220,74 @@ impl RenderedWindow {
         }
 
         {
-            if 1.0 - self.scroll_t < std::f32::EPSILON {
-                // We are at destination, move t out of 0-1 range to stop the animation.
-                self.scroll_t = 2.0;
-                self.snapshots.clear();
+            let scroll_destination = 0.0;
+            let zeta = 1.0;
+            let omega = 4.0 / (zeta * settings.scroll_animation_length);
```
Kethku, line 225: Would it make sense to move this into the animation_utils.rs file? I think I get the basic gist of whats happening, but its not the clearest just inline like this without a label

Kethku: Just a suggestion though, take it or leave it

fredizzimo: Yes, conceptually it would belong better there. I have to think a little bit about the interface though.

And yes, you are also right I should have added some comments. I think I had comments at some point, but must have removed them in one of my countless re-bases. Or it might be in some parts that I did not include here, I had some improvements that uses linear regression to further smooth things out a tiny bit, but I determined that the improvements was so small, that it's not worth including at least not here.

The algorithm is just the `PD` part of the pretty standard https://en.wikipedia.org/wiki/PID_controller

And here's a nice presentation https://gdcvault.com/play/1027059/Math-In-Game-Development-Summit with an associated website to demonstrate it.

tobealive:
Thanks for the hard work over there!

Those are two things I found using the PR under Plasma/X11:

1. Vivid floating blurs 

https://user-images.githubusercontent.com/34311583/223855666-67081ad3-72ec-49a3-80b8-d2034b6e64b8.mov


It's very delicate in floating widows with layered backgrounds, like telescope. With true transparency it becomes obvious (wilder in the video example).

2. No scrolling animations?
For the last two days I couldn't figure why there are no scrolling animations. It wasn't until moving item by item in a floating command or picker dialog that I realized that scrolling animation works for j/k and similar movements. At least for me, it's a movement that never happens in a normal buffer because it's just too inefficient to achieve horizontal scroll editor navigation. For ctrl+d and it's friends, it's probably not just on my box that there are no animations?

fredizzimo:
@tobealive, which graphics card do you have. I have also been testing this in Plasma/X11 in addition to Windows, and I don't see the effect here. Maybe you could try disabling the blur, and see if it's related to that.

2. For the scrolling animations, it looks like it's the "winbar" bug reported above. There's not much I can do about that with the technique of scrolling I'm using here, other than starting to ignore certain lines, but that get's really messy. I will report a bug to the neovim team though at some point, I think the "winbar" should be represented as a floating window, not as a regular line of the buffer. It also does not seem to be possible to make it external, like you can do for many of the other built-in windows.

Kethku:
Another thing I noticed using this branch is that the cursor positioned in the scrolling content. This means that if you jump to the top of the file, there is a brief period where the cursor is at the top while the character its actually connected to isn't visible yet. This looks particularly weird when the cursor is over a character which is rendered inside the cursor.

CursorIssue

Here you can see the issue I'm talking about.

fredizzimo:
@Kethku, I will fix the cursor issue, seems like an easy fix

And for the flickering, I believe that is the same issue as the gaps in the visual mode selection, reported above. I already have that fixed here locally. I just needed to make sure to render the lines to integer coordinates

tobealive:
Also had that assumption that it might be part of the nvidia and linux story. It's an 3050RTX TI.

And yes it's the winbar, scrolling animations work when disabling it :+1: Thanks!

fredizzimo:
@A-Lamia, The `NONE` support has to be implemented later an a separate PR, it's more tricky than I anticipated. The UI docs says the following here https://neovim.io/doc/user/ui.html
>  ["default_colors_set", rgb_fg, rgb_bg, rgb_sp, cterm_fg, cterm_bg]
The first three arguments set the default foreground, background and special colors respectively. cterm_fg and cterm_bg specifies the default color codes to use in a 256-color terminal.
>
>  **The RGB values will always be valid colors, by default. If no colors have been set, they will default to black and white, depending on ['background'](https://neovim.io/doc/user/options.html#'background')**. By setting the ext_termcolors option, instead -1 will be used for unset colors. This is mostly useful for a TUI implementation, where using the terminal builtin ("ANSI") defaults are expected.

This makes it impossible to determine using the protocol if it should be NONE or just another color. It should still be possible, but we need to check the actual highlight groups from. We probably also need a `g:neovide_forground_color` to override the foreground, and allow text transparency. Another possibility is that the neovide colors always overrides these ones if set.

There are also some other corner cases, that needs to be dealt with. For example if I understand correctly, a highlight group with the background NONE does not really seem to imply that it's transparent, instead it means use the colors defined in the `Normal` group, which in turn maps to the default colors in the UI protocol.

A-Lamia:
@fredizzimo all good man better to focus on the more important things now like this patch and the DX implementation, I'm guessing this also means having `NONE` support for the foreground is also going to be difficult.

src/editor/mod.rs
fredizzimo: This is fixed both here, and in the profiling PR, which actually adds the tracy zone

fredizzimo:
I have fixed some things and done a rebase.

@cshuaimin, I think the terminal mode is fixed now. Although I could not exactly replicate it. For me it was only the cursor that was missing and `set cmdheight=0` also works. (The inverted colors were not handled correctly)

@A-Lamia I have also done some transparency fixes. I think the telescope problem you were trying to show is fixed now, or at least looks better now. There's still no `NONE` support, but `g:neovide_transparency` works better in combination with winblend now.

@A-Lamia, @Kethku, The Visual mode artifacts and the gutter flicker should also be fixed now.

@tobealive, The blur is rendered slightly different now, so maybe you can try again?

@Kethku, I can repeat the floating cursor character without my changes as well. It's just a bit easier to see here, due to the fact that it slows down more at the end of the animation. I think the fix belongs in another MR, and it's also not clear how it should behave.

Still to do:
@A-Lamia, I forgot to fix the issues with an animation length of zero
@Kethku, I still haven't refactored the PD controller animation code
Documentation and cleanup of removed parameters

A-Lamia:
@fredizzimo thanks for the work man!

just to go over some things as confirmation.

- Scrolling seems to work fine tested for the float procession issues (black lines) couldn't find any of them in the 3 cases i spotted them.

- Not sure about the transparency issue, I don't think it's something you can fix as it's not a bug but rather functionally how it's working currently, there are 2 highlight groups overlapping each other doubling the blur, this might be able to work in a more desired way once `NONE` is implemented unless you have a way to ignore a stacked blur effect from your end.
image
so there is no placebo i will not say what is what ;).

- this is also not a bug but a side effect, not sure if it's possible to do anything about it but when you use hotkeys that need to "input" a commad quickly there will be a flicker from the transition from the command window scrolling up to cover the status line very quickly (when set cmdheight=0), not sure if it's possible, to basically add a time delay so if it doesn't reach a max time it doesn't get displayed, this could also be completely impossible to resolve as it might be a neovim porblem directly.

tobealive:
@fredizzimo 
> The blur is rendered slightly different now, so maybe you can try again?

No more issues with the blur. Great work :+1: Thanks for that!

xbladesub:
@fredizzimo Thank you very much for this PR.

I noticed couple of things:

1. Which-key panel jumps some reason:
https://user-images.githubusercontent.com/45129790/226102180-38e3144e-d462-4dae-85d2-2010786e3c6f.mp4

2. When I open a new buffer and start to scroll with <C-d><C-u> it seems like scrolling starts to work only after second press of <C-d> or <C-u>.

And is it possible to make all floating windows look like exact the same if I would not enable `multigrid`? (enable only smooth scroll and nothing else by enabling multigrid).

With multigrid I have this:
Screenshot 2023-03-18 at 7 04 37 PM

Without multigrid I have this (I want to achieve this look for floats, but having smooth scroll): 
image

tobealive:
@xbladesub 
Afaik there is no pseudo transparency with multigrid(showing the background behind the neovim client in floating windows). Rather it's a true transparency that shows the neovim content behind the floating window. You fiddle with the `vim.o.winblend` option to find a fitting transparency. Telescope has it's own `winblend` key in it's `config.defaults`.

fredizzimo:
@xbladesub 
1. It indeed looks like it jumps. I think it's the windows animations and you have `postion_animation_length` set to a very low non-zero value, so it animates in just a couple of frames. What makes it worse, is that it animates from the top rather than the bottom. I don't think that's a regression introduced here, but something that's always been there. But still that's a bug.
2. I don't think that's a bug, since the actual text does not change either either, just the cursor moves. Maybe you have re-mapped ctrl-d to mean move the cursor a half screen?

For your last question, @tobealive gave a good answer. Without multi-grid a fake transparency is used, and that's not the same as a real transparency, so it will look slightly different. I'm not sure why it's that much different though.

src/renderer/rendered_window.rs
```diff
@@ -225,26 +220,80 @@ impl RenderedWindow {
...
+            if let Some(background_picture) = &line.background_picture {
+                has_transparency |= line.has_transparency;
+                canvas.draw_picture(background_picture, Some(matrix), Some(&background_paint));
+            }
+        }
+        let mut foreground_paint = Paint::default();
+        foreground_paint.set_blend_mode(BlendMode::SrcOver);
+        for (matrix, line) in &lines {
+            if let Some(foreground_picture) = &line.foreground_picture {
+                canvas.draw_picture(foreground_picture, Some(matrix), Some(&foreground_paint));
```
fredizzimo, line 291: /rant
I hate Skia, it's a big black box, and I have no idea what's going on. The approach I'm taking here is way too slow. 

With the code as it is, and my rendering changes in opengl on an rtx 3060 with a full screen 3840x1600 and a telescope window open, without blur and transparency it can render only around 90FPS. 

Commenting out this line, that is drawing only the background, so jsut a few hundred boxes gives just 160 FPS
Additionally commenting out the the background drawing gives 2000 FPS (that's more like it)

Note that I'm still rendereing from render target, which includes the `clear`  comand, so it shouldn't be some kind of strange synchronization issue between the reneder target and the backbuffer.

And another Skia rant, and upgrading to a newer version of Skia seems to force VSync on in some strange way incompatible with my other rendering changes.
/rant

This function probably does not need to run on every rendered frame. Only when the particular window has changed, which is every frame when scrolling. But at least it doesn't need to be run for the static windows.

But still something else has to be done, even if the FPS is ok for me (still can't render at my display refresh rate of 144 though), it probably won't with slower computers. And even if it is, it's not nice to consume a lot fo power and have loud fans just for rendering with Neovide.

I don't think storing pre-computed textures for each line is the answer either, the work still needs to be done, and I see no reason why this would have to be done. 

The old snapshot based approach is not very good either, it uses a lot of GPU memory, and it complicates a lot of the code.

So I will try to figure out what Skia is doing and see if there's anything to be done. But any ideas are welcome. The profiling I have done does not really help. I also have some very experimental internal profiling of Skia which I used before, but it turned out that is not very useful either.

A-Lamia: @fredizzimo this might sound insane but if skia is a problem is there an alternative to skia that you think would be a better fit, or is it faster to keep skia and optimize around it ?

fredizzimo: @A-Lamia, that's something I have been thinking about multiple times, I have spent so much time fighting with Skia, not just this time. It's just that it finally cooked over for me, when I though that I'm very close to have the rendering changes ready for a PR. It's also not just that it's a blackbox, the documentation is one of the worst I have seen for a software project that big. But on the other hand Skia is written by google for Google using it for Chrome, optimized for their usage patterns and so on.

Glfw is one framework that I have been considering https://www.glfw.org/, with rust support here https://docs.rs/glfw/latest/glfw/. That would also let us get rid of another of our painpoints winit, since it also includes window, keyboard and mouse handling. The problem is that it is only opengl, and from my experience I have not managed to get Windowed opengl smooth in Windows. Even examples by Nvidia are a bit hit and miss, it seems if you load the GPU enough, then it can be smooth, but otherwise there will be some choppyness.

I have not looked deeply at other options, but sdl might be possible https://www.libsdl.org/, that would probably also let us get rid of winit

Even using straight opengl and direct 3d would not be too hard. But all the animation effects would need to be re-done, but could probably be done completly as shaders, give much greater possibilities.

The thing is that all of this is a lot of effort, and still some uncertainy about the results.

A-Lamia: @fredizzimo glfw seems to support vulkan would that be on par with direct 3d ?

fredizzimo: I'm not sure about Vulkan, it's rather new and we might also run into issues with no support for older hardware. I'm also not sure if that would resolve the problems on Windows.

But I think I found a plan now. There seems to be a lot of overhead for each `draw_picture` it creates new command buffers, and flushes a lot of state. So the plan would be to instead do some preprocessing so, I can run a single `draw_vertices` for the backgrounds, and also a single skia call for all the glyps on the screen. And also some preprocesing for all the underlines.

A-Lamia: awesome, let us know how it goes! :)

shaunsingh: Funnily enough this project was originally based on sdl2, but we switched to winit + openGL after it gave us quite a few problems. The rust bindings might be more mature now though

fredizzimo: The thing I really don't understand where this performance regression has come frome. 

I finally got Skia 0.52 working again, after fixing its source code a bit. It stopped compiling from source after I updated Visual Studio. But the problem remains. (There's no binary package available for 0.52 and the the d3d/opengl combination. But there is for 0.54 and onwards, so I updated to that version in my rendering branch).

It was working fine before and super smooth, with almost the same drawing function as I have now in this PR. But now it's running 30 FPS when I open telescope, so even slower than the opengl version.

So I guess I should see if I can find some older versions either in some backup branch that I might have or in the git reflog, if that gives some clue.

I know that Skia should be able to handle this, but the problem is that the almost non-existent documentation and how abstracted it is from the actual hardware rendering that it does in the end. And the source code is also very complex and hard to follow with a lot of strange callbacks.

CertainLach: Offtop: wgpu (https://wgpu.rs/) also supports vulkan and opengl, and has pretty good rust-native API. Not as easy as raw opengl (the pipeline initialization part), but more convenient than raw vulkan. Some examples can be seen here: https://sotrh.github.io/learn-wgpu, https://github.com/gfx-rs/wgpu/tree/master/wgpu/examples

fredizzimo: It turns out that setting a blend mode for `draw_picture`, like `Some(&foreground_paint)` is a massive slowdown. Even if it's just `BlendMode::Src` 

But commenting the background drawing out and and drawing the foreground without a blend mode gives black windows with only the bottom line visible. That seems to indicate that there's a massive overdraw, and that I need to set a clip region. I thought that the picture was only the size I created it with, but apparently not... 

On the other hand, that's not the same as `BlendMode::Src`, which correctly draws each line. So I have no idea what's going on actually. Like I said Skia is a big black box. 

I had a similar experience when trying to fix the transparency for my second pass of this PR, the blend modes and everything seemed to interact in a very strange way. And add to the fact that skia uses premultiplied alpha for all the render target surfaces, at least I think so, although I don't see anything in the official documentation. The chromium doucmentation is a little bit more explicit though https://www.chromium.org/developers/design-documents/graphics-and-skia/. I just eneded up doing trial and error for several hours until I found some magic combination that worked.

Anyway, I will continue this investigation another day.

fredizzimo: @CertainLach wgpu definitely looks interesting. I will take a closer look at it at some point.

Kethku: I've done some playing with wgpu and did some initial ports of zed's shader to it's system. I would LOVE if we could swap to it, but just havent put in the time to really go and do it.

We depend on skia for the text rasterization, font fallback, and rendering you've been discussing above. Text rasterization can be done pretty simply using swash, but font fallback is a much more complicated problem.

If we can address font fallback, I see no reason why we couldn't spend some effort swapping over pretty quickly.

fredizzimo: For the information. I stated experimenting a bit with wgpu here https://github.com/fredizzimo/neovide/tree/fsundvik/wgpu

It currently renders the background, with the correct colors, so you can see the status bar for example. Obviously no text yet, and there's som problems with scaling, the line height is not as tall as it should be.

For the fonts I was planning to keep our current swash shaper (at least for now, although I think we should switch at some pint), and use this https://github.com/alacritty/crossfont as the rasterizer and it looks like it handles the fallbaks as well. Wezterm has a nice atlas implementation https://github.com/wez/wezterm/blob/3666303c7b26c6c966b3f136dbb954686d334cc3/window/src/bitmaps/atlas.rs for caching of the actual rastered glyphs.

Another option is to use almost everyting from Wezterm here https://github.com/wez/wezterm/tree/3666303c7b26c6c966b3f136dbb954686d334cc3/wezterm-font. But I'm not so sure that using freetype on all platforms is a good idea.

I won't be spending a lot of time on this though, so if it's not working out smoothly, someone else can take over, or we can just keep on using Skia.

NOTE: I had to fight a lot with lifetime issues with wgpu, it really wants you to architect everything in a very specific way. But other than that everything has been smooth and quick to get up running so far. On the other hand, I'm still not that confident in writing rust, I really need to read some proper book that goes a bit deeper than "The Book".

Kethku: I'm curious about your desire to move away from swash. Can you elaborate?

I've also been working on a wgpu expirement of my own but using rust-gpu for the shader. Idk if this is a viable cross platform option but it has some really nice side effects like defining the instance types in rust and using them both shader and cpu side. I haven't gotten as far as you, but I've been documenting some of the progress here https://kaylees.dev/trio/maple/day5-quad-shader/

fredizzimo: Regarding swash, mostly because Harfbuzz is the standard almost everywhere and https://github.com/RazrFalcon/rustybuzz is available for Rust, so it would guarantee that the same font is shaped in the same way across all the applications a single user uses. 

We also have some issues where it's clear that we are rendering certain fonts in the wrong way and different than some terminal applications we are compared to. I'm not sure if that's a shaping issue, rasterizing issue, or that we don't use the metrics we get out of the shaper correctly though.

For the shader I would rather implement hot reload, so you can modify the shader on the fly, like shadertoy for example https://webglfundamentals.org/webgl/lessons/webgl-shadertoy.html. But maybe that model is compatible with rust-gpu as well, I will check it out later, and also read your blog.

And a small update on my progress, I have now replaced the skia font loading by font-kit in my branch. But there's at least one drawback, the system font fallback is currently Windows only, in font-kit. Our regular fallbacks should work though, it's just the system part that is missing. It should not be too hard to add support for the other platforms and send a pull request for them, especially since we have both the cross-font and Skia as examples, and it's just a matter of calling the correct system function on each platform. 

So, my next step is actual rasterization and rendering of the fonts. I don't think I will use the atlas from wezterm as I first planned, since our is much more simple with just single or double width characters.

Kethku: We used font-kit in the past. I think at some point we swapped to just skia and swash to reduce dependencies. I also recall fontkit's glyph properties being strange but I don't remember the details.

Kethku: As for rustybuzz, the swap to swash was about performance. For whatever reason the shaping with rustybuzz was pretty slow. We have made some upstream issues in swash about font problems and at least a couple of times gotten them fixed. I dont feel super strongly either way though.

A-Lamia: Am i insane or did smooth scroll stop working ?

Kethku:
I've been using this for the last month and have some mixed feelings. I think the scroll_delta in 0.9 achieves what this is trying to do without as many limitations.

Using the scroll delta like this has the previously highlighted problems with not working for larger jumps, and being explicitly recommended against in the ui documentation. I also have found it to generally fail at random times in ways that I can only explain as nvim deciding arbitrarily to render something without the scroll event.

So my vote would be to not merge this as is. The rendering changes seem worth while though along with the tracing changes in the other branch.

Obviously this is just my two cents. I'd be curious about your thoughts. I can also try to get more specific examples of the scrolling not working. Just haven't been able to track it down so far because it is intermittent...

Kethku:
Aight heres my plan. I will merge the change that updates smooth scrolling to use scroll_delta today and hopefully cut a release so that people can use 0.9. Once that settles down, we should come back to how this PR should get integrated.

I'm also working behind the scenes slowly on a new renderer based on wgpu that removes skia. So its possible that that will get close before this is ready to merge at which point we should consider which is more useful, merging this or pushing as quickly as possible on the new renderer.

@fredizzimo lmk if you have thoughts or concerns. I appreciate your work on this <3

fredizzimo:
I'm not sure if your new scroll using the delta events works any better, after a very brief testing,

1. For large scrolls half the screen goes blank, I personally prefer it to jump to the destination instead of trying to show something that's not available.
2. Similarly for the telescope windows, it scrolls the list up, while tfull list should be visible immediately.
3. It glitches out with folds. If you fold a function, and scroll it in and out a few times, there will be some flickering. But I think that's just a bug, and not something that's impossible to do with that method.

Neither method can deal with scrolling more than a screen at a time, properly, not even in theory. In order to do that we need all the lines between the current position and the destination position. So there are two options, either do a jump or render a blank (fully or partially) screen and start scrolling in the lines. 

I haven't seen that many issues, but it's hard to see because of the base framerate, but I'm sure there are a lot of glitches for anything you might have in the scrolling snapshots in between, there's no guarantee that it's correct. So I suggest that you blank out anything that you can't guarantee, and perhaps add an option to do a hard jump instead.

Finally it should be possible to adapt this MR to use the scroll delta, it's still possible to store individual lines, as long as everything is invalidated when there bigger jumps than the screen. In fact it should not be too hard, and I can do that at some point, but I need to be fully convinced that the method doesn't have any edge cases that are impossible.

Kethku:
I personally prefer the gaps for larger jumps because it implies to me that I have made a larger jump rather than just swapping the content. The animating of the previous content out and the new content in is key for me personally to keep some idea of where I have gone. That said, it would be trivial to add a setting which does a hard cut for these cases.

The fold issue is an interesting one that I need to investigate further. I suspect it will be fixable or might require another change upstream. I will look into it. Its also possible that the change in animation curves is responsible for what I thought was just a lack of animation at all. I haven't dug too deeply into what the changes actually were.

To me the intentional call out that scroll events in the grid rendering api should not be used for scrolls is reason enough to avoid that approach. I also can't explain the exact issue as its intermittent but its clear to me that those scroll events don't cover all of the cases where an animation should happen. The lack of consistency is what had me reaching for a different approach in the first place.

I think we now have the tools necessary to make a truly glitch free scrolling that handles all of the different requirements maybe with some minor config for the corners. At the very least its possible now where the approach wasn't before which has me hopeful for the first time in months :P

fredizzimo:
@Kethku, I have re-implemented this using the `win_viewport::scroll_delta` https://github.com/fredizzimo/neovide/tree/fsundvik/remove-snapshots, but it's still needs some cleanup and finalization of the unit tests. 

Unfortunately it has the same glitches as your implementation, because `scroll_delta` as it's implementated in Neovim currently is flawed and buggy. Both your implementation in the latest Neovide release and my implementation is almost unusable with wrapped or folded lines, it glitches out very frequently, perhaps every 10th scroll event for me, by just scrolling up and down a few times. It's not too bad due to the slow rendering we currently have, but when the rendering and everything is smooth it's very noticible. This is the mappping I'm using, for the scrolling, and you don't even need to let the folded lines go out of the view for it to glich. 
```
nmap <Down> 3<C-e>
nmap <Up> 3<C-y>
```

Furthermore the implementation got quite complex since we are now forced to keep track of two different things.
1. The internal Neovim representation of it's screen, and the one it's modifying
2. The scrollback and screen that we are showing

The reason for the glitches are that when we receive a `win_viewport` event, we are forced to copy what we have to the scrollback buffer. The problem is that the event is not necessarily the first one in a batch, therefore the representation of the screen may not be consistent at this point. In your implementation you have a similar problem, because you take an image snapshot at the time of receiving `win_viewport`. I have not noticed any problems in practice related to this, but my testing is still quite limited., An any case, the Neovim documentation does not specify this at all.

The bigger problem is that the `win_viewport` event in some cases seems to force flush, even when you are in the middle of a scrolling operation. Take a look at the following sequence that we get from Neovim

The first batch
```
GridLine {
    grid: 4,
    row: 0,
    column_start: 126,
    cells: [...] }
WindowViewport { 
    grid: 2,
    top_line: 27.0,
    bottom_line: 113.0,
    current_line: 65.0,
    current_column: 0.0,
    line_count: Some(456.0), 
    scroll_delta: Some(3.0) }
Flush
```
You can ignore the first `grid_line` event, since it belongs to grid 4, which we are not interested in here. Then we have the `win_viewport` event that signals us to update the scrollback buffer, and start the smooth scroll opeartion

But then there's the `flush` event, which is the real problem. We have now scrolled the scrollback buffer, but the current view has not been updated, so everything is in an inconsistent state. And if we update the current view here, there are now three empty lines, that we haven't received yet. 

The second batch
```
Scroll { 
    grid: 2,
    top: 0,
    bottom: 66,
    left: 0,
    right: 137,
    rows: 3,
    columns: 0 }
GridLine {
    grid: 2,
    row: 63,
    column_start: 0,
    cells: [...] })
GridLine { 
    grid: 2,
    row: 64, 
    column_start: 0,
    cells: [...] }
GridLine {
    grid: 2,
    row: 65,
    column_start: 0,
    cells: [...] }
WindowViewport {
    grid: 2,
    top_line: 27.0,
    bottom_line: 113.0,
    current_line: 65.0,
    current_column: 0.0,
    line_count: Some(456.0),
    scroll_delta: Some(0.0) }
CursorGoto {
    grid: 2,
    row: 38,
    column: 4 }
Flush
```

Now it starts with the `grid_scroll` event to update it's internal buffer. Top and left is 0, bottom is the buffer height and right is the buffer width. Rows matches the amount of virtual lines scrolled in the virtual viewport event. These were the same conditions that I use and assume are real scroll events and not internal updates in this PR. I have personally never seen this not being the case, but @Kethku has reported some random problems, so there probably are. Furtermore, as pointed out before in this PR, Neovim documentation also specifically tells  to not rely on it.  In any case the real cause of the problems he observed  could also be some rounding errors or something else.

I have also studied the source code of Neovim, and so far I haven't found a case where this would not be true. Well perhaps deleting all lines in the buffer or pasting at the top, could cause it, but I was not able to repreoduce it that way either. In any case, there would be no big trouble dealing with these two the same way as scrolling, the screen would just animate instead of the operation being instant.

In any case after the scroll, Neovim fills in the missing lines, so nothing special going on there.

We then have a redundant `win_viewport` event. As far as I can see it has no purpose, but perhaps it's left there from how it was working before the `scroll_delta` was added. A really bad hack to work around the problem described here, could be to ignore any flush events between a `win_viewport` with a delta and one without. But that is in no way guranteed, and complicates the implementation even further.

@justinmk. As a member of the Neovide organization, I want to start by saying that I'm sorry, that we pushed you to merge something that we hadn't even done a proof of concept implementation for on the Neovide side. I don't know how that, happened, especially since I already had this PR open at that time, but what's done is done, so now we need to find a way to work with what we have, this message should provide all context needed, so there's no need to read through the various discussions about transparency and other things in the PR

With this MR I planned to see if the approach I took using only the `grid_scroll` events was feasible. And except for some minor bugs it seems like it is. That being the case I had then planned to request for a documentation change, to just standardize the existing behviour for multigrid, so that we don't rely on something undefined.  The scroll event could then have been further extended with the number of virutual lines, and a field telling if it was caused by scrolling or something else, adapting the code that @mdeiml did.

I would have preferred that approach, since then we have an atomic event that scrolls both the scrollback and what's visible on the screen, with the new `scroll_delta`, even if it was implemented correctly, we don't have that, which complicates stuff way more than needed.

But I think we can find a way to work with what we have now, if `win_viewport` with a scroll delta, is guaranteed to be followed by a `grid_scroll` of the whole window, then we can simply just store the delta and use that instead of of the number of lines specified in the `grid_scroll` event, and threat the sequence as one atomic event. I think that's what's happening already, so this might just be a documentation change, otherwise probably some minor bugfixing. I will try an implementation based on this now, and see how it behaves.

Note that this still won't allow us to implement perfect scrolling, which needs all the lines between the current position and the destination to be sent. Or at least all the lines that are going to be rendered. I believe the best way to implement that would be some sort of lua plugin, that processed a scroll event, and sends the needed lines to Neovide, so apart from the scroll event, I don't think it would need much support Neovim's side. But anyway that's a discussion for the future.

And @mdeiml, I'm not blaming you for this, I don't think it was you that pushed for it to be merged. The bug with the extra flush probably does not happen in your unit tests, It's probably a time sensitive bug that depends on a lot of things like asynchronous highlighting. But I do think that next time, we should also have a corresponding Neovide implementation, that we thoroughly test, to verify that what we request is actually what we want, and that it works like it should.

Kethku:
@fredizzimo fwiw I'm a woman. I edited your comment to reflect that.

fredizzimo:
@Kethku, I'm sorry, I should have checked your profile instead of assuming and going by old memory. I will not repeat the mistake 🙂

fredizzimo:
Unfortunately my assumption of receiving a `win_viewport` immediately followed by `grid_scroll` (allowing a flush between) does not hold. 

For example when opening telescope, you first get a `win_viewport` with `scroll_delta`, then just a bunch of `grid_line` events to update the viewport. In this case I did not get a flush between, but based on my earlier investigation, that's in now way guaranteed.

So, it looks like I would have to use the second workaround, of threating the second `win_viewport` without a delta as the actual commit of the scrolling, if we want it to work with the current version of Neovim. 

@Kethku, I think I also found a middle ground between scrolling in all lines when doing a long jump and jump immediately to the destination. We could do a direct jump to some offset close to the destination and then scroll in the missing lines only. That would give some feedback, and make the implementation a lot more simple. 

I realised that much of the complication with my new implementation comes from having to implement some kind of sparse `VecDeque`. But with the one up to a screen worth of empty lines trick, that's not needed, and a normal ringbuffer could be used instead. If you want close to the current behaviour, you would could use a screen, and if you only want a tiny bit of feedback just an one line offset.

fredizzimo:
That hack seems to work, I don't like it, but at least I haven't seen any problems in practice. It's implemented in this commit https://github.com/neovide/neovide/pull/1790/commits/29397eb3f5a7b05fa425fdfd79982a976e9befc6

I also added `g:neovide_scroll_animation_far_scroll_lines`, which controls how many empty lines are scrolled when you scroll more than one screen. It defaults to 1, but you can set it to 9999 for example to emulate the old behaviour of scrolling in a completely empty screen.

fredizzimo:
I have been thinking more about this, if  `win_viewport` is used to implement smooth scrolling and a scrollback buffer we need one of the following requirements: 

1. `win_viewport` is always called before any modifications are done to the destination buffer position, and no flush events should be sent before the destination is completely updated. Theoretically there could be multiple viewport updates in the same batch, but for simplicity that should not be allowed.Also if there is no viewport events, the position from the previous batch is used.
2. `win_viewport` is called as the last event in the batch, this is how it mostly works now. The whole batch should operatate on the same viewport.
3. Is a variation of 2. `win_viewport` is called as the first event in the batch, the whole batch should operate on the same viewport
4. Another variation `win_viewport` is called zero or one times per batch, the whole batch operates on that viewport. If no viewport event is received, the previous viewport is used.

On the Neovide side, 4. is the most difficult to implement and requires buffering of the whole batch before doing anything. So unless really necessary, that should be avoided

1. and 3. Could be implemented mostly in the same way. As soon as `win_viewport` is called, we copy the current viewport state to the destination. The viewport also remains in the scrollback buffer as it is. This is quite inefficient, since the copied viewport will most likely be scrolled and partially updated afterwards. It's not a huge deal, the lines could be some kind of reference counted data structure, which is cheap to copy, much like the `SkiaPicture` that I store in this PR, and even if not it's not a huge amount of data to copy.
2. Either needs to buffer everything to see if there's a viewport event, or always copy the current state at the start of every batch, modify that and finally copy it to the destination. 

So based on this, I would prefer 2, but as mentioned that is an undocumented API breakage, so clients relying on the event being the last, I think even older version of Neovide does,  but I don't remember exactly how it works would break.

The absolute best would be for `win_viewport` and `grid_scroll` to be combined and sent as the first event in the batch. That way only the lines that stay the same needs to be copied to the new destination, for normal scrolls that's most likely none of them. So only the scrollback position needs to be updated, which for the ringbuffer I'm using here is very cheap.

fredizzimo:
I have opened two issues in the Neovim repository.

https://github.com/neovim/neovim/issues/23609
https://github.com/neovim/neovim/issues/23610

fredizzimo:
I improved the rendering speed a bit, it should be acceptable now, although much bigger changes are coming soon with my rendering loop changes.

I also got rid of all the surface textures, that should save a bit more memory, now we only need memory for the back and front buffers. Compared to all the snapshots for each floating windows that we had before this PR.

A-Lamia:
awesome i will give it a go. :)

A-Lamia:
@fredizzimo I'm not sure if it's a placebo but it still feels laggy but it feels like a slightly more consistent less lag rather then the more random nature from the before patch.

fredizzimo:
@A-Lamia, yes, the final lag will only be gone with my next update with the improved rendering loop.

falcucci:
I was trying out those changes and it sounds really good. but some possible bugs:

- the background in selected lines arent visible anymore
- the cursor in the terminal. as you can see in the following video.

https://github.com/neovide/neovide/assets/33763843/65c2c0e7-2f60-4d41-ae44-5dad151626e3

expected:

https://github.com/neovide/neovide/assets/33763843/a1e4b1a7-c05c-4ba7-8016-13b2391e34ad

when I have some free time I can take a look to go deep into this issue.

for now this is just a report.

fredizzimo:
@falcucci, I wonder if you are using some old version of this branch? There was a bug like that, but it was fixed a long time ago, and I can't repeat it either.

I have rebased multiple times, so it's possible that just doing git pull won't update the branch.

fredizzimo:
I have now implemented the rendering loop changes here #1870, which includes these changes, so I suggest that you switch over to testing that instead.

A-Lamia:
@fredizzimo getting an error building this think it might be that skia issue with skia, don't have time right now to investigate it but thought i would just mention it.

fredizzimo:
@A-Lamia, I assume you ment to comment on the other PR. For the error, I think you need to upgrade Visual Studio.

A-Lamia:
@fredizzimo sorry yeah it was supposed to be for the other thread and you are right needed to update MSVC, im so used to using clang and just `scoop update` that i forget about manually updating MSVC zzzz.

falcucci:
@fredizzimo thank you, I have updated now that changes and in fact I have not seen this comment which makes sense.


> @A-Lamia, no keep reporting any issues you find. I'm not sure when I will have time, but I will come back to this and start fixing.
> 
> * The `scroll_animation_length=0`, that should be easy to fix
> * The transparency issues. Did you switch back to the latest of this branch? Do you have `neovide_transparency` set. This setting now applies globally to all backgrounds of all windows, including floating ones and the statusline, so that you always can see your desktop behind. This transparency is multiplied by what neovide internally sets, `winblend`, `pumblend`, highligt groups and so on It's hardfor me to tell if it works like that in your images or not.
>   I'm also not sure if that's the best design, so any suggestions here are welcome. Personally I think you should ignore `neovide_transparency`, and tune your color scheme instead. Complete control through that should now be possible.
> * The visual select artifacts. Yes I see the same, I guess there's some rounding issues somewhere, I'll try to figure that out.
> * The stuttering. It's mostly caused by the way we implement our frame loop, but also a big part is caused by Windows having very poor windowed mode opengl support, on Linux it's much smoother even without further rendering changes. Anyway, this all is going to be fixed with my rendering loop changes and by using Direct3D on Windows, so that everything becomes pixel smooth.
> * **Edit** for the background blur, yes that should be applied for any floating window that has transparency (but it's not currently enabled by just setting `neovide_transparency`

One small thing that I have noticed is that `let g:neovide_background_color = '#000' has not been applied anymore.

check the tittle bar's color with the neovide_background_color setted.

https://github.com/neovide/neovide/assets/33763843/171f7eba-7544-4857-9ce2-32f0ae14418e

curious enough it updates correctly if I untoggle/toggle neovide.

the configs only applies when I (un)toggle neovide transparency mode with the following config.

```vim
    let g:neovide_transparency = 0.0
    let g:transparency = 0.7
    let g:neovide_background_color = '#0f1117'.printf('%x', float2nr(255 * g:transparency))
```

the problem here is that even when applied it kepts a black divider line between the tittle bar and the window.

Screenshot 2023-05-19 at 14 12 25

falcucci:
Another issue is that if you have a file opened and then open another file over it in the same buffer, the previous file stills visible in the background.
___ 
neovide

fredizzimo:
@falcucci, I don't have a Mac to test on, but it looks like this does not work anymore for some reason https://github.com/neovide/neovide/pull/1790/files#diff-3236d45aa6e68d88a718f7a9e523800f3a3da86bfb4114a11c320382149b4b24L513

It would be great if someone else could check what the issue is

falcucci:
@fredizzimo sure I am just reporting for now, but I can take a look in this new changes and try to go deep through it.

fredizzimo:
I'm not sure what to do here. This works decently for normal operation but the terminal is frequently messed up, Neovim tells that one line is scrolled even if multiple lines are scrolled. And that might not even be the only bug there is.
 
I think my first implementation using  the scroll event was more reliable and worked better, although it reverted to non-smooth scrolling more often. But I'm not sure, I did not use it in real world scenarios myself. Does anyone else have some feedback? I know that @Kethku had problems with it. 

Another thing I also noticed is that when you use some terminal application like less, which does it's own internal scrolling it's quite choppy. The reason for that seems to be that for some reasons the timing between the events we get from Neovim varies a lot. On my system I get varying times between around 20 and 50ms for what should be a stable key repeat of 25 ms. But because that furthermore does not match up with the frame rate timing, it can vary even more than that. I think we could use some kind of filtering here, and try to apply the changes at a more fixed rate.

last-partizan:
Maybe we can disable smooth scrolling for terminal? wezterm aren't doing it for a reason

https://github.com/wez/wezterm/issues/3812#issuecomment-1574239625

fredizzimo:
I don't think that smooth scrolling is fundamentally incompatible with a terminal. Yes, you can write and clear any parts of the visible screen, but as soon as a new line is added old lines are scrolled out to the scroll back buffer, and that part can be smoothly scrolled.

A program like `less` or terminal `neovim`  that does it's own internal scrolling is impossible to deal with though. But as explained in my second part of the message, I think we could do better with that. And I think the same improvement applies to running without multigrid as well. If we somehow can smooth out consecutive updates to happen at regular intervals synced with the frame rate, it will look much smoother.

But yes, for now it's probably best to disable smooth scrolling for the terminal, since it's currently more annoying than useful. I think it should be implemented in a generic way like suggested here https://github.com/neovide/neovide/issues/813, but per window, not buffer. Then add an auto-command that disables it for terminal windows in terminal mode, but keeps it for normal mode. At some point it's worth investigating deeper why Neovim seems to send the wrong deltas.

src/renderer/rendered_window.rs
fredizzimo: The code has now been moved to `animation_utils.rs` and also documented.

fredizzimo:
I opened this for review now. I think we can leave the transparency fixes on macOS until later, when @AsPulse or someone else has time to look at them.

According to my testing, this does at least not makes things worse than our current main version, so we should be able to merge before https://github.com/neovide/neovide/pull/1870

falcucci:
@fredizzimo thank you for this amazing work.

I will start to use this version to stress it as much as I can on macOS.

I am trying to replicate the same experience as my current (main) version and for now I have problems with:

- `g:neovide_transparency` is now a bit tricky. if there is any value < 1 the tranparency is applied in the cursor (when use the terminal you can replicate it I guess);
    - the transparency doesnt respect the level applied in the manu bar: if you add 0.5, the content goes to 0.5 but the menu bar goes 0 apparently;
    - the point above can be related with the `neovide_background_color` value. This is mine if you want to replicate:
    ```bash
    let g:neovide_background_color = g:background_color.printf('%x', float2nr(255 * g:neovide_transparency))
    ```

falcucci:
the work around to make the **visual** similar to my current main branch version:

```
let g:neovide_transparency = 0
let g:neovide_background_color = g:background_color.printf('%x', float2nr(255 * 0.6))
```

falcucci:
ops, the `scroll_animation_length` must be changed to ~0.1 also to have the similar smooth scroll, at least for me.

but this is relative to the user as you mentioned in the PR description.

fredizzimo:
@falcucci, you should test https://github.com/neovide/neovide/pull/1870. That one includes these changes, but also makes everything much smoother, they will be merged together.

The transparency issues on macOS are known and @aspulse  has been working on fixing that, but unfortunately those fixes broke the other platforms. https://github.com/neovide/neovide/pull/1870#discussion_r1248488612. I don't know if he is going to attempt to do another fix, or if we need to find someone else to fix that properly.

falcucci:
oh so the plan is to merge #1790 and #1870 together? If so I misunderstood.

I'll check them out.
thank you!

fredizzimo:
@falcucci. Sorry, I was a little bit inaccurate. This will be merged first, to make reviewing the other one easier. But they should still be merged roughly at the same time. And we will never make a release with just this merged and not the other one.

fredizzimo:
Closing this in favour of https://github.com/neovide/neovide/pull/1977

PR #1797, opened 2023-03-13
Encapsulate glutin

fredizzimo:
- Refactor

This moves the Glutin context creation to it's own file. It also passes the winit window around rather than the whole context, so that the code mostly becomes dependent on winit rather than glutin. 

The only direct reference to glutin is in the new `opengl.rs` file, but the context is still used outside of that, so some of the code is still dependent on it at an interface level.

This is done mostly to make it easier to review my upcoming render changes. But it should also be quite easy to switch over to not use a forked version of Glutin anymore. The new Glutin versions, no longer depend on winit, so we can use our own version of it. Instead we could directly do something similar as the Glutin winit module https://github.com/rust-windowing/glutin/tree/master/glutin-winit/src

- No

### DIFF
--- a/src/dimensions.rs
+++ b/src/dimensions.rs
@@ -4,8 +4,8 @@ use std::{
     str::FromStr,
 };
 
-use glutin::dpi::PhysicalSize;
 use serde::{Deserialize, Serialize};
+use winit::dpi::PhysicalSize;
 
 use crate::settings;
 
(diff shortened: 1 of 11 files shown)


shaunsingh:
Code LGTM

Regarding switching to upstream glutin, as per https://github.com/neovide/neovide/pull/1789 the only change we have is https://github.com/neovide/glutin/commit/cf509674b403e0f90f2db41a084e303cbd90fbf5 (apart from one other macOS change that can be upstreamed), our winit fork (https://github.com/neovide/winit/tree/new-keyboard-v3) matches upstream, so we can move everything to upstream if we'd like

(the PR linked above is currently broken on linux, but I believe the changes in this PR would fix it

last-partizan (approved):
Looks good to me!

PR #1192, opened 2022-01-31
Fix modifiers behaviour under X11

last-partizan:
## What kind of change does this PR introduce?
- Fix

## Did this PR introduce a breaking change? 
- No

----

This should fix following issues:

- https://github.com/neovide/neovide/issues/951
- https://github.com/neovide/neovide/issues/1142
- https://github.com/neovide/neovide/issues/942

And, i think, most linux keyboard-related issues on x11.
----

Instead of merging new fixes into `new-keyboard-all`, i created `neovide-x11-nt` for both glutin and neovide, based on this: https://github.com/maroider/winit/pull/6

@Kethku if you remember what else was merged into `new-keyboard-all`, i could check if we still need it, and merge into those new branches.

But, it builds, it works, and i've used it without problems for a day.

My main concern here is not to break something on MacOS/Windows.

PR #1821, opened 2023-03-26
Temporary fix for wayland

rkuklik:
- Fix
- Refactor

_--novsync_ currently doesn't work, some dependencies switched their APIs.

Given that glutin 0.30 completely switched its API, some parts would benefit from partial rewrite (currently there is _Context_ struct in _opengl.rs_ providing glue between old and new parts of code that doesn't look very nice).

This change fixes problems over on wayland, but I wasn't able to test other platforms than linux.

### DIFF
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -59,10 +59,10 @@ fn create_platform_shell_command(command: &str, args: &[&str]) -> Option<StdComm
         let shell = env::var("SHELL").unwrap_or_else(|_| "/bin/sh".to_string());
         let mut result = StdCommand::new(shell);
 
-        result.arg("-c");
         if env::var_os("TERM").is_none() {
             result.arg("-l");
         }
+        result.arg("-c");
         result.arg(format!("{} {}", command, args.join(" ")));
 
         Some(result)
@@ -121,10 +121,10 @@ fn nvim_cmd_impl(bin: &str, args: &[String]) -> TokioCommand {
         .map(|arg| shlex::quote(arg))
         .collect::<Vec<_>>()
         .join(" ");
-    cmd.arg("-c");
     if env::var_os("TERM").is_none() {
         cmd.arg("-l");
     }
+    cmd.arg("-c");
     cmd.arg(&format!("{} {}", bin, args_str));
     cmd
 }
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -411,7 +411,10 @@ pub fn create_window() {
             let monitor_width = monitor_size.width as i32;
             let monitor_height = monitor_size.height as i32;
 
-            let window_position = window.outer_position().ok()?;
+            let window_position = previous_position
+                .filter(|_| !maximized)
+                .or_else(|| window.outer_position().ok())?;
+
             let window_size = window.outer_size();
             let window_width = window_size.width as i32;
             let window_height = window_size.height as i32;


tobealive:
Fixes the keyboard stuck issues on linux x11 for me (related #1192)

The PR might interfere with other ongoing work though.

tobealive:
Some of the recent related works that came to my mind are: #1789 and #1790. The later discusses to move away from winit.

I think it would require to proactively come together to find a good way on how to proceed.

DashieTM:
I tested this out and it works perfectly fine other than some strange things happening with a few keys. 
For example the k key seems to not repeat, but h j and l work fine.
The same thing goes with the arrow keys where right and up work, but down and left are only received once, the repeats are dropped.
Is this perhaps an issue with a certain config? The keys work with xwayland mode, although with lag... and they obviously work in regular nvim.

It is a shame that this wonderful project doesn't work on wayland yet.
It looks and feels amazing, but I just can't justify going back to x11 ever again.

DashieTM:
Seems to be an issue with upstream, there is another fork that tried to integrate newer versions of winit and glutin and the same issue is happening there.

Yesterday17:
> I tested this out and it works perfectly fine other than some strange things happening with a few keys. For example the k key seems to not repeat, but h j and l work fine. The same thing goes with the arrow keys where right and up work, but down and left are only received once, the repeats are dropped. Is this perhaps an issue with a certain config? The keys work with xwayland mode, although with lag... and they obviously work in regular nvim.
> 
> It is a shame that this wonderful project doesn't work on wayland yet. It looks and feels amazing, but I just can't justify going back to x11 ever again.

This seems to be a upstream issue. I left a comment at https://github.com/rust-windowing/winit/commit/639e1d40b13a38d9545e452d9f7e7ae5b77e4740#r108100537.

I tried to fix it by adding `8` as code above. Could you please verify whether the issue still exists on other keys? (I only tried `KeyK`)

DashieTM:
The following keys are also affected: 
left-arrow, down-arrow, up-arrow,m,1,2,3,4,5,6,7
as well as any combination with modifiers -> m and M
It is really weird, as all the other keys work perfectly fine.

fredizzimo:
@rkuklik I cherry-picked your changes into https://github.com/neovide/neovide/pull/1789, but using the upstream glutin-winit instead of embedding it here. It's still unclear to me how we are going to deal with the upstream key repeat bug though.

Could you check that I haven't missed something, and that it still works? And if that's the case, I think we can close this one and start using https://github.com/neovide/neovide/pull/1789 instead.

fredizzimo:
Yes, I will merge @Yesterday17's keyboard repeat fixes. But the other fixes in the PR does not seem to be needed, everything compiles wihtout that.

As for the two version of winit, I also encountered that, but I explictly set the version numbes and used the patch feature of Cargo https://github.com/neovide/neovide/pull/1789/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R96, and ran `cargo update`.

fredizzimo:
The keyboard repeat fix is now included in https://github.com/neovide/neovide/pull/1789

Could someone confirm that everything works there, so we can close this PR?

DashieTM:
I just tested the new-keyboard-v3 branch(#1789), it works! Thank you all for the amazing work!

PR #1829, opened 2023-04-08
Wgpu - Technical preview

fredizzimo:
This is nowhere near ready, but I'm opening a pull request so that we can have some discussions, and to share the progress I'm having towards making a wgpu based renderer.

This also includes a lot of other things, like https://github.com/neovide/neovide/pull/1790, some render loop changes, including a render thread (which I'm not sure is needed anymore). But that will be cleaned up at some point.

It's very late here now, so I won't share much details, but I want to get this up before I go a few for a few days. 

I decided to keep Skia for font loading at the moment and Swash for shaping. And the glyph rasterization from the actual servo web browser engine from here https://github.com/servo/webrender/tree/master/wr_glyph_rasterizer. My initial plan was to use [font-kit](https://github.com/servo/font-kit) for loading and rasterizing. However, it was lacking platform specific fallback code, and for the rasterization, it has the following comment 
>  // TODO(pcwalton): This is woefully incomplete. See WebRender's code for a more complete
>    // implementation.
It also looks nearly unmaintained at the moment. In the long term it might be worth contributing to it and make it do everything though. It's already implemented in may different places, just not as usable libraries.

There's also a lot of other rasterization solutions, including Swash, but I wanted to use something that uses the native method on each platform, so that it's consistent with the rest of the applications. And for that I only know of wr_glyph_rasterizer, font-kit, and Alacritty's crossfont (which unfortunately encapsulates almost all of the functionality, making it impossible to use together with a shaper.

Anyway, I have progressed quite far, and a screenshot can be seen below. It's much closer to correct rendering than it looks, I just need to place the glyphs at the right place with the right scalaing. Now it just stretches out what was rendering to the atlas texture, which varies from glyph to glyph, to fill the whole box. 

I don't recomend to try it out yet, there are so many things that re still work in progress. I will also keep re-basing this branch and write major updates here.

**Edit:** It runs a completely stable 144 Hz on my computer, no micro stuttering at all, not even during the initial scrolling when it's shaping the first time.

image


Kethku:
I'm digging into this and will likely have more thoughts soon.

I have been working over the last week or two on a parallel solution based on rust-gpu. You mentioned concern about not having live reloading of the shaders. I think its possible that live reloading will be an option at some point, but I haven't dug too deeply yet. The main thing I like about the rust-gpu approach is that I can specify instance structs as actual Rust types that get uploaded to the gpu as is and processed on the shader side using the same type. To me this makes the whole stack easier to understand and *hopefully* easy for others to contribute to who don't have experience writing shaders.

To enable this approach, I avoided using vertex buffers and instead upload the instance structs to a uniform buffer which is indexed into in both the fragment and vertex shaders. I'm not sure thats a requirement though as I'm learning as I go.

All that said though, I'm hacking away at this slow and steady so its quite possible you will finish before my version is ready. I think the model I'm working on will be pretty clean but will take a while to be ready. So, one approach would be to land yours when it looks to be ready (which seems like might be pretty soon) and then I can work on adapting your code to a full rust approach and merge it progressively after assuming the benefits I have in mind can be realized.

qinyuhang:
> I'm digging into this and will likely have more thoughts soon.
> 
> I have been working over the last week or two on a parallel solution based on rust-gpu. You mentioned concern about not having live reloading of the shaders. I think its possible that live reloading will be an option at some point, but I haven't dug too deeply yet. The main thing I like about the rust-gpu approach is that I can specify instance structs as actual Rust types that get uploaded to the gpu as is and processed on the shader side using the same type. To me this makes the whole stack easier to understand and _hopefully_ easy for others to contribute to who don't have experience writing shaders.
> 
> To enable this approach, I avoided using vertex buffers and instead upload the instance structs to a uniform buffer which is indexed into in both the fragment and vertex shaders. I'm not sure thats a requirement though as I'm learning as I go.

I agree with using uniform buffer. And maybe we can create unit tests for the renderer part just like `wgpu` repo.

CertainLach:
Regarding font rendering...
What about reusing something some parts of https://github.com/pop-os/cosmic-text?

It looks less complete/correct than similar libraries but is maintained well.
Some projects in the ecosystem are already using/planning to use it, i.e.  https://github.com/iced-rs/iced/pull/1697  https://github.com/grovesNL/glyphon/pull/23 https://github.com/bevyengine/bevy/issues/7616

Kethku:
Cosmic uses swash under the covers and the api is pretty straightforward, so in my cut of the wgpu renderer, I rasterize glyphs using it.

The main add cosmic gives imho is font fallback so I was considering porting their approach but haven't done that yet

fredizzimo:
What I'm most concerned about, with all these rust based rasterizer libraries, is that they all promise speed, but they don't mention the quality, for us as text editor, the quality is the most important, that's why I have been looking at libraries that use the native rendering, DirectWrite on Windows, Core Text on Mac, and FreeType on Linux. This not only gives you a good quality, but more consistent rendering across applications.

That said, I personally prefer modern FreeType to DirectWrite, afer they implemented some of the alogrithms that were patented by Microsoft and now expired. Core Text is obviously great, but we have to remember that it's usually combined with highdpi displays. 

Cosmic Text, will obviously also care about the quality at some point, but I'm not sure that they are at that point yet.

Regarding of my earlier comments about Swash and the shaping part of it, I'm less concerned about it now, they have actually added a great amount of unit tests https://github.com/dfrg/swash/issues/6.

I found out that some of our current issuess could be because we are not rendering at the correct subpixel position vertically (we rounding it to an even pixel), and also that we are wrongly applying clipping to the rendering. But Skia in itself, although using DirectWrite under the hood, also seems to do something strange. Neovide renders the fonts on Windows with a slightly too thich weight compared to other applications. The webrender rasterizer and my renderer has the same thickness as other applications. 

There's still a lot of work to do though, I haven't done anything more than just fixing the scaling issues shown in the screenshot above. But we need at least the following
1. Rebase it again, the branch is currently broken with Neovim 0.9, and that's why I don't include a screenshot in this message
2. Render everything at the correct subpixel position (currently it's ever so slightly blurry because of the wrong position)
3.  Cursor rendering with effects
4. Emoiji rendering
5. Subpixel rendering (tricky especialy with transparency) see this https://github.com/servo/webrender/blob/master/webrender/doc/text-rendering.md
6. Dynamic resizing of the texture atlas

So we need to join our forces and work in the same direction. I'm quite happy with my base architecture code structure, which includes proof of concept hot reloading of shaders. But @Kethku want's to use `rust-gpu` so we should respect her decission. I'm personally not sure about that, it's very early in development, there's almost no online learning material, comparing to the waste amount of examples for wgsl and other shader languages that are very similar. We also need to use rust libraries for simple things like vector and matrix mutliplications, and I'm not sure how optimial that code is, compared to the  shader instructions that are almost 1-1 with the actual hardware. 

If the hot reloading can be done for it, I'm a little bit happier, since that's quite a timesaver when experimenting with effects. Wgpu also has a nice language server that can detect syntax error almost immediatley when you type something, which makes it very pleasant to work with.  As far as I see it the advantages of `rust-gpu` comes when you are building big systems like games, where you have ton's of shaders, or one big shader that does everything, and need to care about re-use and be able to refactor. In Neovide we probably need a handful of shaders at most, so the complexity is not really there.

But it's not a big deal, I can work with either of the technologices.

But before all this can happen, I need to do two things. I strongly believe in the line based rendering, rather than snapshot based, like we had before. So I need get that working in a satisfactory way, see my other MR.

I also need to clean up my render loop changes and fully test it with all configurations. That has the biggest impact on the performace, and should be enough for a solid 60 FPS even with Skia (if you ignore the random stuttering because of OpenGL on Windows). I will make a separate MR out of that.

Kethku:
Let me describe a little of my approach and then we can decide if its worth meeting in the middle or just going with your branch and porting whatever to it. Its unfortunate that we both worked wgpu branches simultaneously, but such is collaborating over the internet.

## Goals

Its helpful to highlight some goals for the effort I've personally been keeping in mind as I think it motivates some of the particular design decisions I've personally been making for my attempt.

1. Cross platform, Rust libraries

Swash appeals in this regard because it will render the same generally on all platforms. Some folks dislike this as it doesn't exactly match their machine, but from a maintenance point of view, its pretty appealing since we only have to get it write once and don't have a change of rendering differently on different platforms. I'm generally VERY impressed with the swash maintainer and from what we've seen so far, any issues we have do get addressed. I also personally appreciate that its written in Rust. Lending our project to testing out a Rust first renderer is a nice bonus.

2. Factor out the Renderer to be used in other projects

A reoccurring request we get in the repo and out is to expose our cursor animations, and rendering platform for other projects such as terminals and editors like Helix. There have even been forks of the codebase hoping to use Neovide's infrastructure to handle rendering for Helix without our involvement. If we can take this refactoring and rewrite as an opportunity to make that easier for folks, that feels like an easy win. I also have a hunch that by paying attention to those use cases, our architecture will become more modular which feels like a worthy goal as well.

3. Render via scenes rather than immediate mode commands

One of my take aways from working at Zed was that its possible to make a more than fast enough renderer which draws primitives from a scene object containing layers, quads, images, and text to render. This approach has a couple of really nice benefits such as serializing the scene struct out on crash for debugging, simplifying our interface (its pretty easy to understand a scene format rather than a bunch of opaque function calls), and writing renderer tests that take a serialized scene and compare the output to some expected baseline. My dev repo https://github.com/Kethku/rusty-2d-renderer implements this approach and supports layers, quads, text (antialiased to respect the background color), and optional blurred backgrounds for each layer.

## Status

So with those goals in mind, heres whats done and what still needs work:

- [x] Text layout and rasterization using swash
- [x] Text rendering from an atlas using the last draw call (including quads) to blend
- [x] Path rendering using lyon to tesselate to triangles and render to an msaa 4 sample antialiased texture
- [x] Instanced quads including transparency
- [x] Clippable layers for rendering each window and with optionally blurred backgrounds
- [ ] Render from a scene object which is passed to the renderer for each frame
- [ ] Scaling out to multiple atlas textures rather than one allocated at the start
- [ ] Implement the higher level grid window api (DrawCommand and WindowDrawCommand in neovide) to make integration easier

I've been doing a very stepwise bottom up approach from scratch rather than forking the Neovide renderer, which means my effort is likely a ways out still. Plus, I have a couple external use cases (some games and helix guis) that I am planning on using this setup for regardless of whether it gets integrated into Neovide, so no harm done if it isn't the right decision to push it through here. I'm going to keep working on it either way.

## Responding

Ok with that out of the way, let me respond to some of your comments directly.

> So we need to join our forces and work in the same direction. I'm quite happy with my base architecture code structure, which includes proof of concept hot reloading of shaders. But @Kethku wants to use rust-gpu so we should respect her decission.

This project isn't my thing, so I'm not making a decision here. I had been working on this renderer on the side for a little while now with Neovide as one possible usecase for it, but that doesn't mean my approach has to be the one we go with. I personally like rust-gpu (despite its youth) because I find rust to be MUCH more productive to work with and understand than wgsl, hlsl, etc even without the live reloading. But thats just me. I also can't stress how nice having shared structs between the gpu and cpu is. Theres a whole class of bugs around format and layout initialization that just go away by using the same struct and data on both sides. However, if thats not what folks on the project want, I'm ok keeping my project just as a side effort for other things.

> But before all this can happen, I need to do two things. I strongly believe in the line based rendering, rather than snapshot based, like we had before. So I need get that working in a satisfactory way, see my other MR.

I agree. One question I have is whether we can get away with recreating the whole scene each frame rather than splitting things up into lines. Today Neovide maintains screen state in two places: the editor which holds the grid of characters, and the renderer which maintains similar windows and information. I would LOVE it if we could keep that data in one place. My theory is that if we make the scene struct high level enough, the grid could create a new one every frame and then the renderer isn't responsible for as much business logic. Thats just a theory at this point though, I haven't gotten that far.

## Summary

All that out of the way, I'm curious about your thoughts. If you feel strongly about your approach and want to see it through, I wont get in your way. Its clear you have a good vision in mind and my slow effort shouldn't get in the way of that. On the other hand, I'm happy to discuss more what I've been working on if that sounds interesting to you. I would be more than happy to work together on it as two pairs of eyes are much better than one. Completely up to you. Its exciting to finally see a way out of the Skia limitations we've delt with for so long. Its great that wgpu and other rust crates are now mature enough to consider such a thing. The future is bright.

src/renderer/pipeline/background.rs
```diff
@@ -0,0 +1,125 @@
...
+        source: ShaderSource::Wgsl(include_str!("../shaders/background.wgsl").into()),
+    });
+
+    let pipeline_layout = device.create_pipeline_layout(&PipelineLayoutDescriptor {
+        label: Some("Background Pipeline Layout"),
+        bind_group_layouts: &[&camera.bind_group_layout],
+        push_constant_ranges: &[],
+    });
+
+    device.create_render_pipeline(&RenderPipelineDescriptor {
```
Yesterday17, line 54:
```
Validation Error

Caused by:
    In Device::create_render_pipeline
      note: label = `Background Pipeline`
    Internal error in VERTEX shader: Metal: program_source:32:19: error: duplicate member 'position'
    metal::float2 position [[attribute(1)]];
                  ^
program_source:31:19: note: previous declaration is here
    metal::float2 position [[attribute(0)]];
                  ^
program_source:41:3: error: type 'vs_mainInput' is not valid for attribute 'stage_in'
  vs_mainInput varyings [[stage_in]]
```

It looks similiar to https://github.com/gfx-rs/naga/issues/1490, which has been open for a year with no response. The same issue exists in glyph pipeline.

src/renderer/renderer.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+impl WGpuRenderer {
+    pub fn new(window: &Window) -> Self {
+        async {
+            let size = window.inner_size();
+
+            // The instance is a handle to our GPU
+            // Backends::all => Vulkan + Metal + DX12 + Browser WebGPU
+            let instance = Instance::new(Backends::DX12);
```
Yesterday17, line 87: Use `Backends::ALL`?

Yesterday17 (commented):
I recently tested out this branch out of curiosity on macOS last week and found some platform-specific issues. Although these are minor issues, I would like to point them out here so that you can take them into account during your future development.

src/window/mod.rs
```diff
@@ -361,165 +386,182 @@ pub fn create_window() {
...
+        log::info!(
+            "window created (scale_factor: {:.4}, font_dimensions: {:?})",
+            scale_factor,
+            renderer.grid_renderer.font_dimensions,
+        );
 
-        let refresh_rate = match focused {
-            FocusedState::Focused | FocusedState::UnfocusedNotDrawn => {
-                SETTINGS.get::<WindowSettings>().refresh_rate as f32
+        let wgpu_renderer = WGpuRenderer::new(&window);
```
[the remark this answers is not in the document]
fredizzimo: Yes, I'm aware, I did it this way for simplicity and because I'm likely going to change the threading model a bit anyway.

src/renderer/pipeline/background.rs
fredizzimo: The two positions are in different structs in the `glyph.wgsl` shader. But it looks like it does not create a proper namespace for the struct on metal. I guess the workaround is to make sure the attribute names are unique, even if they are in different structs.

src/renderer/renderer.rs
fredizzimo: Yes, I forced DX12 for better integration with tools like Pix for Windows, because Vulkan is the default on Windows. 

I think we might want to use DX12 even in the final version on Windows, but with a command line option to switch to Vulkan, so some platform specific code is probably needed here.

fredizzimo:
@Kethku, I will write a more detailed response a bit later with more details,

But just to be able to proceed and avoid further duplication of work, let's use your choice of libraries, Swash and gpu-rs. I don't think we can ever reach a full agreement, but it's far more important to choose something and stick with it.

Kethku:
> @Kethku, I will write a more detailed response a bit later with more details,
> 
> Just to be able to proceed and avoid further duplication of work, let's use your choice of libraries, Swash and gpu-rs. I don't think we can ever reach a full agreement, but you have many valid points, and it's far more important to choose something and stick with it than arguing with the same arguments back and forth.

Ok heres my plan for now. I'm going to keep exploring down the current path implementing the features I know we need for a functioning neovide renderer (font fallback, better atlases etc). Once I am feature complete I will come back and sync up here with the current state of your renderer and really page it in and come up with a list of things that might be good to merge or adapt. That way I don't keep getting in your way.

<3 Said it before, but its exciting to having a way forward!

fredizzimo:
Just a small comment regarding the atlas. I think we can do a really simple implementation for our use case. Almost exactly what I have here already [src/renderer/fonts/atlas.rs](https://github.com/neovide/neovide/pull/1829/files#diff-674eefd333abdc51faaa78fc1ae13c976607276874d3ffc2a192222223dc21db)

That is keep allocating the current row until it's full, and when it's full go to the next row, using the tallest character in the row. The reason for that, is that all the characters have similar heights, and for the ones that don't it does not really matter if we waste a few pixels. I don't see any advanatages of trying to be smart here.

The default altas size should be big enough to easily handle the whole western characterset, at least if you use a reasonable fontsize. But we can't just crash if we run out of space, when using Asian languages for example. So once the atlas becomes full, we flush all the caches, and grow the altas texture, so we always have just a single atlas texture. Again I don't see any reason to try to be smart, we need to be able to deal with an emtpy atlas at the start, so at the most we should drop a few frames when the cache is filled doing this. Trying to evict single glyphs would be slow and/or hard to implement, for very little gain.

Furthermore, if the font size grows to big, we scale it down before putting it into the atlas.

Kethku:
I already did some work integrating etegere which does something smart and pretty fast. It's used all over the place and is pretty well regarded, but I don't have my heart set on it.

The api is pretty damn simple

src/renderer/renderer.rs
Elabajaba: Wgpu DX12 currently uses way too much RAM for an app like neovide (I'll try and get a PR up soon that should be a sizeable reduction, but it'll still be pretty high). If you're just testing with it, you can set the backend with `wgpu::util::backend_bits_from_env` ([example](https://github.com/gfx-rs/wgpu/blob/2571af9557eeba59f00f69649ae6bcd9d5a62392/wgpu/examples/framework.rs#L160)) and use the `WGPU_BACKEND` environment variable (`$env:WGPU_BACKEND="dx12"`, or vulkan/gl/etc in the terminal to temporarily change the env variable for that terminal session and anything launched from that terminal).

Kethku:
Following up as I fell off the face of the planet. I apologize for interfering with the design here. I've worked further with rust-gpu and came to the conclusion you did that its not the right thing for this project. I think its the future of shader work in rust, but still the future and not the present.

fredizzimo:
@Kethku 
No problems :smile:

I have not been working on this either, but managed to optimise Skia and get the smooth renderloop working. Since that works well enough I don't think I will look at wgpu for a while myself. If I continue to work on Neovide things, it will probably be stability fixes and general improvements, there's a quite big need for that when you read the issue list.

But I will probably return to this at some time later, unless you or someone else manages to finsish the work before then. It would be really nice to be able to do all kinds of cool effects, and maybe even run Neovide in the browser at some point. And maybe even `rust-gpu` has stabilised enough by then..

fredizzimo:
I'm closing this for now. We know it's here and I will re-open or create a new PR if I resume working on this.

PR #1832, opened 2023-04-09
Fix opening from Finder in macOS

AThePeanut4:
A fix for a bug introduced in #1758 where opening Neovide from Finder (i.e. directly running the .app) when using the fish shell as your $SHELL fails. Running `neovide` from a terminal works fine.

Previously (in 0.10.3), Neovide would always run `$SHELL -lc "which nvim"` on macOS to find the `nvim` binary. After #1758, it now runs `$SHELL -c -l "which nvim"` only if the TERM environment variable is not set. `bash` and `sh` do not care about the order of the `-c` and `-l` flags, but `fish` does (I haven't tested other shells). This fix simply reorders those flags, fixing the issue.

No


giang-nghg:
Omg thank you I've been going crazy trying to figure this out

Kethku:
Awesome! Thank you so much for the fix <3

PR #1834, opened 2023-04-11
fix: `Download for Windows` link

ErichDonGubler:
- Fix

- No


Kethku:
Thanks for the fix!

PR #1837, opened 2023-04-15
Fixes window position not being remembered on multi-monitor

chomosuke:
- Fix

- No

### The bug
With two monitors and the secondary monitor on the left of the main monitor, neovide would always be positioned on the top left instead of the previous position.

### The cause
On `src/window/mod.rs` line 405, `window.outer_position().ok()?` will always return `PhysicalPosition { x: 0, y: 0 }`. I assume this is because window hasn't been positioned at this point. This led to neovide believing window isn't visible on the current monitor and hence repositioned it to the top left of the current monitor.

### The solution
Use `previous_position` when possible. This fixed the bug.

### My system
I'm using KDE with KWin so this might be a KWin specific problem. But this PR shouldn't break anything as it always falls back to `window.outer_position().ok()?` if `previous_position` can't be used.

### DIFF
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -402,11 +402,10 @@ pub fn create_window() {
             let monitor_width = monitor_size.width as i32;
             let monitor_height = monitor_size.height as i32;
 
-            let window_position = if previous_position.is_none() || maximized {
-                window.outer_position().ok()?
-            } else {
-                previous_position.unwrap()
-            };
+            let window_position = previous_position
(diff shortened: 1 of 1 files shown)


Kethku:
@chomosuke can you fix the clippy error in your change? This fix seems good to go after thats addressed.

chomosuke:
@Kethku Whoops, I fixed it yesterday but forgot to push.

Fixed

Kethku:
Sweet! Thank you so much

PR #1849, opened 2023-04-24
add quake mode accessibility like terminals (with different spaces)

falcucci:
handles Neovide on any screen.

adds a new section to the faq page of the project explaining how one can achieve quake mode accessibility like terminals using hammerspoon.

includes code snippets for adding a key binding to quickly open neovide on the current space. this is done by adding the code to `~/.hammerspoon/init.lua` and binding a specific hotkey to the function. this function will then check if the app is already open and if so, will either open a new window, hide it, or move it to the current space and bring it to the front. finally, if the app is not open, it will open it.

- Documentation

- No

### DIFF
--- a/faq.html
+++ b/faq.html
@@ -250,7 +250,7 @@ any random highlights which have <code>Term</code> in name won't help.</p>
 yours, if it's your own, add it, and if you can't seem to find anything, open an
 issue in the colorscheme's repo.</p>
 
-<h2 id="the-terminal-displays-fallback-colorsterminal-does-not-show-my-colors"><a class="header" href="#the-terminal-displays-fallback-colorsterminal-does-not-show-my-colors">Add Quake mode accessibility like other terminals</a></h2>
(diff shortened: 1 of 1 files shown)


falcucci:
I have noticed that many people like me love this feature (can't live without it anymore) so, as I am using Neovide as my default editor, I am just adding this curious feature to anyone who would like to use it as well.

I am also not sure if this would fit in the FAQ so, feel free to correct me to put it at the most appropriate place. 😊

MultisampledNight:
Hi! Sorry for taking so long.

Two notes:

- The `gh-pages` branch is autogenerated from the `website` folder on the `main` branch. I suggest closing this PR, writing the content in markdown in `website/docs` instead, and opening a new one targeting `main`.
- I'm not sure if the FAQ is the best spot for this, since it's more about questions using Neovide itself. Possibly renaming `Editing w/ External Tools` into `Integration w/ External Tools` and putting the newly added content there instead could be a better fit.

falcucci:
bravo! sounds fair enough. i'll move this PR with the followed fixes.

PR #1850, opened 2023-04-25
Prevent wsl console window shown

nazo6:
- Fix

This PR prevents an additional console window from appearing when starting in `--wsl` mode.

- No

### DIFF
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -54,7 +54,10 @@ fn create_platform_shell_command(command: &str, args: &[&str]) -> Option<StdComm
         result.args(["$SHELL", "-lc"]);
         result.arg(format!("{} {}", command, args.join(" ")));
         #[cfg(windows)]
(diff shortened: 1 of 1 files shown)


src/bridge/command.rs
```diff
@@ -53,6 +53,8 @@ fn create_platform_shell_command(command: &str, args: &[&str]) -> Option<StdComm
         let mut result = StdCommand::new("wsl");
         result.args(["$SHELL", "-lc"]);
         result.arg(format!("{} {}", command, args.join(" ")));
+        #[cfg(windows)]
+        std::os::windows::process::CommandExt::creation_flags(&mut result, 0x0800_0000);
```
fredizzimo, line 57: Could you change this to use the constant from the winapi create, that we are already using?

https://docs.rs/winapi/latest/winapi/um/winbase/constant.CREATE_NO_WINDOW.html

nazo6: Fixed!

fredizzimo:
Thanks, merged.

PR #1885, opened 2023-06-02
Disable srgb on Linux and macOS

fredizzimo:

This disables srgb on Linux and macOs to fix issues with wrong colors being rendered on Wayland with Nvidia. 

It's still enabled on Windows by default  to work around https://github.com/neovide/neovide/issues/907. 

Note that the actual rendering is always non-srgb due to the options we pass to Skia. Enabling srgb only enables the support, so it should really not be needed. 

Disabling it fixes a bug on Wayland with Nvidia cards where the colours are rendered wrong. 

This also adds the full pair of parameters `--srgb`, `--nosrgb`, `--vsync` `--novsync`. So that you always can override what's set in the environment variables. The implementation is not the best due to missing support in clap, see https://github.com/clap-rs/clap/issues/3146 and https://github.com/clap-rs/clap/issues/815

Finally it fixes the environment variables, they have now been renamed to NEOVIDE_VSYNC and NEOVIDE_SRGB,
and now work as they should. Previously `NEOVIDE_NO_SRGB=1` enabled SRGB, so the environment variable value was actually inverted. Note the same fix is included in https://github.com/neovide/neovide/pull/1119 @LoipesMas 

And I almost forgot, it also improves the way the cmd_line tests are run serially and isolates the environment variable setting to affect only one test.

Yes, `NEOVIDE_NO_SRGB=0` is now `NEOVIDE_SRGB=0` and
`NEOVIDE_NO_VSYNC=0` is now `NEOVIDE_VSYNC=1`. 

Srgb is disabled by default on Linux and macOS, previously it was enabled, so it could cause regressions.

NOTE: I have only tested this on X11 and Wayland with KDE Plasma 5.27.5 and Nvidia drivers 525.116.04-1. Wayland did not work properly with this configuration before these changes. Windows should be unchanged, but I haven't tested it, and I don't own a macOS machine.

### DIFF
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -4,6 +4,11 @@ use crate::{dimensions::Dimensions, frame::Frame, settings::*};
 
 use clap::{builder::FalseyValueParser, ArgAction, Parser};
 
+#[cfg(target_os = "windows")]
+const SRGB_DEFAULT: &str = "1";
+#[cfg(not(target_os = "windows"))]
+const SRGB_DEFAULT: &str = "0";
+
 #[derive(Clone, Debug, Parser)]
 #[command(version, about, long_about = None)]
 pub struct CmdLineSettings {
@@ -71,12 +76,12 @@ pub struct CmdLineSettings {
     pub no_tabs: bool,
 
     /// Request sRGB when initializing the window, may help with GPUs with weird pixel
-    /// formats [DEFAULT]
-    #[arg(long = "srgb", env = "NEOVIDE_SRGB", action = ArgAction::SetTrue, default_value = "1", value_parser = FalseyValueParser::new())]
+    /// formats. Default on Windows.
+    #[arg(long = "srgb", env = "NEOVIDE_SRGB", action = ArgAction::SetTrue, default_value = SRGB_DEFAULT, value_parser = FalseyValueParser::new())]
     pub srgb: bool,
 
     /// Do not request sRGB when initializing the window, may help with GPUs with weird pixel
-    /// formats
+    /// formats. Default on Linux and macOs.
     #[arg(long = "nosrgb", action = ArgAction::SetTrue, value_parser = FalseyValueParser::new())]
     pub _nosrgb: bool,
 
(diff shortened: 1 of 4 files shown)


src/cmd_line.rs
```diff
@@ -126,32 +140,33 @@ pub fn handle_command_line_arguments(args: Vec<String>) -> Result<(), String> {
...
     use super::*;
-
-    // Use a mutex to ensure that the settings are initialized and accessed in series
-    lazy_static! {
-        static ref ACCESSING_SETTINGS: Mutex<bool> = Mutex::new(false);
-    }
+    use serial_test::serial;
 
     #[test]
+    #[serial]
```
last-partizan, line 163: Why do we need this serial_test?

fredizzimo: For two reasons. 
1. Settings is a global variable. Previously all the tests used a mutex to run serially, but I made it more obvious this way. It would probably be bettet to refactor the tests to use only local variables, but I did not want to make a big change like that here. It took me some time to figure out why my tests were randomly failing, until I figured it out.
2. Some tests modifies the system enviornment variables and the parsing uses that, so again they can’t be run in parallel

last-partizan: Yeah, it's better this way.

Would be nice to do something like "just run all tests in serial mode", without requiring `#[serial]` on each test, but i don't see a way to do this (except `--test-threads=1`).

last-partizan: https://users.rust-lang.org/t/passing-test-threads-1-to-cargo-test-by-default/87225/13

Looks like there is one option.

But do we have other tests that run in parallel just fine? If so, serial_test for this module seems like good choice. Better than mutex.

last-partizan (approved):

last-partizan:
I think we can resolve conflicts and merge this. Looking good to me.

fredizzimo:
@last-partizan, I rebased it, and also made `_nosrgb` and `_novsync` private as they should be.

last-partizan:
Checked it on linux/wayland. Works fine, merging.

PR #1893, opened 2023-06-11
fix(config): Don't show error when config is missing, instead just log it

last-partizan:
- Fix

- No

### DIFF
--- a/src/settings/config.rs
+++ b/src/settings/config.rs
@@ -76,7 +76,7 @@ impl Config {
 
     fn load_from_path(path: &Path) -> Result<Self, Option<String>> {
         if !path.exists() {
-            return Err(None)
+            return Err(None);
         }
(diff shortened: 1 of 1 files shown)


src/settings/config.rs
```diff
@@ -75,13 +74,14 @@ impl Config {
         }
     }
 
-    fn load_from_path(path: &Path) -> Result<Self, String> {
+    fn load_from_path(path: &Path) -> Result<Self, Option<String>> {
         let toml = std::fs::read_to_string(path).map_err(|e| {
-            format!(
+            log::debug!(
```
LoipesMas, line 79: If you take a look at `src/main.rs`, you'll see that config is being loaded before logger is initialized, so this message won't be logged.

last-partizan: Nice catch, instead of this added path.exists() check.

Issue #1904, opened 2023-06-19
Allow to customize X11 “WM_NAME” (window title)


unclechu:
**Is your feature request related to a problem? Please describe.**

I’m using Rofi to navigate between windows utilizing fuzzy search over their titles. Also I have a currently focused window title on the WM’s bar. It would be more useful if I could customize the window title (`WM_NAME`).

**Describe the solution you'd like**

Window title (`WM_NAME`) is customizable the same way `WM_CLASS` can be customized (via `--x11-wm-class` command-line argument option).

**Describe alternatives you've considered**

Just customizing `WM_CLASS`. It’s not perfect since Rofi renders it in a separate column, and when the names are long it’s kind of shifting everything. All the other apps have a short `WM_CLASS`, just the app name. Also the bar (in my case, with other WMs it can be a part of the window border) shows only the window title, it does not show the class. Using the title to adding different kind of markers seems more suitable.

unclechu:
I was trying to have a look if I can implement it myself. But it seems it’s not that straightforward as there is no dedicated method to set the WM_NAME specifically. And there is no method to set an arbitrary X11 property:

https://docs.rs/winit/latest/winit/platform/x11/trait.WindowBuilderExtX11.html

unclechu:
I probably don’t understand well-enough how X11 works but setting the WM_NAME like this `xprop -set WM_NAME testing` (and clicking on the Neovide window then) sets the property but does not actually change the window title. :thinking: Though this WM_NAME appears on my Rofi navigation between windows list.

I see that the default title is set by `window::WindowBuilder::new().with_title("Neovide")` method call. But I can’t see it anywhere in the props when I do `xprop` inspection on the Neovide window.

unclechu:
To me it seems that any other application (e.g. such as Firefox) set both the WM_NAME property and this window title that does not appear in the list of properties to the same value.

unclechu:
I created a merge request for customizing the window title: https://github.com/neovide/neovide/pull/1907

But the `WM_NAME` is still not set. It (the window title column) renders empty in the `rofi --show window` list of options for instance. So it kind of a step towards closing the issue but it does not closes it yet.

unclechu:
I created an issue for *winit* addressing the `WM_NAME` property customization issue: https://github.com/rust-windowing/winit/issues/2892

PR #1907, opened 2023-06-20
Make window title customizable

unclechu:
Let the user customize the default “Neovide” window title via `--window-title` command-line arguments option.

This change is addressing https://github.com/neovide/neovide/issues/1904 but it doesn’t fully resolve it. It’s still to figure out how to set `WM_NAME` to make the feature complete for X11. Most of the X11 applications seems to set both the title and the `WM_NAME` property.

- Feature

- No

### DIFF
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -97,6 +97,14 @@ pub struct CmdLineSettings {
     #[arg(long = "neovim-bin", env = "NEOVIM_BIN")]
     pub neovim_bin: Option<String>,
 
+    /// Window title
+    #[arg(
+        long = "window-title",
+        env = "NEOVIDE_WINDOW_TITLE",
+        default_value = "Neovide"
+    )]
+    pub window_title: String,
+
     /// The app ID to show to the compositor (Wayland only, useful for setting WM rules)
     #[arg(
         long = "wayland_app_id",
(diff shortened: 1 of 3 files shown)


fredizzimo:
I think this option should be added to the new config file as well, could you add it to `config.rs` as well?

fredizzimo:
Actually, after checking this further. I'm not too sure that this is useful. Neovide follows the [titlestring](https://neovim.io/doc/user/options.html#'titlestring') option. So if you set that, the title will immediately be overwritten. 

Maybe it's better to just use that, would that solve your use case?

unclechu:
@fredizzimo actually this is true, `set title titlestring=smth` works for me. And it also sets the `WM_NAME` :thinking:

fredizzimo:
I think I will close this then. We can always re-open it if the needs change.

unclechu:
Why then `WM_NAME` is not set initially? I assume this is the code that sets the title coming from `titlestring`?

``` rust
    pub fn handle_title_changed(&mut self, new_title: String) {
        self.title = new_title;
    }
```

fredizzimo:
The window might not be fully initialized at that point, and therefore not properly setting `WM_NAME`. That is probably a winit bug, but it could also be caused Neovide doing things wrong. 

As a workaround I think we could call `set_title` again after the window is created. But it's probably better to see what the winit team has to say about it, if you add the additional information to the ticket you created.

PR #1914, opened 2023-06-30
fix: could not draw floating_window with transparent background (when multigrid is enabled)

### LINKED ISSUE #1508: Non-transparent command line when multigrid is enabled
**Describe the bug**
When using `--multigrid` the command line is for some reason not transparent. See screenshot below which shows to equivalent instances where the only difference is that the left uses multigrid.

**To Reproduce**
1. Open neovide with multigrid.
2. Do `:highlight MsgArea guifg=#ffffff guibg=none`
3. See that the command line is black instead of transparent.

**Expected behavior**
Transparent command line.

**Screenshots**
gscreenshot_2022-08-26-093124

 - OS: arch linux 5.19.3-arch1-1
 - Neovide Version 0.10.1
 - Neovim Version v0.8.0-dev-938-g3b2121cedf

**Additional**
Possibly related to #1387

### LINKED ISSUE #1819: Command-line/MsgArea renders with a black background when transparency is enabled and the value is set to none
**Describe the bug**
Unable to make command-line/MsgArea transparent when using Neovide with transparency, 

**To Reproduce**
1. Launch with transparency
2. Attempt to remove the background, screenshot features a minimal init.lua to reproduce

**Expected behavior**
For there to be some way to have no background for the the MsgArea, setting the background to nil/none allows for this in Nvim but in Neovide it renders with a black background when set to a none-value. The fg value is respected though so this seems like a bug to me, or is there something I'm missing? 

**Screenshots**
image

 - OS: 
 macOS Ventura 13.2.1
 - Neovide Version
neovide 0.10.3
 - Neovim Version
NVIM v0.8.3

AsPulse:
**Overview**
- close #1819 
- close #1508
 
image

<br/>
<br/>

### What kind of change does this PR introduce?
- Fix: Changed the handling of colors when grid_renderer draws the background of a floating window.

<br/>

### Description
The original problem was that the `MsgArea` was not transparent, but the immediate cause was that specifying a transparent colour (`guibg=none`) on the Floating Window did not work properly.

In this PR, if a transparent colour is specified, a background colour with an alpha value is specified and returned.
<br/>

### Why I am still Drafting this:
- [ ] I have only installed neovide on my mac and have not tested it enough in other environments.
- [ ] This may have side-effects on the color reversing setting.
- [ ] This may have side-effects on the blur setting.
- [ ] There is room for performance improvements, such as parsing the background_color each time.

<br/>

### Did this PR introduce a breaking change? 
- No

<br/>

I apologize if there are any mistakes in my English writing. Actually, I am a Japanese student learning English ;)

### DIFF
--- a/src/renderer/grid_renderer.rs
+++ b/src/renderer/grid_renderer.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use log::trace;
 use skia_safe::{
-    colors, dash_path_effect, BlendMode, Canvas, Color, Paint, Path, Point, Rect, HSV,
+    colors, dash_path_effect, BlendMode, Canvas, Color, Paint, Path, Point, Rect, HSV, ClipOp,
 };
 use winit::dpi::PhysicalSize;
 
@@ -118,11 +118,24 @@ impl GridRenderer {
             self.paint
                 .set_color(style.background(&self.default_style.colors).to_color());
         }
-
+        let window_transparency = &SETTINGS.get::<WindowSettings>().transparency;
         if is_floating {
-            self.paint
-                .set_alpha((255.0 * ((100 - style.blend) as f32 / 100.0)) as u8);
-        } else if (SETTINGS.get::<WindowSettings>().transparency - 1.0).abs() > f32::EPSILON
+            if self.paint.color() != self.get_default_background() {
+                self.paint
+                    .set_alpha((255.0 * ((100 - style.blend) as f32 / 100.0)) as u8);
+            } else {
+                if let Ok([r, g, b, a]) = &SETTINGS
+                    .get::<WindowSettings>()
+                    .background_color
+                    .parse::<csscolorparser::Color>()
+                    .map(|c| c.to_rgba8())
+                {
+                    self.paint.set_argb(*a, *r, *g, *b);
+                } else {
+                    self.paint.set_argb(( window_transparency * 255.0 ) as u8, 0, 0, 0);
+                }
+            }
+        } else if (window_transparency - 1.0).abs() > f32::EPSILON
             // Only make background color transparent
             && self.paint.color() == self.get_default_background()
         {
--- a/src/renderer/grid_renderer.rs
+++ b/src/renderer/grid_renderer.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use log::trace;
 use skia_safe::{
-    colors, dash_path_effect, BlendMode, Canvas, Color, Paint, Path, Point, Rect, HSV, ClipOp,
+    colors, dash_path_effect, BlendMode, Canvas, ClipOp, Color, Paint, Path, Point, Rect, HSV,
 };
 use winit::dpi::PhysicalSize;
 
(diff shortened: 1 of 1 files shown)


fredizzimo:
Hi, thanks for your contribution, it looks good, but I'm worried that it conflicts with some other changes I have in another pull request https://github.com/neovide/neovide/pull/1790/commits/e96fe45d161278496af0a9dfeb5ddc44ecb02cec

The changes to the transparency drawing does not really belong in that PR though, so I would really appreciate, if you could try to combine my changes and your changes as well as possible.

You can probably not take everything from that commit, since the rendering is already changed quite a bit at that point, but at least the changes to the `grid_renderer.rs` should work, maybe with `self.paint.set_blend_mode(BlendMode::Src)` as an exception.

AsPulse:
Sorry, but it would be great if you could be more specific about what modifications should be made.

As you say, there are significant changes to the rendering system in that pull request and it seems difficult to address the problem by integrating just this file.

I think the best I can do is to keep the signatures, including arguments and return values, as they appear to be, and comment out the annotations, but is it helpful...? or should I do another things...?

(Apologies if I'm misreading the English,
What do you mean by "maybe with `self.paint.set_blend_mode(BlendMode::Src)` as an exception" ......? This code is present in both commits.......)

fredizzimo:
I'm sorry, I did not check very clearly what needs to be done. And yes, you are correct the blendmode line has just been moved.

The reason why I'm asking this, is because I think my changes also fixes this problem and some others as well, by treating floating windows and regular ones the same, and not drawing anything when the style is None or fully transparent. But I'm not completely sure.

That would also avoid the need of platform specific code, your code currently only works on macOS, since `background_color` is only used on that.

You don't need to include the return type, but it does not harm either, since it's going to be used later.

AsPulse:
> your code currently only works on macOS, since background_color is only used on that.

Just to comment here, my code should work without `background_color`. (It's supposed to calculate the color based on the transparency of the window)
```rs
if let Ok([r, g, b, a]) = &SETTINGS
    .get::<WindowSettings>()
    .background_color
    .parse::<csscolorparser::Color>()
    .map(|c| c.to_rgba8())
{
    self.paint.set_argb(*a, *r, *g, *b); // if specified background_color
} else {
    self.paint
        .set_argb((window_transparency * 255.0) as u8, 0, 0, 0); // if not specified background_color
}
```
( This is most likely based on looking at the code, but it may be a different if I actually run it on Win/Lnx...... )


But I think we know what we need to do.
I will try! (Forgive me if I end up asking you for help, lol)

AsPulse:
I found a problem in the fix, so I have reviewed it with the PR that has e96fe45. I apologize for the length of the review, but I hope you would check it out!

fredizzimo:
Sorry, I haven't had time to check it out yet, and it will still take some time until I do. But maybe I will at the end of the day or tomorrow.

Kethku:
@AsPulse is this PR good to close now that the improved render loop PR has merged or is there something that still needs done here?

fredizzimo:
@Kethku, I don't think we ever got this solved. This pull request and the linked issue made some attempt to fix the transparency issues on macOS. But unfortunately the other platforms broke as a consequence of that.

The underlying issue is described here https://github.com/neovide/neovide/issues/2094, on macOS we have a fake background rendered behind the toplevel Neovim buffer, and that messes with the transparency. So I think we need to get rid of that, and then see what else needs to be done.

fredizzimo:
I think the macOS transparency works as well as the other platforms now, so this can be closed.

It's not perfect, but there's a proposal for making it better:
* https://github.com/neovide/neovide/issues/2471

Issue #1912, opened 2023-06-22
Update background light / dark depending on the system theme


mgax:
I'd like Neovide to detect system theme (light / dark) and adjust the ['background' option](https://neovim.io/doc/user/options.html#'background') accordingly.

Neovide should probably listen to winit [`WindowEvent::ThemeChanged`](https://docs.rs/winit/latest/winit/event/enum.WindowEvent.html#variant.ThemeChanged) events and update the `'background'` option.

Apps that support themes (e.g. Discord) generally allow the user to pick light, dark, or auto, where auto follows the system setting. Neovide should probably offer a similar option since not all uesrs will want their `'background'` option to change automatically.

Alternatively, Neovide could provide some sort of hook or event, that a user can subscribe to, by writing code in their configuration file, and they could implement whatever logic to handle it. I don't know enough about neovim to understand if that even makes sense or if it's worth the complexity.

Here's the relevant winit example: https://github.com/rust-windowing/winit/blob/66ff52b0128be9430b32cb5756e124f61d2e67d0/examples/theme.rs#L38-L44. I've tested it on MacOS and it works as expected. Interestingly, if the user overrides the theme by pressing _L_ or _D_, no ThemeChanged events are triggered when the key is pressed, or when the system theme changes; events are only sent when in _Automatic_ mode.

arthsmn:
Hey, I'm using darkman to deal with the light/dark switching and all GTK apps, as well as kitty, respect it, but neovide don't.

PR #1917, opened 2023-07-02
Configuration option to set theme to light/dark/auto

mgax:
Fixes https://github.com/neovide/neovide/issues/1912. I've tested on MacOS and it works as intended.

For existing users, nothing changes; the user must explicitly set ~the `--theme` command-line option, the `NEOVIDE_THEME` env var, or the `theme` option in `config.toml`~  `g:neovide_theme = 'auto'` to enable this new feature.

### DIFF
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -150,7 +150,7 @@ impl ParallelCommand {
                 nvim.command(format!("e {path}").as_str()).await.ok();
             }
             ParallelCommand::SetBackground(background) => {
-                nvim.command(&format!("set background={}", background))
+                nvim.command(&format!("set background={}", background).as_str())
                     .await
                     .ok();
             }
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -150,7 +150,7 @@ impl ParallelCommand {
                 nvim.command(format!("e {path}").as_str()).await.ok();
             }
             ParallelCommand::SetBackground(background) => {
-                nvim.command(&format!("set background={}", background).as_str())
+                nvim.command(format!("set background={}", background).as_str())
                     .await
                     .ok();
             }
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -117,6 +117,7 @@ pub enum ParallelCommand {
     FocusLost,
     FocusGained,
     DisplayAvailableFonts(Vec<String>),
+    SetBackground(String),
     #[cfg(windows)]
     RegisterRightClick,
     #[cfg(windows)]
@@ -147,6 +148,11 @@ impl ParallelCommand {
                 .expect("Focus Gained Failed"),
             ParallelCommand::FileDrop(path) => {
                 nvim.command(format!("e {path}").as_str()).await.ok();
+            }
+            ParallelCommand::SetBackground(background) => {
+                nvim.command(&format!("set background={}", background))
+                    .await
+                    .ok();
             }
             ParallelCommand::DisplayAvailableFonts(fonts) => {
                 let mut content: Vec<String> = vec![
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -2,37 +2,12 @@ use std::{iter, mem};
 
 use crate::{dimensions::Dimensions, frame::Frame, settings::*};
 
-use clap::{
-    builder::FalseyValueParser,
-    builder::{EnumValueParser, PossibleValue},
-    ArgAction, Parser, ValueEnum,
-};
+use clap::{builder::FalseyValueParser, ArgAction, Parser};
 
 #[cfg(target_os = "windows")]
 const SRGB_DEFAULT: &str = "1";
 #[cfg(not(target_os = "windows"))]
 const SRGB_DEFAULT: &str = "0";
-
-#[derive(Debug, Copy, Clone, Eq, PartialEq)]
-pub enum ThemeChoice {
-    Light,
-    Dark,
-    Auto,
-}
-
-impl ValueEnum for ThemeChoice {
-    fn value_variants<'a>() -> &'a [Self] {
-        &[Self::Light, Self::Dark, Self::Auto]
-    }
-
-    fn to_possible_value(&self) -> Option<PossibleValue> {
-        Some(match self {
-            Self::Light => PossibleValue::new("light"),
-            Self::Dark => PossibleValue::new("dark"),
-            Self::Auto => PossibleValue::new("auto"),
-        })
-    }
-}
 
 #[derive(Clone, Debug, Parser)]
 #[command(version, about, long_about = None)]
@@ -145,14 +120,6 @@ pub struct CmdLineSettings {
         default_value = "neovide"
     )]
     pub x11_wm_class_instance: String,
-
-    /// Set background to light, dark, or match system setting with "auto"
-    #[arg(
-        long = "theme",
-        env = "NEOVIDE_THEME",
-        value_parser = EnumValueParser::<ThemeChoice>::new(),
-    )]
-    pub theme: Option<ThemeChoice>,
 }
 
 impl Default for CmdLineSettings {
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -151,7 +151,6 @@ pub struct CmdLineSettings {
         long = "theme",
         env = "NEOVIDE_THEME",
         value_parser = EnumValueParser::<ThemeChoice>::new(),
-        required = false
     )]
     pub theme: Option<ThemeChoice>,
 }
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -2,12 +2,39 @@ use std::{iter, mem};
 
 use crate::{dimensions::Dimensions, frame::Frame, settings::*};
 
-use clap::{builder::FalseyValueParser, ArgAction, Parser};
+use clap::{
+    builder::FalseyValueParser,
+    builder::{EnumValueParser, PossibleValue},
+    ArgAction, Parser, ValueEnum,
+};
 
 #[cfg(target_os = "windows")]
 const SRGB_DEFAULT: &str = "1";
 #[cfg(not(target_os = "windows"))]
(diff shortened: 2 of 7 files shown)


src/bridge/ui_commands.rs
```diff
@@ -148,6 +149,11 @@ impl ParallelCommand {
             ParallelCommand::FileDrop(path) => {
                 nvim.command(format!("e {path}").as_str()).await.ok();
             }
+            ParallelCommand::SetBackground(background) => {
+                nvim.command(&format!("set background={}", background))
```
MultisampledNight, line 153: Minor nit: For the sake of consistency with the `format!` call two lines above, maybe use `.as_str()` instead.

src/cmd_line.rs
```diff
@@ -120,6 +147,14 @@ pub struct CmdLineSettings {
...
     )]
     pub x11_wm_class_instance: String,
+
+    /// Set background to light, dark, or match system setting with "auto"
+    #[arg(
+        long = "theme",
+        env = "NEOVIDE_THEME",
+        value_parser = EnumValueParser::<ThemeChoice>::new())
+    ]
+    pub theme: ThemeChoice,
```
MultisampledNight, line 157: Since you didn't specify a default, the argument will be _always_ required. I suppose you could use `default_value = ThemeChoice::Unset` in the attribute.

mgax: Thanks for the tip with `Option<>`, it did the trick!

mgax:
I've got the tests to pass, the `--theme` option is now truly optional. The _nightly_ CI jobs are failing but I suspect it may not be caused by the PR itself: https://github.com/mgax/neovide/actions/runs/5441501264.

---
I have a stylistic question: would it be better to rewrite the pattern matching like so? https://gist.github.com/mgax/0506394f61f5d47f856803cc52213a7a I'm guessing it's nice to have fewer calls to `set_background`, but maybe the convoluted code is not worth it?

MultisampledNight:
> I have a stylistic question: would it be better to rewrite the pattern matching like so? https://gist.github.com/mgax/0506394f61f5d47f856803cc52213a7a I'm guessing it's nice to have fewer calls to set_background, but maybe the convoluted code is not worth it?

The first hunk looks better to me, I think I'd even split up the matching part and actually setting the background by storing the match result first, shadowing `theme`:

```diff
-                    match theme {
-                        Theme::Light => set_background("light"),
-                        Theme::Dark => set_background("dark"),
-                    };
+                    let theme = match theme {
+                        Theme::Light => "light",
+                        Theme::Dark => "dark",
+                    };
+                    set_background(theme);
```

The second hunk however looks too convoluted to me. I need to backtrack in order to see what braces will actually trigger the background change.

Sidenote: Whatever you end up with, do run rustfmt afterwards, `,` after blocks are deleted by it.

src/cmd_line.rs
```diff
@@ -120,6 +145,15 @@ pub struct CmdLineSettings {
...
     pub x11_wm_class_instance: String,
+
+    /// Set background to light, dark, or match system setting with "auto"
+    #[arg(
+        long = "theme",
+        env = "NEOVIDE_THEME",
+        value_parser = EnumValueParser::<ThemeChoice>::new(),
+        required = false
+    )]
+    pub theme: Option<ThemeChoice>,
```
MultisampledNight, line 156: `required` is not required (ironic) if the target type is `Option`. clap knows automagically that any `Option`al arguments are not required (if unset).

```suggestion
    )]
    pub theme: Option<ThemeChoice>,
```

src/bridge/ui_commands.rs
```diff
@@ -148,6 +149,11 @@ impl ParallelCommand {
             ParallelCommand::FileDrop(path) => {
                 nvim.command(format!("e {path}").as_str()).await.ok();
             }
+            ParallelCommand::SetBackground(background) => {
+                nvim.command(&format!("set background={}", background).as_str())
```
MultisampledNight, line 153: As clippy already mentions, with `.as_str()`, the `&` is not required anymore.

```suggestion
                nvim.command(format!("set background={}", background).as_str())
```

MultisampledNight (commented):
Generally I'm not too sure if passing the argument on the command line is really the way to go, since it could also be desired to be configurable with the casual `g:neovide_*` variable mechanism. But I don't feel strongly about that either way.

MultisampledNight:
Indeed, the CI failure is not your fault, it also happens locally for me, also on `main`. No need to do anything about this, https://github.com/neovide/neovide/pull/1899 fixes this already.

mgax:
> Generally I'm not too sure if passing the argument on the command line is really the way to go, since it could also be desired to be configurable with the casual `g:neovide_*` variable mechanism. But I don't feel strongly about that either way.

Hmm, I see what you mean. I wasn't aware of the `g:neovide_*` variables, but looking at what is already available (fullscreen, transparency, background color, scaling), _theme_ seems to fit right in. I don't mind making the change, in fact it's tempting, but I'd prefer if you, or "the core dev team", make the decision.

> The first hunk looks better to me, I think I'd even split up the matching part and actually setting the background by storing the match result first, shadowing `theme`:

Yep sounds good. I've used `background` as variable name because it's the name used on the vim side.

---
Thanks for reviewing and for the suggestions! Let me know if I should switch to a `g:neovide_theme` variable.

mgax:
I'm sorry for having pressed you, and thanks for the clarification. I've updated the PR to use a setting.

src/window/mod.rs
```diff
@@ -460,6 +479,19 @@ pub fn create_window() {
         renderer.grid_renderer.font_dimensions,
     );
 
+    println!("theme: {}", SETTINGS.get::<WindowSettings>().theme);
+
```
MultisampledNight, line 483: leftover debugging print

MultisampledNight (approved):

MultisampledNight:
I'll take a final look over this in 22 hours or the like, not really in the right state of mind to hit the merge button.

falcucci:
@mgax first, really cool that you have implemented this ***but*** I am not really sure about this feature.

Neovide is fully configurable by neovim setup.

I have the same hook working as expected using [auto-dark-mode.nvim](https://github.com/f-person/auto-dark-mode.nvim).

I think this is the way to go because this also makes neovide totally compatible with neovim (everything should work just fine in case you decide to use neovim directly on terminal IMO).

That's what makes neovide really cool actually. It's fully compatible.

@MultisampledNight what's your thoughts on this?

falcucci:
@mgax really cool feature, I just want to point out that some colorschemes doesnt support light themes or vice-versa. So isnt only about changing the background. 

In that case uou have more components that could be a mess after updating the background as Floaterm and Pmenu.

for example one of the most famous colorscheme [dracula](https://github.com/Mofiqul/dracula.nvim) hasnt support for light mode.

So the user could choose a different one accordingly to the background.

I think that this should be more customizable, as the same as [auto-dark-mode.nvim](https://github.com/f-person/auto-dark-mode.nvim) which for me but is available only for macos.

but look the hook itself

```lua
set_dark_mode = function()
    vim.api.nvim_set_option('background', 'dark')
    vim.cmd('colorscheme gruvbox')
end,
set_light_mode = function()
    vim.api.nvim_set_option('background', 'light')
    vim.cmd('colorscheme gruvbox')
end
```

I think that something similar should be done here.

falcucci:
And yet let's keep in mind that neovide is supposed to be fully compatible with neovim, not saying that this feature would bring incompatibility, but it's something to consider.

As an neovim user I would like to have always as possible my neovim configs working out of any neovide feature (themes is one of them).

falcucci:
to be really concise about it, I think the perfect solution would be a neovim config (lua) or plugin as [auto-dark-mode.nvim](https://github.com/f-person/auto-dark-mode.nvim) but available for all platforms.

MultisampledNight:
Since the `background` option _is_ set, and that's observable from Neovim, I suppose you could try to add an autocmd listening on `OptionSet` on any filetype.

mgax:
@falcucci that's interesting; I'm using [catppuccin](https://github.com/catppuccin/nvim), which has support for both light and dark backgrounds, and flips between them automatically, but it makes sense that you may want more control. In that case @MultisampledNight's suggestion seems like the way to go.

From a purity perspective, _auto-dark-mode.nvim_ polls the system to keep the theme in sync, which is a bit wasteful. The editor already receives events from the system when the theme changes, in a cross-platform compatible way, so it feels natural to automatically update the equivalent option in neovim.

falcucci:
mine would works just fine because I would ignore the `--theme` option and keep my setup with auto-dark-mode.nvim, but for people using linux, windows they should adapt there configs to work properly as you said for any neovide applied changes as the background.

the feature itself is nice, I am just wondering about which side put it, neovim or neovide.

btw I'll test this PR to see if I get something.

falcucci:
@mgax yeah but I think the polling is needed because who receives the event is the terminal (if not used withing neovide), right?

mgax:
@falcucci the implementation changed, it's no longer a command-line option (`--theme`), it's a neovim option, `g:neovide_theme`. I've updated the PR's description to reflect that.

> but for people using linux, windows they should adapt there configs to work properly

FWIW, it looks like winit supports the [ThemeChanged](https://docs.rs/winit/latest/winit/event/enum.WindowEvent.html#variant.ThemeChanged) event on MacOS, Windows and Web. Linux support is discussed here: https://github.com/rust-windowing/winit/issues/1549#issuecomment-1396539233.

In any case, this patch doesn't change behaviour out of the box, it has to be enabled explicitly in the user's config.

falcucci (approved):

falcucci (approved):

MultisampledNight:
Thank you!

mgax:
@MultisampledNight @falcucci thank you both!

CapSel:
What if this does not work on gnome?
For comparison using using gnome-console, gnome-terminal starting nvim causes background to be set automatically on start. Using ghostty with neovim background changes as soon as I change theme in gnome.

But using neovide it's always dark.

Here is a snippet from bottom of init.lua
```
if vim.g.neovide then
  vim.g.neovide_theme = 'auto'
end
```
Theme used is `rmehri01/onenord.nvim`.

PR #1920, opened 2023-07-05
Support font features

### LINKED ISSUE #1249: Fontfeatures support

I've enabled font features for Cascadia Code through the fontconfig.

image

So when using cascadia code italic font, I can get this cursive style.

However it seems like neovide doesn't recognize those feature yet.

image

**Additional context**

TwIStOy:
close #1249 

- Feature

- No

### DIFF
--- a/src/renderer/fonts/font_options.rs
+++ b/src/renderer/fonts/font_options.rs
@@ -25,8 +25,9 @@ impl FontWithFeatures {
     pub fn parse_family_desc(desc: impl AsRef<str>) -> FontWithFeatures {
         let mut features = vec![];
         let mut name = None;
-        // reverse string at first, then split_inclusive at '+' or '-', then
-        // reverse each item in the result
+        // `split_inclusive` leaves the separator at the end of the substring,
+        // but we want it at the beginning of the next substring.
+        // So we reverse the string, split it, and then reverse each substring.
         desc.as_ref()
             .chars()
             .rev()
@@ -35,17 +36,17 @@ impl FontWithFeatures {
             .map(|s| s.chars().rev().collect::<String>())
             .for_each(|s| {
                 if let Some(name) = s.strip_prefix('+') {
-                    features.push((name.to_string(), 1u16));
+                    features.push((name.trim().to_string(), 1u16));
                 } else if let Some(name) = s.strip_prefix('-') {
-                    features.push((name.to_string(), 0u16));
+                    features.push((name.trim().to_string(), 0u16));
                 } else {
-                    name = Some(s);
+                    name = Some(s.trim().to_string());
                 }
             });
 
         FontWithFeatures {
             name: name.unwrap(),
-            features,
+            features: features.into_iter().rev().collect(),
         }
     }
 }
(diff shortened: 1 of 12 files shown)


fredizzimo:
Thank you, if it works this will be a very nice addition, I use a custom built font myself due to the lack of features support.

But before reviewing further, we need some documentation, can you add that to https://github.com/neovide/neovide/blob/main/website/docs/configuration.md? So we can se how you are supposed to use it, and how it ties into all the other options and does not conflict with this, still unmerged https://github.com/neovide/neovide/pull/1227. We also need the documentation to be able to verify that it actually works as documented.

TwIStOy:
config docs updated

website/docs/configuration.md
```diff
@@ -76,6 +77,7 @@ as such it's also documented in `:h guifont`. But to sum it up and also add Neov
     should Hack fail to contain any glyph.
   - `Roboto_Mono_Light:h10` — Roboto Mono Light at size 10.
   - `Hack:h14:i:#e-subpixelantialias:#h-none`
+  - `Iosevka+ss10,Symbols_Nerd_Font:h22` — Iosevka at size 22, and enable stylistic sets `ss10`.
```
fredizzimo, line 81: I was first going to suggest that you move the features to options, which are already parsed, but then it occurred to me that, the features are font-specific, and should be applied per font.

And after even more reflection, I think most of the options should actually be font specific, rather than applying to the whole list. I think at some point we need to come up with another more flexible format, and deprecate the current one, but I don't know how. Anyway, that should probably not be done here in this PR.

src/renderer/fonts/font_options.rs
```diff
@@ -13,17 +21,47 @@ pub struct FontOptions {
     pub edging: FontEdging,
 }
 
+impl FontWithFeatures {
+    pub fn parse_family_desc(desc: impl AsRef<str>) -> FontWithFeatures {
+        let mut features = vec![];
+        let mut name = None;
+        // reverse string at first, then split_inclusive at '+' or '-', then
+        // reverse each item in the result
```
fredizzimo, line 29: I think this comment needs to include why you reverse it. If I understand it correctly it's to keep prefixes rather than suffixes.

src/renderer/fonts/font_options.rs
```diff
@@ -13,17 +21,47 @@ pub struct FontOptions {
     pub edging: FontEdging,
 }
 
+impl FontWithFeatures {
+    pub fn parse_family_desc(desc: impl AsRef<str>) -> FontWithFeatures {
```
fredizzimo, line 25: This functionality is complex enough, that I think you need to add a few new unit tests. It should probably also ignore whitespace at the start and end, and between plus and the feature.

fredizzimo (changes requested):

src/renderer/fonts/font_options.rs
TwIStOy: this comment updated

src/renderer/fonts/font_options.rs
TwIStOy: add some testcases~

website/docs/configuration.md
TwIStOy: Maybe we need a table instead of the simple string? [Font settings in wezterm](https://wezfurlong.org/wezterm/config/font-shaping.html#advanced-font-shaping-options) seems simple and clear.

fredizzimo: Yes, I think something like that makes sense. Maybe we could use the guifont to specify the name and fallbacks, and the size. And then a lua table with the font names for the options.

I’m leaving this open for a bit still, so that we have time to think about it.

fredizzimo:
Sorry for taking so long with this, but we still haven't been able to decide what to do with the font customization setting. 

The current format can break existing clients, since it's against the `guifont` specification https://neovim.io/doc/user/options.html#'guifont'. Or old format is also not strictly conforming, but at least it shouldn't make clients that don't use the options parse the fontname wrongly.

TwIStOy:
@fredizzimo Sorry for my late reply and updates. I introduce a new setting `neovide_font_features` to setup these font features. Font and features can be set as this:

```lua
vim.g.neovide_font_features = {
  ["Iosevka"] = {
    "+ss07",
    "cv49=16",
    "cv94=1",
    "VXLA=2",
    "VXLC=2",
    "cv34=12",
    "cv31=13",
  },
}

vim.o.guifont = "Iosevka,Symbols_Nerd_Font:h22"
```

MultisampledNight:
How would this look like in VimScript? I'd assume something in the direction of a dict, since it's [assumed to be a map](https://github.com/neovide/neovide/pull/1920/files#diff-4563cedfd3717a23a6908d5d31f96e03e6b19db866727f88a1f52df1399984fcR100)? But dicts only support strings as keys.

src/settings/from_value.rs
```diff
@@ -1,76 +1,146 @@
...
+
 use super::Value;
 use log::error;
 
 // Trait to allow for conversion from rmpv::Value to any other data type.
 // Note: Feel free to implement this trait for custom types in each subsystem.
 // The reverse conversion (MyType->Value) can be performed by implementing `From<MyType> for Value`
 pub trait ParseFromValue {
-    fn parse_from_value(&mut self, value: Value);
+    fn parse_from_value(&mut self, value: &Value) -> bool;
```
MultisampledNight, line 10: Instead of returning an opaque `bool`, possibly it could be cleaner to use a direct `Result` with a custom error type or even just a `()` instead. This would also make the control flow of most of the implementors cleaner, since then they don't need to wrap the entire function body in an `if value.is_map() {` with the error case at the end or the like.

MultisampledNight: Totally, thank you!

TwIStOy:
> How would this look like in VimScript? I'd assume something in the direction of a dict, since it's [assumed to be a map](https://github.com/neovide/neovide/pull/1920/files#diff-4563cedfd3717a23a6908d5d31f96e03e6b19db866727f88a1f52df1399984fcR100)? But dicts only support strings as keys, so `["Iosevka"]` won't work (but re-using the fallback list parsing which is already in `guifont` would work).

Vimscript version looks like this:

```vim
let g:neovide_font_features = {
  \ 'Iosevka': [
    \ '+ss07',
    \ 'cv49=16',
    \ 'cv94=1',
    \ 'VXLA=2',
    \ 'VXLC=2',
    \ 'cv34=12',
    \ 'cv31=13',
  \ ]
\ }
```

`["Iosevka"]` just a representation of a key which type is string. the format is be used if there's some special chars in the string like, `["Foo Bar-Bar"]`.

fredizzimo:
@TwIStOy, no we should be sorry for not coming back to you in time. And we are still not fully decided on the format we actually want.

So, don't change anything yet. But I think the following format would make sense for more extensibility in the future and keeping the settings for a single font in one place. But I can see advantages of using separate settings as well.

```lua
vim.g.neovide_font_settings = {
  ["Iosevka"] = {
    features = {
      "+ss07",
      "cv49=16",
      "cv94=1",
      "VXLA=2",
      "VXLC=2",
      "cv34=12",
      "cv31=13",
    },
  },
}
```

fredizzimo:
Actually I don't think the above is enough. I would also like to somehow be able to have different fonts and settings for `normal`, `bold`, `italic` and `underline` Possibly with different fallbacks.

We could perhaps copy or take inspiration from the Wezterm configuration https://wezfurlong.org/wezterm/config/lua/config/font_rules.html

9mm:
Definitely following this so I can set `calt: 0` to send ligatures back to the depths

idnsunset:
Has neovide support font weight settings so far? For some fonts of which the regular weight looks too thin, we have to use their medium or even semibold weights instead.

fredizzimo:
@idnsunset, not yet. But it should be easy to add once we have decided how to configure it. 

@TwIStOy, It seems to be hard for us to decide how we want this configured. But we discussed it a little bit last weekend, and it seems like the best way would be to not include these settings in the neovim configuration at all. It makes a lot of sense to move them to the config file instead https://neovide.dev/config-file.html

The main reason for this, is that the font configuration is quite machine specific, and you might want to use the same configuration even if you remote into different machines.  Additionally, it gives us more flexibility to use the format we really want. `guifont`, would still work as a backup, but it would only support a subset of the features.

The basic format could be something similar to that of [alacritty yml](https://github.com/alacritty/alacritty/releases/download/v0.12.3/alacritty.yml), with sepearte keys for `normal`, `bold`, `italic` and `bold_italic`. 

What do you think? Also @neovide for your comments.

MultisampledNight:
There's two things that make this a bit more difficult:

1. Technically, `Config` works atm by just setting environment variables which are then read by clap. We'd need to serialize the font features this way into JSON or keep it in memory, though that wouldn't be too difficult
2. The config file would eliminate runtime setting, such as trying out fonts with an already opened window, or configuring font features live. Which is something desirable imo, even though it does usually not happen _after_ configuration.

fredizzimo:
For 2. I think we could implement a file watcher and reload the config file on the fly, like a lot of programs, including Alacritty do. But of course that requires some new code, and it might be confusing since not all options can be changed dynamically.

Kethku:
Now that we have a relatively simple way to notify the user of things inside neovim, it seems to me we could detect changes that can't be applied dynamically and warn the user that they wont go into effect until a restart.

So to summarize:

- We need to rework our config file to include dynamic file watching and notifications for settings that can't be reloaded
- We should support a superset of things in guifont influenced by the alacritty yml format
- Settings in the config file should supersede those set in nvim

does that seem right to folks?

TwIStOy:
Re-implement in #2213

PR #1870, opened 2023-05-18
Smooth render loop

fredizzimo:
**NOTE: This works only with Neovim nightly, or the `release-0.9` branch (upcoming 0.9.2 release)**

This completely rewrites the current render and event loop of Neovide to make it almost completely smooth. There are some minor frame drops now and then, caused by the GPU side of Skia suddenly taking a lot longer than usual, but especially with refresh rates over 100, that's almost not noticeable. Note Skia has been updated to the latest version in PR, so it's not some bug that has been fixed later.

I was forced to add a separate thread for rendering thread on Windows because the winit event loop is not suitable for real-time rendering, for that I have reported a bug here https://github.com/rust-windowing/winit/issues/2782

On Windows, I had to work around some bugs with the Windowed mode OpenGL implementation by using a separate thread that calls `DwmFlush` to find a suitable vsync synchronisation points. There seems to be some race conditions in the `SwapBuffer` implementation. It's not perfect, but it's working pretty well. 

On Wayland, frame_callbacks are used for VSync.

On other platforms we assume that `swap_buffers` will wait for VSync. But this might not be the case on macOS, and https://developer.apple.com/documentation/corevideo/cvdisplaylinkoutputcallback might need to be implemented.

If you specify `--novsync`, then a software timer is used on all platforms and `g:neovide_refersh_rate` is used to control the refresh rate.

The redraw scheduler has also been removed and the checks if needing to redraw are now done on the fly, which is possible since the rendering and animation has been split.

This also includes the following PR
https://github.com/neovide/neovide/pull/1790, which removes the scrolling snapshots and  fixes the transparency.

There are two remaining problems, both on macOS, which I can't fix myself due to not owning a mac.
1. The transparency issues that @AsPulse was working on. I think we can fix that later.
2. The vsync option might not work and the maximum refresh rate is used. Someone need to implement https://developer.apple.com/documentation/corevideo/cvdisplaylinkoutputcallback. Alternatively we can force it off on macOS, and then a software timer with `g:neovide_refresh_rate` will be used.

**NOTE:** It's recommended to run this with `--multigrid`, otherwise you wont see many changes. Furthermore there are some glitches caused by Neovim, which are fixed in the nightly builds. So it's recommended to run that. And some more fixes will be included in https://github.com/neovim/neovim/pull/24425, which is now merged into Neovim master as well.

**NOTE2**, that this also includes https://github.com/neovide/neovide/pull/1939, and merging that first will make reviewing this much easier.

**NOTE3:**
[...]

PR #1929, opened 2023-07-08
Release 0.11.0

MultisampledNight:
Just a "formalization" of the 0.11 release, so issues can be linked to it, and we can list what'll be breaking already. ~~I suppose we'll wait with this for some time after #1870 is be merged.~~ Feel free to get involved with testing, feedback and the like!

The list below doesn't list all changes made, but rather what'll be complemented by the autogenerated release notes.

## Breaking

- macOS: `<A-...>` keybindings previously only worked with dead keys. Now it's the other way around, they only work with _non_-dead keys. See https://github.com/neovide/neovide/pull/1899#issuecomment-1622968933. 

## Features

- Config file! https://neovide.dev/config-file.html
- Automatic OS theme change detection, if enabled! https://neovide.dev/configuration.html#theme

## Bug fixes

- Get running on Wayland again!
- IME support!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1372,7 +1372,7 @@ dependencies = [
 
 [[package]]
 name = "neovide"
-version = "0.10.4"
+version = "0.11.0"
 dependencies = [
  "async-trait",
  "backtrace",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "neovide"
-version = "0.10.4"
+version = "0.11.0"
 edition = "2021"
 build = "build.rs"
 description = "Neovide: No Nonsense Neovim Gui"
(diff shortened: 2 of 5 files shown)


sisrfeng:
Look forward to the release! IME support!  (Is there any tweak I can make to get the new IME support before the new release?)

MultisampledNight:
@sisrfeng You can build from source, as covered in https://neovide.dev/installation.html.

fredizzimo:
For this release we should definitely include the deadkey fixes here https://github.com/neovide/neovide/pull/1930 and the fixes for the AppImage here https://github.com/neovide/neovide/pull/1941

And if we can squeeze in the smooth rendering support as well https://github.com/neovide/neovide/pull/1870, it would be great. But that still has some issues on MacOS. But maybe we can document those limitations.
1. The transparency does not work properly on macOS
2. You should run `--novsync` and set `g:neovide_refresh_rate` manually on macOS. Otherwise the frame rate will be unlimited and might consume unnecessary cpu and gpu power.

But we can also wait with that and do another release for the smooth rendering

fredizzimo:
Let's leave the smooth rendering out. That still has some issues to solve.

MultisampledNight:
I believe everything would be ready for the release now, only thing left is actually writing the patch notes and hitting the release button (and maybe a short bitrot period before that in order to squash a few hasty bugs, too).

@fredizzimo would you like to do the release? Or anyone else? I could do it as well, don't have a preference.

fredizzimo (approved):

PR #1930, opened 2023-07-09
fix: Fix deadkey sequences ending with space

fredizzimo:
- Fix

A dead key followed by space will usually result in the character itself. However, Neovide resolved this wrongly, and always treated space as a special key, so the sequence was resolved as space instead.

- No

### DIFF
--- a/src/window/keyboard_manager.rs
+++ b/src/window/keyboard_manager.rs
@@ -67,7 +67,7 @@ impl KeyboardManager {
     }
 
     fn format_key(&self, key_event: &KeyEvent) -> Option<String> {
-        if let Some(text) = get_special_key(&key_event.logical_key) {
+        if let Some(text) = get_special_key(key_event) {
             Some(self.format_key_text(text, true))
         } else {
             self.format_normal_key(key_event)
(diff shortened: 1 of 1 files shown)


fredizzimo:
This fix is for some reason not working in X11, but in Wayland it does. 

I will try to figure out what's wrong.

src/window/keyboard_manager.rs
```diff
@@ -154,10 +154,19 @@ fn use_alt() -> bool {
...
 }
 
-fn get_special_key(key: &Key) -> Option<&str> {
+fn get_special_key(key_event: &KeyEvent) -> Option<&str> {
+    let key = &key_event.logical_key;
     match key {
         Key::Backspace => Some("BS"),
-        Key::Space => Some("Space"),
+        Key::Space => {
+            // Space can finish a dead key sequence, so only threat space as a special key when
```
MultisampledNight, line 162: Typo
```suggestion
            // Space can finish a dead key sequence, so only treat space as a special key when
```

MultisampledNight:
Thank you!

PR #1939, opened 2023-07-20
chore: Extract window_wrapper and the update loop to their own modules

fredizzimo:
- Refactor

This splits the `window/mod.rs` into `update_loop` and `window_wrapper`. The opengl window creation has also been split into two functions, the actual window creation and the context creation.

This will make it easier to review the render loop changes in #1870, and to allow some platforms to run in a thread while other platforms don't. 

- No, functionally there are no changes.

### DIFF
--- a/src/window/update_loop.rs
+++ b/src/window/update_loop.rs
@@ -1,0 +1,117 @@
+use std::time::{Duration, Instant};
+
+use winit::{
+    event::{Event, WindowEvent},
+    event_loop::ControlFlow,
+};
+
+#[cfg(target_os = "macos")]
+use super::draw_background;
+use super::{WindowSettings, WinitWindowWrapper};
+use crate::{
+    profiling::{tracy_create_gpu_context, tracy_zone},
+    running_tracker::*,
+    settings::{save_window_size, SETTINGS},
+};
+
+enum FocusedState {
+    Focused,
+    UnfocusedNotDrawn,
+    Unfocused,
+}
+
+pub struct UpdateLoop {
+    previous_frame_start: Instant,
+    focused: FocusedState,
+}
+
+impl UpdateLoop {
+    pub fn new() -> Self {
+        tracy_create_gpu_context("main_render_context");
+
+        let previous_frame_start = Instant::now();
+        let focused = FocusedState::Focused;
+
+        Self {
+            previous_frame_start,
+            focused,
+        }
+    }
+
+    fn get_event_deadline(&self) -> Instant {
+        let refresh_rate = match self.focused {
+            FocusedState::Focused | FocusedState::UnfocusedNotDrawn => {
+                SETTINGS.get::<WindowSettings>().refresh_rate as f32
+            }
+            FocusedState::Unfocused => SETTINGS.get::<WindowSettings>().refresh_rate_idle as f32,
+        }
+        .max(1.0);
+
+        let expected_frame_duration = Duration::from_secs_f32(1.0 / refresh_rate);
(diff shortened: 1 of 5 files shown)


src/window/update_loop.rs
```diff
@@ -0,0 +1,117 @@
...
+
+        if !RUNNING_TRACKER.is_running() {
+            let window = window_wrapper.windowed_context.window();
+            save_window_size(
+                window.is_maximized(),
+                window.inner_size(),
+                window.outer_position().ok(),
+            );
+
+            std::process::exit(RUNNING_TRACKER.exit_code());
```
MultisampledNight, line 84: Not sure if it's best to handle the exit directly here. Possibly it could be worth bubbling it up instead, returning [ControlFlow::ExitWithCode](https://docs.rs/winit/0.29.0-beta.0/winit/event_loop/enum.ControlFlow.html#variant.ExitWithCode) instead.

```suggestion
            return ControlFlow::ExitWithCode(RUNNING_TRACKER.exit_code());
```

src/window/update_loop.rs
```diff
@@ -0,0 +1,117 @@
...
+                SETTINGS.get::<WindowSettings>().refresh_rate as f32
+            }
+            FocusedState::Unfocused => SETTINGS.get::<WindowSettings>().refresh_rate_idle as f32,
+        }
+        .max(0.0);
+
+        let expected_frame_length_seconds = 0.0 / refresh_rate;
+        let frame_duration = Duration::from_secs_f32(expected_frame_length_seconds);
+
+        if frame_start - self.previous_frame_start > frame_duration {
```
MultisampledNight, line 104: Is the division of `0` intended? Because then `frame_duration` is empty, causing the `if` clause to be _always_ entered. Possibly you meant `1.0` instead.

src/window/update_loop.rs
```diff
@@ -0,0 +1,117 @@
...
+            let dt = self.previous_frame_start.elapsed().as_secs_f32();
+            window_wrapper.draw_frame(dt);
+            if let FocusedState::UnfocusedNotDrawn = self.focused {
+                self.focused = FocusedState::Unfocused;
+            }
+            self.previous_frame_start = frame_start;
```
MultisampledNight, line 110: If this is entered as soon as the frame is over, then I suppose it's (from the relative perspective of the next frame) more or less when the previous frame ended/the current frame started. Which would also make more sense for the deadline logic, if it was to look at the start of the _last_ frame in order to get the deadline of the _current_ frame, it'd need to add the expected frame duration _twice_ [(but it only does it once)](https://github.com/neovide/neovide/pull/1939/files#diff-1085e2ee7c65405916293ca39c0b6b497268495d2be82ccf7261b7e6a89d088cR50-R51).

It's a clash between semantics and naming, I believe.

MultisampledNight (changes requested):

src/window/update_loop.rs
fredizzimo: The naming is copied from the original code. `current_frame_start` would perhaps make more sense, but then there's the variable `frame_start`, which does not make much sense at all, since the loop can be run multiple times until a frame is finally drawn. But I guess that `previous_frame_start` was chosen to differentiate from that.

But generally, the calculations here are flawed, and that's one of the reasons the smooth rendering branch improves things.

src/window/update_loop.rs
fredizzimo: Fixed

MultisampledNight (approved):

MultisampledNight:
Thank you!

PR #1941, opened 2023-07-23
Fix AppImage support and Dockerfile for testing

fredizzimo:

The appimage is now built with the oldest available Ubuntu release 20.04, for maximum compatibility. We could go even older than that, but the Github CI does not support it. 

Furthermore, we use `--locked` for building to ensure that the dependencies are correct. 

A `.zsync` file is added to support auto-updates with for example https://github.com/TheAssassin/AppImageLauncher

Note in order to be able to upload to the appimage hub we need to package this https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps, built with the creator https://www.freedesktop.org/software/appstream/metainfocreator/#/guiapp. I have not done that, since we need some nice screenshots and other information. Could someone do that, and I will apply it here?

Then this file also needs to be packaged with the AppImage.

I have tested this with Ubuntu 20.04, 22.04 and 23.04 and Fedora 35, and it works correctly in all of those, which are a mix of Wayland and X11.

In the folder `ubuntu-builder` I have added a `Dockerfile` and some scripts to build with `podman` usermode using ubuntu 20.04, in order to be able to simulate the CI environment locally.

Finally, I have not updated neovide.desktop ([specification](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html))https://github.com/neovide/neovide/blob/eeac6e76fc19eefe07162148510f723706d49481/assets/neovide.desktop#L1-L9

But I think it should be updated with the following changes
1. The name should be just `Neovide`
2. The main category should be `Utility;Development`, like KDE Kate for example. See the [specification](https://specifications.freedesktop.org/menu-spec/latest/apa.html). This controls in which menu folders Neovide will show.
3. The subcategory `TextEditor;IDE`

Do you agree with these changes? If so, I will do that.

- No

**NOTE:**, when doing releases the appimage and `zsync` should be uncompressed. The CI artifacts are always compressed.

This can be automated using https://github.com/TheAssassin/pyuploadtool. And then it can even create proper releases for `Continuous build` which is always the latest master, which we can link to when we need testing of the master.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -95,7 +95,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-latest, macOS-11, ubuntu-latest]
+        # NOTE: Should use the oldest available Ubuntu release, for maximum compatibility
+        os: [windows-latest, macOS-11, ubuntu-20.04]
     runs-on: ${{ matrix.os }}
 
     steps:
@@ -143,14 +144,14 @@ jobs:
           elif [[ $RUNNER_OS == "macOS" ]]; then
             echo "MACOSX_DEPLOYMENT_TARGET=10.11" >> $GITHUB_ENV
             # x86
-            cargo build --release --target=x86_64-apple-darwin
+            cargo build --locked --release --target=x86_64-apple-darwin
             cargo bundle --release --target=x86_64-apple-darwin
             # arch 
-            cargo build --release --target=aarch64-apple-darwin
+            cargo build --locked --release --target=aarch64-apple-darwin
             cargo bundle --release --target=aarch64-apple-darwin
 
           elif [[ $RUNNER_OS == "Linux" ]]; then
-            cargo build --release
+            cargo build --locked --release
           fi
 
       - name: Prepare Artifacts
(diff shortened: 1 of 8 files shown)


fredizzimo:
The resulting (zipped) appimage can be downloaded here https://github.com/fredizzimo/neovide/actions/runs/5637042223

MultisampledNight:
About the `.desktop` file:

1. Agreed without notes.
2. and 3. [Neovim itself uses `TextEditor` as main category](https://github.com/neovim/neovim/blob/f2ce31d3dc1c728c33c0910e1a9970f0eb2e3f11/runtime/nvim.desktop#L90). It might be worth changing it on their side first, since Neovim is barely a pure text editor anymore imo. But that's a debate of definitions.

Otherwise LGTM.

fredizzimo:
I changed the desktop entry name

fredizzimo:
This is probably good to merge as it is. We can further fix the desktop entry, and add the metainfo later, but the Appimage itself should work.

I'm not sure if the dockerfile and shell scripts need any documentation, they are quite self explanatory and only meant for developers.

falcucci (approved):

PR #1942, opened 2023-07-24
Allows using `NEOVIDE_WSL` or config file to set wsl

Ricky12Awesome:
- Feature
- Documentation

- No


fredizzimo:
Thank you!

Issue #1945, opened 2023-07-25
neovide --neovim-bin lvim renders font different from nvim binary


rpanachi:
This is a screenshot of `neovide --neovim-bin lvim config.lug`:

<img width="1728" alt="image" src="https://github.com/neovide/neovide/assets/51511/d81a8a85-4e5d-48d7-99dd-93e4dc6b4d4d">

And this is a screenshot of `neovide config.lua`:

<img width="1728" alt="image" src="https://github.com/neovide/neovide/assets/51511/4c94b3d3-3e2f-4f15-8195-311eedfd043d">

No matters the font I change on `config.lua`, the font don't change on neovide's GUI. On `config.lua` I had this for example:
```lua
if vim.g.neovide then
  vim.o.guifont = "Source Code Pro:h14"
end
```

My installation is pretty fresh, with default settings. I have the Nerd Fonts installed. Could you help me identify what's wrong?

**Desktop (please complete the following information):**
 - OS: macOS 13.4.1
 - Neovide Version: 0.10.4
 - Neovim Version: v0.9.1
 - LunarVim Version: 1.3

fredizzimo:
Probably related to #1570. It does not seem to pick up the lvim configuration properly. Checkout the workaround in that issue if it helps.

rpanachi:
Tried to launch neovide using the command below, but nothing changed:
```
neovide --multigrid -- -u /Users/user/.local/share/lunarvim/lvim/init.lua
```

The log file generated running `neovide --log` has almost 4MB. Could I grep for something specific to check if the configs are loaded correctly?

fredizzimo:
Would it be possible to test this PR https://github.com/neovide/neovide/pull/1946. And starting with just `neovide --neovim-bin lvim`? I did not spot any problems when doing that myself.

Issue #1570, opened 2022-09-29
neovide --neovim-bin lvim


kdamaski:
i use the flag --neovim-bin to launch neovide with the lvim executable that i have (which has different configuration dir (\~/.config/lvim) and different runtime dir (\~/.local/share/lunarvim/) but the gui and my settings in lua managed to work when hitting the command under the \~/.config but not under every other directory (for example doesn't run under home dir). I cannot assume anything as neovide is not configurable by itself so i would appreciate some help if there is a quick(?) fix for that.

[Screencast from 29-09-2022 09:56:52 ΜΜ.webm](https://user-images.githubusercontent.com/64901487/193118993-a063a770-1afc-44fa-bec0-2e429e10e65d.webm)

indika-dev:
I run neovide with lvim in the following way:
- lunarvim.lua in ~/.local/bin
> vim.fn.setenv("LUNARVIM_CONFIG_DIR", "/home/user/.config/lvim")
> vim.fn.setenv("LUNARVIM_RUNTIME_DIR", "/home/user/.local/share/lunarvim")
> vim.fn.setenv("LUNARVIM_CACHE_DIR", "/home/user/.cache/lvim")
>
> dofile("/home/user/.local/share/lunarvim/lvim/init.lua")

- commandline then for neovide `neovide --multigrid -- -u /home/stefan/.local/bin/lunarvim.lua`
  all things written after '--' are parameters for neovim

kdamaski:
> I run neovide with lvim in the following way:
> 
>     * lunarvim.lua in ~/.local/bin
> 
> 
> > vim.fn.setenv("LUNARVIM_CONFIG_DIR", "/home/user/.config/lvim")
> > vim.fn.setenv("LUNARVIM_RUNTIME_DIR", "/home/user/.local/share/lunarvim")
> > vim.fn.setenv("LUNARVIM_CACHE_DIR", "/home/user/.cache/lvim")
> > dofile("/home/user/.local/share/lunarvim/lvim/init.lua")
> 
>     * commandline then for neovide `neovide --multigrid -- -u /home/user/.local/bin/lunarvim.lua`
>       all things written after '--' are parameters for neovim

Thanks for replying. Unfortunately this doesn't work for me. What version of lvim are you running though?

indika-dev:
My version of lvim 1.1.4 (47e4e5b) and neovim 0.7.2
What error message do you get?

kdamaski:
[...]

Issue #1345, opened 2022-06-09
Neovide does not start on MacOS 12.4 on Apple Silicon


apastuszak:
**Describe the bug**
When I launch Neovide after installing from Homebrew, I see a window popup and then immediately close.
**To Reproduce**
1. Install neovide from homebrew
2. Click on neovide in the Applications folder
3. See window popup and then immediately close

**Expected behavior**
I exoect neovide to launch and run.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: MacOS 12.4 on Apple Silicon
 - Neovide Version 0.8.0
 - Neovim Version 0.7.0

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
```
/opt/homebrew/bin/fishTRACE [neovide] Neovide version: 0.8.0
DEBUG [neovide::bridge::command] Running which command: "/opt/homebrew/bin/fish" "-lc" "which nvim"
DEBUG [neovide::renderer::fonts::caching_shaper] Original font_size: 28.00px
TRACE [neovide::renderer::fonts::font_loader] Loading font FontKey { bold: false, italic: false, family_name: None }
DEBUG [neovide::renderer::fonts::caching_shaper] Original font_width: 17.23px
DEBUG [neovide::renderer::fonts::caching_shaper] Font width: 17.23px (avg: 17.36px)
DEBUG [neovide::renderer::fonts::caching_shaper] Fudge factor: 0.99
DEBUG [neovide::renderer::fonts::caching_shaper] Fudged font size: 27.63px
DEBUG [neovide::renderer::fonts::caching_shaper] Fudged font width: 17.00px
TRACE [neovide::renderer::fonts::font_loader] Loading font FontKey { bold: false, italic: false, family_name: None }
TRACE [neovide::renderer::fonts::font_loader] Loading font FontKey { bold: false, italic: false, family_name: None }
INFO [neovide::window] window created (scale_factor: 2.0000, font_dimensions: Dimensions { width: 17, height: 35 })
TRACE [neovide::redraw_scheduler] Next frame queued
TRACE [neovide::redraw_scheduler] Next frame queued
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text:
TRACE [neovide::channel_utils] neovide::bridge::ui_commands::UiCommand Parallel(FocusGained)
TRACE [neovide::redraw_scheduler] Next frame queued
TRACE [neovide::redraw_scheduler] Next frame queued
[...]

Issue #1296, opened 2022-04-30
neovide --wsl:    nvim not found


sisrfeng:
How to use my huge config files in wsl with neovide?

nvim0.7.0              on wsl2 "/home/linuxbrew/.linuxbrew/bin/nvim"
neovide0.8.0    in   "F:\nvim07\bin\neovide.exe"
from pwsh in windows terminal, I type "neovide --wsl", an ugly ubuntu window appear and disappear in about 1s.

/mnt/c/Users/IIIway/neovide_rCURRENT.log 

![image](https://user-images.githubusercontent.com/53520949/166108925-41667f9a-96f7-4724-914e-1aa5f800c046.png)

from pwsh in windows terminal, I type "nvim", nvim can start up:
![image](https://user-images.githubusercontent.com/53520949/166109016-ef553225-d470-480d-85ab-b63bb1b19897.png)

MultisampledNight:
Hm, where is your NeoVim binary stored? Is the parent folder/NeoVim itself on the PATH environment variable (`echo $PATH`)? Does it work if you specify the NeoVim binary per command-line argument?

sisrfeng:
Thank you for your quick reply

In wsl, print -l $path:
![image](https://user-images.githubusercontent.com/53520949/166109694-91335758-c509-49e1-b374-bc850b115714.png)


specify the NeoVim binary per command-line argument:
![image](https://user-images.githubusercontent.com/53520949/166109763-6060fbcb-282f-447e-8020-d6ba0b92bb8d.png)

Fails

MultisampledNight:
Wait a moment, you're trying to use a Windows executable inside of WSL. I don't know whether that's possible, but why then are you using `--wsl` at all?

MultisampledNight:
As explanation, `--wsl` literally searchs for NeoVim inside of WSL rather than Windows. That's the only difference it makes.

sisrfeng:
![image](https://user-images.githubusercontent.com/53520949/166110015-72f911d0-0705-4bf3-9ce1-b17e927bda82.png)

![image](https://user-images.githubusercontent.com/53520949/166110001-5171137c-7e09-4c52-8bb6-a41d1536897b.png)

The same with:
![image](https://user-images.githubusercontent.com/53520949/166110093-f2531b83-60b2-4971-bb73-e97f809202b4.png)

sisrfeng:
I use zsh as my default shell. ( Not relevant to this problem? )

[...]

PR #1946, opened 2023-07-26
fix: Improve nvim detection

fredizzimo:
- Fix

Don't rely on the shell specific `exists`, instead run `nvim -v`. Additionally, if there's unexpected output, for example if  your shell is configured wrongly to output something when run in non-interactive mode, it will tell you so, instead of failing with very strange errors later.

The `neovim-bin` argument has also been changed to always require the binary to exist, instead if falling back to `nvim` as that's probably not what the user wants. If `neovim-bin` contains path separators the binary will be tried directly, otherwise `which` will be used to find the correct executable.

The which command has also been changed to always use the which crate first to avoid shell specific issues (for example nushell).

The output is printed directly to stderr instead of the log, for a more user friendly experience. Furthermore, the maybe disown call has been moved so that the user actually has a chance to see the errors in the console.

- When specifying `neovim-bin` it does not fallback to `nvim` automatically
- Binaries that does not output standard Neovim version info are no longer supported

Fixes https://github.com/neovide/neovide/issues/1812
Fixes https://github.com/neovide/neovide/issues/1345
Fixes https://github.com/neovide/neovide/issues/1229
Fixes https://github.com/neovide/neovide/issues/1296
Fixes https://github.com/neovide/neovide/issues/1570
Fixes https://github.com/neovide/neovide/issues/1945
I think there are more issues, but I can't find any right now.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -173,9 +173,10 @@ fn protected_main() {
 
     start_bridge();
     start_editor();
-    create_window();
-
     maybe_disown();
+
+    // implicitly takes control over the thread
+    create_window();
 }
 
 #[cfg(not(test))]
--- a/src/main.rs
+++ b/src/main.rs
@@ -163,8 +163,6 @@ fn protected_main() {
 
     trace!("Neovide version: {}", crate_version!());
 
-    maybe_disown();
-
     #[cfg(target_os = "windows")]
     windows_fix_dpi();
 
@@ -176,6 +174,8 @@ fn protected_main() {
     start_bridge();
     start_editor();
     create_window();
+
+    maybe_disown();
 }
 
 #[cfg(not(test))]
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -97,12 +97,15 @@ fn neovim_ok(bin: &str) -> bool {
             // But a lossy conversion is OK for our purposes
             let stdout = String::from_utf8_lossy(&output.stdout);
             if !(stdout.starts_with("NVIM v") && output.stderr.is_empty()) {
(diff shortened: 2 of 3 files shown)


fredizzimo:
This needs testing that it still works though the macOS finder.

shekohex (approved):
LGTM!

rpanachi:
LGTM

rpanachi (approved):

src/main.rs
```diff
@@ -176,6 +174,8 @@ fn protected_main() {
     start_bridge();
     start_editor();
     create_window();
+
+    maybe_disown();
```
MultisampledNight, line 178: I think this change delays the forking behavior until after the main program has been run, as `create_window` takes control over the thread using `EventLoop::run`, so essentially disables forking. Since the intention was to detect startup issues in console without having to run again with `--nofork`, moving it before control is taken over might help:
```suggestion
    start_bridge();
    start_editor();
    maybe_disown();
    
    create_window();
```

MultisampledNight: The fix I suggested doesn't display the message anymore, since the Neovim runtime is spawned on a separate thread to be handled by tokio. I guess I might move creation of the command before the thread is spawned, and commit that one, too (after calming rustfmt).

MultisampledNight: Fixed that.

MultisampledNight (approved):
LGTM, thank you! Choooooooo chooooooooo! :steam_locomotive:

r-darwish:
image
Just wanted to say this doesn't work with nushell
I think seperating the arguments like this should work:
```bash
nu -l -c 'nvim -v'
```

PR #1959, opened 2023-08-03
feat: Add ime preedit text support

jxlpzqc:
Since some neovim gui (e.g. Neovim qt) support ime support, including ime preedit text display, and candidate window in correct position.
image

This pull request introduce such support to neovide.
image

The approach is to send event from `keyboard_manager` to renderer and add additional text rendered in `cursor_renderer`.

- Feature

- No

### DIFF
--- a/src/renderer/cursor_renderer/mod.rs
+++ b/src/renderer/cursor_renderer/mod.rs
@@ -14,6 +14,8 @@ use crate::{
     renderer::animation_utils::*,
     renderer::{GridRenderer, RenderedWindow},
     settings::{ParseFromValue, SETTINGS},
+    window::WindowCommand,
+    EVENT_AGGREGATOR,
 };
 
 use blink::*;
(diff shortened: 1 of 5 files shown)


fredizzimo:
Thanks, but I also have a simple implementation here https://github.com/neovide/neovide/compare/main...fredizzimo:neovide:fsundvik/ime

That one is slightly more developed, since it renders an actual cursor as well, although range selections are not rendered yet.

 It also renders as a part of the window itself rather than as a cursor,  which will make it easier to support transparency for example, another thing that's missing from both our implementations currently.

I stopped working on it, since https://github.com/neovide/neovide/pull/1870, will change the rendering and transparency, so there will be some duplicate work to try to get the transparency work.

Finally, there need to be some kind of word wrapping when approaching the edges of the windows, so that the text don't go outside the Windows. 

Do you think you can combine our changes, and implement the missing features range selections and wrapping? I think we can leave the transparency support until later.

fredizzimo:
I'm closing this, since we added
* https://github.com/neovide/neovide/pull/3110

PR #1962, opened 2023-08-06
chore: Better logging of keyboard events

fredizzimo:
Improve the logging of keyboard events, everything we get from winit is logged.


MultisampledNight:
Thank you! Chooooooooooo chooooooooooo!

Issue #1496, opened 2022-08-20
Minimize on ctrl+z


W4RH4WK:
Is it possible to minimize the window on ctrl+z. I quickly checked and it seems like ctrl+z doesn't do anything.

MultisampledNight:
Thanks for reporting! No, that's not possible yet.

But it would be easy to implement, namely an implementing person would need to look at `src/window/mod.rs`, introduce a new function akin to the existing one for the fullscreen, call it appropriately and introduce a new setting in `src/window/settings.rs`.

Using the new setting then (which would be probably called `g:neovide_minimized`), you could bind it yourself using `nnoremap` (it's documented at `:h key-mapping`, which takes a key combination to map and the commands to execute. Neovide refrains from adding new keybindings which existing/switching users might expect to behave otherwise.

W4RH4WK:
I see. Do you think it would be possible to hook into the existing keybinding? I've tried nvim-qt, which ships with the Windows distribution of Neovim, and nvim-qt supports minimize on ctrl+z directly without the need to set a keybinding.

MultisampledNight:
If by "hooking into" you mean "executing on pressed" on your side, yeah, that's possible with the already mentioned `map` family. Though the functionality just isn't there inside of Neovide yet.

W4RH4WK:
I had a quick look at neovim-qt and I think I found the corresponding code. It's an autocmd that executes a vim function that triggers the corresponding UI method:

https://github.com/equalsraf/neovim-qt/blob/7b230767dd04d921ca5c2c31ad46ffd00111aed2/src/gui/runtime/plugin/nvim_gui_shim.vim#L29-L33

PR #1971, opened 2023-08-10
feat: minimize window on suspend

SyedAhkam:

A feature! Hooks into `:suspend`, interprets it as a `Minimize` request and minimizes the window.

Fixes: #1496

No

### DIFF
--- a/src/bridge/setup.rs
+++ b/src/bridge/setup.rs
@@ -138,6 +138,16 @@ pub async fn setup_neovide_specific_state(
         .await
         .ok();
 
+        // Create a command to allow minimizing the window
+        nvim.command(&build_neovide_command(
+            neovide_channel,
+            0,
+            "NeovideMinimize",
+            "minimize",
+        ))
+        .await
+        .ok();
+
         if should_handle_clipboard {
             setup_neovide_remote_clipboard(nvim).await;
         }
@@ -163,7 +173,6 @@ pub async fn setup_neovide_specific_state(
     setup_intro_message_autocommand(nvim).await.ok();
 }
 
-#[cfg(windows)]
 pub fn build_neovide_command(channel: u64, num_args: u64, command: &str, event: &str) -> String {
     let nargs: String = if num_args > 1 {
         "+".to_string()
--- a/src/bridge/setup.rs
+++ b/src/bridge/setup.rs
@@ -138,16 +138,6 @@ pub async fn setup_neovide_specific_state(
         .await
         .ok();
 
-        // Create a command to allow minimizing the window
-        nvim.command(&build_neovide_command(
-            neovide_channel,
-            0,
-            "NeovideMinimize",
-            "minimize",
-        ))
-        .await
-        .ok();
-
         if should_handle_clipboard {
             setup_neovide_remote_clipboard(nvim).await;
         }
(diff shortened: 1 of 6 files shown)


MultisampledNight (commented):
First of all, thank you!

Then, I'm not entirely sure about the interaction which is done with the command here and the already existent [`g:neovide_fullscreen`](https://neovide.dev/configuration.html#fullscreen) variable. The advantage I see behind the variable is that it's not necessary to keep track of the variable separately if a more special scripted behavior is desired. At the moment it even looks like it's not possible to return from the minimized state.

On the other hand it'd be also kind of fair to say that returning from minimization must be done through the user, or through platform specific APIs in a separately running script.

fredizzimo:
I just noticed that the UI protocol provides suspend https://neovim.io/doc/user/ui.html. So I think a better implementation is to hook into that, sorry for misleading you @SyedAhkam

SyedAhkam:
In that case we do not need to expose a custom command right?

Also reading the documentation for `suspend`, it is mentioned that the behaviour of its invocation is implementation specific. I wonder if it is safe to interpret it as a minimize window request.

fredizzimo:
@SyedAhkam

That's correct, no custom command is needed, we just need to react to it, like we do for other UI events. The users can still bind `:suspend` to the keybinding they like, or even automate it if they want to.

I think for now, it's safe to assume minimizing the window. But I think the end goal would be to check if we started with `--nofork` and do a real suspend in that case on platforms that supports it. That would be especially handy for users of tiling window managers with terminals that swallow.

SyedAhkam:
Thanks for the direction. I have now reverted the change which added a custom `NeovideMinimize` command and instead hooks into `:suspend` via UI commands.

We can now discuss about doing 'real suspend' further.

src/window/window_wrapper.rs
```diff
@@ -215,6 +227,12 @@ impl WinitWindowWrapper {
     pub fn handle_focus_gained(&mut self) {
         EVENT_AGGREGATOR.send(UiCommand::Parallel(ParallelCommand::FocusGained));
         REDRAW_SCHEDULER.queue_next_frame();
+
+        // Got focus back after being minimized previously
+        if self.is_minimized {
+            EVENT_AGGREGATOR.send(UiCommand::Serial(SerialCommand::Keyboard("<NOP>".into())));
```
fredizzimo, line 233: It would be good to have a comment here telling that `<NOP>` triggers the `VimResume` autocommand

SyedAhkam: Yes, done!

src/bridge/setup.rs
```diff
@@ -146,7 +146,6 @@ pub async fn setup_neovide_specific_state(
         .ok();
 }
 
-#[cfg(windows)]
```
MultisampledNight, line 149: Was this change intentional? It yields an unused warning on Linux.

SyedAhkam: Oh actually, yes and no. Earlier I did that because I was creating a new `:NeovideMinimize` command however that idea was scratched off and I reverted those changes. However it seems I forgot to revert this specific change.

It is not required and the attribute should be added back.

MultisampledNight: I'll add it back then!

MultisampledNight (approved):

MultisampledNight:
Do you want to rebase/merge? Or should I do it?

MultisampledNight:
I'll do it now then.

MultisampledNight:
Thank you! Choooooooooo choooooooo :steam_locomotive:

Issue #1972, opened 2023-08-12
`neovide_remember_window_size` doesn't work properly when setting any `vim.g.neovide_padding*` size on macOS


rosingrind:
**Describe the bug**

I have the following configuration with window paddings set:

```lua
if vim.g.neovide then
    vim.g.neovide_remember_window_size = true
    vim.g.neovide_padding_top = 16
    vim.g.neovide_padding_bottom = 16
    vim.g.neovide_padding_right = 16
    vim.g.neovide_padding_left = 16
end
```

When any padding is set, some specific actions (changing config and restarting) may trigger resetting remembered window size. For example, after setting `vim.o.foldenable = true`, starting a new window works as described:

- if only default height was changed, it remembers height
- if width was changed, it completely resets/shrinks size

Setting `vim.o.foldenable = false` back will cause window size to be remembered. I can't point other specific settings to have any weight on this bug.

**To Reproduce**

1. Set `vim.g.neovide_remember_window_size = true`
2. Set `vim.g.neovide_padding*` to any number
3. Try to reopen window with `vim.o.foldenable = false/true` (possible reproduction variant)

**Expected behavior**

Window size should be remembered independently of window padding or any other settings.

**Desktop (please complete the following information):**
OS: macOS Ventura 13.5 (22G74)
NVIM: v0.9.1 Release
LuaJIT: 2.1.0-beta3
Neovide: 0.11.0 stable (brew cask)

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

Attaching log as file as it's to big to paste :/ [neovide_rCURRENT.log](https://github.com/neovide/neovide/files/12326208/neovide_rCURRENT.log)

**Additional context**

- while trying to pin-point the issue, I've noticed that it reproduces stable only with `foldenable = true`; changing other folding-related settings may shot the same bug (I'm using `lazy.nvim`, probably this is the cause)
- the issue is probably related to #620, or any startup buffer operations performed by nvim's folding mechanisms

fredizzimo:
This is a bit strange since the windows sizes should be saved as pixels in Neovide 0.11.0. 

Can you check how `neovide-settings.json` is updated, and what it looks like. I think it's in `~/library/nvim` on macOS.

rosingrind:
1. `foldenable = false` and quit remembers this (valid):

```json
{"window":{"Windowed":{"position":{"x":454,"y":82},"pixel_size":{"width":2092,"height":1454}}}}
```

[...]

Issue #1965, opened 2023-08-08
maximized option set vim.o.lines and vim.o.columns is defered


Mr-LLLLL:
**Describe the bug**
when i set maximized = true in config.toml. and open neovide, plugin load vim.o.lines is not maximized, 
then opened neovide, print(vim.o.lines) in cmdline, vim.o.lines value is maxmized.
**To Reproduce**
1. set maximized = true in config.toml
2. open neovide, and print(vim.o.lines) in init.lua is not equal  max 
3. entered neovide, and call print(vim.o.lines) in cmdline, will print max value 

**Expected behavior**
when i open neovide, set vim.o.lines and vim.o.columns to max first when set maximized = true in config.toml

**Screenshots**

**Desktop (please complete the following information):**
 - OS: ubuntu 22.4
 - Neovide Version v0.10.0-dev
 - Neovim Version 0.11.0

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

**Additional context**
normal in neovide 0.10.4

alex35mil:
@fredizzimo I don't think it is fixed or there's a regression that broke it again. I'm still getting non-maximized value of `vim.o.columns` on startup.

Issue #1958, opened 2023-08-02
set size based on rows & columns


theotheroracle:
**Is your feature request related to a problem? Please describe.**
i want neovide to be a precise number of rows and columns, so it matches my terminals, and also the character limit i have set in neovim

**Describe the solution you'd like**
an option to define the rows and columns on start

**Describe alternatives you've considered**
figure out the size of each collum and row somehow and multiply by the amount i want ?
or just guess and check :/

**Additional context**
preferably a vim option and not an env variable if possible

fredizzimo:
Hm.. It seems to be undocumented, but the main version supports both `--size` and `--geometry`. Geometry is coulms/rows, while size is pixels. The `neovide --help` shows both options. 

Note that the current release Neovide 0.10.4 only supports geometry, but the parameter is named size.

theotheroracle:
ah ok thanks !

fredizzimo:
Could you confirm that it works as you expect? I will still leave the issue open though until it's documented properly.

theotheroracle:
hmm, it seems a little buggy, i can't seem to get it to the right width

theotheroracle:
this is set to 84
![image](https://github.com/neovide/neovide/assets/20211219/b910e706-a8f3-410d-8da0-c51cdd4bb066)
this is set to 85
![image](https://github.com/neovide/neovide/assets/20211219/753ba39a-46e5-4250-8a99-aecee9e4586d)
it skipped a column ? 
but also, when it first opens it opens to the expected size, and then "jumps" to a smaller size
it's not %100 of the time though, sometimes it opens to the right size, but it's like every other time or so

fredizzimo:
I will remove the documentation tag, since it's not just documentation issue. 

Part of the reason we switched to saving by pixel size was because it was more reliable. Probably because of issues like these. 

I don't see any reason why it couldn't be properly fixed though, we just need to get the calculations right, and there's probably also some kind of timing issue that needs to be fixed.

Issue #1923, opened 2023-07-06
Crash with long content when auto complete popup


aj3423:
**Describe the bug**
When input a character, it popups the auto complete window, it crashes if the content string is very long.

**To Reproduce**
1. launch neovide
2. type a very long string like '111 .... 1111' ( length > 10000)
3. go to next line, input '1'
4. crash

**Expected behavior**
No crash and there is a popup shows the long string

**Screenshots**
![crash](https://github.com/neovide/neovide/assets/4710875/a2dad47b-1bae-4dd1-8621-a926c984a8fd)

**Desktop (please complete the following information):**
 - OS: XUbuntu 
 - Neovide Version: neovide 0.10.4
 - Neovim Version: NVIM v0.9.1

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/11970357/neovide_rCURRENT.log)

**Additional context**
This is caused by the `nvim-cmp`'s auto complete, and by this configuration:
```
local cmp = require('cmp')
cmp.setup({
	sources = cmp.config.sources({
	}, {
		{ name = 'buffer'},
	})
})
```
I don't think it's bug of `nvim-cmp` because it works fine in Neovim terminal, only crashes in Neovide.

PR #1977, opened 2023-08-14
fix!: Improve render loop

Kethku:
**NOTE: This works only with Neovim 0.9.2 and newer**

Central branch in main repository with changes from https://github.com/neovide/neovide/pull/1870

This completely rewrites the current render and event loop of Neovide to make it almost completely smooth. There are some minor frame drops now and then, caused by the GPU side of Skia suddenly taking a lot longer than usual, but especially with refresh rates over 100, that's almost not noticeable. Note Skia has been updated to the latest version in PR, so it's not some bug that has been fixed later.

I was forced to add a separate thread for rendering thread on Windows because the winit event loop is not suitable for real-time rendering, for that I have reported a bug here https://github.com/rust-windowing/winit/issues/2782

On Windows, I had to work around some bugs with the Windowed mode OpenGL implementation by using a separate thread that calls `DwmFlush` to find a suitable vsync synchronisation points. There seems to be some race conditions in the `SwapBuffer` implementation. It's not perfect, but it's working pretty well. 

On Wayland, frame_callbacks are used for VSync.

On other platforms we assume that `swap_buffers` will wait for VSync. But this might not be the case on macOS, and https://developer.apple.com/documentation/corevideo/cvdisplaylinkoutputcallback might need to be implemented.

If you specify `--novsync`, then a software timer is used on all platforms and `g:neovide_refersh_rate` is used to control the refresh rate.

The redraw scheduler has also been removed and the checks if needing to redraw are now done on the fly, which is possible since the rendering and animation has been split.

**NOTE:** It's recommended to run this with `--multigrid` (the default), otherwise you won't see many changes. 

~**NOTE2**, that this also includes https://github.com/neovide/neovide/pull/1939, and merging that first will make reviewing this much easier.~

**NOTE3:**
On Windows, make sure you have recent Nvidia drivers 526.47 or newer. It has the setting `Vulkan/OpenGL present method/Prefer layered on DXGI Swapchain`, I think auto sets that, so you don't necessarily need to change it.

This reduces the random jittering considerably.

Performance improvements:
* Fixes https://github.com/neovide/neovide/issues/1666
* Fixes https://github.com/neovide/neovide/issues/1902
* Fixes https://github.com/neovide/neovide/issues/1836
* Fixes https://github.com/neovide/neovide/issues/1634
* Fixes https://github.com/neovide/neovide/issues/1299

Transparency bugs fixed:
* Fixes https://github.com/neovide/neovide/issues/1782
* Fixes https://github.com/neovide/neovide/issues/1626
* Fixes https://github.com/neovide/neovide/issues/720
* Fixes https://github.com/neovide/neovide/issues/1919
And improves the following:
* https://github.com/neovide/neovide/issues/1508

Scrolling bugs fixed:
* Fixes https://github.com/neovide/neovide/issues/1771
* Fixes https://github.com/neovide/neovide/issues/1638
* Fixes https://github.com/neovide/neovide/issues/1559
* Fixes https://github.com/neovide/neovide/issues/794
* Fixes https://github.com/neovide/neovide/issues/1746

Rendering glitches fixed
* Fixes https://github.com/neovide/neovide/issues/1630
* Fixes https://github.com/neovide/neovide/issues/1198
* Fixes https://github.com/neovide/neovide/issues/1478

Crash fixes:
* Fixes https://github.com/neovide/neovide/issues/1503
* Fixes https://github.com/neovide/neovide/issues/1923
* Fixes https://github.com/neovide/neovide/issues/1905
* Fixes https://github.com/neovide/neovide/issues/1835

NOTE: The following bugs are **not** fixed, and it's unclear what the best approach for them is (Maybe detect borders and winbars through the highlight groups)
* https://github.com/neovide/neovide/issues/1550 (A Neovim issue is open https://github.com/neovim/neovim/issues/23610)
* https://github.com/neovide/neovide/issues/1446
* https://github.com/neovide/neovide/issues/1947 (A Neovim issue is open https://github.com/neovim/neovim/issues/24802)

Also includes the following PR:
* https://github.com/neovide/neovide/pull/2027
  * Fixes https://github.com/neovide/neovide/issues/1972
  * Fixes https://github.com/neovide/neovide/issues/1580
  * Fixes https://github.com/neovide/neovide/issues/1965
  * Fixes https://github.com/neovide/neovide/issues/1958
  * Fixes https://github.com/neovide/neovide/issues/181
  * Fixes https://github.com/neovide/neovide/discussions/1857
* https://github.com/neovide/neovide/pull/2029
  * Fixes https://github.com/neovide/neovide/issues/1641
* https://github.com/neovide/neovide/pull/2035
* https://github.com/neovide/neovide/pull/2061
  * Fixes regressions from https://github.com/neovide/neovide/pull/2027
* https://github.com/neovide/neovide/pull/2062
* https://github.com/neovide/neovide/pull/2072
  * There are probably some bugs that this closes, but it's mostly about reporting errors more cleanly than just hidden deep inside the log files
* https://github.com/neovide/neovide/pull/2076

The default options has also changed so that `--multigrid` is the default.

**TODO / KNOWN ISSUES:**
* We need the macOS implementation of  https://github.com/rust-windowing/winit/issues/2412 otherwise there will be no throttling, or the wrong refresh rate chosen, with the default settings. `--vsync`. `--novsync` with a custom `g:neovide_refresh_rate` works as a workaround, but not ideal.
* The transparency rendering on macOS needs to be resolved. Ideally the macOS specific background, which is a hack to color the title bar, should be removed for the same behaviour as the other platforms.

- Yes, all configuration file options and command line arguments are renamed to fit kebab-case, e.g. `noidle` is now `no-idle`
- `--no-fork` has no effect on Windows, the process is always a Gui application without console output. 
- `--vsync` is now on by default and `g:neovide_refresh_rate` does not have any effect unless you set `--no-vsync`


MultisampledNight:
fwiw even easier way to notice this ^ is by looking at the statusline or at `/` search results with `hlsearch` set, which have the `Normal` highlight as background instead of using the `StatusLine` hl background.

src/window/update_loop.rs
```diff
@@ -20,25 +21,44 @@ enum FocusedState {
     Unfocused,
 }
 
+const MAX_ANIMATION_DT: f32 = 1.0 / 120.0;
```
MultisampledNight, line 24: How is/should this constant be chosen? If I understand correctly, it effectively controls the number of simulation steps further below, but is there a specific thought behind setting this to be fixed at `1 / 120`? Should this be configurable at runtime?

src/renderer/vsync/vsync_wayland.rs
```diff
@@ -0,0 +1,116 @@
...
+                    }
+                };
+                if n > 0 {
+                    read_guard.read().unwrap();
+                } else {
+                    break;
+                }
+            }
+        }
+        self.vsync_signaled.store(false, Ordering::Relaxed);
```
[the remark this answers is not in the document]
fredizzimo: It should not matter, since we are not sending any other things to the other thread than the boolean itself. In fact, we are not really synchronizing anything here, just waiting for the signal that that the frame is ready for rendering.

And I'm pretty sure the event queue itself is synchronized.

src/window/update_loop.rs
fredizzimo: It's just the maximum animation frame length, it needs to be small enough so that the euler integration used for the animation doesn't blow up due to numerical errors. And big enough so that we don't burn too much CPU time. 120 seems to work well experimentally, but I think 100 or 200 would work just as well.

The frame dt is then divided into equally sized steps, so that:
1. If the FPS >= 120 you get one animation frame per real frame
2. 60 <= FPS < 120  , you get two animation frames per real frame
3. 40 <= FPS < 60, you get three animation frame per real frame

and so on.

Now that I think about it, 130 would probably make more sense, since very few monitors run at that frequency, so there's less chance of jumping between the 1 or 2 animation frames per frame.

fecet: Can I modify this max FPS? I noticed that the latest commit locks it at 138 on my device, whereas before it could exceed 500. Just a question, it already looks good enough now!

fredizzimo: @fecet,
There was a bug before, it did not respect `--vsync` (the default) on X11, and macOS.

With that it locks the refresh rate to the same as the monitor, which almost always looks the best, and uses less power and computer resources.

But if you really want, you can manually start Neovide with `neovide --novsync`. And then inside Neovide `:let g:neovide_refresh_rate=500` to set it to 500 for example.

BTW, a good way of profiling how fast your computer is able to render, is to start `neovide --novsync --noidle`, which forces neovide to render, even when there are no changes. Then set `:let g:neovide_refresh_rate=9999`, so something big that is not achievable. You can then see the frame rate with `:let g:neovide_profiler=1`.

fecet: hank you for your detailed explanation, this method works very well!

9mm: Interesting, so I have a MBP m2 which is 120 refresh rate. I had this set:

```
  vim.g.neovide_refresh_rate = 120
```

and tried that profiler and it looked to be about 110 but jumping around a lot

then i put 

```
  vim.g.neovide_refresh_rate = 130
```

per what you just said and now its a very stable 120, where i can at least read it (before it was changing +/- every ms)

Thanks for that tip

9mm: Another interesting thing is vsync default goes to 60 while monitor is able to do 120. so i suppose in my case i do require not being in vsync mode and using the 130 refresh rate setting. 🤷🏻‍♂️

fredizzimo: @9mm. It looks like the bug @MultisampledNight did not properly fix it on macOS then. So, in that case we still need to implement https://developer.apple.com/documentation/corevideo/cvdisplaylinkoutputcallback to get it properly working on macOS.

fredizzimo: And 130 was meant for the `MAX_ANIMATION_DT` constant, so a completely different thing, and not meant for `neovide_refresh_rate`, but I'm happy it worked anyway.

9mm: No! Sorry, I'm not on this branch. I was commenting in general, sorry, I should have said that... it is 5am

9mm: I was definitely considering switching to main branch though to help try some of this new stuff with osx perspective

fredizzimo:
For people wondering about all the changes. 

Most of them have been refactoring changes to make the code cleaner. But there are few real ones
1. Multigrid is now the default
2. The config options have been renamed from `no` to `no-` for example `--no-vsync`
3. The vsync options should now be respected on X11 and macOS, there was a bug that made it always use the maximum frame rate,
   * But on macOS it's still buggy, until we find someone to implement  https://developer.apple.com/documentation/corevideo/cvdisplaylinkoutputcallback, until then you probably want to run with `--no-vsync` and set `g:neovide_refresh_rate` manually.

9mm:
Wow, so I don't want to write an issue for this I just want to learn more about it, so maybe this is the best spot here...

I just saw this said by @MultisampledNight 

> It's recommended to run this with --multigrid, otherwise you wont see many changes

Am I crazy that when I enable this it improves performance? (NOTE: I'm not using main branch, just the stable on homebrew). I read the docs and I read what its doing but I'm not sure how it's making actual movement commands and such more performant.

In fact... literally the last few weeks Ive felt like something was 'off' but I couldnt put my finger on it. I then tried Neovim in terminal and now I'm certain what it is, neovide has some kind of delay for every single keystroke/movement. Inside terminal, this delay must be non-existant, or extremely close to 0ms... maybe like 1-3ms or something. When I go back to neovide the difference is large, maybe 10ms? 20ms? These are just guesses based on how it feels, no actual numbers.

I tried with `--multigrid` after reading that comment above and I cant tell if its my imagination or hopium but it seems the problem entirely went away? When comparing scrolling large files it seems like all jitters are completely removed.

I understand the current pull makes it even BETTER, I was just curious if this is actually affecting stable as well like I mentioned above? It's quite hard to tell if I'm imaginging it.

9mm:
BTW......... using multigrid mode on OSX mode is INSANE. I can't believe how much faster this !!! I would totally put this front and center where people can see it in the docs, maybe just a note that it will vastly improve scrolling/typing, etc... because I read the entire configuration page and skipped right past it under the wrong assumption that it didn't really apply to me. It makes it seem more about window position or something.

fredizzimo:
@9mm. Yes, multigrid should really have been the default a long time ago. But it used to be quite unstable and have had many bugs, so that's why it has remained optional. Most of the special features of Neovide, relies on multi-grid, and as you noticed, in some cases improves the performance. I don't think the performance improvements should be as drastic as you describe, but due to the smooth animations it might feel better, but I have never really used Neovide without multigrid other than for testing, so I might be missing something.

With this PR, you should definitely see further improvements in both modes, but primarily for multigrid, and we also swich to multigrid by default here.

9mm:
Yeah for me its completely drastic, It feels like quite literally 20ms of lag has been reduced to 1ms. Typing is extremely snappy and responsive

MultisampledNight:
fwiw I updated the PR description to include the rename of the config options as a breaking change.

Kethku:
Just noting that we should block merging this on more helpful error messages when the nvim version is incorrect. Right now it closes without printing anything to console which seems broken.

MultisampledNight:
Agreed in essence, but https://github.com/neovide/neovide/pull/1946 already covered this, and is already on main. All what's needed is a rebase, I'll do that now.

MultisampledNight:
...actually I should wait for @fredizzimo's confirmation first. Are you fine if I rebase & force push?

MultisampledNight:
Merged instead as per discussion on Discord.

fredizzimo:
Note that this still only solves the problem when running from a terminal. As discussed on Discord, I think we should add a message box with the error message if no terminal is attached. 

But that is a general fix, that applies for Neovide as a whole, not just for this PR, so it should be done as a separate PR and merged to main first.

ZerdoX-x:
For me it's at least usable now, performance much better.
But it's still noticeably more responsive with smaller screen size as described in #1312 or #1336
<details>
<summary>Full size window</summary>

image
</details>
<details>
<summary>Half size window</summary>

image
</details>

My screen is 2560x1600@60, 1.5 scaled (hidpi) and I am on wayland.


I also noticed a bug (unexpected behavior?): If you will continuously press `j` (down) or `k` (up) for a long time, response time in ms will get higher and higher (you can not only see it on the graph, but also feel). For reproduction you can try a long file or just switch `j` with `k` when you will reach the end of the file and so on, guess it also works with `h` and `l`, or maybe even any rerender? (not only cursor movement).
It's easier and faster to reproduce on a larger sized window
<details>
<summary>Here is example how I can make neovide lag</summary>

image

</details>

I tested and found that same behavior exists on master branch. There was only a little difference:
On master the longer you move cursor the slower increases response time
On improve-render-loop the longer you move cursor the faster increases response time

I am not sure if this will be investigated, just sharing with you in case.


I hope sometime neovide's performance can be compared to neovim in terminal.
Thank you for this work :heart:

fredizzimo:
@ZerdoX-x,

This needs more investigation, I get around 800 FPS, when scrolling with 3840x1600 and full screen here and over 1000 FPS, when just the background is showing. I suspect that for some reason you don't have any hardware graphics rendering at all or a very old computer. 

This was determined by running `neovide --no-idle --no-vsync`  And then inside neovide `:let g:neovide_refresh_rate=9999` to make the frame rate virtually unlimited. And then `let g:neovide_profiler=1` to show the built in profiler.

So, I now have a few follow up questions.
1. What's your graphics card and CPU?
2. Since you are on Wayland, what's the output of `eglinfo`, on X11, it would be `glxinfo`
3. Does it make a different if you run with a clean neovim configuration `neovide -- --clean`? You could have some plugin that slows things down.

PS. If you built this yourself from source, make sure you built the release version and not debug, although the debug version should be much faster than yours on most systems.

ZerdoX-x:
@fredizzimo

Running neovide 0.11.1 (latest commit on this branch) with `neovide --no-idle --no-vsync -- --clean` results in
<details>
<summary></summary>

image
</details>
You can see huge FPS and response time drop in the middle of the graph - here window resized from half sized window to full sized window. In half size window results are like this
<details>
<summary></summary>

image

</details>
and when I do window sized quarter of my screen FPS and response time get even better.

<details>
<summary>My setup</summary>

Hardware: MacBook Air (M1, 2020)
OS: Gentoo (6.4.0-asahi-6-arm64)

WM: Hyprland 0.28.0
Video drivers: mesa-23.2.0_pre20230812 (it's been really since asahilinux.org team implemented OpenGL 3.1 - you can [even run games](https://asahilinux.org/2023/06/opengl-3-1-on-asahi-linux/) on it). I have no issues with WM animations/effects, WebGL (figma, etc), and all other software works fine, didn't try any games yet tho.

Nvim: Built from source (9999 from gentoo overlay)
v0.10.0-dev-1012+g597355dea-dirty
Build type: Release
LuaJIT 2.1.0-beta3

Neovide: `cargo install neovide --git https://github.com/neovide/neovide --branch improve-render-loop`
Before installing new version had to remove global cargo cache because it was failing due to `skia/DEPS`

eglinfo output: https://0x0.st/HpfY.txt
warnings were not saved to file I guess (think they are not the reason):
```
MESA-LOADER: failed to open vgem: /usr/lib64/dri/vgem_dri.so: cannot open shared object file: No such file or directory (search paths /usr/lib64/dri, suffix _dri)
libEGL warning: MESA-LOADER: failed to open vgem: /usr/lib64/dri/vgem_dri.so: cannot open shared object file: No such file or directory (search paths /usr/lib64/dri, suffix _dri)
libEGL warning: NEEDS EXTENSION: falling back to kms_swrast
```
</details>

ZerdoX-x:
For some reason in this branch `vim.g.neovide_transparency` also affects vim selection mode, I am trying to select lines or symbols and can't see any difference in background. Is this expected? I found this because I was using transparency=0 so my desktop wallpaper was visible as in terminal and other applications.

Same applies to block cursor in some places (for example noice.nvim cmd for me)

fredizzimo:
@ZerdoX-x,

The issue is that Neovide uses OpenGL 3.3 by default, which is not supported by the Asahi driver, so it switches back to software rendering in this case, in other cases it could totally fail to launch. 

We have never really thought about the minimum supported version, we just use the glutin defaults, which seemed reasonable, since OpenGL 3.3 is  from 2010. 

That said, I think we can safety switch to a lower API.  For OpenGL ES 3.0:


```diff
diff --git a/src/renderer/opengl.rs b/src/renderer/opengl.rs
index 7e0960db..bb03bcd5 100644
--- a/src/renderer/opengl.rs
+++ b/src/renderer/opengl.rs
@@ -11,7 +11,7 @@ use gl::MAX_RENDERBUFFER_SIZE;
 use glutin::surface::SwapInterval;
 use glutin::{
     config::{Config, ConfigTemplateBuilder},
-    context::{ContextAttributesBuilder, GlProfile, PossiblyCurrentContext},
+    context::{ContextApi, ContextAttributesBuilder, GlProfile, PossiblyCurrentContext, Version},
     display::GetGlDisplay,
     prelude::*,
     surface::{Surface, SurfaceAttributesBuilder, WindowSurface},
@@ -111,6 +111,7 @@ pub fn build_context(
         .expect("Failed to create Windows Surface");
 
     let context_attributes = ContextAttributesBuilder::new()
+        .with_context_api(ContextApi::Gles(Some(Version::new(3, 0))))
         .with_profile(GlProfile::Core)
         .build(Some(raw_window_handle));
     let context = unsafe { gl_display.create_context(&config, &context_attributes) }
```

And you can change it to ` .with_context_api(ContextApi::OpenGl(Some(Version::new(3, 1))))` for OpenGL 3.1

If OpengGL 3.1 works, then I think we could make that the default. I'm not sure about the OpenGL ES compatibility elsewhere, although it might work better on Asahi.

fredizzimo:
> For some reason in this branch vim.g.neovide_transparency also affects vim selection mode, I am trying to select lines or symbols and can't see any difference in background. Is this expected? I found this because I was using transparency=0 so my desktop wallpaper was visible as in terminal and other applications.

Yes, that's expected. `g:neovide_transaparency` is just a quick hack to turn all background colors transparent (yes, cursor highlights are technically background colors in Neovim)

The correct way to control the transparency is to use the 'blend` value of the highlight group, which is something that color schemes need to support. But unfortunately, that's something that also needs more work on the Neovim side https://github.com/neovim/neovim/issues/24472

ZerdoX-x:
> The issue is that Neovide uses OpenGL 3.3 by default, which is not supported by the Asahi driver, so it switches back to software rendering in this case, in other cases it could totally fail to launch.

When installing my "new" (haha) drivers, I set this env variables in WM:
```
env = MESA_GL_VERSION_OVERRIDE, 3.3
env = MESA_GLSL_VERSION_OVERRIDE, 330
env = MESA_GLES_VERSION_OVERRIDE, 3.1
```
I cannot say if they are standard or not. I hope some programs use it.

> If OpengGL 3.1 works, then I think we could make that the default.
I am not sure about consequences of downgrading default used OpenGL, maybe we can use these envs? Especially if they are standard (I don't really know if they are, but afair I copied them from Asahi blog). I am not sure how hard is this to implement, but I think it would be better in case it can cause any problems for the rest.

As a temporary solution until "detect OpenGL version" feature implementation I can fork neovide and we can mention this fork in install section of the docs on neovide.dev, so all people using Linux on M1/M2 machines will get acceptable performance. I can apply patch on both branches and keep them up to date, it will boost performance for all Asahi users a lot even before this gets merged.
Do you want to create an issue or we can agree on solution in place? As this is not related to #1977 at all. What do you think?

I'll try to fork, apply patch and recompile neovide and give feedback later if you want to create thread for this.

ZerdoX-x:
> `g:neovide_transaparency` is just a quick hack to turn all background colors transparent

Is there any workaround? Maybe undocumented option or hidden API?
Is this behavior going to be merged into master or it's temporary on this branch while WIP?

I don't even think of neovide controlling any other backgrounds except window's one, as I can tweak others in theme.
Currently I am using catppuccin theme with `transparent_background` set to `true`.

The only workaround I can think of and that one I am using is setting `neovide_transparency` to value > 0.4 (I would like to use 0).

> The correct way to control the transparency is to use the `blend` value of the highlight group, which is something that color schemes need to support. But unfortunately, that's something that also needs more work on the Neovim side

So if this is going to be merged into master, we need to document this and users will need wait for neovim and their color scheme support? Seems pretty complicated for me as for user when I just want to set transparency for window. And it's probably will take long time till it reaches end user.

9mm:
> PS. If you built this yourself from source, make sure you built the release version and not debug, although the debug version should be much faster than yours on most systems.

@fredizzimo can you put the entire full proper command to build?

I'm on OSX, I just got this repo to try out some of these. I've configured my entire neovim config now and now im moving onto neovide itself 😎

9mm:
I added `cargo build --release` and when I run `target/release/neovide` it simply doesn't open. is there something extra I need to do ?

fredizzimo:
@9mm, these Instructions should work https://neovide.dev/installation.html#mac-source. `cargo install` will build a release mode build.

9mm:
Ah... sorry, I should have checked there. Thank you

MultisampledNight:
(chances are it won't show up again, try with `--no-fork` in that case)

9mm:
Hmmm.. it still doesnt appear to to actually open

```
~/.cargo/bin/neovide
```

It just exits immediately with no error, and Neovide.app does not open. FWIW I tried these instructions:

```
Install the latest version of Rust. Using homebrew: brew install rustup-init

Configure rust by running rustup-init

Install CMake. Using homebrew: brew install cmake

git clone https://github.com/neovide/neovide

cd neovide

cargo install --path .

The resulting binary is to be found under ~/.cargo/bin.
```

I also tried the bundle approach and when I open it, it closes immediately

9mm:
Ah... this only happens on the `improve-render-loop` branch. If i run on main it works.

When I try to run debug it just exits:

```
➜  neovide git:(improve-render-loop) RUST_LOG=neovide=debug cargo run --release -- --log
    Finished release [optimized] target(s) in 0.11s
     Running `target/release/neovide --log`
➜  neovide git:(improve-render-loop)
```

9mm:
Ok I installed neovim HEAD, I do notice this is blazingly fast!

These were the main things I notice:

1. I'm sure this is intended behavior, but it looks like treesitter-context's "popup window" or whatever now has blur. I suppose I would need to somehow enable a constant background somewhere for styling

image

2. The bottom bar seems off, color wise:

image

how it looks normally:

image

fredizzimo:
@9mm, It looks like you have `g:neovide_transparency` set to `0`, try setting it to 1 for no transparency.

9mm:
Ahh youre right, that was a hack from the old branch to get the colors to work.

BTW performance wise its............. fucking bananas....... i get over 1000 FPS. is this accurate? LMAO

image

fredizzimo:
@9mm 
> BTW performance wise its............. fucking bananas....... i get over 1000 FPS. is this accurate? LMAO

Yes, that's accurate if you don't limit it in any way.

Normally without any parameters it should limit to the desktop refresh rate, but unfortunately that's not properly implemented on macOS yet. So sometimes it remains unlimited, other times it locks to the wrong rate.

But if you set `--novsync` then it limits to `g:neovide_refresh_rate` , which you can set to whatever you want, but it's usually set to the monitor refresh rate. Or to some really high number like `9999` for almost unlimited FPS.

And for benchmarking, or troubleshooting you can use `--noidle`, but that's not recommended for normal use, since it uses more power, and you should not see any visual differences with it.

9mm:
Awesome, that makes sense. and if I have a new MBP, which is 120HZ, i should probably set to like 120 correct? Or can i just leave it at 9999

9mm:
BTW regarding what you said about background it does show through for me. heres my settings. Not sure if that's what you were referring to, or if that helps at all

EDIT: i actually think i notice what youre saying... both background transparencies play off each other.

image

fredizzimo:
@9mm.
A refresh rate of 120 is probably good if it looks and feels smooth.

The background image was originally added to support giving colors to the title bars https://github.com/neovide/neovide/pull/1459, but that interferes with the normal transparency. So, I think we need an alternative implementation of that.

@falcucci did some initial investigation in this message and the messages above it 
https://github.com/neovide/neovide/pull/1870#issuecomment-1653883061, apparently you can use `--frame-transparent` as a workaround for now.

sidequestboy:
This is awesome and feels very fast! I wonder if it's possible to make the scrolling experience less jumpy still though when scrolling on a trackpad - right now, while scolling animates very smoothly, it still moves in chunks of 3 lines at a time (or the mouse wheel amount?)

Ideally we could scroll with pixel-precision until then maybe the viewport locks onto the "actual" top and bottom line when we let go of the trackpad, rather than line-snapping as the scroll gesture continues.

I'm curious about what your thoughts on this are and what the implementation challenges would be. Thanks for the great work!

jla2000:
I also had a look on the profiler and during this indexing it somehow jumped to very high fps numbers for a blink. Somehow a bit after I get ridiculously high fps numbers displayed there.
neovide

9mm:
This is also so smooth now... 


https://github.com/neovide/neovide/assets/1386207/306c34f6-1c37-4e1b-ba67-704267b00552

fredizzimo:
@sidequestboy.

Setting [mousescroll](https://neovim.io/doc/user/options.html#'mousescroll') to 1 should make it considerably smoother. But the drawback is that the normal mouse scrolling gets slow. Neovim does unfortunately not have separate settings for the two, so that's the best that can be done currently without hacks, but you can make some keybindings for changing it. I think that we should probably create a feature request for that. And I think the setting has to remain on the Neovim side, because you might want to change it or even the scroll wheel bindings depending on the situation.

Additionally, you most likely want to enable [smoothscroll](https://neovim.io/doc/user/options.html#'smoothscroll').

Pixel perfect scrolling on the other hand is considerably more difficult. The main issue is that we can only send requests to scroll to Neovim, but we can't rely on that, it will decide to scroll or not. And it might also scroll due to some other event happening inside Neovim.

I think it's doable by first scrolling the window, so that you scroll in one or more empty lines following your exact touch movements, just like when you scroll too far on a touchscreen. At the same time, we request the scroll of the number of lines we need from Neovim. If we get a confirmation, it's all good, otherwise some spring effect is applied that brings the view back to the last confirmed state with a delay. That ensures that unless you keep scrolling, our view and the actual view will get back in sync. This all is quite tricky, with a lot of special cases, but I don't see any reason why it would not be doable.  `mousescroll=1` will be needed when doing this, so that's the first step in any case.

That is not something I will work on immediately myself, but if someone is interested, feel free to try out that idea, or something else that you think will work. It's still on my list of things I want to work on at some point though.

fredizzimo:
@jla2000,

There could be many reasons for the slow down. If the language server slows down the computer as a whole, then using less cores and/or lower priority processes for it should fix the isssue. But I'm not sure if clangd can be configured that way. If clangd uses all the resources of the system, then there's nothing else that can be done.

If it's Neovim itself that is responsive, then I think it can be fixed at least to some degree. Right now, we are waiting for a confirmation from Neovim for each key press that is pressed. So, if Neovim is unresponsive, then the scrolling can be slow, at least if it's not processing the key repeats in the same pace as they arrive. We should probably use asynchronous events for this, since the only thing that should matter is that the order is correct. But that needs some testing, and possibly fixing before it can be done.

9mm:
@fredizzimo what does `smoothscroll` do? I enabled it and dont see any difference when scrolling. I tried googling and the docs are vague "uses screen lines" whatever that means

these are what im using to test with

```
-- scroll
vim.opt.scrolloff = 4
vim.opt.mouse = 'nv'
vim.opt.mousescroll = 'ver:2,hor:0'
vim.opt.smoothscroll = true

```

fredizzimo:
@9mm,

You can see the difference id you scroll around folded or wrapped lines. Without smoothscroll it can jump several lines instead of just one.

jla2000:
> If it's Neovim itself that is responsive, then I think it can be fixed at least to some degree. Right now, we are waiting for a confirmation from Neovim for each key press that is pressed. So, if Neovim is unresponsive, then the scrolling can be slow, at least if it's not processing the key repeats in the same pace as they arrive. We should probably use asynchronous events for this, since the only thing that should matter is that the order is correct. But that needs some testing, and possibly fixing before it can be done.

Its not really that the animation is slow, its more like the animation time is very fast and instead of smoothly animating the scrolling of the content it directly moves it in a very fast animation. It does not really feel like lag.

> **Edit:** I forgot, 144hz at 4k should not be a problem on most systems, what kind of system specifications do you have, maybe it's not using the correct graphics drivers or something.

I have tested it on a newer windows laptop and on a M2 Macbook, both seem to have the same problem.

sidequestboy:
thanks for the explanation! It does seem quite tricky indeed, but would be very satisfying to solve!

fredizzimo:
@9mm 
> This is also so smooth now...

Nice, but I wonder if the flickering on some of the lines is caused by some Neovide/Neovim bug, or if it's a problem with the plugin that does. Which plugin is that?

9mm:
I think it is the plugin because they're moving so fast

https://github.com/Eandrju/cellular-automaton.nvim

Garoth:
I just wanna say thanks for the incredible effort going on here. Open source developers are often undervalued - and all this focus on details, fps, bindings, and cross platform rendering is paying off. I check this PR regularly because I'm so excited about Neovide's development. It feels like you guys are building the *perfect* development environment - something like a quality hand made armchair chair that you can pass down to your kids

9mm:
Totally, I agree! Thank you so much @Kethku @MultisampledNight @fredizzimo  and everyone else!

I am insanely nitpicky about my environment, or perhaps meticulous not nitpicky, and neovide is just KILLER

website/docs/configuration.md
```diff
@@ -232,7 +232,27 @@ Lua:
 vim.g.neovide_scroll_animation_length = 0.3
 ```
 
-Sets how long the scroll animation takes to complete, measured in seconds.
+Sets how long the scroll animation takes to complete, measured in seconds. Note that the timing is
+not completely accurate and might depend slightly on have far you scroll, so experimenting is
+encouraged in order to tune it to your liking.
+
+#### Far scroll lines
+
```
MultisampledNight, line 240: note for future: should also include `**Unreleased yet.**`

9mm:
I have no idea if this is an issue or not because I purposefully tried to break neovide by launching it 50 times simultaneously, but I did get the following:

```
2023-09-17 13:39:50 - Neovide panicked with the message 'Failed to create OpenGL context: Error { raw_code: None, raw_os_message: None, kind: BadConfig }'. (File: src/renderer/opengl.rs; Line: 121, Column: 10)
   0: __mh_execute_header
   1: __mh_execute_header
   2: __mh_execute_header
   3: __mh_execute_header
   4: __mh_execute_header
   5: __mh_execute_header
   6: __mh_execute_header
   7: __mh_execute_header
   8: __mh_execute_header
   9: __mh_execute_header
  10: __mh_execute_header
  11: __mh_execute_header
  12: __mh_execute_header
```

The only reason I was trying to do that is _very_ rarely I will get neovide opening in a weird state where the tab headers are halfway extending outside the window.

It happens so infrequently though it's not a problem, I figured I would just post it anyway, but I cant reproduce it. When i do ill try to get a screenshot. Not posting this as a problem, only if there was some potential race condition that was avoidable, if not then I highly doubt it would cause problems

fredizzimo:
@9mm. I think you ran out of GPU memory, in that case there's nothing we could do. The failure to create the OpenGL context suggests that might be the cause, since that will need to allocate GPU resources.

But there could also be some other underlying issue, it's hard to tell from that info.

9mm:
Makes sense, I didn't think it was anything but I figure more info is better than less.

This new branch is SWEET.

9mm:
I got a screenshot of what i was referring to. overall though this happens so rarely 

image

how normally works

image

A-Lamia:
Not sure if this issue is because of this branch, it's what I'm using currently but the Normal highlight Fails to fully update the window when set through code and triggered by a hot key.

Example:
image

9mm:
@A-Lamia one thing you could try is `neovide -- --clean` and see if it occurs there. then narrow down if any plugins are causing it.

A-Lamia:
> @A-Lamia one thing you could try is `neovide -- --clean` and see if it occurs there. then narrow down if any plugins are causing it.

Works fine on normal nvim in WT and in Wez ill create a minimal config just to be sure but I don't think it's any plugin.

9mm:
BTW posting 1 more screenshot of the padding issue in case it visually makes it more clear what it might be:

image

image

fredizzimo:
@9mm. I might have a fix on the way. We are not automatically redrawing the Window each time the OS requests a redraw. 

@A-Lamia and @9mm. I'm not sure what causes that issue. But I think both your issues can be fixed by `CTRL-L`, it's not a real fix but at least some workaround.

A-Lamia:
@fredizzimo sadly it doesn't work for me :/

fredizzimo:
@A-Lamia can you create a https://github.com/neovim/neovim/wiki/Minimal-reproduction-template. So it’s easier to test and understand what you are referring to.

9mm:
@A-Lamia out of curiosity, what was it?

A-Lamia:
@9mm well i was triggering a change for the Normal highlight when entering a floating window, NormalNC is the same color as Normal in my theme but is not linked so what was happening is when i moved to an other window NormalNC was doing it's job and setting the inactive window to Normal.

fredizzimo:
@9mm. I think this should fix your problem, are you able to test? https://github.com/neovide/neovide/pull/2035

9mm:
@fredizzimo awesome, i will test that. its really hard to get it to show up though so ill play around with un-lazy loading plugins etc and see if i can get it to be more reproducable so i can make sure its actually fixed. 

Thanks!

9mm:
Hmm i actually did just notice a problem, the bottom area is cut off when in command mode

The text written below is `:w`

image

9mm:
I think i have found another bug on this branch, but Im not sure if this is something ive caused. I dont know how to pass `--clean` into this mess of args so for now it has to be with all my pllugins:

```
cargo run -- ~/Desktop/ab.js --log
```

I noticed when I kept trying to open a file directly on my desktop passed as an argument... nvim dock icon WOULD appear, but an actual neovide window would not appear. This is similar effect i observed here: https://github.com/neovide/neovide/pull/2027#issuecomment-1721738359

I looked in the log and i see this. So maybe if a swap file is present it breaks the window from opening?

If i put some other filename thats new, like `abc.txt` it DOES open properly

```
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: Found a swap file by the name "~/.local/state/nvim/swap//%Users%zesty%Desktop%ab.js.swp"
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: owned by: zesty   dated: Thu Sep 21 09:25:32 2023
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: file name: ~zesty/Desktop/ab.js
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: modified: YES
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: user name: zesty   host name: Zesty-M2.local
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: process ID: 4632
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: While opening file "/Users/zesty/Desktop/ab.js"
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: dated: Thu Sep 21 09:24:33 2023
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: (1) Another program may be editing the same file.  If this is the case,
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: be careful not to end up with two different instances of the same
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: file when making changes.  Quit, or continue with caution.
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: (2) An edit session for this file crashed.
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: If this is the case, use ":recover" or "vim -r /Users/zesty/Desktop/ab.js"
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: to recover the changes (see ":help recovery").
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: If you did this already, delete the swap file "/Users/zesty/.local/state/nvim/swap//%Users%zesty
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: %Desktop%ab.js.swp"
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: to avoid this message.
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: Swap file "~/.local/state/nvim/swap//%Users%zesty%Desktop%ab.js.swp" already exists!
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: [O]pen Read-Only, (E)dit anyway, (R)ecover, (D)elete it, (Q)uit, (A)bort:
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text:  
TRACE [neovide::channel_utils] neovide::bridge::ui_commands::UiCommand Parallel(Resize { width: 101, height: 53 })
```

9mm:
Full log

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/12689094/neovide_rCURRENT.log)

fredizzimo:
It's way tricker than I thought. The error is shown before `VimEnter` is called, and there does not seem to be a good hook point for us to show the window. Using a vim `after` script could work, but I don't see any way to inject that, other than physically copy the file. 

Well, the ui command `grid_clear` is a good indication, to show the window, but at that point we haven't received the information about `columns` or `lines` yet. But perhaps I can use the grid size of `grid 1` instead for the initial size. After that we have to ignore it though, since otherwise there will be a conflict between manually setting the size inside neovim and automatic updates due to resizing the window.

ZerdoX-x:
Got a random `neovide_backtraces.log` today, it spawned when I tried to launch neovide as usual. It was not critical at all, as it successfully launched next time. Not sure if this related to Asahi and not even sure if this related to this branch, but I am using only this branch and I could not find any references to this error in other issues. So.. I post it here. Just in case it can give you any hint on any rare edge-case bug (?)
```
2023-09-26 21:30:51 - Neovide panicked with the message 'Could not setup setting notifier for floating_blur_amount_x: Error sending request 'nvim_command''. (File: src/error_handling.rs; Line: 5, Column: 5)
   0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: <unknown>
   9: <unknown>
  10: <unknown>
  11: <unknown>
  12: <unknown>
  13: <unknown>
  14: <unknown>
  15: <unknown>
```

fredizzimo:
@ZerdoX-x, One way it could fail like that is if Neovim itself crashes or hangs. Or perhaps even when it tries to show an error message at the wrong time. I'm working on fixing some of those edge cases right now.

But it could of course be something totally different.

last-partizan:
I tried building this branch today, and fonts become larger than usual when using xwayland (i'm using `env -u WAYLAND_DISPLAY cargo r` to have window decorations with shadows).

Using `git bisect` i traced this change to commit 5dccb230e441efc6dd79b0c0845192335328b947, but i have no idea what exactly caused this.

fredizzimo:
@last-partizan, probably because of the broken scale factor detection, it has never really worked properly, and the new release might have changed it slightly, see this https://github.com/rust-windowing/winit/issues/2231 and https://docs.rs/winit/latest/winit/dpi/#how-is-the-scale-factor-calculated

> X11: Many man-hours have been spent trying to figure out how to handle DPI in X11. Winit currently uses a three-pronged approach:
>
> * Use the value in the WINIT_X11_SCALE_FACTOR environment variable, if present.
> * If not present, use the value set in Xft.dpi in Xresources.
> * Otherwise, calculate the scale factor based on the millimeter monitor dimensions provided by XRandR.
>
> If WINIT_X11_SCALE_FACTOR is set to randr, it’ll ignore the Xft.dpi field and use the XRandR scaling method. Generally speaking, you should try to configure the standard system variables to do what you want before resorting to WINIT_X11_SCALE_FACTOR.

So you should be able to override it with `WINIT_X11_SCALE_FACTOR`

glibg10b:
1. Open a file with many lines
2. `:set cursorline`
3. H
4. Hold \<C-e\>
5. Notice that the highlight spans from the current line to the top of the screen.

I would expect only the current line to be highlighted, as when using j and k in the middle of the screen.

fredizzimo:
@glibg10b. The cursorline as currently implemented in Neovim is fundamentally incompatible with the way we are doing the scrolling. I think we need to open an issue on the Neovim repository for it.

orlandosh:
"We need the macOS implementation of Implement VSync source rust-windowing/winit#2412 otherwise there will be no throttling, or the wrong refresh rate chosen, with the default settings. --vsync. --novsync with a custom g:neovide_refresh_rate works as a workaround, but not ideal."
I saw in the linked issue that it's not being worked on as there are only a couple of macOS developers. While I still don't know how to contribute to neovide codebase or rust winit, I'd like to say that if anyone decides to implement this blindly, I'm okay with building neovide locally to test it, how many times necessary.

9mm:
@orlandosh it still works if you hand specify the refresh rate though, no? I was under the impression that was only the default setting

fredizzimo:
@orlandosh, there's a known exit crash, I think your issue might be the same although it usually crashes instead of hangs.

Hopefully we will be able to fix it soon.

MOldtime:
It works great. Just now I noticed that horizontal scrolling is still the same as before.

fredizzimo:
@orlandosh, I think the hang should be fixed by this https://github.com/neovide/neovide/pull/2076

fredizzimo:
@MOldtime
> It works great. Just now I noticed that horizontal scrolling is still the same as before.

Yes, horizontal scrolling is quite tricky to do, so it's out of the scope for this PR.

I think it would be possible to use the same kinds of techniques for that. But since each line is stored as a set of characters  it gets more tricky. And even more so when considering characters that takes multiple grid position, like ligatures, symbols, or some Asian characters for example. 

I have been planning to propose a new UI protocol to Neovide, which would make this and also things like variable font support easier. But I have not yet found a design that I'm happy with even myself.

fredizzimo:
We are soon planning to merge this, but I need some help from macOS users/developers before we can do that. I don't have a Mac myself, so I can't do anything about these. 

1. I don't think we can require users to set `--novsync` and manually set the refresh rate. Ideally we should implement this for macOS https://github.com/rust-windowing/winit/issues/2412, this was started here https://github.com/neovide/neovide/issues/2073, but needs to be finished, and ideally implemented in Winit.

   If that can't be done, we need some kind of hack, perhaps forcing it off, and using the refresh rate that Winit detects. You can get the monitor from the window, and the monitor handle exposes the refresh rate https://docs.rs/winit/latest/x86_64-apple-darwin/winit/monitor/struct.MonitorHandle.html#method.refresh_rate_millihertz

   As a last resort, if nothing can be done we need to properly document the workaround
2. We need to fix the documentation for the transparency. The old hack with the background color does not work. Ideally the background color support should be completely removed and replaced by something better that actually works. I know that @falcucci had some ideas how to document in some of the earlier versions of this PR.

If needed, we can create pinned issues with these instructions, but if we can, we would like to avoid it. The release notes for the actual release can also contain some instructions, but only as a last resort.

9mm:
Hmm so Im not sure what changed but i pulled and try to do:

```
cargo install --path .
```

And it wont open (immediate crash, no log)

When I do `cargo run -- --log` it does open though. Would you mind refreshing my mind how to log in release mode?

fredizzimo:
@9mm. Hm, it should be vey close to some of the PRs you already tested.

And `cargo run —release — —log` should log in release mode.

9mm:
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/13064154/neovide_rCURRENT.log)

fredizzimo:
@9mm, do you have a neovide_backtraces.log file. 

It look like the the window creation fails, the last log is just before that inside `create_window`. You could try to add more logging there.

9mm:
No backtrace log, I will have a look later today to see if I can add some more logging, just about to eat.

fredizzimo:
@9mm and everyone else using macOS, you can work around the bug for now using `--no-fork`. On macOS it seems that we need to fork/exec, instead of just fork https://stackoverflow.com/questions/2707572/fork-within-cocoa-application. 

But it looks like the library we use don't support that https://github.com/immortal/fork, even if it was specifically made for macOS.

I'm looking into how to best fix this, or if we should just revert the changes to `maybe_detach` on macOS for now.

9mm:
Just to make sure I understand, `--no-fork` will change the fundamental way neovide "blocks" the terminal right? I am asking because I often launch neovide from command line in the current project folder, and then use that terminal for the project itself. having multiple terminals to do neovim would be a pain, so my vote is to revert on macos only

This is a random comment that's probably not related, but I have always noticed how Neovide will open multiple instances of the app every time its launched:

image

The normal macOS behavior is that there's must multiple windows that are launched from the same app instance.

I remember months ago I wrote that as an issue, but have really come to like multiple app instances a lot better, as I can entirely isolate separate instances of code and can simply do `command+Q` to quit that instance of Neovide without ruining my other sessions.

fredizzimo:
@9mm, Yes, I will revert it, tomorrow I think. And it seems like that’s the best solution we have.

For the multiple icons, that’s exactly what I was hoping that the new way of forking would solve. Previously, and also after I revert, we launch Neovide twice, the first instance quits almost immediately, to be able to detach from the terminal.

9mm:
Yes, that makes sense. I did also notice that when Neovide launches on OSX it "boots twice", once where the app animates for about 50ms then "restarts", so that fits what you're saying with the first instance exiting immediately.

So it sounds like the multiple-app instances are actually just a side effect of the workaround to detach from the terminal.

Oh well, even if it goes away it isn't end of world. But how would you handle that then, would you just have multiple windows within the same app instance?

fredizzimo:
* The problematic fork behaviour has been reverted so macOS should work again.
* Winit has also been updated, so IME should work on Windows

9mm:
OSX workin again 🙏🏼 Let me know any other ways I can assist, have been insanely busy but Neovide is just so freakin good I like to help

fredizzimo:
I guess we are just waiting for a plan to deal with the 2 remaining macOS issues (transparency and vsync) before we can merge. Since I don't own a mac, it's hard for me to judge how serious those are if we are going to get flooded with issues if we merge now or not.

There's also no approving review, but it was reviewed in the earlier iterations and new PRs that has been merged into this, and I think I have taken all feedback into account. Still it's probably good to at least do a quick pass.

MOldtime:
When I used the version of this branch, I found that when `neovide_transparency` was set, there was a problem with the transparency using the [noice](https://github.com/folke/noice.nvim) plug-in. I tried the official version of neovide and there was no such problem.
**formal edition**：
图片
**This branch**：
图片

ZerdoX-x:
> When I used the version of this branch, I found that when `neovide_transparency` was set, there was a problem with the transparency using the [noice](https://github.com/folke/noice.nvim) plug-in.

Please search through the thread.

https://github.com/neovide/neovide/pull/1977#issuecomment-1703775821

fredizzimo:
I just merged in the latest main, and we will merge this PR tomorrow, if no new unexpected issues are found.

Note, that we still haven't resolved the macOS issues, but as soon as this is merged we will create a sticky issue giving an overview of the change and the remaining problem.  I will also create dedicated issues for each of the remaining problems, so it's easier to find all the relevant information and come up with proper solutions.

ZerdoX-x:
Could you please create a comment here where you'll link remaining issues? Would love to see transparency issue being fixed :heart:

fredizzimo:
Thank you all for the amazing testing work! 

I'm very happy to finally see this merged, it's been a rather big project to get all the pieces into place, and still not everything is fixed, see the following issues
* https://github.com/neovide/neovide/issues/2093
* https://github.com/neovide/neovide/issues/2094
* https://github.com/neovide/neovide/issues/2095

Any new discussion should be done in those issues, or if you notice new bugs, open regular issues like normal

papricasix:
Thank you to everyone involved in you great work on Neovide. You are awesome!

Issue #1877, opened 2023-05-27
Add option to map only the `left option` key to meta on macOS


esamattis:
**Is your feature request related to a problem? Please describe.**

At the moment I'm unable to use Neovide because I cannot get the meta key working with being able type certain special characters. I use a finnish keyboard layout where for example `Option-8` is used to type `[`, `Option-shift-8` for `{` etc. 

To get the `<M-a>` style keybindings working I need to set `neovide_input_macos_alt_is_meta` to true but this makes me unable to type the above special characters because the Option keys now send the meta key to Neovide causing it to just exit from the insert mode.

**Describe the solution you'd like**

With iTerm2 and terminal Neovim I was able to solve this sending the meta key only from the Left Option key so I could use Right Option to type the special characters.

I'd like to be able to this in Neovide too.

**Describe alternatives you've considered**

I'm open to other workarounds / solutions too.

Ideally I'd like to be able tell Neovide/Neovim to bind actions directly to the "left option key" and not to the "meta key".

Or maybe allow configuring the macOs `fn` key as the meta key? I could just swap it with the left option using Karabiner Elements.

**Additional context**

iTerm2 options screen

<img width="335" alt="image" src="https://github.com/neovide/neovide/assets/225712/84acc407-ca14-4e84-beaa-cdfd47d43bf3">

---------------

Thank you for creating such a awesome tool! I'd really like be able make work for me :)

esamattis:
Figured out a workaround: Neovide / Neovim supports binding keys to `<D-S-[key]>` which means `command-shift-[key]` so I just configured Karabiner Elements to emit `command-shift` from the Left Option key when Neovide is the "frontmost application" and used this helper to bind the keys

```lua
function meta_key(key)
    if vim.g.neovide then
        return "<S-D-" .. key .. ">"
    else
        return "<M-" .. key .. ">"
    end
end
```

so the Left Option key works the same in terminal Neovim and Neovide.

fredizzimo:
We need to implement winit `OptionAsAlt`  https://github.com/rust-windowing/winit/pull/2887, instead of our custom `neovide_input_macos_alt_is_meta`, just call `window.set_option_as_alt`

With that we should also be able to remove all the customization for MacOS in https://github.com/neovide/neovide/blob/main/src/window/keyboard_manager.rs

JamesWidman:
[...]

Issue #1866, opened 2023-05-16
<A-n> fails to interprets in Macos. I set `neovide_input_macos_alt_is_meta `


OliverChao:
In Macos, I set `vim.g.neovide_input_macos_alt_is_meta = true`,  keys works fine except <A-n>.

I don't know if someone can reproduce it.  But other keys like <A-m> and <A-p> are ok. 

**To Reproduce**
<A-n> can not be matched.

**Expected behavior**

**Screenshots**

https://github.com/neovide/neovide/assets/39361033/19c3ec45-f4a7-46ab-8e6e-50952b6bb603

**Desktop (please complete the following information):**
 - OS: 13.3.1 (a) (22E772610a)
 - Neovide Version 0.10.4
 - Neovim Version 0.9.0

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

**Additional context**

richmeta:
I also have this too. 

In neovide:

`:map <a-n>`
No mapping found

`:map <a-n> :echo "hello"<cr>`
Pressing `<a-n>` - nothing happens

But 
`:map <a-m> :echo "hello"<cr>`
and `<a-m>` works as expected

I even tried with `<m-n>` which doesn't work either. 

neovide 0.10.4
NVIM v0.8.2

fredizzimo:
Could you try the latest master version, the keyboard handling was changed a lot here https://github.com/neovide/neovide/pull/1789

OliverChao:
> Could you try the latest master version, the keyboard handling was changed a lot here #1789

Thanks for your reply. However, there also some errors for converting mappings.
I write `<C-v>` then `<A-h>` and it prints `<M-`>`.
Here are some conversions:
```
<A-n > --> <M-n>  right
<A-m> --> <M-µ>
<A-i> --> <M-i>  right
<A-p> --> <M-π>
<A-h> --> <M-˙>
<A-l> --> <M-¬>
```

fredizzimo:
This should be fixed by #1899, it would be great if you can test that.

OliverChao:
Thanks for your work. Sorry to say that it some errors exists. 
The <Alt> mappings #1789 works, #1899 fails, and vice versa.

`<A-n>` and `<A-i>` fail to map `<M-n>` and `<M-i>`.
`<A-m>` maps `<M-m>` successfully.

I think that is because some keymaps(like `<Option-n>` and `<Option-i>`) have special functions in MacOS input method.

[...]

PR #1979, opened 2023-08-15
macOS configuration options for using only the left or right alt key …

apettel:
…as meta

- Feature

I'm using the [EurKEY Keyboard Layout](https://eurkey.steffen.bruentjen.eu/?lang=en). To access certain characters (e.g. ä, ö, ü, ß) I need the alt key. But I also need the alt key as meta key for certain terminal applications. 
I configured the right alt key in my terminal program to switch the characters in the EurKEY-Layout and the left alt key is used as meta key. I created a patch for neovide to allow the separate configuration of the right and the left alt key (branch: macos-left-right-alt-meta).

- No

### DIFF
--- a/src/window/keyboard_manager.rs
+++ b/src/window/keyboard_manager.rs
@@ -79,7 +79,7 @@ impl KeyboardManager {
         // the whole modifier state. Otherwise send the resulting character with "S-" and "M-"
         // removed.
         #[cfg(target_os = "macos")]
-        if self.modifiers.state().alt_key() && use_alt() {
+        if modifier_meta_pressed(&self.modifiers) {
             return key_event
                 .key_without_modifiers()
                 .to_text()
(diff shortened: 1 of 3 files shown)


fredizzimo:
Thanks, but this should be implemented using the native Winit `OptionAsAlt`  https://github.com/rust-windowing/winit/pull/2887

The only reason why it wasn't implemented like that before is that it came when this was almost finished https://github.com/neovide/neovide/pull/1899. And because I don't have a mac to test on, I did not want to risk anything.

If you implement it using that, I think all, at least almost all the macOS specific stuff in `keyboard_manager.rs` can be removed.

fredizzimo:
This is the relevant issue for this PR https://github.com/neovide/neovide/issues/1877

And it might also fix this issue (that needs to be tested)
https://github.com/neovide/neovide/issues/1866

uloco:
omg i need this so badly, typing on a german layout makes using neovide impossible right now without this. hope this gets merged soon. thank you!

fredizzimo (changes requested):
I'm making it more explicit that this won't be approved until it's re-implemented with `OptionAsAlt ` as described in the first message.

Note, that I'm very happy to see this being done and worked on and I don't see anything specifically wrong with this implementation. It's just that we don't want to maintain and duplicate work that's already done in Winit, and we also want to avoid special cases for different operating systems where possible.

fredizzimo:
An alternative implementation was done here
* https://github.com/neovide/neovide/pull/2486

Issue #620, opened 2021-05-03
How to get neovim's default buffer?


bonecountysheriff:
Neovide seems to open a scratch file by default instead of the help screen in vim/neovim. How do you change this?

Kethku:
Thats a good question. I suspect this is a bug in the init logic.

Kethku:
Shouldn't be hard to fix. THe relevant code is in the bridge mod.rs file

ErikMayrhofer:
I would like to offer my help with this issue.

At first glance, the issue seems to be that the screen is redrawn before anything is input by the user (e.g because of a resize or something).
nvim's `:h :into` tells us that:
```
When Vim starts without a file name, an introductory message is displayed (for
those who don't know what Vim is).  It is removed as soon as the display is
redrawn in any way.  To see the message again, use the ":intro" command (if
there is not enough room, you will see only part of it).
```

I will now investigate if my suspicion is correct and if startup-redraws can be avoided.

Kethku:
Wow thanks so much for the in depth investigation. This is super useful. As a side note: did you find logging the rpc calls to slow down the system? I wonder if we should change the --log function to do this all the time. For bonus points we could also have it log every command between the different neovide threads. That would turn it into a truly useful log rather than the hit or miss experience it is today.

As for the issue at hand, I suspect that the two resizes happen because we resize the screen when we get the font from neovim. I don't remember for sure, but I think its possible that the window tries to preserve the current size as close as possible when the font changes. So we start with the default font at a given size, and then on receiving the configured font, resize again to match the new font dimensions.

ErikMayrhofer:
I did not notice any slowdowns while logging. However this isn't pretty representative, because first I didn't do much more than starting it up, look if the intro is shown and then `:q`uitting instantly and second I just used a `println!` for the sake of brevity and third my system's pretty beefy. 

I simply inserted a `println!("Method CALL: {} {:?}", method, args);` as the first line in [`nvim-rs/src/neovim.rs:Neovim.call`](https://github.com/Kethku/nvim-rs/blob/109feea9e345fcbb2674384b0d6914ba4b8fc61e/src/neovim.rs#L125).

I did however notice that there weren't that many RPC calls at all so maybe a simple `trace!` call wouldn't be a problem. Maybe that call would even get removed in release builds. I can draft up a PR to your `nvim-rs` repo if you deem it useful.

---

[...]

PR #1980, opened 2023-08-15
feat: Show intro message

fredizzimo:
## What kind of change does this PR introduce

This emulates the standard Neovim `:intro` message at startup. It's shown if the current buffer is empty and not backed by a file. The message is always centered and does not close when the window is resized, unlike the standard message. Another difference is that this message closes directly on any mode chance except changes from/to cmd mode, while the standard message stays visible if you go into insert mode, until you type something.  

NOTE: That this is not a maintainable solution, so the goal is to get this or something similar implemented in Neovim itself, but for now I think this is good enough for us.

Fixes: https://github.com/neovide/neovide/issues/620

For the corresponding Neovim issue see https://github.com/neovim/neovim/issues/24705

- No

### DIFF
--- a/lua/intro.lua
+++ b/lua/intro.lua
@@ -43,6 +43,7 @@ local function show_intro()
 
     -- Show the sponsor and register message one out of four times, the Uganda
     -- message two out of four times.
+    math.randomseed(os.time())
     local help_type = math.random(4)
     if help_type <= 2 then
         vim.list_extend(lines, help_messages[help_type])
--- a/lua/intro.lua
+++ b/lua/intro.lua
@@ -1,0 +1,149 @@
+local function show_intro()
+    -- Don't show the intro if a buffer is loaded
+    if vim.api.nvim_buf_line_count(0) > 1 or vim.api.nvim_buf_get_name(0) ~= "" then
+        return
+    end
+
+    -- Create a new unlisted buffer
+    local buf = vim.api.nvim_create_buf(false, true)
+    local version_string = vim.api.nvim_exec2("version", { output = true }).output
+    -- we only need the first line of the version
+    version_string = version_string:sub(2, version_string:find("\n", 2) - 1)
+    local version = vim.version()
+
+    local lines = {
+        version_string,
+        "",
+        "Nvim is open source and freely distributable",
+        "https://neovim.io/#chat",
+        "",
+        "type  :help nvim<Enter>       if you are new! ",
+        "type  :checkhealth<Enter>     to optimize Nvim",
+        "type  :q<Enter>               to exit         ",
+        "type  :help<Enter>            for help        ",
+        "",
+        string.format("type  :help news<Enter> to see changes in v%d.%d", version.major, version.minor),
+        "",
+    }
+
+    local help_messages = {
+        {
+            "Sponsor Vim development!",
+            "type  :help sponsor<Enter>    for information ",
+        },
+        {
+            "Become a registered Vim user!",
+            "type  :help register<Enter>   for information ",
+        },
+        {
+            "Help poor children in Uganda!",
+            "type  :help iccf<Enter>       for information ",
+        },
+    }
+
+    -- Show the sponsor and register message one out of four times, the Uganda
+    -- message two out of four times.
+    local help_type = math.random(4)
+    if help_type <= 2 then
+        vim.list_extend(lines, help_messages[help_type])
+    else
+        vim.list_extend(lines, help_messages[3])
+    end
+
+    local longest_line = 0
+    for i, line in ipairs(lines) do
+        if line:len() > 0 then
+            -- translate
+            line = vim.fn.gettext(line)
+        end
+        longest_line = math.max(longest_line, line:len())
+        lines[i] = line
+    end
+
+    -- add padding to the lines
+    for i, line in ipairs(lines) do
+        local padding = math.floor((longest_line - line:len()) / 2) + 1
+        lines[i] = string.rep(" ", padding) .. line
+    end
+
+    -- add highlight to the special keys
+    vim.api.nvim_buf_set_lines(buf, 0, -1, false, lines)
+    for i, line in ipairs(lines) do
+        local s, e = line:find("<.*>")
+        if s ~= nil then
+            vim.api.nvim_buf_add_highlight(buf, -1, "SpecialKey", i - 1, s - 1, e)
+        end
+    end
+
+    local width = longest_line + 2
+    local height = #lines
+
+    local function get_config()
+        return {
+            relative = "editor",
+            width = width,
+            height = height,
+            col = math.max(0, math.ceil((vim.o.columns - width) / 2)),
+            row = math.max(0, math.ceil((vim.o.lines - height) / 2)),
+            style = "minimal",
+            border = "none"
+        }
+    end
+
+    local win = vim.api.nvim_open_win(buf, false, get_config())
+    vim.api.nvim_win_set_option(win, "winhighlight", "NormalFloat:Normal")
+
+    local win_autogroup = vim.api.nvim_create_augroup("neovide_intro_win", { clear = true })
+
+    local function close()
+        vim.api.nvim_del_augroup_by_id(win_autogroup)
+        if vim.api.nvim_buf_is_valid(buf) then
+            vim.api.nvim_buf_delete(buf, { force = true })
+        end
+        if vim.api.nvim_win_is_valid(win) then
+            vim.api.nvim_win_close(win, true)
+        end
+    end
+
+    vim.api.nvim_create_autocmd({ "VimResized" }, {
+        pattern = "*",
+        group = win_autogroup,
+        callback = function()
+            vim.api.nvim_win_set_config(win, get_config())
+        end
+    })
+
+    local first_cursor_move = false
+
+    vim.api.nvim_create_autocmd(
+        {
+            "BufEnter",
+            "ModeChanged",
+            "CursorMoved", "CursorMovedI",
+            "TextChanged", "TextChangedI", "TextChangedP", "TextChangedT"
+        },
+        {
+            pattern = "*",
+            group = win_autogroup,
+            callback = function(param)
+                -- Allow changing from and to command mode
+                if param.event == "ModeChanged" and param.match:find("c") ~= nil then
+                    return
+                end
+                -- The first cursor move is ignored, since it happens after the message is shown
+                if not first_cursor_move and param.event == "CursorMoved" then
+                    first_cursor_move = true
+                    return
+                end
+                close()
+            end
+        })
+end
+
+
+local autogroup = vim.api.nvim_create_augroup("neovide_intro", { clear = true })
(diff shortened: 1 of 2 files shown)


fredizzimo:
I have only done some basic testing myself, so this needs a bit more testing before merging. We need to test at least for compatibility with plugin managers like lazy vim. It should not show there, since they have their own startup screen.

src/bridge/setup.rs
```diff
@@ -30,6 +30,8 @@ const REGISTER_CLIPBOARD_PROVIDER_LUA: &str = r"
         cache_enabled = 0
     }";
 
+const INTRO_MESSAGE: &str = include_str!("../../lua/intro.lua");
```
Kethku, line 33: Maybe call this INTRO_MESSAGE_LUA. As is the variable could be construed as the message itself

fredizzimo: This is fixed now

rosingrind:
> this needs a bit more testing before merging

## Startup message

I can confirm that it's showing an intro message on my system with `--clean`:

Screenshot 2023-08-16 at 12 33 24 AM

And with my config (`lazy.nvim`, `lualine`, etc.):

Screenshot 2023-08-16 at 12 34 00 AM

## Invoking `:intro`

But, behavior of invoking `:intro` command haven't changed, neither on `--clean`:

Screenshot 2023-08-16 at 12 38 13 AM

Nor on my config:

Screenshot 2023-08-16 at 12 38 32 AM

It's an invisible buffer, and the cursor still moves to `bottom-right` of this buffer (regarding of having a starup message on screen or not, it's the same):

Screenshot 2023-08-16 at 12 41 48 AM

fredizzimo:
> But, behavior of invoking :intro command haven't changed,

Yes, I did not want to override that command and I'm not sure if it's even possible. It works without multi-grid, but when using multigrid it gets rendered behind the main window.

rosingrind:
Oh, I have multigrid enabled indeed. Then I think the goals are achieved 👍

By the way, startup intro obviously closes when you enter insert mode, so I think it's good to go. I didn't mention that, but it may be important :)

fredizzimo:
> By the way, startup intro obviously closes when you enter insert mode, so I think it's good to go. I didn't mention that, but it may be important :)

That's actually slightly different from the original behaviour, but I think it's better. With standard terminal Neovim, entering insert mode does not close the message, it only closes when you start to type.

The message here also closes when switching to other modes. on movement, on buffer switching, on modifications and so on. Basically, anything else than going into the command mode, so you can see the message when typing `:help iccf`, once you press enter it disappears and the help is shown instead.

One thing that still needs testing is compatibility with LazyVim for example, which has its own startup screen.

rosingrind:
> With standard terminal Neovim, entering insert mode does not close the message, it only closes when you start to type

You're right, I forgot that it's behaving like that in a terminal nvim :)

> compatibility with LazyVim

That's how it is on a fresh start with pre-install (intro message haven't showed up):

Screenshot 2023-08-16 at 1 25 14 AM

This is a restarted instance (same):

Screenshot 2023-08-16 at 1 26 27 AM

Invoking `:intro` changes absolutely nothing (there's no changes in a cursor position, or anything else)

rosingrind:
It could be nice if someone from a linked issue may verify that for integrity on their machines by the way, but years have passed. So, if you need anything else - feel free to ask/post a checklist

fredizzimo:
I have tested passing both directories and files on the commandline, and everything seems to be consistent with the default intro screen, it's not shown when you do that.

It even works correctly with `neo-tree` which is opened instead of the built-in `netrw` when passing directories.

lua/intro.lua
```diff
@@ -0,0 +1,149 @@
...
+
+    local first_cursor_move = false
+
+    vim.api.nvim_create_autocmd(
+        {
+            "BufEnter",
+            "ModeChanged",
+            "CursorMoved", "CursorMovedI",
+            "TextChanged", "TextChangedI", "TextChangedP", "TextChangedT"
+        },
```
MultisampledNight, line 124: The original intro in TUI Neovim vanishes as soon as any character is typed in command mode. The intro here atm however does not, and persists even after a command has been run (excluding buffer-interacting stuff like `:normal`).

```suggestion
        {
            "BufEnter",
            "ModeChanged",
            "CursorMoved", "CursorMovedI",
            "TextChanged", "TextChangedI", "TextChangedP", "TextChangedT", "CmdlineChanged",
        },
```

lua/intro.lua
```diff
@@ -0,0 +1,149 @@
...
+        {
+            "Help poor children in Uganda!",
+            "type  :help iccf<Enter>       for information ",
+        },
+    }
+
+    -- Show the sponsor and register message one out of four times, the Uganda
+    -- message two out of four times.
+    local help_type = math.random(4)
```
MultisampledNight, line 46: I've never spotted any other message in `help_messages` than the Uganda one. Possibly that stems from `math.random` in Neovim's lua inheriting the C semantics (`:help math.random`):

  ```help
  math.random([{m} [, {n}]])                                       *math.random()*
          This function is an interface to the simple pseudo-random generator
          function `rand` provided by ANSI C. (No guarantees can be given for
          its statistical properties.)
  
          When called without arguments, returns a pseudo-random real number in
          the range `[0,1)`. When called with a number {m}, `math.random`
          returns a pseudo-random integer in the range `[1, m]`. When called
          with two numbers {m} and {n}, `math.random` returns a pseudo-random
          integer in the range `[m, n]`.
  ```

  Which means that `math.random`, just like C's `rand`, returns the same sequence on startup unless seeded (see [cppreference on `rand`][cppreference]. And that happens to be `4` at least in my testings. [cppreference] also lists a usable workaround: seeding the RNG by time before sampling.

[cppreference]: https://en.cppreference.com/w/c/numeric/random/rand

```suggestion
    -- math.random returns the same on every startup if not seeded explicitly
    math.randomseed(os.time())
    local help_type = math.random(4)
```

lua/intro.lua
fredizzimo: I left it like that so you don’t have to remember the full help command to type. I think it makes more sense that way.

lua/intro.lua
fredizzimo: Yes, I was doing almost all the testing without closing Neovide, so I did not notice. I was also confused with the documentation of the built in rand which says an internal seed is used.

I will fix

fredizzimo: Ok fixed.

I would have hoped that Neovim set the random seed itself though, since now each plugin has to initialize it separately and possibly make the random numbers less random in the progress.

fredizzimo: I did a few test runs and this is what I got
```
uuuuuuuuuuuuuuuuuuuuuuuu
rrrrrrrrrrrrr
ssssssssssssssss
```
So it seems to be pretty decent.

MultisampledNight: I'm not exactly sure how desirable non-deterministic random numbers inside an editor really are anyway. Thank you either way!

MultisampledNight (approved):

9mm:
Is there something special to getting `shortmess = 'I'` to work?

```
vim.opt.shortmess = 'I'
```

It doesnt appear to skip it

MultisampledNight:
At the moment it's ignored. Chances are the logic here will be reworked sooner or later for listening to Neovim itself. A quick workaround for Neovide (and not the TUI) would be this in `init.vim` though, since entering insert mode removes the message in Neovide:

```vim
startinsert
stopinsert
```

9mm:
Is there a way to get it to work with init.lua instead of init.vim? this doesnt appear to do anything:

```
vim.cmd('startinsert')
vim.cmd('stopinsert')
```

fredizzimo:
Yes, this was an oversight, I did not know about the `shortmes` option. But I think the workaround does not work because it happens too early. So you probably need something like this, but I haven't tested it
```lua
local autogroup = vim.api.nvim_create_augroup("disable_intro", { clear = true })
vim.api.nvim_create_autocmd({ "VimEnter" }, {
    pattern = "*",
    group = autogroup,
    callback = function()
        vim.cmd('startinsert')
        vim.cmd('stopinsert')
    end
})
```

9mm:
Hmm yeah that doesnt appear to work for me either, I'll have to play around with it later.

asharkhan3101:
`vim.opt.shortmess:append({I = true})` seems to be getting ignored.
I also don't know how to get rid of this message since its not going away.

I am downgrading but thought to post it here.
Screenshot 2023-08-23 at 10 22 41 PM

fredizzimo:
@casrass 
I'm looking into it, but meanwhile you can use https://github.com/neovide/neovide/pull/1985, which let you disable the message with the `shortmess+=I` option.

fredizzimo:
I don't know why that's the cause, but I found out that using packer to lazy load `dashboard-nvim` causes the `VimEnter` autocommand to fire twice.

A workaround is to remve `event = 'VimEnter'` and change
```
use {
  'glepnir/dashboard-nvim',
  event = 'VimEnter',
  config = function()
    require('dashboard').setup {
      -- config
    }
  end,
  requires = {'nvim-tree/nvim-web-devicons'}
}
``` 
to 
```
use {
  'glepnir/dashboard-nvim',
  config = function()
    require('dashboard').setup {
      -- config
    }
  end,
  requires = {'nvim-tree/nvim-web-devicons'}
}
```

I don't use lazy.nvim, but the cause is probably the same with that as well.

I can fix it in `Neovide` by ignoring the second `VimEnter`, but I don't know why the  package manager causes the off-spec behaviour of firing the event twice in the first place.

fredizzimo:
Hm. packer does indeed seem to fire the event again
https://github.com/wbthomason/packer.nvim/blob/ea0cc3c59f67c440c5ff0bbe4fb9420f4350b9a3/lua/packer/load.lua#L118

I still have no idea why, but there's probably a good reason.

I also don't have any idea why `startup-nvim` recommends to lazy load on `VimEnter`, rather than just loading it normally during the startup, to me there doesn't seem to be any benefits at all, since both are done before anything is shown.

Anyway, I will fix the issue by only listening for `VimEnter` once

9mm:
I don't know if it matters but I don't use any dashboard at all, nothing fancy, and I still tried multiple ways to get it working without much luck. Heres the current way i have commented out that also doesnt work:

```
local autogroup = vim.api.nvim_create_augroup("disable_intro", { clear = true })
vim.api.nvim_create_autocmd({ "VimEnter" }, {
    pattern = "*",
    group = autogroup,
    callback = function()
        vim.cmd('startinsert')
        vim.cmd('stopinsert')
    end
})
```

fredizzimo:
@9mm If you are talking abou tthe issue of not being able to close it at all, then you probably have some other lazy loaded plug that causes `VimEnter` to be called multiple times. This is fixed in https://github.com/neovide/neovide/pull/1985.

That PR also fixes the `shortmess+=I` option not working, so you should be able to completely turn it off.

9mm:
It closes OK, I was just also trying to find a way to not have it open to start with. Im considering switching to master soon though as all the good fixes I want are there :D

Issue #1996, opened 2023-08-27
What could be the reason that `<S-Enter>` is not being received by neovim as `<S-Enter>`


fredizzimo:
### Discussed in https://github.com/neovide/neovide/discussions/1994

<div type='discussions-op-text'>

<sup>Originally posted by **IndianBoy42** August 27, 2023</sup>
`<C-v><S-Enter>` is showing `^M`. The same as `<C-v><Enter>` </div>

This is a bug, I have confirmed that `shift+enter` sends just enter.

fredizzimo:
#1997 Should fix this.

PR #1997, opened 2023-08-27
fix: Add missing Enter special key

fredizzimo:
This fixes the shift+enter key combination, enter was missing from the list of special keys. 

Additionally, the keys are now sorted alphabetically.

Fixes https://github.com/neovide/neovide/issues/1996

- No


IndianBoy42:
Can confirm this works on my end, thanks!

MultisampledNight:
Thank you! Only tested on Wayland, but change looks innocent enough to me.

PR #2009, opened 2023-09-02
chore: Update to winit 0.29.1-beta

fredizzimo:

This updates to Winit 0.29.1-beta, which had a few breaking changes, but they were easy to work around.

I also fixed our exit code, so that it exits cleanly and waits for Neovim to exit instead of just terminating directly. This might fix some issues and cause some others.  But at least with my testing, it properly terminates when Neovim is killed unexpectedly, and of course when closing Neovide the normal ways.

This also fixes https://github.com/neovide/neovide/security/dependabot/14

- Yes there was a regression in Winit, forcing X11 by undefining `WAYLAND_DISPLAY` does not work. It will be fixed here https://github.com/rust-windowing/winit/pull/3058

### DIFF
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -8,10 +8,14 @@ mod window_wrapper;
 #[cfg(target_os = "macos")]
 mod draw_background;
 
+use std::process::ExitCode;
+
 #[cfg(target_os = "linux")]
 use std::env;
 
 use winit::{
+    error::EventLoopError,
+    event::Event,
     event_loop::EventLoop,
     window::{Icon, WindowBuilder},
 };
@@ -52,7 +56,7 @@ pub enum WindowCommand {
     ListAvailableFonts,
 }
 
-pub fn create_window() {
+pub fn create_window() -> ExitCode {
     let icon = {
         let icon = load_from_memory(ICON).expect("Failed to parse icon data");
         let (width, height) = icon.dimensions();
@@ -143,7 +147,16 @@ pub fn create_window() {
 
     let mut update_loop = UpdateLoop::new();
 
-    let _ = event_loop.run(move |e, _window_target, control_flow| {
-        *control_flow = update_loop.step(&mut window_wrapper, e);
-    });
+    match event_loop.run(move |e, _window_target, control_flow| {
+        if e != Event::LoopExiting {
+            *control_flow = update_loop.step(&mut window_wrapper, e);
+        }
+    }) {
+        Ok(()) => 0,
+        // All error codes have to be u8, so just do a direct cast with wrap around, even if the value is negative,
+        // that's how it's normally done
+        Err(EventLoopError::ExitFailure(code)) => code as u8,
+        _ => 1,
+    }
+    .into()
 }
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -63,7 +63,7 @@ pub fn create_window() {
         Icon::from_rgba(rgba, width, height).expect("Failed to create icon object")
     };
 
-    let event_loop = EventLoop::new();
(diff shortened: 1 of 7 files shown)


src/window/mod.rs
```diff
@@ -52,7 +56,7 @@ pub enum WindowCommand {
     ListAvailableFonts,
 }
 
-pub fn create_window() {
+pub fn create_window() -> ExitCode {
```
MultisampledNight, line 59: Could it be cleaner to instead return an error enum in `Result<Infallible, ...>`?

src/window/mod.rs
```diff
@@ -143,7 +147,16 @@ pub fn create_window() {
...
-        *control_flow = update_loop.step(&mut window_wrapper, e);
-    });
+    match event_loop.run(move |e, _window_target, control_flow| {
+        if e != Event::LoopExiting {
+            *control_flow = update_loop.step(&mut window_wrapper, e);
+        }
+    }) {
+        Ok(()) => 0,
+        // All error codes have to be u8, so just do a direct cast with wrap around, even if the value is negative,
+        // that's how it's normally done
```
MultisampledNight, line 157: Why do all error codes have to be `u8`? Is consistency with errors on Unix vs. Windows from concern if internals error out?

src/window/window_wrapper.rs
```diff
@@ -30,9 +30,11 @@ const MIN_WINDOW_WIDTH: u64 = 20;
 const MIN_WINDOW_HEIGHT: u64 = 6;
 
 pub struct WinitWindowWrapper {
-    pub windowed_context: WindowedContext,
-    skia_renderer: SkiaRenderer,
+    // Don't rearrange this, unless you have a good reason to do so
+    // The destruction order has to be correct
     renderer: Renderer,
+    skia_renderer: SkiaRenderer,
+    pub windowed_context: WindowedContext,
```
MultisampledNight, line 37: Does "has to be correct" express an implicit unsafe invariant to uphold? Or just a logical error if changed without much thought?

src/window/mod.rs
fredizzimo: I don't think so, the function can return both Ok and an Error, we could move the error code translation up to main, and perhaps extract it to a function though.

The `create_window` function should perhaps be named to `create_window_and_run` to indicate that it does more than just creates the window. Or perhaps split into two separate functions, one that creates the event loop (which owns everything it needs) and one that runs it.

src/window/mod.rs
fredizzimo: Since Unix really only support U8, rust made the decision to only support that. https://doc.rust-lang.org/stable/std/process/struct.ExitCode.html

We could change all our internal uses to u8, but unfortunately winit uses integer exit codes, and the exit codes we receive from Neovim are also integers. So the most straightforward approach is to use int32 everywhere and convert it to u8 just before returning it.

src/window/window_wrapper.rs
fredizzimo: It hangs if you delete the opengl context before Skia, The rest are not that important, but this is the order I have tested with and verified at least quickly to work.

src/window/mod.rs
MultisampledNight: The docs do mention https://doc.rust-lang.org/std/process/fn.exit.html as a usable alternative, we could call that as the last line in main.

fredizzimo: The problem is that it does not call any destructors, so we will not properly wait for Neovim to terminate for example. Of course, it's possible to wrap the function to ensure that, but I prefer a totally clean exit if possible. 

https://doc.rust-lang.org/std/process/fn.exit.html

fredizzimo:
Replaced by #2035

Issue #2017, opened 2023-09-09
can't map <c-s-v>


Mr-LLLLL:
**Describe the bug**
can't map <ctrl+shift+v> in any mode, the behavior is normar only nvim.

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04
 - Neovide Version 0.11.2
 - Neovim Version v0.9.2

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

**Additional context**

MultisampledNight:
Hi, thanks for reporting! Could you try `<C-V>` (note the capitalization of the `V`)?

Mr-LLLLL:
> Hi, thanks for reporting! Could you try `<C-V>` (note the capitalization of the `V`)?

i try it;
if i mapped `<c-V>`, and i found `<c-v>` is worked like `<c-V>` in neovide. but it's goods in nvim.

by a way:
i test in 0.10.4, `<c-s-v>` is work goods like nvim.
0.11.x (all version) will have the bug above.

fredizzimo:
`<C-V>` and `<C-v>` are actually the same in Neovim due to legacy reason, and that's not something we can do about. The actual mapping actual overwrites the previous one if you try to do that.

From https://neovim.io/doc/user/intro.html (the notation section)

> CTRL-{char} {char} typed as a control character; that is, typing {char}
while holding the CTRL key down. The case of {char} is
ignored; thus CTRL-A and CTRL-a are equivalent. But in
some terminals and environments, using the SHIFT key will
produce a distinct code (e.g. CTRL-SHIFT-a); in these
environments using the SHIFT key will not trigger commands
such as CTRL-A

But Neovide produces `<C-S-V>` as a different mapping, so you should be able to use that. `<C-S-v>` might also work, but I'm not sure. 

I'm just testing this myself now, on Windows, and yes it's broken for some reason. I'm pretty sure it used to work.

Mr-LLLLL:
> -a are equivalent. But in



> `<C-V>` and `<C-v>` are actually the same in Neovim due to legacy reason, and that's not something we can do about. The actual mapping actual overwrites the previous one if you try to do that.
> 
> From https://neovim.io/doc/user/intro.html (the notation section)
> 
> > CTRL-{char} {char} typed as a control character; that is, typing {char}
> > while holding the CTRL key down. The case of {char} is
> > ignored; thus CTRL-A and CTRL-a are equivalent. But in
[...]

Issue #1237, opened 2022-03-04
CMD + SHIFT  works as CMD only


goldcoders:
Ive enable CMD in neovide...

it works but i cant bind anything with SHIFT.

CTRL+SHIFT Works

also If i try to bind <S-D-anykey> it is only registering the key as <D-anykey>

this is not the case on other gui vim like VimR

MultisampledNight:
See https://github.com/neovide/neovide/issues/992#issuecomment-1019569183.

MultisampledNight:
Partly closed by #1483. `<S-{A,M,D,C}-*` is implemented now, but `<S-*` is still to be done.

fredizzimo:
This should be fixed by #1899, it would be great if you can test that.

mgax:
> This should be fixed by #1899, it would be great if you can test that.

@fredizzimo It doesn't seem to be fixed in 340571b. On Mac OS, I have this pair of key bindings:

```lua
vim.keymap.set("n", "<D-]>", "gt")
vim.keymap.set("n", "<D-[>", "gT")
```

They get triggered when pressing either ⌘-[ or ⌘-shift-[. On the other hand, if I write them as follows, neither key combination will trigger them:

```lua
vim.keymap.set("n", "<D-S-]>", "gt")
vim.keymap.set("n", "<D-S-[>", "gT")
```

mgax:
@MultisampledNight I don't think the bug is resolved, https://github.com/neovide/neovide/issues/1237#issuecomment-1622948873 is still true.

fredizzimo:
> Would this work for <D-]> vs <D-}>

Yes, it should, but we probably need https://github.com/rust-windowing/winit/pull/2887. 

On other plarforms we are able to do similar mappings.

fredizzimo:
It seems like winit might not be reporting the shifted character at all in combination with `CMD`. If that's the case we should report it to the Winit team, since all other platforms does that, and it would be better if it's consistent. As a last resort the following code could be modified to not strip out shift (and probably not alt either) if `CMD` is held on `macOS`
https://github.com/neovide/neovide/blob/eeac6e76fc19eefe07162148510f723706d49481/src/window/keyboard_manager.rs#L119-L134

technicalpickles:
Can someone confirm I am understanding this problem correctly? To map to, say, command-shift-p:

- `<D-S-p>` **does not** work currently
- `<D-P>` **should** work

[...]

PR #2018, opened 2023-09-09
fix: Fix some mappings involving shift

fredizzimo:
**NOTE:** this needs careful testing before merging, so that something else does not break,

This fixes https://github.com/neovide/neovide/issues/2017  - "can't map `<c-s-v>`". This is done by sending shfit when it's combined with CTRL and an ascii alpha character. See the corresponding logic here https://github.com/neovim/neovim/blob/c422722b2e94b94d7f9374dbae12f17580cd1d41/src/nvim/keycodes.c#L764-L799 

It also partially fixes the following issue https://github.com/neovide/neovide/issues/1237 - "CMD + SHIFT works as CMD only". Partially because it's likely not dealing with all the cases. The real issue is in Winit. The fix is to normalize shifted asicii alpha characters already on the Neovide side. This is always done one the Neovim side anyway, so it should be a safe thing to do. 

Note that only the following form of mapping work: `<D-A>`.  There appears to be a bug in Neovim https://github.com/neovim/neovim/issues/25068, preventing the alternative forms `<D-S-A>` and `<D-S-a>` from working, but once that's fixed all forms should work. The `<D-A>` form was chosen because that's the most basic form, for example `<M-S-a>` also simplifies down to `<M-A>` in Neovide, so it's consistent with that. Furthermore, this format of mapping will work, even after the Winit bug https://github.com/rust-windowing/winit/issues/3078 is fixed.

The same non `<S->` form should also always be use for non-alpha mappings for all modifiers. The only special case is CTRL + SHIFT + alpha, which should be mapped like `<C-S-A>` due to legacy reasons.
From https://neovim.io/doc/user/intro.html (the notation section)

> CTRL-{char} {char} typed as a control character; that is, typing {char}
while holding the CTRL key down. The case of {char} is
ignored; thus CTRL-A and CTRL-a are equivalent. But in
some terminals and environments, using the SHIFT key will
produce a distinct code (e.g. CTRL-SHIFT-a); in these
environments using the SHIFT key will not trigger commands
such as CTRL-A

No

### DIFF
--- a/src/window/keyboard_manager.rs
+++ b/src/window/keyboard_manager.rs
@@ -135,16 +135,16 @@ impl KeyboardManager {
         // In all other cases the resulting character is enough.
         // Note that, in Neovim <C-a> and <C-A> are the same, but <C-S-A> is different.
         // Actually, <C-S-a> is the same as <C-S-A>, since Neovim converts all shifted
-        // lowercase alphas to uppercase internally in its mapppings.
+        // lowercase alphas to uppercase internally in its mappings.
         // Also note that mappings that do not include CTRL work differently, they are always
         // normalized in combination with ascii alphas. For example <M-S-a> is normalized to
         // uppercase without shift, or <M-A> .
-        // But in combination with ohter characters, such as <M-S-$> they are not,
+        // But in combination with other characters, such as <M-S-$> they are not,
         // so we don't want to send shift when that's the case.
         let include_shift =
             is_special || (self.modifiers.state().control_key() && is_ascii_alphabetic_char(text));
 
-        // Always send meta (alt) togehter with special keys, or when alt is meta on macOS
+        // Always send meta (alt) together with special keys, or when alt is meta on macOS
         let include_alt = use_alt() || is_special;
 
         let state = self.modifiers.state();
--- a/src/window/keyboard_manager.rs
+++ b/src/window/keyboard_manager.rs
@@ -98,7 +98,7 @@ impl KeyboardManager {
     }
 
     fn format_key_text(&self, text: &str, is_special: bool) -> String {
-        let modifiers = self.format_modifier_string(is_special);
+        let modifiers = self.format_modifier_string(text, is_special);
         // < needs to be formatted as a special character, but note that it's not treated as a
         // special key for the modifier formatting, so S- and -M are still potentially stripped
         let (text, is_special) = if text == "<" {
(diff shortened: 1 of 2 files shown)


JamesWidman:
> Note that only the following form of mapping work: `<D-A>`. There appears to be a bug in Neovim [neovim/neovim#25068](https://github.com/neovim/neovim/issues/25068), preventing the alternative forms `<D-S-A>` and `<D-S-a>` from working, but once that's fixed all forms should work.

What about `<S-D-a>`?  (In neovide 0.10.4 and in the patch that i mentioned [on issue 2017](https://github.com/neovide/neovide/issues/2017#issuecomment-1712803712), this is what neovide/nvim inserts for me when i type `<ctrl-v>` followed by `cmd + shift + a`, and my `cmd + shift` mappings have been working fine with it.)

fredizzimo:
> What about <S-D-a>? (In neovide 0.10.4 and in the patch that i mentioned https://github.com/neovide/neovide/issues/2017#issuecomment-1712803712, this is what neovide/nvim inserts for me when i type <ctrl-v> followed by cmd + shift + a, and my cmd + shift mappings have been working fine with it.)

`<S-D-a>` and `<D-S-a>` are the same, it's just the order that's different. 

When https://github.com/neovim/neovim/issues/25068, is fixed, it will also be the same as  `<D-A>`, and `<D-S-A>`, but currently those two are different from `<D-S-a>` inside Neovim.

The reason for not staying backwards compatible with 0.10.4 is simply because reporting `<S-D-a>` is caused by a winit bug. Furthermore, if that Winit bug is fixed, it will break all user mappings, until additionally https://github.com/neovim/neovim/issues/25068 is fixed. We don't want to do braking changes like that ,0.11.0 already had the breaking change of breaking it completely.

the only two reasonable ways of reporting it is `<D-A>` or `<D-S-A>`, but since the normalized format inside Neovim will be `<D-A>`, that's the way I chose to report it here as well.

JamesWidman:
Also: on a windows machine, with a keyboard with the windows-logo key, wouldn't `shift + win + a` produce the same behavior as what we see on macOS for `shift + cmd + a`?  I mean, `cmd` and `win` both map to `super` in winit, right?  

(i guess the window manager might intercept it, but hopefully microsoft didn't define any mappings that aren't documented in their [list of shortcuts](https://support.microsoft.com/en-us/windows/keyboard-shortcuts-in-windows-dcc61a57-8ff0-cffe-9796-cb9706c75eec))

fredizzimo:
For the windows key, there are very few keys that are available and I'm not sure if the ones that are forwarded from Winit, during my quick testing it seems like they are not. But if it's available and winit forwards it, then it should be mappable.

Usually, you need to register an actual hotkey to capture the windows key events, and some are not even overridable by using global hotkeys.

JamesWidman:
fwiw, LGTM (though i'm not sure if this matters since i'm not a regular contributor)

technicalpickles:
> It also partially fixes the following issue https://github.com/neovide/neovide/issues/1237 - "CMD + SHIFT works as CMD only".

Pulled this branch and tested locally. Can confirm that this PR fixes this issue

fredizzimo:
@technicalpickles, Can you test with some special characters as well, like `<D-%>` for example? I think that we still need https://github.com/rust-windowing/winit/issues/3078, for the complete fix.

MultisampledNight (approved):

MultisampledNight:
Works perfect from what I tested on Wayland with Bone as layout.

What was important for me that most characters were indeed mapped as the resulting characters. For example, <kbd>+</kbd> on QWERTY results in `´` as compose, but <kbd>Shift</kbd> + <kbd>+</kbd> results in `~` as compose. So mapping `<C-~>` is triggered by _both_ <kbd>Caps Lock</kbd> + <kbd>v</kbd> (which types a literal `~`) and <kbd>Shift</kbd> + <kbd>+</kbd> hit **twice** — which is completely fine imo. <kbd>Shift</kbd> (and any other modifier) + <kbd>Enter</kbd> still results in `<S-Enter>` (or the respective modifier), which works fine, too.

technicalpickles:
@fredizzimo Just tested, and that one does not seem to be working. Using these mappings:

```lua
vim.api.nvim_set_keymap("n", "<D-5>", "<cmd>echo 'D-5'<CR>", { noremap = true })
vim.api.nvim_set_keymap("n", "<D-%>", "<cmd>echo 'D-%'<CR>", { noremap = true })

vim.api.nvim_set_keymap("n", "<D-i>", "<cmd>echo 'D-i'<CR>", { noremap = true })
vim.api.nvim_set_keymap("n", "<D-I>", "<cmd>echo 'D-I'<CR>", { noremap = true })
```

These all work except for `<D-%>`

fredizzimo:
@technicalpickles, thanks, we will not close #1237 just yet then.

PR #2025, opened 2023-09-13
feat: configurable mouse-as-touch interpretation

MultisampledNight:
- Feature

Introduces a configuration option `g:neovide_mouse_as_touch`, which makes the mouse act like a finger for a touchscreen.

Personally I want this for QEMU, which doesn't really have a way to pass through touchscreens, and only USB tablets which are basically mouses with more freedom. Also this allows usage on smaller screens or with real USB tablets (think of using a pen on a convertible, for example).

It could be argued that this is a job of the OS, and I'd agree. But on the other hand this solution isn't very intruding and extends the already existing touchscreen infrastructure, and I don't know any software that is capable of remapping the mouse to a touchscreen (neglecting Android/scrcpy) either.

- No


Kethku:
I do understand the desire for this, but it does feel a little niche. Are there many users who would want this? If the answer is no, I'm not sure the cost of maintenance is worth it 😓

MultisampledNight:
The only use case I can think of atm apart from QEMU is also testing the touchscreen/scrolling/mouse infrastructure, like ad-chaos could use in #2057.

However, I'm biased as it's tailored to what I need, so ultimately it's your call. I'd be happy to semi-maintain a fork with this applied, if that is a preferable option.

fredizzimo:
I think some of these changes are quite generic, like the support for finger id. We will need something like that if we implement zooming (perhaps changing font size). I also think that is needed for more reliable scrolling.

So perhaps a compromise would be adding those changes in but leave the mouse support out for now. Although, it does not seem like there would be that much maintenance burden left after that.

MultisampledNight:
The finger ID is just extracted into its own struct, it was previously tracked already ad-hoc in a tuple (on main):

https://github.com/neovide/neovide/blob/a8cec7e16f9f8bd2e3282a9c13928b585d5d596f/src/window/mouse_manager.rs#L80-L81

Either way, after thought I believe this is indeed not worth the maintenance effort. I'll close this and make it a patch in my system configuration instead.

Issue #1580, opened 2022-10-03
window position is always saved and restored, even when `let g:neovide_remember_window_position=v:false`


correabuscar:
Neovide always saves window position and thus (since it's there, saved) it always restores it!
Saves it as `0,0` when `let g:neovide_remember_window_position = v:false`.

ie. 
`DEBUG [neovide::settings::window_geometry] Saved Window Settings: {"window":{"Windowed":{"position":{"x":0,"y":0},"size":{"width":100,"height":50}}}}`
`DEBUG [neovide::settings::window_geometry] Loaded window settings: Windowed { position: PhysicalPosition { x: 0, y: 0 }, size: Dimensions { width: 100, height: 50 } }`

Workaround patch (ie. don't always restore position):
```patch
diff --git a/src/window/mod.rs b/src/window/mod.rs
index 0dccb94..8f74de3 100644
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -295,7 +295,15 @@ pub fn create_window() {
                 maximized = true;
             }
             PersistentWindowSettings::Windowed { position, .. } => {
-                previous_position = Some(position);
+                // don't position window to 0,0 when: let g:neovide_remember_window_position = v:false
+                // but rather, let user's window manager position it
+                let remember_window_position = { SETTINGS.get::<WindowSettings>().remember_window_position };
+                if remember_window_position {
+                    // ugly hack because position is always saved: as 0,0 when this is false
+                    // FIXME: don't save window position when remember_window_position is false!
+                    // seen in settings/window_geometry.rs
+                    previous_position = Some(position);
+                }
             }
         }
     }
```

Fix: TODO: don't [always save position](https://github.com/neovide/neovide/blob/0b6e20c1994c761814eb6a8ad33dea90ade97573/src/settings/window_geometry.rs#L93-L98) ? but rather only when `remember_window_position` is true.

Side fix (unrelated): `repositioned window: true` was always printed even if it were `false`
```patch
diff --git a/src/window/mod.rs b/src/window/mod.rs
index 0dccb94..8f74de3 100644
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -376,9 +384,10 @@ pub fn create_window() {
                 || window_position.y > monitor_position.y + monitor_height
             {
                 window.set_outer_position(monitor_position);
+                Some(())
+            } else {
+                None
             }
-
-            Some(())
         })
         .is_some();
 
```
[...]

PR #2027, opened 2023-09-15
fix: Improve windows creation and sizing

fredizzimo:
**NOTE: This code targets the `improve-render-loop` branch**

* The window is now created as hidden at startup with the correct size from the start, unless `geometry` is specified. 
* Adds support for `:columns` and `:lines`, to change the geometry at any time, including from `init.vim/lua`
* The resizing during startup happens once at `VimEnter` to make it robust against apparent race conditions like https://github.com/neovide/neovide/issues/1972 and https://github.com/neovide/neovide/issues/1580
* https://github.com/neovide/neovide/issues/1965
* https://github.com/neovide/neovide/issues/1958
* https://github.com/neovide/neovide/issues/181
* https://github.com/neovide/neovide/discussions/1857
* Refactors the code to separate `create_window` and `main_loop`.

TODO: Check for open issues about to see other ones that this possibly fixes. NOTE: Don't mark anything to close the issue before the `improve-render-loop` branch has been merged.

- No

### DIFF
--- a/src/bridge/setup.rs
+++ b/src/bridge/setup.rs
@@ -3,7 +3,10 @@ use nvim_rs::Neovim;
 use rmpv::Value;
 
 use super::setup_intro_message_autocommand;
-use crate::{bridge::NeovimWriter, error_handling::ResultPanicExplanation};
+use crate::{
+    bridge::NeovimWriter, error_handling::ResultPanicExplanation, settings::SETTINGS,
+    CmdLineSettings,
+};
 
 const REGISTER_CLIPBOARD_PROVIDER_LUA: &str = r"
     local function set_clipboard(register)
@@ -156,6 +159,16 @@ pub async fn setup_neovide_specific_state(
     )
     .await
     .ok();
+
+    if let Some(geometry) = SETTINGS.get::<CmdLineSettings>().geometry {
+        let lines = geometry.height;
+        let columns = geometry.width;
+        nvim.command(&format!(
+            "autocmd VimEnter * ++once ++nested set lines={lines} | set columns={columns}"
+        ))
+        .await
+        .ok();
+    }
 
     setup_intro_message_autocommand(nvim).await.ok();
 }
--- a/src/bridge/setup.rs
+++ b/src/bridge/setup.rs
@@ -62,7 +62,7 @@ const SETUP_GEOMETRY_LUA: &str = r"
         end
     })";
 
-pub async fn setup_neovide_remote_clipboard(nvim: &Neovim<NeovimWriter>, neovide_channel: u64) {
+pub async fn setup_neovide_remote_clipboard(nvim: &Neovim<NeovimWriter>) {
     // Users can opt-out with
     // vim: `let g:neovide_no_custom_clipboard = v:true`
     // lua: `vim.g.neovide_no_custom_clipboard = true`
@@ -76,9 +76,6 @@ pub async fn setup_neovide_remote_clipboard(nvim: &Neovim<NeovimWriter>, neovide
         return;
     }
 
-    nvim.set_var("neovide_channel_id", Value::from(neovide_channel))
-        .await
-        .ok();
     nvim.execute_lua(REGISTER_CLIPBOARD_PROVIDER_LUA, vec![])
         .await
         .ok();
@@ -135,6 +132,10 @@ pub async fn setup_neovide_specific_state(
             neovide_channel
         );
 
+        nvim.set_var("neovide_channel_id", Value::from(neovide_channel))
+            .await
+            .ok();
+
         // Create a command for registering right click context hooking.
         #[cfg(windows)]
         nvim.command(&build_neovide_command(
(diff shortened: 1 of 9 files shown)


fredizzimo:
@9mm, could you test this, if it solves your startup resizing issues?

9mm:
I will!

src/bridge/setup.rs
```diff
@@ -31,6 +36,32 @@ const REGISTER_CLIPBOARD_PROVIDER_LUA: &str = r"
...
+        pattern = '*',
+        once = true,
+        nested = true,
+        callback = function()
+            if force then
+                vim.o.columns = initial_columns
+                vim.o.lines = initial_lines
+            else
+                -- Just set the values again to trigger the OptionSet callback
+                -- It's not automatically run when running init.vim/lua
```
fredizzimo, line 53: This would not be needed if Neovim sent `columns` and `lines` as UI options

src/bridge/setup.rs
```diff
@@ -142,9 +173,33 @@ pub async fn setup_neovide_specific_state(
...
+    nvim.command(
+        "autocmd OptionSet columns call rpcnotify(g:neovide_channel_id, 'neovide.columns', str2nr(v:option_new))",
+    )
+    .await
+    .ok();
+    nvim.command(
+        "autocmd OptionSet lines call rpcnotify(g:neovide_channel_id, 'neovide.lines', str2nr(v:option_new))",
+    )
+    .await
+    .ok();
```
fredizzimo, line 191: I will create a Neovide feature request for sending `columns` and `lines` as UI options.  So that this wouldn't be needed. 

Another thing to note here is that these don't trigger on automatic resizes with `nvim_try_resize`, which is ideal for avoiding strange race conditions. The user is unlikely to both set those and resize the window manually at the same time.

9mm:
This is a super dumb question.... sorry. How do I test a pull request if its not on the actual branch. should i just pull this?

https://github.com/fredizzimo/neovide/tree/fsundvik/refactor-window-creation

9mm:
nevermind i got it, thats coool... i didnt know you could do that

fredizzimo:
@9mm. I don't know what you did, but I recommend that you install the [Github cli](https://cli.github.com/).

Then you can do `gh pr checkout 2027`

9mm:
Yep thats what i did, and then i switched back to lazygit.

Hmm, so I did

```
cargo install --path .
```

```
  Installing neovide v0.11.1 (/Users/zesty/Code/neovide)
    Updating crates.io index
    Finished release [optimized] target(s) in 1.00s
   Replacing /Users/zesty/.cargo/bin/neovide
    Replaced package `neovide v0.11.1 (/Users/zesty/Code/neovide)` with `neovide v0.11.1 (/Users/zesty/Code/neovide)` (executable `neovide`)
warning: be sure to add `/Users/zesty/.cargo/bin` to your PATH to be able to run the installed binaries
➜  neovide git:(pr/2027)
```

And then ran `/Users/zesty/.cargo/bin/neovide` and it opens the mac Icon but theres no actual neovide window. the icon only appears on the dock. I tried using my standard way of opening with the following:

```
export NEOVIDE_MULTIGRID=true
export NEOVIDE_FRAME=buttonless
export NEOVIDE_VSYNC=0

export ENABLE_EDGE=1

if [ "$ENABLE_EDGE" = "1" ]; then
  neovide_bin="/Users/zesty/.cargo/bin/neovide"
else
  neovide_bin="neovide"
fi
```

But also i get the same thing just directly running `/Users/zesty/.cargo/bin/neovide`

I looked in `~` and theres no stack traces

LMK anything else i should try, maybe i will try to figure out how to build again in non release mode

fredizzimo:
You can try `cargo run --release` first. If that does not work `cargo run`, and finally `cargo run --log` to create the log file.

9mm:
`cargo run -- --log`

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/12623218/neovide_rCURRENT.log)

fredizzimo:
Ah I see the problem now, `neovide_channel_id` is nil, because it's set from Windows specific code. I will fix that

fredizzimo:
@9mm It should be fixed now

9mm:
**Yessss** Works perfect!

Awesome!

9mm:
I also tested saving different sizes and it works as well. It obviously doesnt work with command+Q becuase of the othr reason, but works perfect with `:wq`

MultisampledNight (approved):

MultisampledNight:
Tested on Sway and it works perfectly. Only thing I noticed was with

1. `columns` or `lines` being set in `init.vim`/manually in a session/through `--geometry`
2. Sway _not_ setting the window to floating by default (which is the default if not explicitly requested)

that the actual rendered size is be smaller than that with the rest being transparent:

Neovide window filling half of the screen, where only a small part inside the window is rendered. Everything around it is transparent and shows the wallpaper instead

Upon switching focus though, the rendered size always snapped back to the actual window size. Imo this is okay, I can't think of any case where the user would like to combine manual sizing and the sizing of the WM.

Merging, thank you! Chooooooooooo chooooooooo :steam_locomotive:

PR #2028, opened 2023-09-16
Add drop shadow to floating windows

Kethku:
Adds drop shadows to floating windows to help differentiate them from the background.

image

Remaining Tasks:
- [x] Rebase on improved render branch
- [x] Update skia to newest release that obviates the need for the fork
- [x] Add configuration to turn it off
- [x] Add documentation

- Feature

- No


Kethku:
This PR is still waiting on a release from skia and agreement on the way to configure this behavior. Neovim does have a shadow border in it's list of existing borders, but in my opinion this effect works best when there aren't any added border characters. I'm personally inclined to merge this with a setting of some sort (even just a global setting for now to see how people like it) and then think about integrating with neovim directly after it settles, but its an open question.

MultisampledNight:
Fwiw [skia-safe 0.67.0](https://github.com/rust-skia/rust-skia/releases/tag/0.67.0) is out, including the previously missing shadow flags. Only thing open now is the neovim integration.

last-partizan:
Btw, maybe we can use the same `draw_shadow` function to add shadow to the main window under wayland?

We're using sctk-adwaita, but it uses tiny-skia, which doesn't have `draw_shadow` as far as i can tell, so i cannot add that directly. There was some attempts to draw shadows using other technics, but so far none is ready.

Kethku:
> Btw, maybe we can use the same `draw_shadow` function to add shadow to the main window under wayland?
> 
> We're using sctk-adwaita, but it uses tiny-skia, which doesn't have `draw_shadow` as far as i can tell, so i cannot add that directly. There was some attempts to draw shadows using other technics, but so far none is ready.

Honestly I have no idea. I dont understand linux window compositing and dont use linux personally, so somebody would have to do give it a try. My gut is that it wouldn't work as skia needs a canvas to draw onto, but thats only a guess

Kethku:
This is about ready to merge. Only remaining issues are documentation of the new configuration settings and a fix for the failing macos build. I'm not sure whats up with the mac ci pipeline and am too tired to figure it out myself rn

fredizzimo:
I think need to figure out the mac CI issues first, it seems like there's some problems with the ARM toolchain, even if the logs says that it's installed properly.

As for the window shadows, I don't think we should give in on Gnome's agenda to refuse to implement the server side window decorations - https://wayland.app/protocols/xdg-decoration-unstable-v1 protocol and force every application to draw their own client-side decorations. At least not without using a common library, like `sctk-adwaita`, which we already use. 

I don't see any technical reason's why it can't be done there, it's just that none one have had the motivation to do the work until now, but just yesterday, there's a new draft PR up https://github.com/PolyMeilex/sctk-adwaita/pull/43

Furthermore, the more complete adwaita library has moved under the Gnome organization here https://gitlab.gnome.org/World/Rust/libadwaita-rs. So maybe someone could convince winit to support that.

src/renderer/rendered_window.rs
```diff
@@ -285,25 +285,28 @@ impl RenderedWindow {
...
-        if self.anchor_info.is_some() {
+        if self.anchor_info.is_some() && settings.floating_shadow {
             root_canvas.save();
             let shadow_path = Path::rect(pixel_region, None);
             // We clip using the Difference op to make sure that the shadow isn't rendered inside
             // the window itself.
             root_canvas.clip_path(&shadow_path, Some(ClipOp::Difference), None);
+            // The light angle is specified in degrees from the vertical, so we first convert them
+            // to radians and then use sin/cos to get the y and z components of the light
+            let light_angle_radians = settings.light_angle_degrees * std::f32::consts::PI / 180.;
```
MultisampledNight, line 296: Could use `f32::to_radians`.

```suggestion
            let light_angle_radians = settings.light_angle_degrees.to_radians();
```

Issue #1641, opened 2022-11-17
Neovide hover popup overflows the window edge to the right/bottom and not staying within window boundaries


juhaku:
**Describe the bug**
When hovered over method / function / struct etc declaration the hover box overflows over the right screen edge when opened with neovide. But when same operation is performed with neovim using terminal or in this case KDE's konsole the hover will not overflow the screen edge but stays withing the window boundaries.

I am aware there was previously bug related it overflowing to the left, but did not find a issue related it bleeding to the right of the window.

**To Reproduce**
1. Open any file which has language support protocol enabled with support for hovering documentation. 
2. Either make window small enough (for example splitting) or make sure you have big enough hover documentation available.
3. Open hover e.g. hovering a function / method / struct / class etc. declaration.
4. See if the hover window stays withing the window (and within split window).

**Expected behavior**
The hover should stay within the window boundaries and not overflow to the right.

**Screenshots**
![image](https://user-images.githubusercontent.com/26358664/202572587-09d9ac13-b605-4b74-ba1b-fe3c6048b845.png)
![image](https://user-images.githubusercontent.com/26358664/202573288-175bce6e-d73b-4442-8853-f4a1adbbbb70.png)

**Desktop (please complete the following information):**
 - OS: Manjarjo KDE / Endeavour OS KDE / v. 5.26 ->
 - Neovide Version 0.10.3
 - Neovim Version 0.8.0 ->

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

**Additional context**
Attached log files since the content would be too large to paste in this thread.
[neovide_r2022-11-18_00-39-32.log](https://github.com/neovide/neovide/files/10036250/neovide_r2022-11-18_00-39-32.log)
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/10036251/neovide_rCURRENT.log)

juhaku:
It also overflows to the bottom as seen in latter image.
![image](https://user-images.githubusercontent.com/26358664/202877004-08152ccf-fcff-421a-8199-a990f3a7435f.png)
![image](https://user-images.githubusercontent.com/26358664/202876972-0d8ed9a6-8a1c-41ef-8589-dda4688aefc5.png)

fredizzimo:
This is probably fixed by https://github.com/neovide/neovide/pull/1786

juhaku:
@fredizzimo  Just tested with master branch and result is still the same.
[...]

PR #2029, opened 2023-09-16
fix: Reposition the floating windows inside the main window

fredizzimo:
* Note: builds on top of https://github.com/neovide/neovide/pull/2027, so this should be merged and reviewed after that.

The floating windows are now re-positioned to fit inside the main window, as long as it's possible. And the same thing is done for the messages window, so that you don't see a blank screen if there are errors at startup.

* https://github.com/neovide/neovide/issues/1641 will be fixed once it's completely merged to main.

- No

### DIFF
--- a/src/editor/window.rs
+++ b/src/editor/window.rs
@@ -9,6 +9,7 @@ use crate::{
     renderer::{LineFragment, WindowDrawCommand},
 };
 
+#[derive(Clone, Copy, Debug, PartialEq)]
 pub enum WindowType {
     Editor,
     Message,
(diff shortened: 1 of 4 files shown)


MultisampledNight (approved):

fredizzimo:
I rebased to fix the conflicts due to git selecting the wrong base.

fredizzimo:
I’m fixing a few issues related to paddings. So I temporarily set the draft status.

fredizzimo:
I fixed the problems with the padding.

And I also did some refactoring to store the window padding only once instead of three times the same value was stored in `renderer` , `rendered_window` and `winit_window_wrapper`.

PR #2032, opened 2023-09-17
Added numpad support

plague-spreader:
- Fix

- No

### DIFF
--- a/src/window/keyboard_manager.rs
+++ b/src/window/keyboard_manager.rs
@@ -67,29 +67,25 @@ impl KeyboardManager {
         }
     }
 
-    fn my_some(s: &str) -> Option<&str>{
-        Some(s)
-    }
-
     fn handle_numpad_numkey<'a>(
         is_numlock_enabled: bool, numlock_str: &'a str,
         non_numlock_str: &'a str)
         -> Option<&'a str> {
         if is_numlock_enabled {
-            return KeyboardManager::my_some(numlock_str);
+            return Some(numlock_str);
         }
-        return KeyboardManager::my_some(non_numlock_str);
+        return Some(non_numlock_str);
     }
 
     fn handle_numpad_key(key_event: &KeyEvent) -> Option<&str> {
         let is_numlock_key = key_event.text.is_some();
         match key_event.physical_key {
-            KeyCode::NumpadDivide   => KeyboardManager::my_some("<kDivide>"),
-            KeyCode::NumpadStar     => KeyboardManager::my_some("<kMultiply>"),
-            KeyCode::NumpadSubtract => KeyboardManager::my_some("<kMinus>"),
-            KeyCode::NumpadAdd      => KeyboardManager::my_some("<kPlus>"),
-            KeyCode::NumpadEnter    => KeyboardManager::my_some("<kEnter>"),
-            KeyCode::NumpadDecimal  => KeyboardManager::my_some("<kDel>"),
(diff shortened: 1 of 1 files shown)


src/window/keyboard_manager.rs
```diff
@@ -67,8 +67,57 @@ impl KeyboardManager {
...
+            KeyCode::Numpad0 => self.handle_numpad_numkey(is_numlock_key,
+                                                          "<k0>", "<Insert>"),
+            _ => None
+        }
+    }
+
     fn format_key(&self, key_event: &KeyEvent) -> Option<String> {
-        if let Some(text) = get_special_key(key_event) {
+        if key_event.location == KeyLocation::Numpad {
+            self.handle_numpad_key(key_event)
```
fredizzimo, line 119: I think we should deal with the numpad keys the same way as special keys, so that they can properly be combined with modifiers.

src/window/keyboard_manager.rs
```diff
@@ -67,8 +67,57 @@ impl KeyboardManager {
...
+        Some(s.to_string())
+    }
+
+    fn handle_numpad_numkey(&self, is_numlock_enabled: bool,
+                            numlock_str: &str, non_numlock_str: &str)
+        -> Option<String> {
+        if is_numlock_enabled {
+            return KeyboardManager::my_some(numlock_str);
+        }
+        return KeyboardManager::my_some(non_numlock_str);
```
fredizzimo, line 80: I think this is easier to read without `my_some`, because I have no idea what it does without checking the implementation.

fredizzimo (changes requested):
I think we want to deal with the numpad keys

src/window/keyboard_manager.rs
plague-spreader: Also changed this, thanks for the reviews :)

plague-spreader: I actually have one question: in my code I respected the 80 lines margin, is this the code style wanted for the repository? Sorry for the questions, I have some programming knowledge but I never worked in an community-driven project.

fredizzimo: We use `cargo fmt` for the formatting with the default options https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=, so line length is 100. 

Actually, the linter failed now, when I enabled the CI build.

src/window/keyboard_manager.rs
```diff
@@ -67,6 +67,60 @@ impl KeyboardManager {
...
 
+    fn handle_numpad_numkey<'a>(
+        is_numlock_enabled: bool,
+        numlock_str: &'a str,
+        non_numlock_str: &'a str,
+    ) -> Option<&'a str> {
+        if is_numlock_enabled {
+            return Some(numlock_str);
+        }
+        return Some(non_numlock_str);
```
fredizzimo, line 78: You can remove return from this line, otherwise `cargo clippy` will complain.

fredizzimo (approved):

fredizzimo:
I have not yet tested this, but the code looks good.

plague-spreader:
Re-requesting check due to a bug introduced by erroneously adding < and > signs within the special key formatting

fredizzimo (approved):
Yes, that's good. I still haven't tested it myself though.

fredizzimo:
Thank you!

I did not get the all the modifiers to work in combination with the keypad on Windows. But I think that's more of a problem with Winit not reporting it correctly, or even not supported by the operating system.

Issue #1477, opened 2022-08-12
Neovide is ignoring `set autoread`, i.e. reloading the currently edited file when activated


SirVer:
**Describe the bug**
MacVim and gVim automatically reload the buffer from disk when they are activated (i.e. clicked into). Neovide is not doing this for me.

**To Reproduce**
Make sure `:set autoread` is in your vimrc.

~~~
echo "Hello World" > /tmp/blub.txt
neovide /tmp/blub.txt
vim /tmp/blub.txt -> and change a bunch of stuff, save the file
Click on neovide -> it still shows the original content
~~~

**Expected behavior**
The buffer should be updated from the contents from disk, as long as it has not been changed in Neovide. If it has, an error should be shown.

**Desktop (please complete the following information):**
 - OS: Mac OS, but I doubt it matters
 - Neovide Version: not sure, the latest version downloaded 22-08-12
 - Neovim Version: 0.6.0

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
`/Volumes/Neovide/Neovide.app/Contents/MacOS/neovide --log` does not run for me - due to CodeSign it seems. How do I get the commandline to run?

0x00002a:
neovim bug: https://github.com/neovim/neovim/issues/20082

alexventuraio:
I added this line to my config and it works fine according to this comment https://github.com/neovim/neovim/issues/20082#issuecomment-1288228566:
```vim
autocmd FocusGained * checktime
```

Kethku:
closing as upstream not related to us.

drmikehenry:
The `'autoread'` feature does not work out-of-the-box for both `nvim-qt` and `neovide`:
- <https://github.com/equalsraf/neovim-qt/issues/846>
- <https://github.com/neovide/neovide/issues/1477>

The `'autoread'` feature does work for `nvim`'s TUI.

The reason it works for the TUI is mentioned in this now-closed Neovim issue:
<https://github.com/neovim/neovim/issues/20082#issuecomment-1236324274>
> No. I just noticed that neovide already triggers `FocusGained`, and that triggering `FocusGained` from `:doautocmd` doesn't actually trigger a timestamp check. With TUI a timestamp check is done the same time as `FocusGained`, but it is not an autocommand callback, but something triggered the same time as the autocommand.

In that issue, Neovim maintainer @justinmk has commented on the `'autoread'` feature:
[...]

PR #2033, opened 2023-09-17
Tell neovim focus was gained/lost via 'ui_set_focus'

9mm:
- Fixes https://github.com/neovide/neovide/issues/1477

This will use the newly added [nvim_ui_set_focus](https://neovim.io/doc/user/api.html#api-ui) function to notify neovim of focus gained/loss.

- No

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1491,9 +1491,9 @@ dependencies = [
 
 [[package]]
 name = "nvim-rs"
-version = "0.5.0"
+version = "0.6.0"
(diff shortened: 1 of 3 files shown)


fredizzimo (approved):
If it works as documented, then this looks good, thanks.

9mm:
I have tested basic functionality by opening a file, opening a new window with a different editor, saving that file. Upon returning to the nvim window I can see it automatically updates without needing the following in my nvim config:

```
autocmd FocusGained * checktime
```

fredizzimo:
Thank you!

PR #2035, opened 2023-09-18
Update to Winit 0.29 master

fredizzimo:
This does a few different things:
* Updates the following dependencies
  * Winit master 0.29.1-beta#2422ea39d0e97fd43698391f84d9300cd169d8cd
  * Glutin 0.4.1-beta
* Removes `vsync_wayland`, and uses the built-in winit frame throttling instead
* Ensures that a rendering happens when the window needs to be re-painted. There were some situations where this could occur previously, like moving an off-screen window into view.
* Fixes the scale factor updates on Windows
* A clean exit is now performed, which gives Neovim a chance to properly exit as well

NOTE:  This also includes the changes from
  *  https://github.com/neovide/neovide/pull/2029
  * https://github.com/neovide/neovide/pull/2034
  * And replaces https://github.com/neovide/neovide/pull/2009

- No


9mm:
I'm now on this branch testing, I'll let you know if it occurs. given that its pretty rare i should know in the next day or so.

Also - it looks like this was just merged as well, so thats cool its not blocking anymore https://github.com/rust-windowing/winit/pull/3091

9mm:
so far workin perfectly

fredizzimo:
@9mm, is that something that happened just once, or soemthing you can repeat?

9mm:
It is repeatable every time i open it

fredizzimo:
Can you send a log file.

9mm:
Sure

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/12688890/neovide_rCURRENT.log)

fredizzimo:
I’m pretty sure it has something to do with

* https://github.com/neovide/neovide/pull/2029

I should probably take the margins into account or some5in.

9mm:
You know... i found out i can reproduce it much better if i open a file from command line:

```
n ~/Desktop/deposit-options.js
```

I can get it to occur about 1 out of 10 times it seems, so that should help testing a fix

image


image

9mm:
I dont know if this helps narrow it down, but i have animations disabled (set to 0) and when i open a file from command line, i can see the cursor for probably ... 15ms ... it does a animation "jump" as if I had opened the terminal and the cursor animate/zooms and leaves a trail

Im pretty sure whats happening is the padding is initially not there, and then when the paddnig is animated, visually i see the cursor animate diagonally down about 20px distance

I can try to record a loom

9mm:
The cusor thing is irrelevant i just only mentioned that because maybe it explains "where" the problem is happening to narrow it down, ie, before config options are set. it happens at around 0:16

As far as the initial problem, you can see it happen in the loom on the first and last time before video ends

https://www.loom.com/share/37fa34229cb443ab8beaba86c85b2cd8

fredizzimo:
Yes, it does seem like it does not re-position the windows after the padding have been applied. Do they snap back to the correct place if you type something, or do they stay like that?

9mm:
It will remain broken through any typing, command mode, opening/closing tabs.

What does fix it however, is if i resize the window 1px

fredizzimo:
Ok, that seems to indicate a pure logic error somewhere, since we redraw everything each time you type. 

It could be due to my changes, but it could also go as far back as to the PR that added the padding support
* https://github.com/neovide/neovide/pull/1632

And anything between

9mm:
I see, well aside from the fact that typing doesnt 'fix' it, is that the same problem as it occuring in the first place?

In other words, is the actual fix for it not to occur simple, even if there is a separate bug where typing doesnt redraw, or are those one and the same issue?

9mm:
I also notice in this branch the window is blocked from appearing if something is active which is annoyance, for example if lazy.nvim is installing a plugin.

fredizzimo:
@9mm. I think I have fixed all three of your previous problems
1. The command line being clipped on the left side, when you have window padding enabled
2. The window padding was not updated properly, causing the main window to overlap with the status bar
3. The window was not shown when a swap message was shown at startup
  * It still a bit buggy in the sense that Neovide crashes if you press `quit` or `abort`. But fixing that needs some changes to the way we do the error handling in general.
 
But unfortunately, I don't think these sets of fixes the new problem you reported about  lazy.nvim

9mm:
@fredizzimo you are a based god, let me check right quick

fredizzimo:
These new changes includes everything from main, I just merged a few ones including yours @9mm.

9mm:
Aweome, I see 3 things fixed, but 1 small quirk introduced

1. ✅ **FIXED** - The command line being clipped on the left side, when you have window padding enabled
2. ✅ **FIXED (I tried about 50 times and cant see any padding issues)** - The window padding was not updated properly, causing the main window to overlap with the status bar
3. ✅ **FIXED** The window was not shown when a swap message was shown at startup

The quirk i see is that now any new time a window or popup is introduced it slides "down and to the right" because its animating the padding amount

For example if i open neotree, or telescope window.

I assume whats happening is, even though i have animations disabled, its opening the window at 0,0, adding padding, then transforming the window to (0+x, 0+y) but ignoring the animation setting (so x, y is the current padding values I have).

fredizzimo:
@9mm, can you try if disabling these changes fixes what you see https://github.com/neovide/neovide/pull/2035/commits/cf8e85200c639830a3816b89eb37257f85d76636

9mm:
That fixes it @fredizzimo

fredizzimo:
@9mm,

Thanks, I removed that commit. It can probably be done in a better way later.

9mm:
Awesome!! Thank you, this is epic. Let me know if you need more OSX testing anywhere, even on unrelated branches. I'm happy to help.

src/bridge/mod.rs
```diff
@@ -81,12 +87,19 @@ async fn start_neovim_runtime(instance: NeovimInstance) {
...
         }
     }
-
     let settings = SETTINGS.get::<CmdLineSettings>();
 
     let should_handle_clipboard = settings.wsl || settings.server.is_some();
-    setup_neovide_specific_state(&nvim, should_handle_clipboard).await;
+    setup_neovide_specific_state(&session.neovim, should_handle_clipboard).await;
 
+    start_ui_command_handler(session.neovim.clone());
```
MultisampledNight, line 95: It's not quite clear here that in fact the `Arc` shallow copied.

```suggestion
    start_ui_command_handler(Arc::clone(session.neovim));
```

MultisampledNight (approved):

MultisampledNight:
Thank you! Choooooo choooooooooooooooo

PR #2046, opened 2023-09-23
Fix option as meta

JamesWidman:
- Fix
This re-implements support for the alt_is_meta feature for macOS using winit's `OptionAsMeta` interface, [as requested](https://github.com/neovide/neovide/issues/2017#issuecomment-1712955629).

It also introduces the ability to map only the left or only the right 'option' key to the virtual key 'meta',  which leaves the other 'option' key free so that it can be used to access the additional two layers of the macOS input source. (E.g. with the u.s. english input source selected, option-c produces "ç" and option-shift-c produces "Ç".) This is done by setting the nvim variable:

```
vim.g.neovide_input_macos_option_key_is_meta
```

Possible values (defined by winit::platform::macos::OptionAsAlt) are:
```
    "Both"
    "OnlyLeft"
    "OnlyRight"
    "None"
```

The preëxisting variable `vim.g.neovide_input_macos_alt_is_meta` can still be used, and if true, it will internally result in the use of OptionAsAlt::Both.

This PR also fixes an issue where, on macOS, format_modifier_string() [would include "S-" for "meta + shift + cmd + y"](https://github.com/neovide/neovide/issues/2017#issuecomment-1712827024) (producing `<M-S-D-Y>` instead of `<M-D-Y>`)

if the user changes the nvim variable `vim.g.neovide_input_macos_alt_is_meta` after startup, it will now be ignored. This is not ideal, but from within `WinitWindowWrapper::synchronize_settings()`, there isn't a way to tell whether the legacy variable setting happened before or after a use of the new variable (`vim.g.neovide_input_macos_option_key_is_meta`). The workaround is to use the new variable.

Potentially, this breaking change could be avoided; see the comment in `WinitWindowWrapper::new()`

Fwiw, this is my first attempt at using the Rust programming language (outside of small toy examples from several years ago), so I probably made some mistakes that a beginner would make.

Also note that this PR is based on [fredizzimo:fsundvik/fix-ctrl-shift](https://github.com/neovide/neovide/pull/2018).

### DIFF
--- a/src/window/keyboard_manager.rs
+++ b/src/window/keyboard_manager.rs
@@ -3,7 +3,11 @@ use crate::{
     event_aggregator::EVENT_AGGREGATOR,
 };
 #[cfg(target_os = "macos")]
-use crate::{settings::SETTINGS, window::KeyboardSettings};
+use {
+    crate::{settings::SETTINGS, window::KeyboardSettings},
+    winit::{keyboard::ModifiersKeyState, platform::macos::OptionAsAlt},
+};
+
 #[allow(unused_imports)]
 use winit::platform::modifier_supplement::KeyEventExtModifierSupplement;
 use winit::{
@@ -18,6 +22,7 @@ fn is_ascii_alphabetic_char(text: &str) -> bool {
 pub struct KeyboardManager {
     modifiers: Modifiers,
     ime_preedit: (String, Option<(usize, usize)>),
+    meta_is_pressed: bool, // see N.B. on 'meta' below
 }
 
 impl KeyboardManager {
@@ -25,6 +30,7 @@ impl KeyboardManager {
         KeyboardManager {
             modifiers: Modifiers::default(),
             ime_preedit: ("".to_string(), None),
+            meta_is_pressed: false,
         }
     }
 
@@ -66,6 +72,25 @@ impl KeyboardManager {
                 // text
                 log::trace!("{:?}", *modifiers);
                 self.modifiers = *modifiers;
+
+                #[cfg(target_os = "macos")]
+                {
+                    let ks = SETTINGS.get::<KeyboardSettings>();
+                    self.meta_is_pressed = match ks.macos_option_key_is_meta.0 {
+                        OptionAsAlt::Both => self.modifiers.state().alt_key(),
+                        OptionAsAlt::OnlyLeft => {
+                            self.modifiers.lalt_state() == ModifiersKeyState::Pressed
+                        }
+                        OptionAsAlt::OnlyRight => {
+                            self.modifiers.ralt_state() == ModifiersKeyState::Pressed
+                        }
+                        OptionAsAlt::None => false,
+                    };
+                }
+                #[cfg(not(target_os = "macos"))]
+                {
+                    self.meta_is_pressed = self.modifiers.state().alt_key();
+                }
             }
             _ => {}
         }
@@ -134,11 +159,11 @@ impl KeyboardManager {
         // the whole modifier state. Otherwise send the resulting character with "S-" and "M-"
         // removed.
         #[cfg(target_os = "macos")]
-        if self.modifiers.state().alt_key() && use_alt() {
+        if self.meta_is_pressed {
             return key_event
                 .key_without_modifiers()
                 .to_text()
-                .map(|text| self.format_key_text(text, true));
+                .map(|text| self.format_key_text(text, false));
         }
 
         key_event
@@ -191,33 +216,22 @@ impl KeyboardManager {
         // uppercase without shift, or <M-A> .
         // But in combination with other characters, such as <M-S-$> they are not,
         // so we don't want to send shift when that's the case.
-        let include_shift =
-            is_special || (self.modifiers.state().control_key() && is_ascii_alphabetic_char(text));
-
-        // Always send meta (alt) together with special keys, or when alt is meta on macOS
-        let include_alt = use_alt() || is_special;
-
         let state = self.modifiers.state();
+        let include_shift = is_special || (state.control_key() && is_ascii_alphabetic_char(text));
+
+        #[cfg(target_os = "macos")]
+        let have_meta = self.meta_is_pressed || is_special && state.alt_key(); // e.g. non-meta 'option' with <F1> yeilds <M-F1>
+
+        #[cfg(not(target_os = "macos"))]
+        let have_meta = self.meta_is_pressed;
+
         let mut ret = String::new();
         (state.shift_key() && include_shift).then(|| ret += "S-");
         state.control_key().then(|| ret += "C-");
-        (state.alt_key() && include_alt).then(|| ret += "M-");
+        (have_meta).then(|| ret += "M-");
         state.super_key().then(|| ret += "D-");
         ret
     }
-}
-
-#[cfg(not(target_os = "macos"))]
-fn use_alt() -> bool {
-    true
-}
-
-// The option or alt key is used on Macos for character set changes
-// and does not operate the same as other systems.
-#[cfg(target_os = "macos")]
-fn use_alt() -> bool {
-    let settings = SETTINGS.get::<KeyboardSettings>();
-    settings.macos_alt_is_meta
 }
 
 fn get_special_key(key_event: &KeyEvent) -> Option<&str> {
@@ -287,3 +301,57 @@ fn get_special_key(key_event: &KeyEvent) -> Option<&str> {
         _ => None,
     }
 }
+
+// N.B. on 'meta', and on the macintosh key 'option':
+//
+// 'Meta' can be thought of as a virtual key. On a macintosh, either or both of
+// the physical keys labeled 'option' (⌥) may be configured to map to the
+// virtual key 'meta' by using the neovide setting:
+//         vim.g.neovide_input_macos_option_key_is_meta
+// ..where possible values (defined by the 'winit' crate) are:
+//
+//    "Both"
+//    "OnlyLeft"
+//    "OnlyRight"
+//    "None"
+//
+// (On a Windows PC, or on non-mac POSIX platforms with a Windows PC keyboard,
+// the physical key labeled 'alt' always maps to the virtual key 'meta'.)
+//
+// When an 'option' key is:
+//
+//     - not mapped to meta, and
+//
+//     - used with a printable character (like "y")
+//
+// ...the option key behaves like a second kind of 'shift' key in the sense that
+// it transforms the printable character into a different printable character;
+// for example, just as shift+y transforms 'y' into 'Y', option+y (on a u.s.
+// english layout) transforms 'y' into '¥' (U+00A5 YEN SIGN), and shift+option+y
+// transforms 'y' into 'Á' (U+00C1 LATIN CAPITAL LETTER A WITH ACUTE). See also:
+//
+// https://en.wikipedia.org/wiki/Option_key
+//
+// And like the 'shift' key, the non-meta 'option' key is not represented in the
+// string returned by format_modifier_string() (because it would be redundant
+// next to the transformed printable character).
+//
+// But when a non-meta 'option' key is used with a special key (see
+// get_special_key() above), we may as well treat it as 'meta' because:
+//
+//    - This is how we behaved before the behavior was documented (so a user
+//      could have had alt_is_meta=false and still have used <M-CR>, <M-F1>,
+//      etc.); and
+//
+//    - There is no secondary layer of special keys.
+//
+// Note on 'option' vs 'alt':
+// On macintosh keyboards made until 2018, the 'option' key additionally bears
+// the label 'alt'. But from 2018 on, 'alt' no longer appears on this key; it is
+// labeled only with the word 'option' and the symbol '⌥' (U+2325 OPTION KEY).
+//
+// Both before and after 2018, this key has been consistently labeled 'option'
+// (at least as far back as the earliest macintosh that neovide supports). So to
+// avoid confusing users who have a post-2017 keyboard and are not aware of this
+// history, it is probably best to refer to this physical key as the 'option'
+// key, and not as the 'alt' key.
--- a/src/window/settings.rs
+++ b/src/window/settings.rs
@@ -1,3 +1,14 @@
+#[cfg(target_os = "macos")]
+use {
+    log::error,
+    rmpv::{Utf8String, Value},
+    serde::{
+        de::{value, IntoDeserializer},
+        Deserialize,
+    },
+    winit::platform::macos::OptionAsAlt,
+};
+
 use crate::{cmd_line::CmdLineSettings, settings::*};
 
 #[derive(Clone, SettingGroup)]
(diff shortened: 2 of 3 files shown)


JamesWidman:
[updated the "breaking changes" section]

JamesWidman:
re `src/window/settings.rs`: I'm not sure if I implemented serialization/de-serialization correctly (or if there was a way to avoid doing anything explicit here); comments on this would be very much appreciated!

JamesWidman:
cc @clason (another mac user)

fredizzimo:
@JamesWidman. Thank you. I will review the code a bit later, and also see if we can get support for updating it at any time.

@apettel, here's an alternative implementation of what you did. 
@9mm, here's a macOS PR to test if you want to help

9mm:
I can confirm that with the following setting:

```
```

Option (right) + c produces `ç`

Meta (left) + Command + y produces `<M-D-y>` instead of `<M-S-D-Y>`

I tested `OnlyLeft, OnlyRight, Both, None` and all work as expected

I dont use `M` in hotkeys much so if there is more specific test of what to look for please let me know

JamesWidman:
> I dont use `M` in hotkeys much so if there is more specific test of what to look for please let me know.

The test results you gave (particularly the output for meta + shift + command +x, and the output for (non-meta) option + c) should be enough; thank you for testing!

JamesWidman:
Aaaaand i just realized that this part of the git commit message:

> This is now fixed by relying on winit to supply the transformed printable character (i.e., winit produces the text "Y" instead of "y").

is not true in the case of e.g. shift + command + y.  In this case, winit actually gives us: 
```
TRACE [neovide::window::keyboard_manager] KeyEvent {
    physical_key: KeyY,
    logical_key: Character(
        "y",
    ),
    text: Some(
        "y",
    ),
    location: Standard,
    state: Pressed,
    repeat: false,
    platform_specific: KeyEventExtra {
        text_with_all_modifiers: Some(
            "y",
        ),
        key_without_modifiers: Character(
            "y",
        ),
    },
}
```
...however, in the case of shift + y, winit gives us:
```
TRACE [neovide::window::keyboard_manager] KeyEvent {
    physical_key: KeyY,
    logical_key: Character(
        "Y",
    ),
    text: Some(
        "Y",
    ),
    location: Standard,
    state: Pressed,
    repeat: false,
    platform_specific: KeyEventExtra {
        text_with_all_modifiers: Some(
            "Y",
        ),
        key_without_modifiers: Character(
            "y",
        ),
    },
```
... and in the first case, we produce the string `<D-Y>` (instead of `<D-y>`) only because neovide's format_key_text() does this:
```rust
 let text = if self.modifiers.state().shift_key() && is_ascii_alphabetic_char(text) {
     text.to_uppercase()
 } else {
     text.to_string()
 };
```
...which is also why we produce the wrong string for shift + command + 5. (i.e., `to_uppercase()`  doesn't transform '5' into '%', and in general we don't want to be using `to_uppercase()` to get the result of transforming the character by any combination of `shift` and `option` in the presence of `command`).

So like @fredizzimo said elsewhere, it looks like we'll need a fix in winit to resolve this.

fredizzimo:
Yes, the winit issue for that is the following
* https://github.com/rust-windowing/winit/issues/3078

fredizzimo:
The main issue connected to this PR is this
* https://github.com/neovide/neovide/issues/1877

But could someone also test if this fixes?
* https://github.com/neovide/neovide/issues/1866

@JamesWidman. The ctrl shift fix is merged to main now, so once you rebase the PR should be easier to review. The CI also complains about some formatting issues, you can run `cargo fmt` to fix them.

JamesWidman:
> @JamesWidman. The ctrl shift fix is merged to main now, so once you rebase the PR should be easier to review. The CI also complains about some formatting issues, you can run `cargo fmt` to fix them.

done!

src/window/settings.rs
```diff
@@ -49,10 +60,36 @@ impl Default for WindowSettings {
...
+            Ok(oa) => *self = OptionAsMeta(oa),
+            Err(e) => error!("Setting neovide_input_macos_option_key_is_meta expected one of OnlyLeft, OnlyRight, Both, or None, but received {:?}: {:?}", e, value),
+        };
+    }
+}
+
+#[cfg(target_os = "macos")]
+impl From<OptionAsMeta> for Value {
+    fn from(oam: OptionAsMeta) -> Self {
+        let s = serde_json::to_string(&oam.0).unwrap();
```
JamesWidman, line 82: i feel a little uncertain about this (i.e. whether it's appropriate to use the JSON serializer here).

more broadly, i don't really know if i took the right approach to serialization/deserialization.

JamesWidman: ...but maybe this is secondary to the topic of whether we want to try a different approach to settings updates as mentioned in the comment in `WinitWindowWrapper::new()`?

9mm:
On this most recent code from 1hr ago, `<M-n>` returns `Pattern not found: eventloopbuilder` (weird)

However, if I actually map a key to it, it works (I tested both insert/normal mode)

```
vim.keymap.set({ 'n', 'i' }, '<M-n>', function() vim.cmd('Lazy') end, { desc = 'LazyVim' })
```

JamesWidman:
> On this most recent code from 1hr ago, `<M-n>` in insert mode echos an error to the console: `Pattern not found: eventloopbuilder` (weird)

That *is* weird...

I'm having trouble reproducing it (and i made sure that `<M-n>` is not mapped in insert mode, and also not mapped in normal mode); here's my environment:

- commit: `99a18f2`
- OS: macOS Ventura 13.6 (22G120)
- physical keyboard: macbook, US english 
- input source ("System Settings" -> "Keyboard" -> "Text Input"): US english
- cargo: 1.72.0 (103a7ff2e 2023-08-15)
- rustc: 1.72.0 (5680fa18f 2023-08-23)

I also tried `git grep 'Pattern not found'` in both the neovide repo and the winit repo; no hits. I'll try looking in the local cache of all other dependencies...

JamesWidman:
@9mm 
> I also tried git grep 'Pattern not found' in both the neovide repo and the winit repo; no hits. I'll try looking in the local cache of all other dependencies...

...and... no hits under `~/.cargo/registry/src`...

I'm stumped. Is anyone else able to reproduce this?

9mm:
I’ll check tomorrow to make sure it’s not some plugin Sent from my iToaster—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>

JamesWidman:
@9mm i think it's probably not coming from the iToaster though, because from a local clone of the winit repository, when we do:
```sh
git log -p -SEventLoopBuilder
```
...we see some relatively-recent commits involving `winit::event_loop::EventLoopBuilder`.  

So winit is probably involved somehow. But I don't see how `<M-n>` (or any other key event) would lead to that by default...

You probably already checked, but (without the mapping that you mentioned) when you invoke the vim command:
```vim
:map <M-n>
```
...nvim replies "No mapping found", right?

fredizzimo:
I'm pretty sure `Pattern not found: eventloopbuilder` comes from the built-in search. More specifically search next - `n`. I'm not sure how that triggers from insert mode though. Maybe there's some strange mapping going on causing it to run a normal mode command?

JamesWidman:
> I'm pretty sure `Pattern not found: eventloopbuilder` comes from the built-in search.

Ah... this seems to be it.

So, I've confirmed that I don't have any mappings (insert mode or otherwise) for `<M-n>`.  Then, in insert mode, I issue `<M-n>`, and nvim behaves as if I had typed `esc` `n`. (i.e. it repeats the last "/" search.)

I still don't see the error message on the console, but I see the normal-mode `n` behavior.

(this is with nvim 0.9.2.)

Presumably nvim also behaves this way on other platforms (in the event of `alt` + `n`)...?

JamesWidman:
Maybe there is some part of nvim that, in insert mode, would translate "`<M-n>`" into the terminal control character for "escape" (U+001B escape) followed by `n`?

fredizzimo:
> Presumably nvim also behaves this way on other platforms (in the event of alt + n, assuming the input front-end translates that to "<M-n>")...?

Yes, and that's actually a documented behaviour in `intro.txt`
> - When a key is pressed using a meta or alt modifier and no mapping exists for
  that keypress, Nvim may behave as though <Esc> was pressed before the key.

9mm:
Man thats weird... I thought it might be in the clipboard, and I closed and opened a new tab and it showed up again. I didnt think to actually check that it was the search buffer. duh

fredizzimo:
@9mm 
* https://neovim.io/doc/user/usr_21.html#21.3
* https://neovim.io/doc/user/starting.html#shada-file

9mm:
Pretty awesome, thanks for posting that

JamesWidman:
update: waiting on steps to reproduce [this odd behavior](https://github.com/neovide/neovide/issues/1866#issuecomment-1733595121).

JamesWidman:
> That one will probably be hard to fix, so maybe we should rebase this, review and merge?

will do!

src/window/window_wrapper.rs
```diff
@@ -112,6 +117,40 @@ impl WinitWindowWrapper {
...
+                // window.option_as_alt() is arguably the source of truth for this setting).
+                // Repeat this for 'ime', and then we wouldn't need the KeyboardSettings
+                // struct at all, and we wouldn't need the duplicate values stored in
+                // WinitWindowWrapper::{macos_option_is_meta, ime_enabled}.
+                //
+                // Come to think of it... do we really need to synchronize all settings once
+                // per video frame? It seems like we'd generally only need a synchronization
+                // once per "setting_changed" notification from nvim (and then only for one
+                // setting, not all of them)...?
+            }
```
fredizzimo, line 150: I think it would be cleaner if this was moved into it's own `get_option_as_alt_setting` function. 

And then called at synchronise settings, that way it can also be updated on the fly, which I don't think is possible with your current implementation.

I also don't think you need to call it here, you can just set it to `None`, and let the synchronise settings take care of actually setting the value.

I agree that using events would be a better approach, but our settings system currently does not support that. So, I guess we will have to live with `synchronise_settings`  approach for now. 

I don't believe that the order matters, since we want to deprecate the old `macos_option_is_meta` option. So it's really a user fault if both options are set, so if the new one always overrides, I think that's ok.

src/window/settings.rs
fredizzimo: I don't think this is needed, we don't really need to call `set`, just read the values and use them. `ParseFromValue` is still needed though.

fredizzimo (changes requested):
The documentation in website/docs still need to be updated. Both to document the new option and deprecate the old one.

JamesWidman:
> The documentation in website/docs still need to be updated. Both to document the new option and deprecate the old one.

i'm not sure what to write for the "available since" line; e.g. the documentation for `neovide_input_macos_alt_is_meta` says:

```markdown
**Available since 0.10.**
```

(I would write `**Available since 0.11.3**` for the new option, but i don't know if that's what the next version number is going to be.)

fredizzimo:
@JamesWidman. We need to add some actual contributor documentation too, but we do have these instructions in the maintainer documentation https://neovide.dev/maintainer-cookbook.html#actually-releasing

> website/docs/*.md and update Unreleased yet to Available since $tag (where $tag is the tag name)

So, you need to tag it with `**Unreleased yet.**`

JamesWidman:
[sorry for the lag; i've been busy with several things that are *not* neovide, but i expect to push a new commit soon]

src/window/keyboard_manager.rs
```diff
@@ -66,6 +72,25 @@ impl KeyboardManager {
...
+                    self.meta_is_pressed = match ks.macos_option_key_is_meta.0 {
+                        OptionAsAlt::Both => self.modifiers.state().alt_key(),
+                        OptionAsAlt::OnlyLeft => {
+                            self.modifiers.lalt_state() == ModifiersKeyState::Pressed
+                        }
+                        OptionAsAlt::OnlyRight => {
+                            self.modifiers.ralt_state() == ModifiersKeyState::Pressed
+                        }
+                        OptionAsAlt::None => false,
+                    };
```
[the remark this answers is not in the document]
JamesWidman: > Hm.. shouldn't the winit `alt_key()` function take care of these checks?

`ModifiersState::alt_key()` is insensitive to whether the right or left <kbd>option</kbd> key is pressed; this is why we can use it when the user specifies "Both", but not when they specify "OnlyLeft" or "OnlyRight"; see also the comment above the definition of `ModifiersKeys` (in `winit/src/keyboard.rs`):

```
// The reason this API is even exposed is mostly to provide a way for users
// to treat modifiers differently based on their position, which is required
// on macOS due to their AltGr/Option situation.
```

src/window/window_wrapper.rs
```diff
@@ -112,6 +117,40 @@ impl WinitWindowWrapper {
 
         let ime_enabled = { SETTINGS.get::<KeyboardSettings>().ime };
 
+        #[cfg(target_os = "macos")]
+        let mut ks = SETTINGS.get::<KeyboardSettings>();
```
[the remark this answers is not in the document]
JamesWidman: but `ks` is also used further down (at the end of the initializer for `let mut window_wrapper`, to initialize `WinitWindowWrapper::macos_option_is_meta`)

src/window/keyboard_manager.rs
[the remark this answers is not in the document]
fredizzimo: Yes, it looks like you are correct. I just thought that this is some abstraction they should provide, but it does not seems like they do.

src/window/window_wrapper.rs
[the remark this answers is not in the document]
fredizzimo: Ah, yes I missed that.

9mm:
Hey just checking on the status of this one, I'm stoked for this one to get put into main now that the one hard issue is no more of an issue it seems (the issue linked above about <A-n>)

src/window/window_wrapper.rs
fredizzimo: We do have a settings listener now. So I think that should clean this up.

fredizzimo:
@JamesWidman, are you still planning to work on this, or should we find someone else to finish it. 

Also note, that we have `*SettingsChanged` events now, so most of the custom code can be removed.
https://github.com/neovide/neovide/blob/879172ec071a9d02279a50aaeb38072fe258efe7/src/settings/mod.rs#L188-L195

And 
https://github.com/neovide/neovide/blob/879172ec071a9d02279a50aaeb38072fe258efe7/src/window/window_wrapper.rs#L196-L231

9mm:
What still has to be done on this one? Ive spent the last 4 weeks learning rust in extreme detail for another project so I'm actually ready (at least I think) to make better commits now 😎

fredizzimo:
@9mm, thank you. 

The first step is to rebase this on the latest master to get it to compile again.

Then most of the code in `settings/mod.rs` and `window_wrapper` can be removed and replaced with the `SettingsChanged` event.

fredizzimo:
Thank you @JamesWidman, I'm closing this in favour of:
* https://github.com/neovide/neovide/pull/2486

PR #2055, opened 2023-09-27
Custom Menu Bar

9mm:
**Currently WIP**

This PR seeks to solve the problem of needing certain native hotkeys, such as `⌘-Q` on mac to quit. Up until now, winit includes a default menu which handles this functionality, but it's impossible to capture that event, so we can't properly shut down neovide down and simultaneously acknolwedge config options like `vim.g.neovide_confirm_quit`.

So instead, we create our own menu using the `menubar` crate.

I'm new to rust so please feel free to correct anything and everything aggressively as I chip away at these issues. You wont offend me.

Todo:
- [ ] implement for `windows` and `linux` targets. all items below applies to `macos`
- [ ] on mac, use the `⌘` key instead of the globe key for all the default system hotkeys
- [ ] implement basic core functionality:
  - [ ] about
  - [ ] hide / hide others
  - [ ] quit
- [ ] implement window menu functionality
  - [ ] implement minimize which doesn't seem to be in the default
- [ ] implement help menu, similar to macvim:
  - [ ] neovide website
  - [ ] I'm open to others... below is a screenshot of macvim
- [ ] change `neovide` to `Neovide` proper capitalization on Mac (maybe related to the process name?)

image

- Feature

The breaking changes to my knowledge is that because some portions of the menu are now being handled natively, certain people's keymaps may need to be made sure they still work.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,6 +45,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "ahash"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2c99f64d1e06488f620f932677e24bc6e2897582980441ae90a671415bd7ec2f"
+dependencies = [
+ "cfg-if",
+ "once_cell",
+ "version_check",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -228,7 +239,16 @@ version = "0.1.0-beta.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0fa55741ee90902547802152aaf3f8e5248aab7e21468089560d4c8840561146"
 dependencies = [
- "objc-sys",
+ "objc-sys 0.2.0-beta.2",
+]
+
+[[package]]
+name = "block-sys"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2dd7cf50912cddc06dc5ea7c08c5e81c1b2c842a70d19def1848d54c586fed92"
+dependencies = [
+ "objc-sys 0.3.1",
 ]
 
 [[package]]
@@ -237,8 +257,18 @@ version = "0.2.0-alpha.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8dd9e63c1744f755c2f60332b88de39d341e5e86239014ad839bd71c106dec42"
 dependencies = [
- "block-sys",
- "objc2-encode",
+ "block-sys 0.1.0-beta.1",
+ "objc2-encode 2.0.0-pre.2",
+]
+
+[[package]]
+name = "block2"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "15b55663a85f33501257357e6421bb33e769d5c9ffb5ba0921c975a123e35e68"
+dependencies = [
+ "block-sys 0.2.0",
+ "objc2 0.4.1",
 ]
 
 [[package]]
@@ -899,7 +929,7 @@ dependencies = [
  "glutin_glx_sys",
  "glutin_wgl_sys",
  "libloading 0.7.4",
- "objc2",
+ "objc2 0.3.0-beta.3.patch-leaks.3",
  "once_cell",
  "raw-window-handle",
  "wayland-sys 0.30.1",
@@ -954,7 +984,7 @@ version = "0.12.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8a9ee70c43aaf417c914396645a0fa852624801b24ebb7ae78fe8272889ac888"
 dependencies = [
- "ahash",
+ "ahash 0.7.6",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


fredizzimo:
So far this looks good!

9mm:
Thanks! I had a super busy week but I was hoping to do some more here soon. I'm at a complex part because I need to now trigger those window events to winit, so I just need to study other code to see how to do that (like sending the hide vs minimize event from that other PR I can look at for code examples)

9mm:
Ill probably try to do this a little differently per his advice here https://github.com/madsmtm/menubar/issues/5

I think we can start with OSX menu only i suspect, and use the default menu for the others

9mm:
Ok I have looked at this quite a bit, I will revisit it later. My cat currently I found out is going to die so I have been super sad and not interested in doing anything. 😿

According to the issue above, he basically says swap out `MenuItemWrapper` with its icrate equivalent, however when I do that, all the types fail, so that leads to a bit of an unraveling situation where I can't use part of `menubar` (for example the `set_menubar` function which does a lot of work) and then also mix that with my own custom implementation of `NsMenuItem`. 

So next I checked out winit source code for creating the default menu: https://github.com/rust-windowing/winit/blob/af247eac0f62fc4a1dfb025470f809d3413536d9/src/platform_impl/macos/menu.rs

This actually seems like a good approach, the only problem is he references lots of files within `appkit` folder which aren't public. I don't know enough about rust but it seems weird to re-create 4-5 different files or functions that are already there, and duplicate them over to Neovide.

If this is an acceptable approach, to copy over all of these:

```
use super::appkit::{NSApp, NSEventModifierFlags, NSMenu, NSMenuItem};
```

Then should macos get its own folder or something?

Note: No need to answer these questions or spend too much time on it, more thinking out loud, as well as keeping  notes to myself for later

fredizzimo:
@9mm That's sad to hear :crying_cat_face: 

I converted this PR to draft, to indicate that it's not yet ready. But there's no rush take all the time you need, I don't think that many people are waiting for this fix, so it should be fine if it's delayed.

I think copying part of the code from winit and modifying it is a good approach, since winit is not planning to provide any customisation of that feature themselves. We could even extract the individual functionality into rpc functions that can be called and mapped inside Neovim for people who don't want to use the menu.

However, you should not need to copy those files, just part of the actual menu implementation in Winit. The same functionality can be imported through `use cocoa::appkit::{NSApp, NSEventModifierFlags, NSMenu, NSMenuItem};` in Neovide, se the `draw_background.rs` file for example.

9mm:
I know, it is super depressing :(

Thank you for changing it to a draft, I meant to do that as I didn't want to create clutter, sorry.

Thanks for showing me how to import those, that helps a lot. If I can just access those same files I should be able to do it pretty easily. I will try that out soon

9mm:
So I looked at this again, if I import `NSMenu` etc directly from cocoa I get argument errors etc, which appears that winit created sorts of wrapper functions, like this:

https://github.com/rust-windowing/winit/blob/af247eac0f62fc4a1dfb025470f809d3413536d9/src/platform_impl/macos/appkit/menu.rs

I don't know rust well enough, but IF you have not (not important if you dont), how would you take this winit menu creation code:

```rust
pub fn initialize() {
    let menubar = NSMenu::new();
    let app_menu_item = NSMenuItem::new();
    menubar.addItem(&app_menu_item);

    let app_menu = NSMenu::new();
    let process_name = NSProcessInfo::processInfo().processName();

    // About menu item
    let about_item_title = ns_string!("About ").stringByAppendingString(&process_name);
    let about_item = menu_item(&about_item_title, sel!(orderFrontStandardAboutPanel:), None);
```

And rewrite it w/o using their wrapper (and just using the direct import from `coacoa`)? (or even a very small part of it)

If the answer is too much no worries

fredizzimo:
@9mm. I think this was closed by accident, because the target branch was deleted. I'm not sure if you can re-open it and change the target branch (I can't). Otherwise, I guess you have to create a new PR.

9mm:
@fredizzimo ugh youre right. last night i was trying to rebuild main so i could use newest version and there were like 3 copies of main. i didnt want to use the github app just so i could better learn git in the context of the way open source and PRs work. i just nuked it and started over and apparently nuked this PR.

If you have any thoughts about my comment above ill see if i can change ttarget branch.

fredizzimo:
@9mm. I'm not sure how much I can help you remotely. But I can create a branch for you as a starting point, with your changes rebased onto the latest master. I'll do that a bit later today, if you want me to do it.

Otherwise, if you want to try it yourself, I think you need to use the `git rebase <from> <to> --onto main´ variant and resolve the potential commits.

9mm:
@fredizzimo oh sorry, don't worry about the git stuff I will figure that out.

I meant my question above about the imports not actually seeming to work the same way.

Ultimately from my limited POV it looks like duplicating a lot of code over is necessary because the ones being imported still have to be rewritten (in the same way he did in the original library)

fredizzimo:
@9mm those are indeed private in winit, but the menubar crate exposes an api in a slightly different way https://github.com/madsmtm/menubar/blob/master/src/appkit/menu.rs

I guess you should be able to use it a bit like the example https://github.com/madsmtm/menubar/blob/master/examples/hello_world.rs

9mm:
Hmm so the problem with that way was what I mentioned here:

https://github.com/madsmtm/menubar/issues/5

You can see his replies there, thats why I was like... instead of using menubar I can just use winit directly in an attempt to simply recreate it how they do here, but with my own changes:

https://github.com/rust-windowing/winit/blob/af247eac0f62fc4a1dfb025470f809d3413536d9/src/platform_impl/macos/menu.rs

But yes, then its a bit of a chicken and egg because now reverting back to this, it's not public so it seems to be re-creating everything yet again which basically brings it back to menu bar

fredizzimo:
Yes, maybe we need to wait until the menubar stuff is ready to be used and find another workaround for the original problem. Perhaps dealing with the keypress directly.

9mm:
Ok thanks, for now I suppose I will wait so i dont take away from more important issues asking for help etc. The main benefit of the menu bar is that command+Q will work properly and listen to the setting about confirming before quit.

PR #2056, opened 2023-09-28
fix: run :edit command with nvim_cmd to escape special characters

### LINKED ISSUE #1983: drag-n-drop to open file that has "()" in its PATH
**Describe the bug**

On Windows, I cannot drag-n-drop files with paths like this:

`C:/(this-path)/file.md`

Windows native notepad can open the file successfully.

This is similar but different to the issue:  https://github.com/neovide/neovide/issues/1608

**To Reproduce**
1. Create a folder structure containing "()" in the folder name
2. Create a file in the folder
3. Drag-n-drop to open that file from neovide

**Expected behavior**
The file can be opened successfully

 - OS: Windows
 - Neovide Version: 0.11.0
 - Neovim Version 0.9.1

nothing

ZyX-II:
I tested this commit by dragging and dropping file named `W:\tmp\neovide\(abc)\example#abc%def.txt` and it was opened successfully.

Fixes #1983

- Fix

- No

### DIFF
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -152,21 +152,9 @@ impl ParallelCommand {
             ParallelCommand::FileDrop(path) => {
                 nvim.cmd(
                     vec![
-                        (
-                            "cmd".into(),
-                            "edit".into(),
-                        ),
-                        (
-                            "magic".into(),
-                            vec![(
-                                "file".into(),
-                                false.into(),
-                            )].into(),
-                        ),
-                        (
-                            "args".into(),
-                            vec![Value::from(path)].into(),
-                        ),
+                        ("cmd".into(), "edit".into()),
+                        ("magic".into(), vec![("file".into(), false.into())].into()),
+                        ("args".into(), vec![Value::from(path)].into()),
                     ],
                     vec![],
                 )
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -4,7 +4,7 @@ use std::sync::Arc;
 use log::error;
 use log::trace;
 
-use nvim_rs::{call_args, rpc::model::IntoVal, Neovim};
+use nvim_rs::{call_args, rpc::model::IntoVal, Neovim, Value};
 use tokio::sync::mpsc::unbounded_channel;
 
 #[cfg(windows)]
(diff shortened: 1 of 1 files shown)


ZyX-II:
Wondering whether somebody knows less verbose way to spell

    :call nvim_cmd({'cmd': 'edit', 'args': [file], 'magic': {'file': v:false}}, {})

in Rust. I only found [Neovim::cmd](https://docs.rs/nvim-rs/0.6.0/nvim_rs/neovim/struct.Neovim.html#method.cmd) which removes one nesting level.

ZyX-II:
Note about `magic` and missing setting `bar` to false: I have tested running

    call nvim_cmd({'cmd': 'edit', 'args': ['abc#def|ghi'], 'magic': {'file': v:false}}, {})

and

    call nvim_cmd({'cmd': 'edit', 'args': ['abc#def|ghi'], 'magic': {'file': v:false, 'bar': v:true}}, {})

and it looks like `bar` is ignored even though it is not mentioned in documentation (both commands open `abc#def|ghi`). Makes sense though.

fredizzimo:
I think it would be less verbose to use `nvim.command` instead of `nvim.cmd` and use string formatting to build the command.

You could also use `nvim.execute_lua` with a single arg the filename. In that case no string formatting is needed, you just write the lua code.

ZyX-II:
@fredizzimo `nvim.command` requires proper escaping and it is exactly it being used without escaping is what #1983 is about. As for `lua`: using

```
                nvim.exec_lua(
                    "vim.cmd.edit{ args = { ... }, magic = { file = false } }",
                    vec![Value::String(path.to_owned().into())],
                )
                .await
                .ok();
```

works and is certainly shorter, but I do not like extra roundtrips. Also `rg lua` finds basically three occurrences of neovide using lua: as an implementation detail of `<CR>` shortcut of `DisplayAvailableFonts` commands, in clipboard provider implementation and in intro message implementation.

---

There is one additional thing though: while determining what is wrong with faulty variants of the above expression I found out that expressions executed like this do not show errors, neither directly from them, nor from my autocommand which automatically switches to the directory of the entered file (and fails due to the deficiency of `fnameescape()`: `fnameescape('W:\tmp\neovide\(abc)')` yields `W:\tmp\neovide\(abc)` which is treated as `W:\tmp\neovide(abc)` which does not exist). In my opinion this is wrong.

Note that errors do not show in log either. Not that I consider having them only in a log a good thing, in my opinion it makes sense to display errors in main UI.

fredizzimo:
This will add a bit more lua https://github.com/neovide/neovide/pull/2062. I think the absolutely cleanest solution would be to add a new function there that can be called through RPC from neovide.

For the error handling:
You need to deal with the errors inside the function you create, so that the error can be reported to neovim instead of neovide. All the `nvim.*` calls return the error back to neovide, which is not what we want here, since we can't do anything with them other than appending them to the log. The `.ok()` you have at the end just tells it to ignore any errors.

ZyX-II:
I think it makes sense to merge this as-is, then (not) change how errors are dealt with later and basically for everything in ui_commands: I just realised that errors for almost every command there can appear and are silenced: it is rather common to have autocommands triggering at commands like `split`, but I just remembered that there is `OptionSet` autocommand which can trigger for `setlocal` commands too.

Also, possibly not with function in lua, but with replacement for `ok` on Rust side which will call something like `nvim.err_write`.

MultisampledNight:
Looks like we don't even support drag-and-drop on Wayland atm, but tested through Xwayland and it works perfectly fine, thanks!

I'll take care of making the callback a bit more readable with ad-hoc placed closures, then I'd merge this.

MultisampledNight (approved):

MultisampledNight:
Turns out that wasn't even necessary, [`nvim_rs::Value`] handily offers `From`/`Into` implementations for its contained values.

Thank you! Chooooooooooo chooooooooooooo :steam_locomotive: 

[`nvim_rs::Value`]: https://docs.rs/nvim-rs/0.6.0/nvim_rs/enum.Value.html

PR #2057, opened 2023-09-29
Make mouse more responsive

ad-chaos:
- Fix
supersedes #1511 

- No

### DIFF
--- a/src/window/mouse_manager.rs
+++ b/src/window/mouse_manager.rs
@@ -67,6 +67,29 @@ struct TouchTrace {
     left_deadzone_once: bool,
 }
 
+struct ScrolledPixels {
+    x: f32,
+    y: f32,
+}
+
+impl ScrolledPixels {
+    fn new() -> ScrolledPixels {
+        ScrolledPixels { x: 0., y: 0. }
+    }
+
+    fn reset(&mut self) {
+        self.x = 0.;
+        self.y = 0.;
+    }
+
+    fn to_lines(&self, font_width: u64, font_height: u64) -> (f32, f32) {
+        (
+            self.x.round() / font_width as f32,
+            self.y.round() / font_height as f32,
+        )
+    }
+}
+
 pub struct MouseManager {
     dragging: Option<String>,
     drag_position: PhysicalPosition<u32>,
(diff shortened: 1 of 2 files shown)


fredizzimo (changes requested):
This scrolls 6 lines instead of 3 for each tick on Windows when `mousescroll=ver:3,hor:6` and 2 lines when `mousescroll=ver:1,hor:1`.

It worked perfectly before.

ad-chaos:
I was blocked from discord...

I am not even kidding, I have no idea why and what, I was just typing a response to your comment and suddently kabaam, I get a white screen :((

Till I try figuring that out, Do you agree with the approach taken here? Or do you believe its a winit bug?

fredizzimo:
@ad-chaos. I think discord is partially down or something, my message also got the red color. There's no hurry to figure it out, we can continue when it's up. But we need to find the root cause of the issue.

But floating points are problematic due to drift, on the other platforms as well, so here's what I wrote in discord
>  think at least on Windows we should convert all the values we get from winit to fractions of 120 and do all the calculations as integer. On Wayland and X11, at least the way winit handles it, the input is integers, should the values could multipled by 120 to keep the same logic. 
>
> The macOS implementation is a mystery though, since a single value of 0.1 does not make any sense. A tick should always add up to 1, or close to it allowing floating point errors.

ad-chaos:
I see, I'll read up a bit more on mouse handling till then

src/window/mouse_manager.rs
```diff
@@ -67,6 +67,29 @@ struct TouchTrace {
...
 }
 
+struct ScrolledPixels {
+    x: f32,
+    y: f32,
+}
+
+impl ScrolledPixels {
+    fn new() -> ScrolledPixels {
+        ScrolledPixels { x: 0., y: 0. }
```
MultisampledNight, line 77: Could maybe just be `Default::default`.

src/window/mouse_manager.rs
```diff
@@ -67,6 +67,29 @@ struct TouchTrace {
...
+    }
+
+    fn reset(&mut self) {
+        self.x = 0.;
+        self.y = 0.;
```
MultisampledNight, line 82:
```suggestion
        *self = Self::new();
```

doesn't require updating another place for theoretical future fields, defaults or type changes.

Kethku:
I'm going to close this for now. @ad-chaos feel free to reopen if you get back to this

PR #2061, opened 2023-10-01
fix!: window not showing up

fredizzimo:
* Fixes the window not showing up when there are error messages or `lazy.nvim` is updating for example. This was reported by @9mm here https://github.com/neovide/neovide/pull/2035#issuecomment-1732360990
* The initial size can now be specified by `--grid`, if a size is given then that's used, otherwise the grid will be read from `init.vim/lua`
* The arguments `--size`, `--maximized` and `--grid` are now mutually exclusive

The implementation is a bit hackier than I would want and shows the window when `win_viewport` or `grid_line` is recieved. This is due to the neovim limitations discussed here:
* https://github.com/neovim/neovim/issues/25377

And the main reason for supporting an empty `--grid` argument is because of this:
* https://github.com/neovim/neovim/issues/25378

**NOTE:** Includes the changes from, so they should be reviewed and merged first
* https://github.com/neovide/neovide/pull/2035

* Yes, `--geometry` has been renamed to `--grid`. But this argument was not documented and does not work properly in our release versions.

### DIFF
--- a/src/bridge/handler.rs
+++ b/src/bridge/handler.rs
@@ -96,9 +96,6 @@ impl Handler for NeovimHandler {
                     EVENT_AGGREGATOR.send(WindowCommand::Lines(lines));
                 }
             }
-            "neovide.ui_ready" => {
-                EVENT_AGGREGATOR.send(WindowCommand::UIReady);
-            }
             #[cfg(windows)]
             "neovide.register_right_click" => {
                 EVENT_AGGREGATOR.send(UiCommand::Parallel(ParallelCommand::RegisterRightClick));
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -15,8 +15,8 @@ use tokio::{
 };
 
 use crate::{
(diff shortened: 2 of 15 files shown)


9mm:
Hey this is sweet. Is there a PR or branch which soon will combine this + improve render loop + update winit?  I'm assuming its all eventually going to shuffle into improve render loop? I could then use that branch for daily testing to find regressions. Maybe it is obvious looking at the commits but my git-fu is not that adept when comparing all the branching, forced pushes, interactive rebases etc (I usually work alone so this is all a bit new)

fredizzimo:
@9mm this one includes all the changes, and will be merged to the render branch at some point.

Well, this one includes one additional commit, but that’s just refactoring and no functional changes https://github.com/neovide/neovide/pull/2062

MultisampledNight (approved):
Best viewed through `git diff improve-render-loop fredizzimo/fsundvik/fix-window-positioning`.

PR #2062, opened 2023-10-01
chore: Move the initial setup to lua

fredizzimo:
Moves the setup code to lua, which makes the code easier to read and it also makes it slightly faster.

**NOTE:** Depends on https://github.com/neovide/neovide/pull/2061 and should be reviewed and merged after that

* No

### DIFF
--- a/lua/init.lua
+++ b/lua/init.lua
@@ -1,0 +1,91 @@
+---@class Args
+---@field neovide_channel_id integer
+---@field register_clipboard boolean
+---@field register_right_click boolean
+---@field enable_focus_command boolean
+
+---@type Args
+local args = ...
+
+
+vim.g.neovide_channel_id = args.neovide_channel_id
+
+-- Set some basic rendering options.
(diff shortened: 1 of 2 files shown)


lua/init.lua
```diff
@@ -0,0 +1,86 @@
...
+    once = false,
+    nested = true,
+    callback = function()
+        rpcnotify("neovide.columns", tonumber(vim.v.option_new))
+    end
+})
+
+vim.api.nvim_create_autocmd({ "OptionSet" }, {
+    pattern = "lines",
+    once = false,
```
max397574, line 71: I'd remove default values

fredizzimo: I did not remove it since, I prefer to make things explicit, and people might not know what the defaults are without having to look at the documentation.

MultisampledNight (approved):

MultisampledNight:
Thank you! Choooo choooooooooooo :steam_locomotive:

Issue #1859, opened 2023-05-04
Neovide doesn't opens on MacOS


emakei:
**Describe the bug**
Nothing happens when I'm try to open app from Finder or panel. But if use command "open Neovide.app" from shell then works fine

**To Reproduce**
1. Install or build from source app
2. Bundle if builded from sources
3. Try open app

**Expected behavior**
App should opens as regular application

**Desktop (please complete the following information):**
 - OS: 13.3.1 (22E261)
 - Neovide Version 0.10.4
 - Neovim Version 0.9.0

**Additional context**
Ctrl+Click doesn't help

dashkb:
This is affecting me as of yesterday after a `brew upgrade` - have been using Neovide as daily driver successfully for months, but also tried the solution(s) in #1175 just in case. Works when run as `neovide` or `open /Applications/Neovide.app` from shell.

How can I enable/view logs when launching from Finder/Alfred/etc?

MPWrobel:
If it runs when you open it from shell, it's most likely caused by the `PATH` variable. Be careful where you add the path containing `nvim` executable to the `PATH`. When opened from shell, the process inherits the variable from shell, so it works just fine. E.g. neovide won't work if you add it in `.zshrc`, but it will work fine if you add it in `.zshenv`. That's because .zshrc is read only by interactive shells.

bnwa:
Getting the same issue, however I use `fish`, not `zsh`, and there's no difference in setting the homebrew path in `config.fish` or under `conf.d`

bnwa:
Also exported `NEOVIM_BIN` directly as a global var to avoid the path lookup, same issue

netfizz-se:
> If it runs when you open it from shell, it's most likely caused by the `PATH` variable. Be careful where you add the path containing `nvim` executable to the `PATH`. When opened from shell, the process inherits the variable from shell, so it works just fine. E.g. neovide won't work if you add it in `.zshrc`, but it will work fine if you add it in `.zshenv`. That's because .zshrc is read only by interactive shells.

Great tip and good best practice, corrected to match this (correct) dotfiles configuration and works great.  Thanks!

dashkb:
@bnwa and other `fish` users - looks like Finder is using `zsh`'s environment, irrespective of configured login shell. Adding `/opt/homebrew/bin` to `.zshenv` has everything working again for me, even though I've only ever used `zsh` on this machine for the minute it took to get `fish`. :shrug:

[...]

PR #2063, opened 2023-10-05
Tokenize command arguments in `$NEOVIM_BIN` (fix #2060)

### LINKED ISSUE #2060: ERROR: NEOVIM_BIN was not found.
**Describe the bug**
Neovide reports an error that NEOVIM_BIN has not been found , but the path shown
is valid.
```
ERROR: NEOVIM_BIN='/usr/bin/env nvim' was not found.
```
```
$ /usr/bin/env nvim    
Successfully launches nvim
```

**To Reproduce**
Start neovide as normal `$ neovide`

**Expected behavior**
Neovide to start

 - OS: NixOS 23.11.20230925.6500b45 (Tapir) x86_64
 - neovide 0.11.2
 - Neovim Version
    ```    
    NVIM v0.10.0-dev-e46f5aa
    Build type: Release
    LuaJIT 2.1.1693350652
    Run "nvim -V1 -v" for more info
    ```

```
❯ neovide --log
ERROR: NEOVIM_BIN='/usr/bin/env nvim' was not found.
```

rhysd:
- Fix #2060

- No

I confirmed this fix on Ubuntu 22.04 and macOS 13.6.

### DIFF
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -32,21 +32,8 @@ fn set_windows_creation_flags(cmd: &mut TokioCommand) {
 
 fn build_nvim_cmd() -> TokioCommand {
     if let Some(cmdline) = SETTINGS.get::<CmdLineSettings>().neovim_bin {
-        if let Some((bin, args)) = shlex::split(&cmdline)
-            .filter(|tokens| !tokens.is_empty())
-            .map(|mut tokens| (tokens.remove(0), tokens))
-        {
-            // if neovim_bin contains a path separator, then try to launch it directly
-            // otherwise use which to find the fully path
-            if bin.contains('/') || bin.contains('\\') {
-                if neovim_ok(&bin, &args) {
-                    return build_nvim_cmd_with_args(bin, args);
-                }
-            } else if let Some(bin) = platform_which(&bin) {
-                if neovim_ok(&bin, &args) {
-                    return build_nvim_cmd_with_args(bin, args);
-                }
-            }
+        if let Some((bin, args)) = lex_nvim_cmdline(&cmdline) {
+            return build_nvim_cmd_with_args(bin, args);
         }
 
         eprintln!("ERROR: NEOVIM_BIN='{}' was not found.", cmdline);
@@ -125,6 +112,19 @@ fn neovim_ok(bin: &str, args: &[String]) -> bool {
         }
     }
     false
+}
+
+fn lex_nvim_cmdline(cmdline: &str) -> Option<(String, Vec<String>)> {
+    let mut tokens = shlex::split(cmdline).filter(|t| !t.is_empty())?;
+    let (mut bin, args) = (tokens.remove(0), tokens);
+
+    // if neovim_bin contains a path separator, then try to launch it directly
+    // otherwise use which to find the fully path
+    if !bin.contains('/') && !bin.contains('\\') {
+        bin = platform_which(&bin)?;
+    }
+
+    neovim_ok(&bin, &args).then_some((bin, args))
 }
 
 fn platform_which(bin: &str) -> Option<String> {
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -31,28 +31,29 @@ fn set_windows_creation_flags(cmd: &mut TokioCommand) {
 }
 
 fn build_nvim_cmd() -> TokioCommand {
-    if let Some(path) = SETTINGS.get::<CmdLineSettings>().neovim_bin {
-        if let Some(path) = shlex::split(&path) {
-            if let Some((bin, args)) = path.split_first() {
-                // if neovim_bin contains a path separator, then try to launch it directly
-                // otherwise use which to find the fully path
-                if bin.contains('/') || bin.contains('\\') {
-                    if neovim_ok(bin, args) {
-                        return build_nvim_cmd_with_args(bin, args);
-                    }
-                } else if let Some(bin) = platform_which(bin) {
-                    if neovim_ok(&bin, args) {
-                        return build_nvim_cmd_with_args(&bin, args);
-                    }
+    if let Some(cmdline) = SETTINGS.get::<CmdLineSettings>().neovim_bin {
+        if let Some((bin, args)) = shlex::split(&cmdline)
+            .filter(|tokens| !tokens.is_empty())
+            .map(|mut tokens| (tokens.remove(0), tokens))
+        {
+            // if neovim_bin contains a path separator, then try to launch it directly
+            // otherwise use which to find the fully path
+            if bin.contains('/') || bin.contains('\\') {
+                if neovim_ok(&bin, &args) {
+                    return build_nvim_cmd_with_args(bin, args);
+                }
+            } else if let Some(bin) = platform_which(&bin) {
+                if neovim_ok(&bin, &args) {
+                    return build_nvim_cmd_with_args(bin, args);
                 }
             }
         }
 
-        eprintln!("ERROR: NEOVIM_BIN='{}' was not found.", path);
+        eprintln!("ERROR: NEOVIM_BIN='{}' was not found.", cmdline);
         std::process::exit(1);
     } else if let Some(path) = platform_which("nvim") {
         if neovim_ok(&path, &[]) {
-            return build_nvim_cmd_with_args(&path, &[]);
+            return build_nvim_cmd_with_args(path, vec![]);
         }
     }
     eprintln!("ERROR: nvim not found!");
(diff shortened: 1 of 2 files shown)


src/bridge/command.rs
```diff
@@ -32,23 +32,27 @@ fn set_windows_creation_flags(cmd: &mut TokioCommand) {
...
-        // otherwise use which to find the fully path
-        if path.contains('/') || path.contains('\\') {
-            if neovim_ok(&path) {
-                return build_nvim_cmd_with_args(&path);
-            }
-        } else if let Some(path) = platform_which(&path) {
-            if neovim_ok(&path) {
-                return build_nvim_cmd_with_args(&path);
+        if let Some(path) = shlex::split(&path) {
+            if let Some((bin, args)) = path.split_first() {
```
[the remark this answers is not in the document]
rhysd: No. `shlex::split` returns only when it failed to tokenize the input. So it can return an empty array when the input is a valid empty command line such as `""`. This check is necessary.

```rust
dbg!(shlex::split("'foo")); // -> None
dbg!(shlex::split("")); // -> Some([])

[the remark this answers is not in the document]
Comment (rhysd) [reply src/bridge/command.rs:36]: I understood your intention. I'll try to reduce indentation. I also noticed I can remove one redundant clone so I'll remove it as well.

File: src/bridge/command.rs [hunk 2]
```diff
@@ -31,24 +31,29 @@ fn set_windows_creation_flags(cmd: &mut TokioCommand) {
...
-            if neovim_ok(&path) {
-                return build_nvim_cmd_with_args(&path);
-            }
-        } else if let Some(path) = platform_which(&path) {
-            if neovim_ok(&path) {
-                return build_nvim_cmd_with_args(&path);
+    if let Some(cmdline) = SETTINGS.get::<CmdLineSettings>().neovim_bin {
+        if let Some((bin, args)) = shlex::split(&cmdline)
+            .filter(|tokens| !tokens.is_empty())
+            .map(|mut tokens| (tokens.remove(0), tokens))
```
rhysd, line 37: I changed to split the tokens into the first argument and the rest within this method to reduce the indent. As a bonus, I could remove one redundant clone comparing to my previous implementation since now it modifies the `Vec` instance directly.

src/bridge/command.rs
```diff
@@ -176,8 +181,8 @@ fn nvim_cmd_impl(bin: &str, args: &[String]) -> TokioCommand {
     }
 }
 
-fn build_nvim_cmd_with_args(bin: &str) -> TokioCommand {
-    let mut args = vec!["--embed".to_string()];
+fn build_nvim_cmd_with_args(bin: String, mut args: Vec<String>) -> TokioCommand {
```
rhysd, line 184: I changed `&[String]` to `Vec<String>` so that the `Vec` instance can be reused. It can remove one redundant clone here.

src/bridge/command.rs
```diff
@@ -147,27 +154,25 @@ fn platform_which(bin: &str) -> Option<String> {
...
+fn nvim_cmd_impl(bin: String, mut args: Vec<String>) -> TokioCommand {
     let shell = env::var("SHELL").unwrap_or_else(|_| "/bin/sh".to_string());
+    args.insert(0, bin);
+    let args = shlex::join(args.iter().map(String::as_str));
     let mut cmd = TokioCommand::new(shell);
-    let args_str = args
-        .iter()
-        .map(|arg| shlex::quote(arg))
-        .collect::<Vec<_>>()
-        .join(" ");
```
rhysd, line 157: I noticed that this can be simplified with `shlex::join`.

src/bridge/command.rs
```diff
@@ -147,27 +154,25 @@ fn platform_which(bin: &str) -> Option<String> {
 }
 
 #[cfg(target_os = "macos")]
-fn nvim_cmd_impl(bin: &str, args: &[String]) -> TokioCommand {
+fn nvim_cmd_impl(bin: String, mut args: Vec<String>) -> TokioCommand {
     let shell = env::var("SHELL").unwrap_or_else(|_| "/bin/sh".to_string());
+    args.insert(0, bin);
```
rhysd, line 159: By inserting the executable path into the head of `args`, I could avoid `format!("{} {}", bin, args)` later. By this, one redundant clone was removed.

src/bridge/command.rs
```diff
@@ -147,27 +154,25 @@ fn platform_which(bin: &str) -> Option<String> {
...
 }
 
 #[cfg(not(target_os = "macos"))]
-fn nvim_cmd_impl(bin: &str, args: &[String]) -> TokioCommand {
+fn nvim_cmd_impl(bin: String, mut args: Vec<String>) -> TokioCommand {
     if cfg!(target_os = "windows") && SETTINGS.get::<CmdLineSettings>().wsl {
+        args.insert(0, bin);
         let mut cmd = TokioCommand::new("wsl");
-        cmd.args(["$SHELL", "-lc", &format!("{} {}", bin, args.join(" "))]);
+        cmd.args(["$SHELL", "-lc", &args.join(" ")]);
```
rhysd, line 175: Ditto. One redundant clone was removed here.

rhysd (commented):
@fredizzimo Thanks for your review. I simplified my changes and removed 4 redundant clones. Would you review again?

rhysd:
BTW I got interested in why it wraps the command execution with a shell on macOS though it doesn't do it on Linux. Does executing the command directly on macOS cause any issue?

https://github.com/neovide/neovide/blob/a8cec7e16f9f8bd2e3282a9c13928b585d5d596f/src/bridge/command.rs#L149-L164

rhysd:
@fredizzimo I have another idea to split off the logic to lex Neovim command and args into a new function. Thanks to early returns, it can avoid nested blocks. Do you prefer it?

```diff
diff --git a/src/bridge/command.rs b/src/bridge/command.rs
index 0ec0cff..d4b760e 100644
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -32,21 +32,8 @@ fn set_windows_creation_flags(cmd: &mut TokioCommand) {
 
 fn build_nvim_cmd() -> TokioCommand {
     if let Some(cmdline) = SETTINGS.get::<CmdLineSettings>().neovim_bin {
-        if let Some((bin, args)) = shlex::split(&cmdline)
-            .filter(|tokens| !tokens.is_empty())
-            .map(|mut tokens| (tokens.remove(0), tokens))
-        {
-            // if neovim_bin contains a path separator, then try to launch it directly
-            // otherwise use which to find the fully path
-            if bin.contains('/') || bin.contains('\\') {
-                if neovim_ok(&bin, &args) {
-                    return build_nvim_cmd_with_args(bin, args);
-                }
-            } else if let Some(bin) = platform_which(&bin) {
-                if neovim_ok(&bin, &args) {
-                    return build_nvim_cmd_with_args(bin, args);
-                }
-            }
+        if let Some((bin, args)) = lex_nvim_cmdline(&cmdline) {
+            return build_nvim_cmd_with_args(bin, args);
         }
 
         eprintln!("ERROR: NEOVIM_BIN='{}' was not found.", cmdline);
@@ -127,6 +114,27 @@ fn neovim_ok(bin: &str, args: &[String]) -> bool {
     false
 }
 
+fn lex_nvim_cmdline(cmdline: &str) -> Option<(String, Vec<String>)> {
+    let Some(mut tokens) = shlex::split(cmdline).filter(|t| !t.is_empty()) else {
+        return None;
+    };
+    let (bin, args) = (tokens.remove(0), tokens);
+
+    // if neovim_bin contains a path separator, then try to launch it directly
+    // otherwise use which to find the fully path
+    if bin.contains('/') || bin.contains('\\') {
+        if neovim_ok(&bin, &args) {
+            return Some((bin, args));
+        }
+    } else if let Some(bin) = platform_which(&bin) {
+        if neovim_ok(&bin, &args) {
+            return Some((bin, args));
+        }
+    }
+
+    None
+}
+
 fn platform_which(bin: &str) -> Option<String> {
     let is_wsl = SETTINGS.get::<CmdLineSettings>().wsl;
 
```

src/bridge/command.rs
fredizzimo: Nice, yes this looks good to me.

fredizzimo (approved):
I'm happy with the code now, but I have not tested it yet, so I'm not merging until that's done.

fredizzimo:
> @fredizzimo BTW I got interested in why it wraps the command execution with a shell on macOS though it doesn't do it on Linux. Does executing the command directly on macOS cause any issue?
> 
> https://github.com/neovide/neovide/blob/a8cec7e16f9f8bd2e3282a9c13928b585d5d596f/src/bridge/command.rs#L149-L164

macOS does some strange things when not launched from a shell. and that were multiple attempts to fix it, but I don't think it's enough, because people still report problems like https://github.com/neovide/neovide/issues/1859. We need a better solution for this, but unfortunately, I don't own a mac, so I have no idea how it's supposed to work, and what's the correct way to fix it is.

Here are some of those attempts:
* https://github.com/neovide/neovide/pull/1832
* https://github.com/neovide/neovide/pull/1758
* https://github.com/neovide/neovide/pull/309

rhysd:
@fredizzimo Thank you for the answer. I understood the reason.

When starting an application by clicking `Foo.app` in Finder or clicking an application icon in LaunchPad, it is intentional that no environment variable is set. This is the same as `MacVim.app` (by opening MacVim from LaunchPad and checking `:echo $PATH`).

Users need to set `$PATH` in their config files (e.g. `.gvimrc`). However this workaround would be reasonable as well for usability.

fredizzimo:
Thank you. I did some additional tests with WSL on Windows and everything seems to work fine.

rhysd:
@fredizzimo Thank you. It's very helpful since I don't have those environments.

Issue #2069, opened 2023-10-07
Docs zoom function returns error


Slushee-a:
## Bug explanation:
I've always used the [scale shortcut function from the docs](https://neovide.dev/faq.html?highlight=scale#how-can-i-dynamically-change-the-scale-at-runtime) (the lua one) and it's always worked.

It suddenly stopped working and displays this message when I press either the zoom in or out shortcuts:
```
Error detected while processing function NeovideNotifyscale_factorChanged:
line 1:
E121: Undefined variable: g:neovide_channel_id
```

#### I have:
 - Done a full system upgrade (native package manager, flatpak, rustup, cargo, pip3, packer.nvim)
 - Removed all old cached packages and orphans
 - Recompiled neovide
 - Reinstalled neovim
 - Rebooted
 
And I still get the same error.

## To reproduce
1. Add the lua runtime scaling shortcut from [the docs](https://neovide.dev/faq.html?highlight=scale#how-can-i-dynamically-change-the-scale-at-runtime) to the lua nvim shortcuts config file
2. Open neovide. Error occurs with and without `--multigrid`
3. Use the defined shortcuts (`Ctrl` + `-`) and (`Ctrl` + `=`)

## Expected behavior
The function from the docs working as expected, zooming in when (`Ctrl` + `=`) and out when (`Ctrl` + `-`) is pressed instead of showing an error.

## Screenshots
> ![image](https://github.com/neovide/neovide/assets/55996847/dd6d8d71-a27f-4bce-b7a4-de16c8281463)
> What pops up after one of the zoom shortcuts is pressed.

## Desktop
 - **OS:** Void Linux (x86_64, glibc) Kernel 6.5.6_1
 - **Neovide Version:** 0.11.2
 - **Neovim Version:** 0.9.2

## Log
Full log can be seen here: https://gitlab.com/-/snippets/3607601

fredizzimo:
Hm. are you using the release version or build from master? 

Because this was changed yesterday here https://github.com/neovide/neovide/pull/2067. I don't have that much time to investigate deeper right now, but maybe @mopp has some ideas?

Slushee-a:
I used `cargo install --git https://github.com/neovide/neovide --force` for the reinstall, which I assume uses the main branch.

mopp:
oops, I reproduced the same error by inputting 

```vim
:let g:neovide_channel_id = 1.1
```

on command line :sob:

fredizzimo:
[...]

PR #2070, opened 2023-10-08
set neovide_channel_id always

mopp:

- Fix

- No

## Overview

The variable `neovide_channel_id` was NOT initialized in some cases.
This PR sets the variable when neovide gets the channel.

It resolves https://github.com/neovide/neovide/issues/2069


fredizzimo:
Thank you for fixing this.

PR #2072, opened 2023-10-09
fix!: Report errors at startup

fredizzimo:
**NOTE:** Depends on and includes https://github.com/neovide/neovide/pull/2062

Startup errors are now properly reported. If a terminal is attached, then they are reported to the terminal and otherwise in a GUI window, which supports, resizing, scrolling and copying, much like a pager like `less`.

On Windows the subsystem is always `windows` and no console is ever attached. So all errors and help messages are reported through the GUI. This was done because `AttachConsole` does not really properly attach to the console, and it does not return to the command line until the user press enter, which is confusing.

On Linux and MacOS, a double fork is done instead of launching a completely new process, unless  `--no-fork` is specified. This happens immediately after the commandline is processed, so any commandline helps is still printed to the terminal.

* Yes, some errors that previously were silently ignored are now reported and hard errors. This mainly concerns errors in the `ginit.vim`, which were completely silently ignored before.
* `--no-fork` have no effect on Windows anymore.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -33,18 +33,16 @@ extern crate derive_new;
 extern crate lazy_static;
 
 use anyhow::Result;
-use clap::error::Error as ClapError;
 use log::trace;
 use std::env::{self, args};
 use std::fs::{File, OpenOptions};
 use std::io::Write;
 use std::panic::{set_hook, PanicInfo};
 use std::process::ExitCode;
-use std::sync::Arc;
 use std::time::SystemTime;
 use time::macros::format_description;
 use time::OffsetDateTime;
-use winit::{error::EventLoopError, event_loop::EventLoop};
+use winit::error::EventLoopError;
 
 #[cfg(not(test))]
 use flexi_logger::{Cleanup, Criterion, Duplicate, FileSpec, Logger, Naming};
@@ -53,11 +51,12 @@ use backtrace::Backtrace;
 use bridge::NeovimRuntime;
 use cmd_line::CmdLineSettings;
 use editor::start_editor;
-use renderer::{cursor_renderer::CursorSettings, GlWindow, RendererSettings};
+use error_handling::handle_startup_errors;
+use renderer::{cursor_renderer::CursorSettings, RendererSettings};
 use settings::SETTINGS;
 use window::{
     create_event_loop, create_window, determine_window_size, main_loop, KeyboardSettings,
-    UserEvent, WindowSettings, WindowSize,
+    WindowSettings, WindowSize,
 };
 
 pub use channel_utils::*;
@@ -93,15 +92,7 @@ fn main() -> ExitCode {
     let event_loop = create_event_loop();
 
     match setup() {
-        Err(err) => {
-            if let Some(clap_error) = err.downcast_ref::<ClapError>() {
-                let _ = clap_error.print();
-                clap_error.exit_code() as u8
-            } else {
-                eprintln!("ERROR: {}", err);
-                1
-            }
-        }
+        Err(err) => handle_startup_errors(err) as u8,
         Ok((window_size, _runtime)) => {
             maybe_disown();
             start_editor();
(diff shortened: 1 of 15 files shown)


fredizzimo (commented):
I found a couple of things to clean up and will do that a bit later today.

src/window/error_window.rs
```diff
@@ -0,0 +1,490 @@
...
+        }
+    }
+
+    fn render(&mut self) {
+        let size = Size::new(self.size.width as f32, self.size.height as f32);
+        let (message_rect, help_message_rect) = self.layout();
+
+        self.paragraphs.message.layout(size.width - 2.0 * PADDING);
+        let (offset, possible_scroll_direction) =
+            self.handle_scrolling(message_rect.height() as f64);
```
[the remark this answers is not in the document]
fredizzimo: done

src/main.rs
```diff
@@ -81,11 +81,36 @@ fn main() -> ExitCode {
...
+            match main_loop(window, window_size, event_loop) {
+                Ok(()) => 0,
+                // All error codes have to be u8, so just do a direct cast with wrap around, even if the value is negative,
+                // that's how it's normally done on operating systems that don't support negative return values.
+                Err(EventLoopError::ExitFailure(code)) => code as u8,
+                _ => 1,
+            }
+        }
+    }
+    .into()
```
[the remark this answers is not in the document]
fredizzimo: done

src/window/error_window.rs
```diff
@@ -0,0 +1,488 @@
...
+}
+
+struct ErrorWindow<'a> {
+    skia_renderer: SkiaRenderer,
+    context: WindowedContext,
+    font_collection: FontCollection,
+    size: PhysicalSize<u32>,
+    scale_factor: f64,
+    paragraphs: Paragraphs,
+    message: &'a str,
```
MultisampledNight, line 69: It might be useful to add context about the fact that Neovide just crashed and this in fact represents an error which might be relevant to report, instead of just the error itself. Maybe something like this could be prepended?

> Neovide just crashed :(
> This is the error that caused the crash. In case you don't know what to do with this, please feel free to report this on https://github.com/neovide/neovide/issues!

src/window/error_window.rs
```diff
@@ -0,0 +1,488 @@
...
+    window::{load_icon, SkiaRenderer, UserEvent},
+};
+
+const TEXT_COLOR: Color4f = WHITE;
+const BACKGROUND_COLOR: Color4f = BLACK;
+const FONT_SIZE: f32 = 12.0 * 96.0 / 72.0;
+const PADDING: f32 = 10.0;
+const MAX_LINES: i32 = 9999;
+const MIN_SIZE: PhysicalSize<u32> = PhysicalSize::new(800, 600);
+const DEFAULT_SIZE: PhysicalSize<u32> = PhysicalSize::new(500, 500);
```
MultisampledNight, line 33: Mixed up. Their values and their effect are correct, but their names should be exchanged.

```suggestion
const DEFAULT_SIZE: PhysicalSize<u32> = PhysicalSize::new(800, 600);
const MIN_SIZE: PhysicalSize<u32> = PhysicalSize::new(500, 500);
```

src/window/error_window.rs
fredizzimo: Done!

It's not perfect, since it also prints the message for user errors like giving the wrong `neovim-bin`. But regular wrong command line arguments are printed without that message.

src/window/error_window.rs
fredizzimo: Fixed, I must have made some mistake when I refactored the variable names.

fredizzimo:
I need to check closer what happened, but it looks like these changes were already included and merged in
* #2076

fredizzimo:
That was expected, the order of merging should just have been the opposite. Closing this

PR #2076, opened 2023-10-12
fix: Fix some crashes that can happen during exit

fredizzimo:
Depends on https://github.com/neovide/neovide/pull/2072

Some of the errors, like a channel being closed, are just ignored. While others are logged instead of panicking. This also gives better chance of recovery if something goes wrong. It could still be possible to save and exit from Neovim for example.

- No

### DIFF
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -4,7 +4,8 @@ use std::sync::Arc;
 use log::error;
 use log::trace;
 
-use nvim_rs::{call_args, rpc::model::IntoVal, Neovim};
+use anyhow::{Context, Result};
+use nvim_rs::{call_args, error::CallError, rpc::model::IntoVal, Neovim};
 use tokio::sync::mpsc::unbounded_channel;
 
 #[cfg(windows)]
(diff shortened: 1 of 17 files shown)


fredizzimo:
I put it as draft due to build failures on Windows

src/bridge/ui_commands.rs
```diff
@@ -129,108 +138,130 @@ pub enum ParallelCommand {
     },
 }
 
+async fn display_awailable_fonts(
```
MultisampledNight, line 141: Not sure if this is introduced by this PR, but — typo.

```suggestion
async fn display_available_fonts(
```

MultisampledNight:
Works almost completely fine, thank you!

Only thing I noticed is that issuing an exit request from the WM results in this on stderr at exit, but no panic anymore and no other side effects either:

```
ERROR [neovide::bridge::ui_commands] ConfirmQuit failed

Caused by:
    0: Error decoding response to request 'nvim_command'
    1: Error while reading from Reader
    2: EOF
```

MultisampledNight:
Probably worth noting that ^ only seems to happen with a debug build, release builds seem to be fine, so it's probably a race condition.

fredizzimo:
I will check the debug exit problem before we merge.

src/bridge/ui_commands.rs
fredizzimo: fixed

fredizzimo:
It turned out that the exit crash is normal, neovide requests neovim to exit, and expects a response back. But neovim actually exits instead of sending a response back.

So I just disabled any error reporting for it.

MultisampledNight (approved):

MultisampledNight:
Thank you! Choooo chooooooooo :steam_locomotive:

Issue #2080, opened 2023-10-13
Double-click + Drag


9mm:
**Is your feature request related to a problem? Please describe.**

On MacOS (and Im pretty sure windows as well), if you double click in the middle of text it will auto-highlight the word. This already words correctly, in Neovide.

But... if you don't let go the second time (ie... click + clickHOLD), if you drag the cursor around it should select by word.

I realize mouse is stupid in vim, but a lot of times I will use mouse when needing to copy lines of variable names out into wherever (env variables, etc).

**Describe the solution you'd like**

To follow the standard OS functionality where click+clickHold will highlight by word instead of highlighting by character

fredizzimo:
Hm. It works for me as documented https://neovim.io/doc/user/term.html#mouse-using

Single click drag selects by character, double click drag selects by word, triple click selects by line, and quadruple clicking selects a rectangular region.

We don't do anything special about this, but Neovide should take care of it.

Maybe your https://neovim.io/doc/user/options.html#'mousetime' is set too fast? 

Note, this was broken in Neovim for a long time, but should have been fixed last year https://github.com/neovim/neovim/issues/4754

9mm:
Hmm, I dont have mouse time set (ill experiment with it) but i did see that the problem doesnt happen in Neovim, only in Neovide

45 seconds:

https://www.loom.com/share/88116f6c5df84943a62f712402ee7f99

fredizzimo:
I can repeat it on this computer to some degree, it seems like it sometimes works, but most of the times it does not.

I wonder if we should send `left` and `drag`  instead of `move` and empty https://neovim.io/doc/user/api.html#nvim_input_mouse()

But I  don't understand why Neovim would need drag at all, since it should be able to track button down and up itself.

fredizzimo:
It turns out that it works in debug mode, but not in release mode. But the fix seems to be easy, only send the move event when not dragging. Currently we actually send both events, and that seems to confuse Neovim it they arrive too close together.

I will create a PR

9mm:
Sweet!! Glad that was easy one

fredizzimo:
It should be fixed here, https://github.com/neovide/neovide/pull/2082. But, note that the PR targets main and not the rendering branch.

PR #2082, opened 2023-10-16
fix: Fix mouse dragging

fredizzimo:
Fixes mouse dragging
* https://github.com/neovide/neovide/issues/2080

Mouse move should not be sent when the mouse is dragged. It looks like Neovim can't deal with both events happen simultaneously.

This also fixes a clippy issue.

- No

### DIFF
--- a/neovide-derive/src/lib.rs
+++ b/neovide-derive/src/lib.rs
@@ -7,7 +7,7 @@ pub fn setting_group(item: TokenStream) -> TokenStream {
     let input = parse_macro_input!(item as DeriveInput);
     let prefix = setting_prefix(input.attrs.as_ref())
         .map(|p| format!("{p}_"))
-        .unwrap_or_else(|| "".to_string());
+        .unwrap_or_default();
     stream(input, prefix)
 }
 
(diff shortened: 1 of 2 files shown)


9mm:
Just tested on OSX works amazing

Out of curiosity, do you know what style token I would use to style the part UNDER the mouse?

Its hard to explain but if you click+drag, the actual cursor moves to the end of the word, but underneath the mouse it turns black. Im assuming becuase my theme isnt styling whatever that is called

fredizzimo:
@9mm, I don't think there are any https://neovim.io/doc/user/syntax.html#highlight-groups, so it might be a Neovim bug. 

The cursor is displayed like normal at the end of the selection, and that is controlled by https://neovim.io/doc/user/options.html#'guicursor'. But I don't think it affects the fake cursor under the mouse pointer.

There's also a https://neovim.io/doc/user/options.html#'mouseshape' option, but that is currently not implemented, even if it's documented https://github.com/neovim/neovim/issues/21458

9mm:
Gotcha. Totally not a big deal just curious. Thanks, this highlighting bug will help me a lot. 99% of the times i have to use the mouse but other times I do need to do significant copying pasting into other things outside the editor which can be tedious

Thanks you so much 💯

PR #2089, opened 2023-10-23
Report error when setting a wrong guifont value

### LINKED ISSUE #2077: Setting the wrong guifont should report errors
We have many issues related to people mistyping the font names, or just otherwise use the guifont format the wrong way for example: 
* https://github.com/neovide/neovide/issues/2054
* https://github.com/neovide/neovide/issues/1944 (this one is because a Windows bug, but we should still report that we fail to load the font, instead of showing something else)
* https://github.com/neovide/neovide/issues/1872
* https://github.com/neovide/neovide/issues/1864
* https://github.com/neovide/neovide/issues/1396

And even more issues that are already closed, for example
* https://github.com/neovide/neovide/issues/2071
* https://github.com/neovide/neovide/issues/1973
* https://github.com/neovide/neovide/issues/1882
* https://github.com/neovide/neovide/issues/1500
* https://github.com/neovide/neovide/issues/1471
* https://github.com/neovide/neovide/issues/1274
* https://github.com/neovide/neovide/issues/846
* https://github.com/neovide/neovide/issues/260

So, we clearly need to report that loading the font failed to the user, and not just in the log. This should be done with `echomsg` or similar, so that it actually shows up in Neovim.

Finally, we should also check that all fallback fonts can be loaded properly and not just the main font, like we do currently.

Tmpod:

It addresses #2077 by fist implementing a new parallel `UiCommand`, `ShowError { message: String }` which gets translated to neovim's [`echomsg`](https://neovim.io/doc/user/eval.html#%3Aechom), and then using it in tandem with `log::error!` in `CachingShaper::update_font` (when `self.font_loader.get_or_load` fails).

- Yes
  - Invalid [`guifont`](https://neovim.io/doc/user/options.html#'guifont') values will now error out instead of silently being accepted. To pick no hinting or aliased edging, you should now use `#h-none` and `#e-alias`, respectively.
    - *Actually, just checked the docs, and they already specified those options, even though the code didn't check for them specifically, they were just being handled on the wildcard match. Still this is a breaking change because of the new errors.*

Closes #2077

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -59,6 +59,13 @@ pub async fn show_intro_message(
     nvim.exec_lua(INTRO_MESSAGE_LUA, args).await
 }
 
+pub async fn show_error_message(
+    nvim: &Neovim<NeovimWriter>,
+    message: &String,
+) -> Result<(), Box<CallError>> {
+    nvim.command(&format!("echomsg \"{message:?}\"")).await
+}
+
 #[tokio::main]
 async fn start_neovim_runtime(instance: NeovimInstance) {
     let handler = NeovimHandler::new();
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -6,8 +6,9 @@ pub mod session;
 mod setup;
 mod ui_commands;
 
-use std::{process::exit, sync::Arc, thread};
+use std::{process::exit, sync::Arc, thread, ops::Add};
 
+use itertools::Itertools;
 use log::{error, info};
 use nvim_rs::{error::CallError, Neovim, UiAttachOptions, Value};
 
@@ -61,9 +62,22 @@ pub async fn show_intro_message(
 
 pub async fn show_error_message(
     nvim: &Neovim<NeovimWriter>,
-    message: &String,
+    lines: &[String],
 ) -> Result<(), Box<CallError>> {
-    nvim.err_writeln(message).await
+    nvim.echo(
+        lines
+            .iter()
+            .map(|l| {
+                Value::Array(vec![
+                    Value::String(l.clone().add("\n").into()),
+                    Value::String("ErrorMsg".into()),
+                ])
+            })
+            .collect_vec(),
+        true,
+        vec![],
+    )
+    .await
 }
 
 #[tokio::main]
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -63,7 +63,7 @@ pub async fn show_error_message(
     nvim: &Neovim<NeovimWriter>,
     message: &String,
 ) -> Result<(), Box<CallError>> {
-    nvim.command(&format!("echomsg \"{message:?}\"")).await
+    nvim.err_writeln(message).await
 }
 
 #[tokio::main]
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -6,11 +6,12 @@ pub mod session;
 mod setup;
 mod ui_commands;
 
-use std::{process::exit, sync::Arc, thread, ops::Add};
+use std::{ops::Add, process::exit, sync::Arc, thread};
 
 use itertools::Itertools;
 use log::{error, info};
 use nvim_rs::{error::CallError, Neovim, UiAttachOptions, Value};
+use rmpv::Utf8String;
 
 use crate::{
     cmd_line::CmdLineSettings, error_handling::ResultPanicExplanation, running_tracker::*,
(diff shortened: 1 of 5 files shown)


Tmpod:
At the moment, this is not producing any message, though the log is going through just fine.

I'm not familiar with Neovim's API, so perhaps I'm not using `echomsg` right?

*PS: I'm aware of the extra `:` there, my bad, will remove in later commit.*

src/bridge/mod.rs
```diff
@@ -59,6 +59,13 @@ pub async fn show_intro_message(
     nvim.exec_lua(INTRO_MESSAGE_LUA, args).await
 }
 
+pub async fn show_error_message(
+    nvim: &Neovim<NeovimWriter>,
+    message: &String,
+) -> Result<(), Box<CallError>> {
+    nvim.command(&format!("echomsg \"{message:?}\"")).await
```
fredizzimo, line 66: The code might be cleaner if you use the api `nvim_echo`. Or perhaps even an error `nvim_err_writeln`. 

So, `nvim.echo` or `nvim.err_writeln` instead of `nvim.command`

Tmpod: aah makes sense, thank you!

Tmpod:
Alright, this is now working. I have two questions regarding the code though (sorry, first contribution!)

1. Since `show_error_message` is just a forward call to `Neovim::err_writeln`, is it really worth having? Perhaps it would just be best to use that neovim API call directly in the command match arm.
2. Should I write a test for this, and if so, where?

Thank you!

fredizzimo:
@Tmpod. 
1. We use the nvim api directly elsewhere in the file, so I think removing the function makes sense. It's clear what writeline does. 
2. I think tests are quite hard for this, since they would rely on the fonts installed on the system or require some mocking. Additionally, the actual functionality is quite trivial, so I don't think we need any tests.

The only thing left to do would be to also verify the fallback fonts. But if you are unsure how to do that, we can leave it for later in another PR. Basically, you need to try loading all of the fonts in the guifont list.

BTW, I'm also on Discord so if you have further questions, you can ask there too.

Tmpod:
> 1. We use the nvim api directly elsewhere in the file, so I think removing the function makes sense. It's clear what writeline does. 
> 2. I think tests are quite hard for this, since they would rely on the fonts installed on the system or require some mocking. Additionally, the actual functionality is quite trivial, so I don't think we need any tests.

Makes sense!

> The only thing left to do would be to also verify the fallback fonts. But if you are unsure how to do that, we can leave it for later in another PR. Basically, you need to try loading all of the fonts in the guifont list.

That doesn't sound too difficult, I can try do that as well.
Essentially, walk the font list and try loading them until we get one
that does work? Or try load them all and report which ones failed?

> BTW, I'm also on Discord so if you have further questions, you can ask there too.

o7

src/renderer/fonts/font_options.rs
```diff
@@ -15,47 +29,44 @@ pub struct FontOptions {
...
+                font_options.size = parse_pixels(part).map_err(|_| INVALID_SIZE_ERR)?;
+            } else if part.starts_with(FONT_WIDTH_PREFIX) && part.len() > 1 {
+                font_options.allow_float_size |= part[1..].contains(ALLOW_FLOAT_SIZE_OPT);
+                font_options.width = parse_pixels(part).map_err(|_| INVALID_WIDTH_ERR)?;
+            } else if part == FONT_BOLD_OPT {
                 font_options.bold = true;
-            } else if part == "i" {
+            } else if part == FONT_ITALIC_OPT {
                 font_options.italic = true;
             }
```
fredizzimo, line 66: It would be good to give an error for an unknown part. But I'm not sure if we can do that, since that might be valid for another UI.

Tmpod: mmm, yeah not sure about that, but maybe it is a good decision.  
On one hand, as you say, it may be valid for another UI and invalid for neovide. On the other hand though, perhaps it is the user's responsibility to adjust their `guifont` setting according to the UI being used. Even though it isn't really standard, pretty much every alternative UI I know exposes some way of knowing if their active or not, and such settings can be conditionally tweaked that way.

@Kethku thoughts?

fredizzimo: I think we can add more validation later if needed.

fredizzimo:
Thank you!

Tmpod:
Glad to help! :3

BatmanAoD:
This is honestly a good change (I've struggled way too much with font configuration in Vim before), but since it's a breaking change, it should probably be a little more prominent in the release notes or something. I upgraded from v0.11 and was very confused.

PR #2102, opened 2023-11-05
feat!: use display link api to implement vsync on macos.

### LINKED ISSUE #2073: macos, laggy if move mouse on startup (window being created).
**Describe the bug**
Neovide becomes laggy when I move mouse during the startup.
As Neovide will create a maximized window during startup (after click the app icon in dock), during the window being created and zooming out in the screen, if I move mouse, Neovide becomes laggy (maybe losing frames?) and does not recover even after performing some actions.
But when I deliberately keep the mouse not moved and wait for the window created, zoomed out and completely stable, the problem disappears.
It can be reproduced stably in my environment.

**To Reproduce**
1. Click the App icon in Dock
2. Immediately move mouse in the window being created and zooming out.
3. After the window created completely, Neovide becomes laggy (maybe losing frames) and never recovers.

**Expected behavior**
Running fluently of course.

**Screenshots**
Not really useful I think. But if you need one, I'll provide.

 - OS: macOS Ventura 13.6 (22G120), Intel CPU, macbook pro 13inch.
 - Neovide Version 0.11.2 (cask from homebrew)
 - Neovim Version (from homebrew)
```
NVIM v0.9.2
Build type: Release
LuaJIT 2.1.0-beta3
```
Logs:
There are two files after run with --log. Both are two long, so I make a zip of them.
Dir `good` for the good one
Dir `bad` for the bug one
[log.zip](https://github.com/neovide/neovide/files/12856695/log.zip)

Feel free to ask me for more info if you need! I love Neovide! Thanks so much!

### LINKED ISSUE #2093: Frame rate stuttering on macOS
On some macOS systems the frame rate stutters a bit due to a buggy vsync implementation. If you have that issu, it's recommended to:
1. Start Neovide with the command `--no-vsync` to disable the vsync and switch to a software based timer. 
2. Set `g:neovide_refresh_rate` https://neovide.dev/configuration.html#refresh-rate to the same as your current monitor refresh rate. This combination should make it reasonably smooth, but if not, you might want to try different refresh rates to see what works for you.

This will be properly fixed by:
* https://github.com/rust-windowing/winit/issues/2412

But currently the winit team don't have the manpower to implement it, so contributors are welcome. @crupest is also working on a quick fix for Neovide itself, at least that's the status at the time of writing, see
* https://github.com/neovide/neovide/issues/2073, so any additional effort should be discussed with them first.

The vsync issues are well known and discussed here:
* https://github.com/libsdl-org/SDL/issues/4918
* https://github.com/glfw/glfw/issues/2249
* https://github.com/libsdl-org/SDL/issues/4471
* https://github.com/libsdl-org/SDL/issues/1871
* https://github.com/glfw/glfw/issues/2249

And also on or Discord server: 
* https://discord.com/channels/896132176091955261/896263985614168125/1132361827095953408
* https://discord.com/channels/896132176091955261/896263985614168125/1142096375178342480
* https://discord.com/channels/896132176091955261/896263894086066216/1152521461152628736

crupest:
Fix #2073, fix #2093 

This pr uses display link api on macos to correctly do vsync. The opengl vsync seems broken on macos.

~BLOCKED by #2097~

UNDER WORKING: Don't merge now. Just to stop from duplicate work.

BREAKING CHANGE: This may totally change the structure of vsync usage.

TODOS:
- [x] After #2097 fixed, have a leading test of whether this works.
- [x] Use the right way to wait for vsync.
- [x] Create a context to share sync primitives between main thread and display link thread.
- [x] Make VsyncMacos an individual enum value instead of using Windows one.
- [x] Add logging.
- [x] Run tracy to get test data.
- [ ] Examine the result.

@fredizzimo I have some questions about vsync usage. In current code, vsync seems owned by the update loop. But as I pointed out before, the display link is specific to a screen. So the structure of vsync usage should be changed as I think. To be simple, I make vsync owned by a window. And when window moves and the screen is changed, vsync updates. Do you have any ideas about this?

### DIFF
--- a/src/renderer/vsync/vsync_macos.rs
+++ b/src/renderer/vsync/vsync_macos.rs
@@ -1,0 +1,98 @@
+use std::sync::{Arc, Condvar, Mutex};
+
+use log::{error, trace, warn};
+
+use crate::renderer::WindowedContext;
+
+use super::macos_display_link::{
+    core_video, get_display_id_of_window, MacosDisplayLink, MacosDisplayLinkCallbackArgs,
+};
+
+struct VSyncMacosDisplayLinkUserData {
+    vsync_count: Arc<(Mutex<usize>, Condvar)>,
+}
+
+fn vsync_macos_display_link_callback(
+    _args: &mut MacosDisplayLinkCallbackArgs,
+    user_data: &mut VSyncMacosDisplayLinkUserData,
+) {
+    let (lock, cvar) = &*user_data.vsync_count;
+    let mut count = lock.lock().unwrap();
+    *count += 1;
+    cvar.notify_one();
+}
+
+pub struct VSyncMacos {
+    old_display: core_video::CGDirectDisplayID,
+    display_link: Option<MacosDisplayLink<VSyncMacosDisplayLinkUserData>>,
+    vsync_count: Arc<(Mutex<usize>, Condvar)>,
+    last_vsync: usize,
+}
+
+impl VSyncMacos {
+    pub fn new(context: &WindowedContext) -> Self {
+        let mut vsync = VSyncMacos {
+            old_display: 0,
+            display_link: None,
+            vsync_count: Arc::new((Mutex::new(0), Condvar::new())),
+            last_vsync: 0,
+        };
+
+        vsync.display_link = vsync.create_display_link(context);
+
+        vsync
+    }
+
+    fn create_display_link(
+        self: &mut Self,
+        context: &WindowedContext,
+    ) -> Option<MacosDisplayLink<VSyncMacosDisplayLinkUserData>> {
+        self.old_display = get_display_id_of_window(context.window());
+
+        let vsync_count = self.vsync_count.clone();
+
+        match MacosDisplayLink::new_from_display(
+            self.old_display,
+            vsync_macos_display_link_callback,
+            VSyncMacosDisplayLinkUserData { vsync_count },
+        ) {
+            Ok(display_link) => {
+                trace!("Succeeded to create display link.");
+                match display_link.start() {
+                    Ok(did) => match did {
+                        true => {
+                            trace!("Display link started.");
+                        }
+                        false => {
+                            warn!("Display link already started. This does not affect function. But it might be a bug.");
+                        }
+                    },
+                    Err(code) => {
+                        error!("Failed to start display link, CVReturn code: {}.", code);
+                    }
+                }
+                Some(display_link)
+            }
+            Err(code) => {
+                error!("Failed to create display link, CVReturn code: {}.", code);
+                None
+            }
+        }
+    }
+
+    pub fn wait_for_vsync(&mut self) {
+        let (lock, cvar) = &*self.vsync_count;
+        let count = cvar
+            .wait_while(lock.lock().unwrap(), |count| *count < self.last_vsync + 1)
+            .unwrap();
+        self.last_vsync = *count;
+    }
+
+    pub fn update(&mut self, context: &WindowedContext) {
+        let new_display = get_display_id_of_window(context.window());
+        if new_display != self.old_display {
+            trace!("Window moved to a new screen, try to re-create the display link.");
+            self.display_link = self.create_display_link(context);
+        }
+    }
+}
--- a/src/main.rs
+++ b/src/main.rs
@@ -51,6 +51,7 @@ use cmd_line::CmdLineSettings;
 use editor::start_editor;
 use error_handling::{handle_startup_errors, NeovideExitCode};
 use renderer::{cursor_renderer::CursorSettings, RendererSettings};
+#[cfg_attr(target_os = "windows", allow(unused_imports))]
 use settings::SETTINGS;
 use window::{
     create_event_loop, create_window, determine_window_size, main_loop, WindowSettings, WindowSize,
--- a/src/renderer/opengl.rs
+++ b/src/renderer/opengl.rs
@@ -123,7 +123,7 @@ pub fn build_context(window: GlWindow, srgb: bool, vsync: bool) -> Context {
     // NOTE: We don't care if these fails, the driver can override the SwapInterval in any case, so it needs to work in all cases
     // The OpenGL VSync is always disabled on Wayland and Windows, since they have their own
     // implementation
-    let _ = if vsync && env::var("WAYLAND_DISPLAY").is_err() && OS != "windows" {
+    let _ = if vsync && env::var("WAYLAND_DISPLAY").is_err() && OS != "windows" && OS != "macos" {
         surface.set_swap_interval(&context, SwapInterval::Wait(NonZeroU32::new(1).unwrap()))
     } else {
         surface.set_swap_interval(&context, SwapInterval::DontWait)
--- a/src/renderer/vsync/macos_display_link.rs
+++ b/src/renderer/vsync/macos_display_link.rs
@@ -1,0 +1,225 @@
+use std::{ffi::c_void, pin::Pin};
+
+use crate::profiling::tracy_zone;
+
+use self::core_video::CVReturn;
+
+use cocoa::{
+    appkit::{NSScreen, NSWindow},
+    base::{id, nil},
+    foundation::{NSAutoreleasePool, NSDictionary, NSString},
+};
+use objc::{rc::autoreleasepool, *};
+
+use raw_window_handle::{HasRawWindowHandle, RawWindowHandle};
+use winit::window::Window;
+
+// Display link api reference: https://developer.apple.com/documentation/corevideo/cvdisplaylink?language=objc
+#[allow(non_upper_case_globals, non_camel_case_types)]
+pub mod core_video {
+    use std::ffi::c_void;
+
+    pub type CGDirectDisplayID = u32;
+
+    pub type CVReturn = i32;
+    pub const kCVReturnSuccess: CVReturn = 0;
+    pub const kCVReturnDisplayLinkAlreadyRunning: CVReturn = -6671;
+    pub const kCVReturnDisplayLinkNotRunning: CVReturn = -6672;
+    // pub const kCVReturnDisplayLinkCallbacksNotSet: CVReturn = -6673;
+
+    type SInt16 = i16;
+    type UInt32 = u32;
+    type uint32_t = u32;
+    type int32_t = i32;
+    type uint64_t = u64;
+    type int64_t = i64;
+    type double = f64;
+
+    #[repr(C)]
+    #[allow(non_snake_case)]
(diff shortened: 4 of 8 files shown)


src/renderer/vsync/vsync_macos.rs
```diff
@@ -0,0 +1,76 @@
...
+    flagsOut: *mut u64,
+    displayLinkContext: *mut c_void,
+) -> core_video::CVReturn {
+    tracy_zone!("VSyncDisplayLinkCallback");
+
+    // The display link should be dropped before vsync, so this should be safe.
+    let vsync = displayLinkContext as *mut VSyncMacos;
+
+    unsafe {
+        (*vsync).barrier.wait(); // TODO: Maybe there is a safer way.
```
fredizzimo, line 30: I think it would be better to use a synchronisation primitive that does not block the callback thread. For example a condition variable like the windows implementation. 

That also makes it possible to wait for vsync with a timeout in the wait function. I think that might be needed to to allow the main thread to wake up when there are problems, like the display link getting invalid because of a monitor disconnect or something.

fredizzimo:
> @fredizzimo I have some questions about vsync usage. In current code, vsync seems owned by the update loop. But as I pointed out before, the display link is specific to a screen. So the structure of vsync usage should be changed as I think. To be simple, I make vsync owned by a window. And when window moves and the screen is changed, vsync updates. Do you have any ideas about this?

The changes you have made makes perfect sense and also simplifies the code without the need to pass the vsync object around. I don't remember the exact reason why it was owned by the update loop in the first place, it might have been needed in some of the earlier versions of the render loop, or just because I thought it fitted better there than in the window.

src/renderer/vsync/vsync_macos.rs
crupest: Ok. I'll take a look this night. The display link callback is called in a separate thread according to apple's document. So the implementation should be quite similar.

crupest: @fredizzimo Do you think it is still necessary to use `Arc` here like on Windows? The display link thread should be destroyed once the vsync object is dropped. But I'm not sure. Apple's documentation is very unclear.

fredizzimo: I think the arc was used just to be able to be able to share the same Mutex and Condvar  and in the update loop. So I believe it's still required.

fredizzimo: Hm, now that I think about it, you probably don't need the arc. I think it's enough to store the objects directly on self, like you did for the barrier.

crupest: I think the arc used on windows is to avoid that lifetime of vsync thread does not lie in lifetime of main thread. So arc can avoid use-after-release. On macos, the issue is the same. We can't promise the lifetime relationship of two threads.
However, the barrier should be wrong. Because if the display link fails, main thread will block forever since the barrier never counts down to 0. Actually I have marked this in code at `wait_for_vsync`.

crupest: I think a little bit more. If I need to share sync primitives between two threads, use arc is the correct way in rust. This unsafe cast of vsync just breaks the semantic of rust though it might work well.

fredizzimo: Yes, I'm too inexperienced with rust to give any direct recommendations. But you probably need to create a struct containing the arcs, cast that and send it to the callback.

crupest: That's what I think. I'll update the code tomorrow. Now I just mark this as resolved.

crupest:
@fredizzimo I think coding is completed and it's time to examine it.

Actually when I follow up the main branch, the bug already disappears. And even using the display link api, there are still some frames lost. I don't feel there is a big difference between using or not using display link. And current performance is good enough as far as I'm concerned .  I don't know this makes some sense any more.

Whatever, here are the new tracy:
1. main -> main branch, no command line arguments
2. display-link -> this pr branch, no command line arguments
3. no-vsync -> main branch, with `--no-vsync`

[tracy-log.zip](https://github.com/neovide/neovide/files/13326149/tracy-log.zip)

fredizzimo:
@crupest, thank you.

I think we need more testing before merging, I will ask for that both in the issue and on Discord, after one remaining issue is fixed as described below.

But from the traces it looks reasonable.
1. main -> main branch, no command line arguments - This jitters between a fast and a slow frame
2. display-link -> this pr branch, no command line arguments - Mostly looks good, but there are periods of jitter (more on that later)
 3. no-vsync -> main branch, with `--no-vsync` -- No jitter, but seems to have two modes 8 ms and 16 ms

I think the display-link case can be improved by forcing the swap interval to 0, in and always taking this branch https://github.com/neovide/neovide/blob/b79b7cc0be18fd36b469713e55c558835178dad5/src/renderer/opengl.rs#L129. The reason for the jitter seems to be that `swap_buffers` sometimes blocks after the displaylink wait, and disabling that should hopefully fix it.

NOTE: I have not yet had time to review the actual code.

crupest:
@fredizzimo I have pushed the new code to force swap interval to 0.

Here is the new tracy: [tracy.zip](https://github.com/neovide/neovide/files/13365257/tracy.zip)

Just take no hurry. Let's make it more mutual before merge.

By the way, I'll try to take a look on tracy by myself. I think it's good for me to learn on this.

Thank you very much!

fredizzimo:
At least the tracy log looks good now with a stable 60 FPS. Except for the long red peaks, where nothing seem to be happening. But maybe you did not run it with `--no-idle`? In that case it's expected to not render when nothing happens.

crupest:
@fredizzimo With `--no-idle`, [tracy.zip](https://github.com/neovide/neovide/files/13365550/tracy.zip)

Looks much better as I can see.

fredizzimo:
If you ran with `--no-idle` it's a bit weird. It looks like the update loop is actually called regularly during those frame drops, but it never calls `draw_frame` to actually render. I wonder why that happens?

It should be forced here with `--no-idle` https://github.com/neovide/neovide/blob/3e83b031defacb6906e7afe8229eab275a4f80ea/src/window/update_loop.rs#L136-L140

crupest:
Sorry I need to rest now. I'll take a look another day. I understand what you mean and I'll do some debug then.

fredizzimo:
No problems, thank you for the work so far, it already seems like a huge improvement!

crupest:
@fredizzimo Hey, I'm back again.

> If you ran with `--no-idle` it's a bit weird. It looks like the update loop is actually called regularly during those frame drops, but it never calls `draw_frame` to actually render. I wonder why that happens?

I take a look into the tracy log of `--no-idle` myself. I can't find frames where `draw_frame` does not run. Would you mind pointing it out? Or I misunderstand something?

fredizzimo:
I mean the log from this message https://github.com/neovide/neovide/pull/2102#issuecomment-1812546460

Do you see the big red bars? If you zoom in on those, then you can see that draw frame is not called, even if the loop is run regularly.

On the other hand this one https://github.com/neovide/neovide/pull/2102#issuecomment-1812580149 looks much better.

crupest:
> I mean the log from this message [#2102 (comment)](https://github.com/neovide/neovide/pull/2102#issuecomment-1812546460)
> 
> Do you see the big red bars? If you zoom in on those, then you can see that draw frame is not called, even if the loop is run regularly.
> 
> On the other hand this one [#2102 (comment)](https://github.com/neovide/neovide/pull/2102#issuecomment-1812580149) looks much better.

Yeah, now I get what you mean! Debugging now!

crupest:
@fredizzimo Please help me analyze. I have made following extra log:

Screenshot 2023-11-25 at 22 22 15

Screenshot 2023-11-25 at 22 22 43

The log file and tracy file is in following zip.

[new-logs.zip](https://github.com/neovide/neovide/files/13464331/new-logs.zip)

crupest:
@fredizzimo I think I know the cause now. Because there is actually NOTHING to redraw. I was keeping acting before with some little stop in between. So the frames are "lost".

[nothing.tracy.zip](https://github.com/neovide/neovide/files/13464375/nothing.tracy.zip)

fredizzimo:
I'm on the phone and can't check it with the computer right now.

But if you use `--no-idle` it should always redraw, unless that is broken somehow

crupest:
Sorry I'm a little hysteric about this result.

To summarize, between the "lost" frames,
1. neovim does not tell you that you need to redraw because you didn't perform any actions
2. the settings are not changed
3. the window does not resize
4. ...

So during these time, `prepare_frame` does NOT return `true` to let you call `render`, which leads to the "lost" frames in tracy.

> @fredizzimo I think I know the cause now. Because there is actually NOTHING to redraw. I was keeping acting before with some little stop in between. So the frames are "lost".
> 
> [nothing.tracy.zip](https://github.com/neovide/neovide/files/13464375/nothing.tracy.zip)

I did a test on this later. I keep no moving and the frame is lost for a long time as I suppose.

> I'm on the phone and can't check it with the computer right now.
> 
> But if you use `--no-idle` it should always redraw, unless that is broken somehow

Actually `--no-idle` works correctly.

> @fredizzimo With `--no-idle`, [tracy.zip](https://github.com/neovide/neovide/files/13365550/tracy.zip)
> 
> Looks much better as I can see. **EDIT:** I mean the tracy log.

From this test data, `render` is called no matter what `prepare_frame` returns, so it keeps synchronized with vsync, with only few frame lost as the vsync lost.

fredizzimo:
It looks like everything works correctly then. The fact that it does not draw anything when there's nothing to do is by design to conserve power, especially on laptops. `--no-idle`, disables that and is mostly there for debugging purposes, performance profiling and and problem situations.

fredizzimo:
Thank you, I merged this to get more testing.

crupest:
@fredizzimo Thank you too!

From the beginning of this issue, you have taught me so many with plenty of patience. I really appreciate it. Never had I imaged I can complete this before. This is the first big contribution to one of my favorite projects. And I couldn't make it without your help.

By the way, the implementation of the display link might have some potential bugs since it connects the native platform and the project's rust codes. If anyone runs into a problem on this, please leave a message and I'll be very pleasant to fix them!

Whatever, cheers 🍻!

fredizzimo:
Thank you, it's been a pleasure to work with you. 

Open source is all about working together to achieve something, and a lot of patience is needed, since people usually have other things things in life, so things can take a long time to complete. It's a lot different from regular programming at a company. 

This particular issue has been something we have wanted for several months, to make all platforms work roughly equally well when releasing the smooth rendering changes. But due to the lack of macOS developers, and the nature of the task, you have been the only one accepting the challenge, and I really appreciate that.

So again, thank you for this work!

Issue #1936, opened 2023-07-16
Editor background transparency renders, but isn't using the compositor effects for the window background


alerque:
[This comment](https://github.com/neovide/neovide/pull/1870#discussion_r1264691807) reminded me that I haven't tried transparency lately since I started running a build of PR#1870 that is fast enough to actually use as a daily driver (love it!). The result on my end is roughly what is mentioned in that comment, but it lacks a notable feature of my Wayland compositor environment: namely the Window is not reporting it's transparency via the compositor properly so it isn't getting the proper composite effects back. My compositor renders everything behind the foreground window in transparent zones with a blur that makes it much easier to not be distracted by text and lines.

Look for example at the difference here between Neovide and Alacritty+Neovim:

![20230716_18h10m12s_grim](https://github.com/neovide/neovide/assets/173595/34f5605c-40b9-4b95-bd90-ed476347c9ea)

This effect is not unique to Alacritty (where my setting is just setting an opacity level). The effect comes from the compositor and works for *all* windows that report transparency. I don't know what the exact mechanism is that they do this, but it works for quite a few other things and of course if I set the whole window opacity. It apples to more than just the background to, it is anything under the current window.

I'm using Neovide 0.10.4.r49.g56e27b7-1 built from the smooth render loop PR a while back. I can update it to the latest but having watched it I don't think there are any relevant changes. The compositor is Hyprland 0.27.0.

fredizzimo:
This was also reported here https://github.com/neovide/neovide/issues/1918, but that issue is closed and not the main point of the issue.

fredizzimo:
There's now a Winit PR open, which I believe will fix the issue https://github.com/rust-windowing/winit/pull/3060

fredizzimo:
This is now merged into Winit in https://github.com/rust-windowing/winit/pull/3060, so if someone wants to integrate this into Neovide feel free to do it.

For now it's best to target the "Improve Render Loop" branch https://github.com/neovide/neovide/pull/1977, since we have updated winit to the latest master version there and fixed some breaking changes.  But one more version bump is still required to enable this feature.

Once winit is updated, it should just be a matter of calling `set_blur`.

fredizzimo:
This will be fixed by:
* https://github.com/neovide/neovide/pull/2440

PR #2104, opened 2023-11-06
feat: Add acrylic blur option for background

Kethku:
- Feature

- No


src/error_handling.rs
```diff
@@ -25,7 +25,7 @@ macro_rules! error_msg {
         let msg = format!($($arg)+);
         log::error!("{}", msg);
         EVENT_AGGREGATOR.send(UiCommand::Parallel(ParallelCommand::ShowError {
-            lines: msg.split('\n').map(|s| s.to_string()).collect_vec(),
+            lines: itertools::Itertools::collect_vec(msg.split('\n').map(|s| s.to_string())),
```
MultisampledNight, line 28: Was this change intentional? It only makes things longer. If the intention was to avoid the `Itertools` dep, this could be done using `.collect::<Vec<_>>()`.

src/window/window_wrapper.rs
```diff
@@ -158,18 +161,44 @@ impl WinitWindowWrapper {
...
+                blur_color.r(),
+                blur_color.g(),
+                blur_color.b(),
+                blur_color.a(),
+            );
+            #[cfg(target_os = "windows")]
+            window_vibrancy::apply_acrylic(window, Some(blur_color))
+        } else {
+            #[cfg(target_os = "windows")]
+            window_vibrancy::clear_acrylic(window)
```
MultisampledNight, line 178: [It looks like macOS could be supported as well](https://docs.rs/window-vibrancy/0.4.2/window_vibrancy/index.html#example), although we'd need with a macOS machine to confirm that.

src/error_handling.rs
Kethku: It was intentional to avoid requiring a using statement on the user side of the macro.

src/window/window_wrapper.rs
Kethku: Yeah I mentioned this in the issue. I dont know how the macos arguments work and it was late so I just didnt bother. This PR isn't really mergeable as is and should probably be done with a winit upstream change anyway. I just threw it together as an example

fredizzimo:
On Wayland we should use https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_blur where supported

fredizzimo:
macOS also supports blur through winit in the latest release https://github.com/rust-windowing/winit/releases/tag/v0.29.3, so on that there's now two alternatives, the `window-vibrancy` crate and the winit one. The vibrancy crate is a bit more flexible, but personally I would prefer if as much as possible was handled by winit, and maybe they could add some extra configuration options.

regxxxman:
image
can you help me build

falcucci:
@Kethku @fredizzimo I guess we could proceed with he macOS blurring I've implemented at #2273 after our fix #2272 so we could merge your blurred windows OS based afterwards?


the macOS blurring option is really simple to enable and as far as I've using it, works really well with our floating blurred windows.

fredizzimo:
@Kethku, I think we should implement this totally with winit now. Acryllic backdrop support was added here, and as mentioned earlier, both macOS and Wayland has blur support.
* https://github.com/rust-windowing/winit/pull/3257

fredizzimo:
Once wayland is added it should also close this issue
* https://github.com/neovide/neovide/issues/1936

Kethku:
Ok I looked into this and the feature isn't released yet on winit. So I will wait on updating this PR for a little while longer

Kethku:
The relevant function hasn't actually been released yet so I'm going to hold off on updating this till that gets in.

Kethku:
> Any news regarding this? I don't want to be pushy, just curious. It's also a minor thing, but it's the main thing stopping me from using Neovide, and sticking to using the terminal.

This is still waiting on an upstream winit release. The necessary code has been merged but not released yet. Plus winit updates in the past have been... tricky. So its gonna be a while yet. Out of our hands for the moment

AntonDankov:
I've made the change and created pull request:
https://github.com/neovide/neovide/pull/3021

PR #2108, opened 2023-11-06
chore!: Update the macOS CI to use macOS 12 instead of 11

fredizzimo:
Updates the CI images to macOS 12. MacOS 12 - Big Sur is no longer supported according to https://en.wikipedia.org/wiki/MacOS. I don't see anything in the Rust release notes or in the Gitlab CI release notes about macOS 11 being unsupported though.

But in any case this seem to fix the CI problems we have compiling the `aarch64-apple-darwin` image, which broke after this was released https://github.com/actions/runner-images/releases/tag/macOS-11%2F20231030.1 and can be seen here https://github.com/neovide/neovide/actions/runs/6768373367/job/18392914175?pr=2106 for example. This has been verified on my fork https://github.com/fredizzimo/neovide/actions/runs/6773128253/job/18407618899.

This possibly breaks Neovide on macOS 11 and earlier.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-latest, macOS-11, ubuntu-latest]
+        os: [windows-latest, macOS-12, ubuntu-latest]
         toolchain: [stable, nightly]
     runs-on: ${{ matrix.os }}
     continue-on-error: ${{ matrix.toolchain == 'nightly' }}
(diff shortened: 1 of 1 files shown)


Kethku:
I'm going to merge this. If it breaks things we will hear about it but without devices to test i think its better to get our ci unblocked

tssm:
I can confirm this breaks the app in macOS 11, as in it refuses to run. My guess it the macOS 12 image doesn't include the right SDK by default. I could try to fix it, but maybe the right thing to do is to update the README, as I would need some help with the fix (not sure where to start, but I have the machine to test)

fredizzimo:
@tssm, I don't think we should do anything about that. macOS 11 Big Sur has reached it's end of life https://endoflife.date/macos, and does not even receive security updates anymore.

Users on that can try to compile from source, but there's no guarantee that it will work.

Maybe the readme could list the minimum supported OS versions though.

PR #2120, opened 2023-11-10
Fix remember window size

Kethku:
The new option observation starts sending notifications earlier than the previous solution. So we need to ignore the columns and lines change notifications until the ui has finished initializing or we don't properly set the window size

- Fix

- No

### DIFF
--- a/lua/init.lua
+++ b/lua/init.lua
@@ -3,6 +3,8 @@
 ---@field register_clipboard boolean
 ---@field register_right_click boolean
 ---@field enable_focus_command boolean
+---@field global_variable_settings string[]
+---@field option_settings string[]
 
 ---@type Args
 local args = ...
@@ -62,23 +64,29 @@ vim.api.nvim_create_user_command("NeovideFocus", function()
     rpcnotify("neovide.focus_window")
 end, {})
 
-vim.api.nvim_create_autocmd({ "OptionSet" }, {
-    pattern = "columns",
-    once = false,
-    nested = true,
-    callback = function()
-        rpcnotify("neovide.columns", tonumber(vim.v.option_new))
-    end
-})
+vim.api.nvim_exec([[
+function! WatchGlobal(variable, callback)
+    call dictwatcheradd(g:, a:variable, a:callback)
+endfunction
+]], false)
 
-vim.api.nvim_create_autocmd({ "OptionSet" }, {
-    pattern = "lines",
-    once = false,
-    nested = true,
-    callback = function()
-        rpcnotify("neovide.lines", tonumber(vim.v.option_new))
+for _,global_variable_setting in ipairs(args.global_variable_settings) do
+    local callback = function()
+        rpcnotify("setting_changed", global_variable_setting, vim.g["neovide_" .. global_variable_setting])
     end
-})
+    vim.fn.WatchGlobal("neovide_" .. global_variable_setting, callback)
+end
+
+for _,option_setting in ipairs(args.option_settings) do
+    vim.api.nvim_create_autocmd({ "OptionSet" }, {
+        pattern = option_setting,
+        once = false,
+        nested = true,
+        callback = function()
+            rpcnotify("option_changed", option_setting, vim.o[option_setting])
+        end
+    })
+end
 
 -- Create auto command for retrieving exit code from neovim on quit.
 vim.api.nvim_create_autocmd({ "VimLeavePre" }, {
--- a/neovide-derive/src/lib.rs
+++ b/neovide-derive/src/lib.rs
@@ -40,19 +40,37 @@ fn struct_stream(name: Ident, prefix: String, data: &DataStruct) -> TokenStream 
         Some(ref ident) => {
             let vim_setting_name = format!("{prefix}{ident}");
 
-            let location = match option(field) {
-                Ok(option) => match option {
-                    Some(option_name) => quote! {{ crate::settings::SettingLocation::NeovimOption(#option_name.to_owned()) }},
-                    None => quote! {{ crate::settings::SettingLocation::NeovideGlobal(#vim_setting_name.to_owned()) }},
-                },
+            let option_name = match option(field) {
+                Ok(option_name) => option_name,
                 Err(error) => {
                     return error.to_compile_error();
                 }
             };
 
+            let location = match &option_name {
+                Some(option_name) => quote! {{ crate::settings::SettingLocation::NeovimOption(#option_name.to_owned()) }},
+                None => quote! {{ crate::settings::SettingLocation::NeovideGlobal(#vim_setting_name.to_owned()) }},
+            };
+
             let field_ident = field.ident.as_ref().unwrap();
             let case_name = field_ident.to_string().to_case(Case::Pascal);
             let case_ident = Ident::new(&case_name, field_ident.span());
+
+            // Only create a reader function for global neovide variables
+            let reader = if option_name.is_none() {
+                quote! {
+                    fn reader(settings: &crate::settings::Settings) -> Option<rmpv::Value> {
+                        let s = settings.get::<#name>();
+                        Some(s.#ident.into())
+                    }
+                }
+            } else {
+                quote! {
+                    fn reader(_settings: &crate::settings::Settings) -> Option<rmpv::Value> {
+                        None
+                    }
+                }
+            };
 
             quote! {{
                 fn update(settings: &crate::settings::Settings, value: rmpv::Value) {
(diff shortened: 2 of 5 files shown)


9mm:
Hmm.. I hate to be bringer of bad news but it didn't seem to fix issue.

image

This is how it starts

Also did `cargo clean` and `cargo run --release`

Let me know how else I could help if you want me to test other stuff

9mm:
A log if that helps
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/13315427/neovide_rCURRENT.log)

Kethku:
And you for sure dont have a line in your config that sets the lines and columns?

9mm:
I'm 99% sure, but what is the exact config variable name and ill search for it

```
➜  nvim git:(master) ✗ ack colu
init.lua
58:vim.opt.signcolumn = 'yes:1' -- auto:1'

lua/zesty/plugins/mini.lua
26:          -- align first column by default
➜  nvim git:(master) ✗ pwd
/Users/zesty/.config/nvim
```

There's also nothing for lines except these which arent it:

```
➜  nvim git:(master) ✗ ack lines
lua/zesty/keymaps.lua
14:-- move lines
17:vim.keymap.set('x', '<D-k>', ":m '<-2<CR>gv=gv", { desc = 'Move lines up'})
18:vim.keymap.set('x', '<D-j>', ":m '>+1<CR>gv=gv", { desc = 'Move lines down'})
54:-- dont skip wrapped lines
61:-- delete empty lines
62:vim.keymap.set('n', '<Leader>del', ':g/^$/d<CR>', { desc = 'Delete empty lines' })

lua/zesty/plugins/theme.lua
42:            underlines = {

lua/zesty/plugins/lsp_cmp.lua
35:      max_lines = 1000,

lua/zesty/plugins/mini.lua
90:          trailspace.trim_last_lines()

lua/zesty/plugins/repl.lua
8:        ignore_blank_lines = true,
27:      desc = 'REPL lines',

lua/zesty/plugins/treesitter.lua
107:        max_lines = 2,
```

Kethku:
its super weird. Your log file indicates that the lines and columns are getting set which triggers a window resize

9mm:
Could that be set in the config file? As in, its reading from the config file properly but breaking after?

Because one thing I do notice (like last time) is the config is actually saved properly, just reading it seems to be the problem. (indicated by it being broken for awhile, and when I use a compiled version that works, suddenly it restores back how it was)

Kethku:
The config file does have a maximized setting but I dont think that would conflict here. The persistent file is in the nvim-data directory and should be read just like before.

9mm:
here are my main settings:

```
--                          __
-- .-----.-----.-----.--.--|__.--------.
-- |     |  -__|  _  |  |  |  |        |
-- |__|__|_____|_____|\___/|__|__|__|__|

-- essentials
vim.g.mapleader = ','
vim.g.maplocalleader = ','
vim.opt.termguicolors = true

-- lazy
local lazypath = vim.fn.stdpath('data') .. '/lazy/lazy.nvim'
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    'git',
    'clone',
    '--filter=blob:none',
    'https://github.com/folke/lazy.nvim.git',
    '--branch=stable',
    lazypath,
  })
end
vim.opt.rtp:prepend(lazypath)
require('lazy').setup('zesty.plugins', {
  change_detection = {
    enabled = true,
    notify = false,
  },
})

-- gui
require('zesty.gui')

-- keymaps
require('zesty.keymaps')

-- autocommands
require('zesty.autocommands')

-- aesthetic
vim.opt.guifont = 'JetBrainsMono Nerd Font:h14' -- AnonymicePro Nerd Font:h15

-- indentation
vim.opt.autoindent  = true -- continue indentation to new line
vim.opt.smartindent = true -- add extra indent when it makes sense
vim.opt.smarttab    = true -- <tab> at the start of a line behaves as expected
vim.opt.expandtab   = true -- <tab> inserts spaces
vim.opt.shiftwidth  = 2    -- >>, << shift line by 4 spaces
vim.opt.tabstop     = 2    -- <tab> appears as 4 spaces
vim.opt.softtabstop = 2    -- <tab> behaves as 4 spaces when editing

-- cursor
vim.opt.guicursor = ''
--vim.opt.virtualedit = 'all'

-- line numbers
vim.opt.number = true
vim.opt.signcolumn = 'yes:1' -- auto:1'

-- wrap
vim.opt.wrap = false
vim.opt.linebreak = true
vim.opt.breakindent = true

-- scroll
vim.opt.scrolloff = 4
vim.opt.mouse = 'nv'
vim.opt.mousescroll = 'ver:2,hor:0'

-- splits
vim.opt.splitright = true
vim.opt.splitbelow = true

-- folds
vim.opt.foldtext = 'v:lua.vim.treesitter.foldtext()'

-- search
vim.opt.hlsearch = true
vim.opt.incsearch = true
vim.opt.ignorecase = true
vim.opt.smartcase = true

-- recovery
vim.opt.backup = false
vim.opt.undolevels = 1000
vim.opt.undofile = false
vim.opt.undodir = os.getenv('HOME') .. '/.local/state/nvim/undo//'
vim.opt.swapfile = true
vim.opt.directory = os.getenv('HOME') .. '/.local/state/nvim/swap//'

-- misc
vim.opt.updatetime = 50
vim.opt.timeout = true
vim.opt.timeoutlen = 600
vim.opt.shortmess:append('I')

-- filetypes
vim.filetype.add({
  extension = {
    pcss = 'css',
  },
})
```

```
  -- padding
  vim.g.neovide_padding_top    = 20
  vim.g.neovide_padding_bottom = 20
  vim.g.neovide_padding_right  = 20
  vim.g.neovide_padding_left   = 20

  -- display
  --vim.g.neovide_transparency = 0.8
  --vim.g.transparency = 0.8
  --vim.g.neovide_background_color = '#0f0f12' .. alpha()
  vim.g.neovide_refresh_rate = 9999 -- fps peaks over 1000+ (requires NEOVIDE_VSYNC=0)
  vim.g.neovide_refresh_rate_idle = 5
  vim.g.neovide_hide_mouse_when_typing = true
  vim.g.neovide_underline_automatic_scaling = true

  -- window
  vim.g.neovide_remember_window_size = true
  vim.g.neovide_confirm_quit = true

  -- animation
  vim.g.neovide_cursor_animation_length = 0
  vim.g.neovide_cursor_trail_size = 0
  vim.g.neovide_scroll_animation_length = 0
  vim.g.neovide_cursor_animate_command_line = false

  -- https://github.com/neovide/neovide/issues/1838
  vim.keymap.set('t', '<MouseMove>', '<NOP>')

```

Kethku:
Yeah doesn't look like it gets set at all. Very strange...

9mm:
Is there any more insight here maybe?

https://github.com/neovide/neovide/issues/1580
https://github.com/neovide/neovide/pull/2027

9mm:
More specifically the last one, 2027, is where he fixed it the first time

Kethku:
Ill look into it more tomorrow. The timing is subtle i think

fredizzimo:
@Kethku, based on the log the problem is that it reads the initial columns and lines.

The resizing implementation is based on the fact that the columns and lines autocommand is only called when the user sets the column or line manually But with your changes it's also called with the initial value.

And a couple of other things that I did not had time to review in the original PR
1. The `reader_func` or setting the initial values was really useful for at least me. I use the autocompletion all the time. I also frequently check the existing value of the options typing it's name. So this feels like a regression, but obviously we should only do it for our settings where the initial values are controlled by us and not for neovim options.
2. I'm ok with combining `KeyboardSettings` and `WindowSettings`, but it feels arbitrary to just do it for those two and not for `CursorSettings` and `RenderSettings` as well.
3. This was already addressed here, original autocommands were not deleted.

Kethku:
Okeydokey. I will put it back. Thanks for taking a look.

fredizzimo:
And
4. This is a bigger change, and did not get much worse in your PR,  but I think we should move the setting vimscript code to our Neovide `init.lua` and just pass the list of settings to that. This would improve the startup times on low latency connections. With the current implementation there are two separate RPC calls for each setting reading the initial value and adding the listener. So with a round-trip of just 10 ms and 50 settings, it would mean a full second extra second startup time.

Kethku:
I'm working on #4. Small hiccup here is that I dont think we can call dictwatcheradd from lua. Theres a fair amount of discussion in the neovim repo about using metatables to do something similar but I'm not sure how that would work with the way we do global variables.

I feel like maybe I'm missing something obvious. @fredizzimo do you know of another way to get global variable notifications?

Kethku:
Aight @9mm let me know if the current set of changes works for you.

@fredizzimo I ended up creating a global function via an vimscript exec and then calling that to work around not being able to pass the g: dictionary from the lua side. I think its possible theres a better way to do this but thats the best I've got for now. Other than that I addressed your other concerns. Should be good to go I think.

9mm:
@Kethku that fixed it!

PR #2121, opened 2023-11-10
Add support for variable font weights

Kethku:
Adds the option to add a number after b to represent the configured font weight. (We should decide if we would prefer to change to w instead of b for this. I went with b as it was part of what we already had)
Add an option to configure what weight bold should be

be7ab14a-0cb2-4714-aa3f-90c211d56651

- Feature

- No

### DIFF
--- a/src/editor/mod.rs
+++ b/src/editor/mod.rs
@@ -277,6 +277,7 @@ impl Editor {
             EditorCommand::RedrawScreen => {
                 tracy_zone!("EditorRedrawScreen");
                 self.redraw_screen();
+                self.draw_command_batcher.send_batch();
             }
         };
     }
--- a/src/renderer/cursor_renderer/mod.rs
+++ b/src/renderer/cursor_renderer/mod.rs
@@ -10,7 +10,7 @@ use crate::{
     bridge::EditorMode,
     editor::{Cursor, CursorShape},
     profiling::tracy_zone,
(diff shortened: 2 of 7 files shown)


fredizzimo:
I don't think we should extend `Guifont` any more. It's simply not flexible enough, so we need to find a better alternative. We have had some discussions on Discord, and some more here https://github.com/neovide/neovide/pull/1920, but never reached a conclusion.

The only reason for using the inflexible `guifont` was to be compatible with other GUIs, but I don't think our `guifont` longer is, and definitely won't be if we start to interpret options differently,

In particular for this font weight option, I would like to be able to specify different weights for `normal` `bold` and `italic`. Actually I would like to see the possibility to use different font's for all of those, especially the `italic` could be would be nice to override, since some font's have way too stylzed italics, and and oblique style or even different font would be handy.

I like the base format of Alacritty https://github.com/alacritty/alacritty/releases/download/v0.12.3/alacritty.yml. Where each style has it's own configuration and overrides things from the base. It does not necessarily need to be a separate yml config file, although the advantage of that it is machine specific, independent on where the remote neovim runs.

But we could also implement it as a lua function for example `neovide_set_guifont`, which takes a dictionary with that format. The advantage of doing it this way would be that it's easy to change just the size for example dynamically by calling it again with just the size field, if it's implemented correctly.

Issue #1331, opened 2022-05-25
neovide copy/paste is not working when connect to remote neovim


jjpark78:
i am really happy to using neovide at my workflow. 
before using neovide, i was a emacs user for 10 years. 
now, i am using neovim + lua config + neovide for 1 years. 

most of my workflow i use neovide on linux x11 desktop
but sometime i have to use windows for specific workload. 
in this situation i have following problems. 

**Describe the bug**
in Windows 10, neovide can connect to remote neovim (running on ubuntu) but, cannot paste windows clipboard, also cannot copy remote neovim's yanked text to windows clipboard

**My neovim config**
```lua
vim.g.mapleader = ' '
vim.g.encoding = "utf8"

vim.cmd([[
...
set clipboard=unnamedplus
...
]])
```

start neovim headless with this command 

```sh
ssh -L 6666:localhost:6666 jltech-lab.iptime.org -p 9022 nvim --headless --listen localhost:6666
```

using neovide connect to remote neovim
```pwsh
C:\neovide.exe --remote-tcp=localhost:6666
```

in neovide, yank & paste not working. 

how can i make this scenario working ??

cajomar:
I have a very similar setup, although with a macbook client instead on Windows. 
I believe neovim is using the remote machine's clipboard for yank/put, rather than your local clipboard. 
I'm looking in solutions for this, and would be happy to make a PR if this is a feature that could be added to Neovide.

j-xella:
Same problem here. Inability to use copy paste is a major obstacle

j-xella:
One dirty workaround is to install a local X-Server on your windows machine, and ensure that it is the one used by the nvim host. From what I understand, nvim can use X-Server clipboard if available.

1. X-Server is running on your Windows box
2. X connection is forwarded by your SSH session
3. Check the `$DISPLAY` variable in your ssh shell
4. Make sure the same value is set in nvim session

khjorth:
I suspect that `set clipboard=unnamedplus` counts as a clipboard operation with respect to https://github.com/neovim/neovim/issues/8017. This is relevant because Neovide relies on `g:clipboard` to support clipboard functionality for remote nvim instances. In support of this suspicion, I can break remote clipboard support in my (Linux) setup by adding that line to my `init.vim`.

What happens if you run
```
:unlet g:loaded_clipboard_provider
:runtime autoload/provider/clipboard.vim
```
after you connect Neovide?

N8Brooks:
[...]

PR #2123, opened 2023-11-11
Wayland clipboard

cshuaimin:

- Fix

I first fixed a missing return in the clipboard initialization code as mentioned [here](https://github.com/neovide/neovide/issues/1331#issuecomment-1801900152).

Then I make use of wayland clipboard of copypasta crate.

### Motivation

Recently on GNOME the `wl-clipboard` cli works poorly for me. Every time I use `dd` or `cw` there's an annoying notification says "wl-clipboard is ready".

In the TUI I switched to recently merged OSC 52 clipboard on nightly. In Neovide I found the copypasta works pretty well, unless it's using X11 clipboard.

PS: `g:clipboard` config to use Neovide clipboard instead of wl-clibpard cli:

```lua
if vim.g.neovide then
  local function neovide_rpc(method, ...) return vim.rpcrequest(vim.g.neovide_channel_id, method, ...) end

  local function neovide_copy(lines) return neovide_rpc("neovide.set_clipboard", lines) end

  local function neovide_paste() return neovide_rpc("neovide.get_clipboard") end

  vim.g.clipboard = {
    name = "neovide",
    copy = {
      ["+"] = neovide_copy,
      ["*"] = neovide_copy,
    },
    paste = {
      ["+"] = neovide_paste,
      ["*"] = neovide_paste,
    },
    cache_enabled = 0,
  }
end
```

- No


src/window/mod.rs
```diff
@@ -309,6 +311,8 @@ pub fn main_loop(
     let cmd_line_settings = SETTINGS.get::<CmdLineSettings>();
     let mut window_wrapper = WinitWindowWrapper::new(window, initial_window_size);
 
+    crate::clipboard::init(&event_loop);
```
fredizzimo, line 314: Could this be moved before calling `main_loop` so that we don't need to duplicate the call on Windows and the other platforms?

cshuaimin: Fixed!

fredizzimo:
Thanks!

At some point I think we should investigate if we could make copypasta the default. It would also solve the problem on Windows with the WSL, that you need an external tool like `win32yank` installed. It could be the default and set before `inti.lua/vim` is run, so the user could still set it manually and revert to default neovim by unsetting it.

khjorth:
I'm not sure how copypasta works, but won't this break the clipboard functionality for X11 users on Linux?

Issue #1553, opened 2022-09-22
Set window title


bigH:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to customize Neovide's title bar to include the `pwd`.

**Describe the solution you'd like**
Ideally, I'd expect that `vimrc` or equivalent can simply set:
```viml
let g:neovide_window_title = $PWD
```

Furthermore, it would also be nice to change the title while Neovide is running. This could be useful to indicate that `:terminal`s are waiting to complete some command or any number of things.

**Describe alternatives you've considered**
The only alternative I know is to use a terminal emulator instead.

**Additional context**
N/A

MultisampledNight:
Thanks for reporting!

This is already possible, see the `title` and `titlestring` options. They both are respected by Neovide and can also be changed at runtime.

Though for the terminal, I'm not sure how this would go. Maybe the shell can set the current working directory and NeoVim just propagates it upwards, but this is either way something that's outside of Neovide's play area.

fredizzimo:
@gsouf, they are Neovim options https://neovim.io/doc/user/options.html#'titlestring'

PR #2126, opened 2023-11-12
fix: prevent empty window title

raggi:
If Neovim attempts to send an empty window title, it is now set to
Neovide instead, rather than the empty string.

This improves two user scenarios:

- Users who utilize assistive technologies will now always observe a
  window title by default.
- Users of window managers that rely on window titles will have improved
  integration.

The precipitating use case was using Komorebi, a tiling window manager
for Windows. Komorebi uses an empty window title as a strong signal that
the software is a hidden window, as this is a common convention on the
platform.

Updates https://github.com/neovide/neovide/issues/1553

### DIFF
--- a/src/editor/mod.rs
+++ b/src/editor/mod.rs
@@ -81,8 +81,11 @@ impl Editor {
     pub fn handle_editor_command(&mut self, command: EditorCommand) {
         match command {
             EditorCommand::NeovimRedrawEvent(event) => match event {
-                RedrawEvent::SetTitle { title } => {
+                RedrawEvent::SetTitle { mut title } => {
                     tracy_zone!("EditorSetTitle");
+                    if title.is_empty() {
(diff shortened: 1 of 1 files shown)


fredizzimo:
I don't think this should be configurable, since the configuration can already be done with `title` and `titlestring`. But I think hardcoding it to Neovide when it's empty would be ok, that way the title can also be set immediately when the window is created and not at a later time when the configuration is read.

raggi:
> I don't think this should be configurable, since the configuration can already be done with title and titlestring. But I think hardcoding it to Neovide when it's empty would be ok, that way the title can also be set immediately when the window is created and not at a later time when the configuration is read.

Done.

Kethku:
Thanks for the fix! I noticed a similar issue recently and hadn't tracked down what was going on.

PR #2129, opened 2023-11-14
fix: event loop shouldn't join the render thread

sid-6581:
- Fix

- No

On Windows, the event loop calls `join()` on the render thread, but if the render thread is trying to do something that requires the Windows message pump to be running, this results in a deadlock. The easiest way to reproduce is to run `neovide -- -c "qa"`. On my machine, the event loop will try to join the render thread while the render thread is trying to set the window title, which completely hangs neovide. Since we're already polling `RUNNING_TRACKER` I just added a test for the render thread completion and only exit the event loop once the render thread is finished.

### DIFF
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -237,44 +237,48 @@ pub fn main_loop(
     event_loop: EventLoop<UserEvent>,
 ) -> Result<(), EventLoopError> {
     let cmd_line_settings = SETTINGS.get::<CmdLineSettings>();
-    let render_thread_done = Arc::<AtomicBool>::new(AtomicBool::new(false));
-    let render_thread_done2 = render_thread_done.clone();
+    let render_thread_done = Arc::new(AtomicBool::default());
     let (txtemp, rx) = channel::<Event<UserEvent>>();
     let mut tx = Some(txtemp);
 
-    let render_thread_handle = thread::spawn(move || {
-        let mut window_wrapper = WinitWindowWrapper::new(window, initial_window_size);
-        let mut update_loop = UpdateLoop::new(
-            cmd_line_settings.vsync,
-            cmd_line_settings.idle,
-            &window_wrapper.windowed_context,
-        );
-        #[allow(unused_assignments)]
-        loop {
-            let (wait_duration, _) = update_loop.get_event_wait_time();
-            let event = if wait_duration.as_nanos() == 0 {
-                rx.try_recv()
-                    .map_err(|e| matches!(e, TryRecvError::Disconnected))
-            } else {
-                rx.recv_timeout(wait_duration)
-                    .map_err(|e| matches!(e, RecvTimeoutError::Disconnected))
-            };
+    let render_thread_handle = thread::spawn({
+        let render_thread_done = Arc::clone(&render_thread_done);
+        move || {
+            let mut window_wrapper = WinitWindowWrapper::new(window, initial_window_size);
+            let mut update_loop = UpdateLoop::new(
+                cmd_line_settings.vsync,
+                cmd_line_settings.idle,
+                &window_wrapper.windowed_context,
+            );
 
-            if update_loop.step(&mut window_wrapper, event).is_err() {
-                break;
+            #[allow(unused_assignments)]
+            loop {
+                let (wait_duration, _) = update_loop.get_event_wait_time();
+                let event = if wait_duration.as_nanos() == 0 {
+                    rx.try_recv()
+                        .map_err(|e| matches!(e, TryRecvError::Disconnected))
+                } else {
+                    rx.recv_timeout(wait_duration)
+                        .map_err(|e| matches!(e, RecvTimeoutError::Disconnected))
+                };
+
+                if update_loop.step(&mut window_wrapper, event).is_err() {
+                    break;
+                }
             }
+
+            let window = window_wrapper.windowed_context.window();
+            save_window_size(
+                window.is_maximized(),
+                window.inner_size(),
+                window_wrapper.get_grid_size(),
+                window.outer_position().ok(),
+            );
+            render_thread_done.store(true, Ordering::Relaxed);
         }
-        let window = window_wrapper.windowed_context.window();
-        save_window_size(
-            window.is_maximized(),
-            window.inner_size(),
-            window_wrapper.get_grid_size(),
-            window.outer_position().ok(),
-        );
-        render_thread_done.store(true, Ordering::Relaxed);
     });
 
-    let result = event_loop.run(move |e, window_target| {
+    let result = event_loop.run(|e, window_target| {
         // We need to wake up regularly to check the running tracker, so that we can exit
         window_target.set_control_flow(ControlFlow::WaitUntil(
             std::time::Instant::now() + std::time::Duration::from_millis(100),
@@ -298,7 +302,7 @@ pub fn main_loop(
             }
         }
 
-        if render_thread_done2.load(Ordering::Relaxed) {
+        if render_thread_done.load(Ordering::Relaxed) {
             window_target.exit();
         }
     });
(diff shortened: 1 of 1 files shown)


src/window/mod.rs
```diff
@@ -235,9 +237,12 @@ pub fn main_loop(
     event_loop: EventLoop<UserEvent>,
 ) -> Result<(), EventLoopError> {
     let cmd_line_settings = SETTINGS.get::<CmdLineSettings>();
+    let render_thread_done = Arc::<AtomicBool>::new(AtomicBool::new(false));
```
Kethku, line 240: AtomicBool isn't necessary because the type can be inferred by the constructor argument
```suggestion
    let render_thread_done = Arc::new(AtomicBool::new(false));
```

src/window/mod.rs
```diff
@@ -235,9 +237,12 @@ pub fn main_loop(
...
+    let render_thread_done2 = render_thread_done.clone();
     let (txtemp, rx) = channel::<Event<UserEvent>>();
     let mut tx = Some(txtemp);
-    let mut render_thread_handle = Some(thread::spawn(move || {
+
+    let render_thread_handle = thread::spawn(move || {
```
Kethku, line 245: Rather than carrying a temporary variable here with a clumsy 2 suffix, I'd wrap the clojure which uses it in a block and shadow the original variable name like this:

```suggestion
    let render_thread_handle = thread::spawn({
        let render_thread_done = render_thread_done.clone();
        move || {
        ...
    }});
```

Then you don't need the second variable.

Kethku:
I added some minor style comments, but this should wait for @fredizzimo to weigh in since he wrote the current iteration of the update loop.

src/window/mod.rs
sid-6581: Ah, I will try that. I basically copied the style in that vsync thread for this, but I don't like the 2 suffix much either.

src/window/mod.rs
sid-6581: Good point, I will simplify that!

sid-6581: Simplified!

src/window/mod.rs
sid-6581: I fixed this one as well. I used `Arc::clone()` instead of `.clone()` since it's more idiomatic and doesn't trigger clippy's `clone_on_ref_ptr` lint.

sid-6581:
> I added some minor style comments, but this should wait for @fredizzimo to weigh in since he wrote the current iteration of the update loop.

Definitely appreciate it. I know from painful experience that multiple threads and the Windows message pump is a minefield. The PRs I submitted are the ones I've run into so far, but from looking at the code I suspect there are more issues waiting to surface. I'll tackle them as I come across them.

fredizzimo:
I don't think this is safe. 

Before  the `event_loop` quits, all windows and the message loops are automatically closed. But the `winit_window_wrapper` can potentially be closed after that. There's a small time window when that can happen, since the event loop is quit in parallel with the render thread shutdown, after `render_thread_done` is set. So in that case the `winit_window_wrapper`, the opengl context and skia stuff is shut down too late.

I would simplify this, by:
1. having the render thread check for `RUNNING_TRACKER.is_running()` instead of the event loop, and automatically quit when the quit is triggered. 
2. The event loop can then then check for `render_thread_handle::is_ready`, and call `window_target.exit()` at that point.
3. And the `render_thread_handle.join` happens in `Event::LoopExiting`. 
4. I don't think we can reach this point without the `RUNNING_TRACKER` knowing about it, but maybe as an additional safety we could call `RUNNING_TRACKER.quit()` before the join.

So this way there's no for any atomic bool either.

NOTE: That neither of the threads are completely blocking at the moment and should wake up regularly, but if either of them is changed to block, then the above will not work. In that case we can change the running tracker to use a custom `UserEvent`, which is sent to the event loop queue and will wake both of them up. But there's a few other changes we need in regards to dealing with our own internal events as well, all events likely need to be sent using UserEvent, or we need to wait for multiple channels.

sid-6581:
I'm not sure what you mean is not safe? The winit wrapper will be dropped only after both the render thread and event loop are done, which is how it should be?

1. I'm not sure this is needed, since the event loop already correctly shuts down the render loop by dropping tx, which it needs to do anyway. It shouldn't hurt, but I don't see how it adds anything.

2. Good point, I didn't realize rust had `is_finished` on the join handle. Looks like that was added in 1.61 and I never noticed. I'll definitely make that change, it will eliminate the bool.

3. I don't think this should hurt if `is_finished` is true at that point, but I still feel a little uneasy about having a thread join inside an event loop on Windows. I'd worry that if future modifications were to be made in this function, this situation could happen again. Any objections to leaving the join outside? I think it's more clear (and safer) that way. 

4. I don't think this is needed since neither the render thread nor the event loop can exit without the running tracker knowing about it in any of the current or proposed scenarios.

fredizzimo:
1. is to avoid the original problem you have. The main event loop needs to keep running until the render thread is shut down to avoid deadlocks. By not dropping the tx until it actually have exited we keep the communication channel open until it's completely shut down. I also think it simplifies the code by not having to deal with the dropped channels.

3. We absolutely need to do a join here to properly shut down opengl and skia before the buffers and window handles they are using are destroyed by the operating system.

fredizzimo:
For 4. There might be some internal winit event or error condition that triggers `window_target.exit()`, and therefore triggers `Event::LoopExiting`

sid-6581:
I see what you mean regarding the initial AtomicBool solution flagging the thread as done before the winit wrapper being dropped. `is_finished` should fix that, since all user code (including drops) will be completely finished by the time `is_finished` returns true. So if we check for `is_finished` inside the event loop and don't exit the event loop until then, we are guaranteed that the drop has occurred, so there's no need for a join inside the event loop. I'll make those changes and see what else I can simplify. I agree I don't think we need to rely on the channel drops.

Speaking of dropping the winit wrapper, should that really be the render thread's responsibility? It seems more natural to move that outside the thread, so it happens after the event loop.

fredizzimo:
You are correct, I guess with the `is_finished` is does not really matter where you perform the join, the thread has to complete before it can be joined anyway.

> Speaking of dropping the winit wrapper, should that really be the render thread's responsibility? It seems more natural to move that outside the thread, so it happens after the event loop.

The winit wrapper contains some thread specific opengl context, so they need to be created and deleted on the thread that renders them, or moved across the threads. And winit further restricts it by declaring some of the stuff as `!send`. I don't remember the exact details anymore.

But if that wasn't the case, more of the code could be shared between the platforms.

sid-6581:
Yes, I was worried about that. I figured I might run into problems, even if I made sure to properly move it so it's not accessed by multiple threads at the same time. You're right, a lot of stuff is `!send`. I'll keep that as-is and rework the rest according to what we discussed.

sid-6581:
@fredizzimo I pushed some changes that I believe covers what we discussed and simplifies it quite a bit.

src/window/mod.rs
```diff
@@ -235,69 +234,68 @@ pub fn main_loop(
...
             let (wait_duration, _) = update_loop.get_event_wait_time();
-            let event = if wait_duration.as_nanos() == 0 {
-                rx.try_recv()
-                    .map_err(|e| matches!(e, TryRecvError::Disconnected))
-            } else {
-                rx.recv_timeout(wait_duration)
-                    .map_err(|e| matches!(e, RecvTimeoutError::Disconnected))
-            };
+            let event = rx
+                .recv_timeout(wait_duration)
```
fredizzimo, line 254: I seem to recall that I had some problems with a zero duration and `recv_timeout` when I wrote the code originally. But I can't find anything in the documentation or in the rust bug reports. I also can't reproduce it, so maybe either I had a different code structure when I noticed the problem, or my code had some other problem at that time.

sid-6581: I also looked around for reasons not to call it with a zero duration, just in case, but didn't see anything. I traced through the `recv_timeout` source code as well to make sure, and didn't see anything that would lead me to think that anything bad would happen. Everything appears to be running fine this way.

fredizzimo:
Thank you!

PR #2130, opened 2023-11-14
fix: don't wait forever for missing nvim process

### LINKED ISSUE #1823: Neovide hangs when nvim exits unexpectedly
#### Describe the bug
On Windows, when the `nvim` service exits, instead of closing the window, NeoVide hangs and becomes unresponsive.

#### To Reproduce
The easiest way is to have swap-files enabled and open the same file in two different windows, which causes the classic "Swap file...already exists" error, then pressing `a` for "(A)bort".

I'm not sure if this happens on all platforms, so I've used Windows path separators.

1. Open a file: `neovide ~\tmp\foo`
2. Open a second Neovide window with the same file: `neovide ~\tmp\foo`
3. Press `a` to abort the `nvim` process

#### Expected behavior
The Neovide window connected to the nvim session should close.

#### Screenshots
Immediately after nvim exits:
image

After Windows notices the process isn't responding:
image

#### Desktop (please complete the following information):
 - OS: Windows 11 (stable)
 - Neovide Version: 0.10.3 (installed with Chocolatey)
 - Neovim Version: v0.8.0-1210-gd367ed9b2

#### Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:
I've attached this as a file.

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/11133231/neovide_rCURRENT.log)
199-f44e6731-b76d-44cd-a0fe-b3e916fe2a7b.png)

sid-6581:
- Fix

- No

A lot of the time after the nvim.exe process is gone on Windows, neovide doesn't detect that the process is gone. `session.io_handle.await` just waits forever, even when the neovim process is clearly gone (not visible in Process Explorer), including `conhost`. This means that when `NeovimRuntime` drops and blocks on the tasks to finish, neovide hangs forever.

I'm not really happy with how crude this solution is, but I'm not quite sure where the problem really lies after a bit of debugging. Not sure if it's a problem with `nvim-rs`, or `tokio`, or some other internal deadlock. This will at least work for now, because these hangs happen very, very frequently for me. It should also be a safe enough hack, since `NeovimRuntime` should only be dropped when neovide is exiting anyway.

I'm guessing this will also fix #1823

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -11,11 +11,8 @@ use itertools::Itertools;
 use log::{error, info};
 use nvim_rs::{error::CallError, Neovim, UiAttachOptions, Value};
 use rmpv::Utf8String;
-use std::{io::Error, ops::Add, sync::Arc};
-use tokio::{
-    runtime::{Builder, Runtime},
-    task::JoinHandle,
-};
+use std::{io::Error, ops::Add, sync::Arc, time::Duration};
+use tokio::runtime::{Builder, Runtime};
 
 use crate::{cmd_line::CmdLineSettings, dimensions::Dimensions, running_tracker::*, settings::*};
 use handler::NeovimHandler;
@@ -30,14 +27,8 @@ pub use ui_commands::{start_ui_command_handler, ParallelCommand, SerialCommand, 
 const INTRO_MESSAGE_LUA: &str = include_str!("../../lua/intro.lua");
 const NEOVIM_REQUIRED_VERSION: &str = "0.9.2";
 
-enum RuntimeState {
-    Idle,
-    Attached(JoinHandle<()>),
-}
-
 pub struct NeovimRuntime {
-    runtime: Runtime,
-    state: RuntimeState,
+    runtime: Option<Runtime>,
 }
 
 fn neovim_instance() -> Result<NeovimInstance> {
@@ -154,25 +145,23 @@ impl NeovimRuntime {
         let runtime = Builder::new_multi_thread().enable_all().build()?;
 
         Ok(Self {
-            runtime,
-            state: RuntimeState::Idle,
+            runtime: Some(runtime),
         })
     }
 
     pub fn launch(&mut self, grid_size: Option<Dimensions>) -> Result<()> {
-        assert!(matches!(self.state, RuntimeState::Idle));
-        let session = self.runtime.block_on(launch(grid_size))?;
-        self.state = RuntimeState::Attached(self.runtime.spawn(run(session)));
+        let runtime = self.runtime.as_ref().unwrap();
+        let session = runtime.block_on(launch(grid_size))?;
+        runtime.spawn(run(session));
         Ok(())
     }
 }
 
 impl Drop for NeovimRuntime {
     fn drop(&mut self) {
-        if let RuntimeState::Attached(join_handle) =
-            std::mem::replace(&mut self.state, RuntimeState::Idle)
-        {
-            let _ = self.runtime.block_on(join_handle);
-        }
+        self.runtime
+            .take()
+            .unwrap()
+            .shutdown_timeout(Duration::from_secs(1));
     }
 }


fredizzimo:
I need more time to look at this, in particular since I haven't seen this personally.  Also while 1 second  should be enough, I'm a little bit afraid of potential issues, since the the saving of the Neovim shada file, and closing of all external channels (LSP servers for example) are done after the quit request. So if those are not done in time, the process might get killed with corruption as a result.

sid-6581:
I believe the current rpc notification to signal a quit is done on VimLeavePre, should that be changed to VimLeave instead? I'm not sure if that could introduce other issues.

I tried to track down this bug for a while, and I don't really see anything in neovide that should be causing this. I'm suspecting either nvim-rs or tokio, but it's hard to say because no threads give any clues when this happens. The nvim processes are completely gone, yet the await never returns.

I can easily reproduce it on any of my systems, because it happens more than 1 in 10 times I shut down neovide. I'm not happy with a hacky timeout like this, but I don't see an easy way to fix it in neovide itself. We could also increase the one second to something higher, but it would still unfortunately be arbitrary.

fredizzimo:
> I believe the current rpc notification to signal a quit is done on VimLeavePre, should that be changed to VimLeave instead? I'm not sure if that could introduce other issues.

I changed it to `VimLeavePre` based on this https://github.com/neovim/neovim/issues/7727. And it makes sense it gives both Neovim and Neovide more time to finish the quit process independently. The notification in itself should not actually be required, other than for communicating the exit code when you quit with `:cquit`. The fallback is here https://github.com/neovide/neovide/blob/3e83b031defacb6906e7afe8229eab275a4f80ea/src/bridge/mod.rs#L149, and that should be triggered when neovim quits/crashes for local processes and for a broken streams for other connections, for example due to a network error.

If that is not triggered, then I suspect it's a bug in nvim-rs.

sid-6581:
@fredizzimo that's definitely not triggered, so I agree it's most likely a bug in nvim-rs. I'm going to try a different approach, see if I can propagate up the tokio child process and await both the nvim-rs future and the tokio process together until one of them resolves. If tokio notices that the child process is gone we should be safe to exit. I'll let you know how that goes.

sid-6581:
No luck, tokio doesn't detect that the process is gone.

fredizzimo:
Strange, I tried killing Neovim on both Windows 11 and Linux and in both cases it breaks out of that loop. Sometimes it triggers this before that though https://github.com/neovide/neovide/blob/f093080206a5aa96feba7d1b41962da8ae2d55ec/src/bridge/ui_commands.rs#L114-L115

sid-6581:
I haven't been able to get neovide to lock up when forcibly killing the nvim process either. So far I've only seen it when exiting cleanly with `qa`. I can reproduce it easily by running neovide from the command line and immediately hitting my keyboard shortcut that maps to `<CMD>qa<CR>`. If I do that a couple of times in a row, neovide hard locks and I can attach the debugger.

The only neovide code I can see in the threads is the wait for `block_on()` in `NeovimRuntime::drop()` and `blocking_recv()` in `start_editor`. Besides that it appears to mainly be tokio worker threads doing their thing. There's one thread that's locked trying to read, which I assume is `nvim-rs` still trying to read from the process.

tokio task dumping only works on linux, so I'm trying to do some tracing to see if the runtime can tell me what's going on with the tasks.

fredizzimo:
Yes, that has never happened to me. 

It does seem like it's tokio that does not detect the process exit then. I don't think the `start_editor` thread will exit until the `EVENT_AGGREGATOR` is destructed after main has exited.

fredizzimo:
I guess we need to figure out which (or both) of these two are blocking

https://github.com/KillTheMule/nvim-rs/blob/0d2b1c884f3c39a76b5b7aac0b429f4624843954/src/neovim.rs#L109-L111

fredizzimo:
This loop looks a bit suspicious https://github.com/KillTheMule/nvim-rs/blob/0d2b1c884f3c39a76b5b7aac0b429f4624843954/src/rpc/model.rs#L60-L74

I'm not sure if it will ever return if a partial message is decoded and the stream is broken.

sid-6581:
In theory the stream breaking should cause the `read` to return an `Err` immediately, thus breaking out of the loop in the match below.

I tried a `tokio::JoinSet` where I waited for the first to complete of `io_handle` or the `tokio::Child` for the nvim command. When neovide hangs, neither of those two return at all, so I don't know if `nvim-rs` is to blame here. I'm not seeing any `nvim-rs` code in the debugger call stacks, but I assume it's in some internal tokio queue waiting for a wakeup.

While I've debugged a lot of multithreading and async issues in other languages, I've never really had to debug tokio hangs before, so I'm learning this as I go along. I tried `tokio-console`, but I don't see anything that stands out (I updated tokio to the latest version for testing, just to eliminate the possibility of a bug being fixed):

image

sid-6581:
I've been doing some more searching, and I've found a lot of people with hangs like this:

https://github.com/tokio-rs/tokio/issues/2318

sid-6581:
@fredizzimo another thing I tried that seemed to work was adding an autocmd on `VimLeave` to call `chanclose()` on the neovim side. That seems to reliably force `nvim-rs` to break out of the await, and allows neovide to close. I haven't found a cleaner way yet, but it seems reasonable. Thoughts?

Also, is there a reason the runtime is created manually with all the extra code needed to deal with that as opposed to just using `tokio::main`? I wasn't sure if there was a reason it was done this way.

sid-6581:
In the latest commit using the latest nightly I actually can't reproduce this issue, so maybe some package update fixed it or something was fixed in neovim itself. I'll close this for now since it appears to not be needed anymore.

Issue #2132, opened 2023-11-15
Panic closing the file 


dev-ardi:
**Describe the bug**
Panic closing the file 

**To Reproduce**
1. Close neovide

**Expected behavior**
Closes normally but instead it panics

**Desktop (please complete the following information):**

`Linux arch 6.6.1-arch1-1 #1 SMP PREEMPT_DYNAMIC Wed, 08 Nov 2023 16:05:38 +0000 x86_64 GNU/Linux`
```
NVIM v0.9.4
Build type: Release
LuaJIT 2.1.1699392533
  ```
`neovide 0.11.2`

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide_backtraces.log](https://github.com/neovide/neovide/files/13366657/neovide_backtraces.log)
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/13366658/neovide_rCURRENT.log)

**Additional context**

dev-ardi:
Solved with `sudo chown ardi /home/ardi/.local/share/nvim/neovide-settings.json`.

I think that at some point I ran neovide as sudo because it had root permissions

fredizzimo:
Edit: I saw your message after writing this, but I think the below can still be an issue.

Are you running neovim from an AppImage or something else where Neovide would not have write permission to? Remote connections to Neovim could also cause the issue.

We save our data files to the data path of Neovim for some reason (I have no idea why).

dev-ardi:
> We save our data files to the data path of Neovim for some reason

That's not where you should save it!

agraven:
This issue should presumably be considered addressed by #2133 and #2135

fredizzimo:
I agree closing this

PR #2133, opened 2023-11-15
Extra context if can't save to file

dev-ardi:
- Fix

- No

Gives more context for errors such as #2132 so that the user can fix them.

### DIFF
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -203,12 +203,14 @@ mod tests {
     #[test]
     #[serial]
     fn test_files_to_open_with_passthrough() {
-        let args: Vec<String> = ["neovide",
+        let args: Vec<String> = [
+            "neovide",
             "--no-tabs",
             "./foo.txt",
             "./bar.md",
             "--",
-            "--clean"]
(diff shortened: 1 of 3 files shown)


fredizzimo (changes requested):
Can you format the code with `cargo fmt`?

There are also a couple of clippy errors, see the "files changed" tab for the output, or run `cargo clippy` locally.

dev-ardi:
I should have checked before, my bad

dev-ardi:
No idea why it was failing on those files, I didn't change them

fredizzimo:
Thanks you!

> No idea why it was failing on those files, I didn't change them

Each new version of clippy adds a few more checks, and there's no way to lock to a specific version, the newest version is always used. So unfortunately that's expected.

dev-ardi:
I meant fmt was failing! maybe someone commited directly to master without passing through CI?

fredizzimo:
Your `cargo clippy --fix` commit actually touched those files. 

Rust 1.74 was released the 16th November, so that's why you started to see new clippy warnings.

PR #2135, opened 2023-11-16
fix: Use own data directory instead of nvim's

agraven:
Use a separate directory for storing application data, i.e. window settings, instead of re-using Neovim's.

- Fix
    - See #2132

- Arguably

Existing window settings will be lost when using a neovide version with this change, but that data is arguably somewhat ephemeral in nature, or at least trivial to recreate.

### DIFF
--- a/src/settings/window_size.rs
+++ b/src/settings/window_size.rs
@@ -40,21 +40,12 @@ struct PersistentSettings {
     window: PersistentWindowSettings,
 }
 
-#[cfg(windows)]
-fn neovim_std_datapath() -> PathBuf {
-    let mut data_path = dirs::home_dir().unwrap();
-    data_path.push("AppData/local/nvim-data");
-    data_path
-}
-
-#[cfg(unix)]
-fn neovim_std_datapath() -> PathBuf {
-    let xdg_dirs = xdg::BaseDirectories::with_prefix("nvim").unwrap();
-    xdg_dirs.get_data_home()
+fn neovide_std_datapath() -> PathBuf {
+    dirs::data_local_dir().unwrap().join("neovide")
 }
 
 fn settings_path() -> PathBuf {
-    let mut settings_path = neovim_std_datapath();
+    let mut settings_path = neovide_std_datapath();
     settings_path.push(SETTINGS_FILE);
     settings_path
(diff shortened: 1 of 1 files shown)


sid-6581:
I agree the location should change, but the function name `neovim_std_datapath` is a bit misleading after this change. Also, I don't think the file should be saved directly in the root of `AppData/Local` on Windows. I also don't know if `.config/neovide` is the right place for what I agree is ephemeral data. Maybe `XDG_STATE_HOME`?

As an aside, is the `xdg` crate really needed? Seems like `dirs` handles both Windows and Linux, while `xdg` overlaps in functionality and handles only Linux.

agraven:
Reasonable points all around. I updated the function names and unified the implementation to use `dirs` everywhere. I didn't remove `xdg` as it's still used in other places, and I figured a larger refactor would be out of scope.

re: it's location on unix, it will be going to `.local/share/neovide` with the current implementation, not `.config/neovide`

Kethku:
Hey thanks for the change! I'm spending some time looking into updating our configuration file and data path and this looks like the right way to go. Further, the dirs crate which we've used elsewhere seems to supersede the xdg crate so I'm inclined to use it instead everywhere we need known directories.

But I can snag that change while I'm in the area. I'll merge this in the meantime

dev-ardi:
> Reasonable points all around. I updated the function names and unified the implementation to use `dirs` everywhere. I didn't remove `xdg` as it's still used in other places, and I figured a larger refactor would be out of scope.
> 
> re: it's location on unix, it will be going to `.local/share/neovide` with the current implementation, not `.config/neovide`

I'm not a fan of applications unnecesarily polluting my `.config`. Thanks for using `.local/share`!

fredizzimo:
Another thing that wasn't really done was to refactor all our configuration to use common functions instead of being scattered all around the place.

https://github.com/neovide/neovide/pull/1119#discussion_r1218465807

fredizzimo:
The above two comments are mostly for @Kethku since she is going to look more into this.

sid-6581:
Another thing that should be fixed is that the json file is loaded multiple times instead of being loaded once and cached.

PR #2149, opened 2023-11-27
fix: Don't poll the event loop on Wayland to reduce the power usage

fredizzimo:
When the winit vsync is used, the event loop should never be polled, otherwise the CPU usage is just wasted. Winit will send the redraw request and wake the loop up when it's time to render anyway.

All other ways of doing vsync, just blocks inside the loop itself, so polling should be used there.

This also updates the tracy profiler to the latest version.

- No


Kethku:
LGTM

Issue #2136, opened 2023-11-17
Windows with a height of 0 cause Neovide to freeze when using multigrid


Ontonator:
**Describe the bug**
When multigrid is enabled, causing a window with a height of 0 cause Neovide to freeze and become unresponsive.

**To Reproduce**
1. Open neovide
2. Run `:set winminheight=0`
3. Split the window with `<C-W>s`
4. Increase the size of the active window such that the inactive window's height becomes 0 using `1000<C-W>_`

**Expected behavior**
The active window should expand to fill as much space as it can. The inactive window should shrink to a height of 0. Neovide should remain responsive.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows 11 22H2 with WSL Ubuntu 22.04
 - Neovide Version: 0.11.2 (f093080206a5aa96feba7d1b41962da8ae2d55ec)
 - Neovim Version: NVIM v0.10.0-dev

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
```log
2023-11-17 00:21:37 - Neovide panicked with the message 'attempt to calculate the remainder with a divisor of zero'. (File: /rustc/79e9716c980570bfd1f666e3b16ac583f0168962\library\core\src\num\mod.rs; Line: 455, Column: 5)
   0: backtrace::backtrace::dbghelp::trace
             at C:\Users\[redacted]\.cargo\registry\src\index.crates.io-6f17d22bba15001f\backtrace-0.3.69\src\backtrace\dbghelp.rs:98
   1: backtrace::backtrace::trace_unsynchronized<backtrace::capture::impl$1::create::closure_env$0>
             at C:\Users\[redacted]\.cargo\registry\src\index.crates.io-6f17d22bba15001f\backtrace-0.3.69\src\backtrace\mod.rs:66
   2: backtrace::backtrace::trace<backtrace::capture::impl$1::create::closure_env$0>
             at C:\Users\[redacted]\.cargo\registry\src\index.crates.io-6f17d22bba15001f\backtrace-0.3.69\src\backtrace\mod.rs:53
   3: backtrace::capture::Backtrace::create
             at C:\Users\[redacted]\.cargo\registry\src\index.crates.io-6f17d22bba15001f\backtrace-0.3.69\src\capture.rs:176
   4: backtrace::capture::Backtrace::new
[...]

Issue #1669, opened 2022-12-09
Crash when maximizing buffer


fdev31:
**Describe the bug**
neovide seems to crash when maximizing the current buffer

**To Reproduce**
1. Open Neovide
2. :split
3. Ctrl+W, _
4. *Crash*

**Expected behavior**
Selected buffer is maximized

**Screenshots**
Not applicable

**Desktop (please complete the following information):**
 - OS: Archlinux @ 9/12/2022
 - Neovide Version 0.10.3
 - Neovim Version 0.8.1
 
**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

```
DEBUG [neovide::bridge::command] Starting neovim with: Command { std: "/usr/bin/nvim" "--embed" "-p", kill_on_drop: false }
INFO [neovide::bridge::setup] Neovide registered to nvim with channel id 1
INFO [neovide::bridge] Neovim process attached
DEBUG [neovide::settings::window_geometry] Loaded window settings: Windowed { position: PhysicalPosition { x: 2, y: 530 }, size: Dimensions { width: 95, height: 24 } }
DEBUG [neovide::bridge::events] Ignored style attribute: "nocombine"
DEBUG [neovide::bridge::events] Ignored style attribute: "nocombine"
DEBUG [neovide::bridge::events] Ignored style attribute: "nocombine"
DEBUG [neovide::bridge::events] Ignored style attribute: "nocombine"
DEBUG [neovide::renderer::fonts::caching_shaper] Original font_size: 18.67px
DEBUG [neovide::renderer::fonts::caching_shaper] Original font_width: 11.49px
DEBUG [neovide::renderer::fonts::caching_shaper] Font width: 11.49px (avg: 11.49px)
DEBUG [neovide::renderer::fonts::caching_shaper] Fudge factor: 0.96
DEBUG [neovide::renderer::fonts::caching_shaper] Fudged font size: 17.88px
DEBUG [neovide::renderer::fonts::caching_shaper] Fudged font width: 11.00px
INFO [neovide::window] window created (scale_factor: 1.0000, font_dimensions: Dimensions { width: 11, height: 22 })
DEBUG [neovide::renderer::fonts::caching_shaper] Updating font: 
DEBUG [neovide::renderer::fonts::caching_shaper] Font updated to: 
DEBUG [neovide::renderer::fonts::caching_shaper] Original font_size: 18.67px
[...]

PR #2150, opened 2023-11-27
Fix crash with zero height windows

fredizzimo:
Fixes 
* https://github.com/neovide/neovide/issues/1669
* https://github.com/neovide/neovide/issues/2136

- No


Kethku:
LGTM. Merge when ready

PR #2151, opened 2023-11-27
fix: Try to render with vsync even when animations are disabled

fredizzimo:
TODO: More info

- No

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2849,9 +2849,9 @@ checksum = "c06d3da6113f116aaee68e4d601191614c9053067f9ab7f6edbcb161237daa54"
 
 [[package]]
 name = "tracy-client-sys"
-version = "0.19.0"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7fcbdba03a3cfc5f757469fd5b6d795fc461484c97e47e94b0fc7db93261d9c5"
+checksum = "3db0b1cc1bb12a70457300d9affc07acb587390d971a796dac2f4d9bca8df776"
 dependencies = [
  "cc",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,13 @@ default = []
 embed-fonts = []
 profiling = ["dep:tracy-client-sys"]
 gpu_profiling = ["profiling"]
+# Corresponds to https://github.com/nagisa/rust_tracy_client/blob/main/FEATURES.mkd
+tracy-fibers = ["tracy-client-sys?/fibers"]
+tracy-system-tracing = ["tracy-client-sys?/system-tracing"]
+tracy-context-switch-tracing = ["tracy-client-sys?/context-switch-tracing"]
+tracy-sampling = ["tracy-client-sys?/sampling"]
+tracy-code-transfer = ["tracy-client-sys?/code-transfer"]
+tracy-callstack-inlines = ["tracy-client-sys?/callstack-inlines"]
 
 [dependencies]
 anyhow = { version = "1.0.75", features = ["backtrace"] }
@@ -58,7 +65,7 @@ time = "0.3.9"
 tokio = { version = "1.25.0", features = ["full"] }
 tokio-util = { version = "0.7.4", features = ["compat"] }
 toml = "0.7.3"
-tracy-client-sys = { version = "0.19.0", optional = true }
+tracy-client-sys = { version = "0.22.0", optional = true, default-features = false, features = ["broadcast", "delayed-init", "enable", "manual-lifetime"] }
 unicode-segmentation = "1.9.0"
 which = "4.2.5"
 winit = { version = "=0.29.4", features = ["serde"] }
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -245,7 +245,7 @@ pub fn main_loop(
                 break;
             }
 
-            let (wait_duration, _) = update_loop.get_event_wait_time();
+            let (wait_duration, _) = update_loop.get_event_wait_time(&window_wrapper.vsync);
             let event = rx
                 .recv_timeout(wait_duration)
                 .map_err(|e| matches!(e, RecvTimeoutError::Disconnected));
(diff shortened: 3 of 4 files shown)


9mm:
hmm, its slightly better but still not nearly what just explicitly setting the FPS to something high is. its definitely an imporvement, i would say its less noticeable now, whereas before it was very noticeable

fredizzimo:
Could you take two different captures? One with the high refresh rate and the other one with this. Both without the `--no-idle` flag.

This should allow me to compare the two scenarios and hopefully see what's wrong.

9mm:
Hmm yeah, actually on this branch even the vsync=0 999 fps isnt smooth anymore, im not sure if thats expected 

So for that version i went to the other branch from about 5 hours agin stead of this one (#2149)

# Version 1: `vsync1_default` - this is from this branch, VSYNC is off via ENV variable, and my "refresh_rate" line config is commented out

# Version 2: with `960` set, this config like:

  vim.g.neovide_refresh_rate = 960 -- fps peaks over 1000+ (requires NEOVIDE_VSYNC=0)

and also the env var set to `1`

TBH the results seemed the same... it was hard for me to tell them apart, there was lagging on both, so maybe the change that affects the vsync0 version is on that other PR as well?

If I just go like this however...

```
NEOVIDE_VSYNC=0 /Users/zesty/.cargo/bin/neovide --no-fork --no-idle
```

^ this also is using the 960, from a 'main' compiled version. It is **exceptionally** smooth. Theres literally not even a single stutter

Whether that has anything to do with the main branch vs this branch, or the fact maybe the profiler slows it down im not sure...
[Archive.zip](https://github.com/neovide/neovide/files/13477993/Archive.zip)

fredizzimo:
Hm, I wonder if you provided the correct trace for the 960 refresh rate since both run at 120 FPS?

Both traces also seem to use the display link, which indicates that vsync is on (that's the default). 

BTW, the frame rate is actually very stable (just a couple of microseconds off the vsync) in the captures, it's just measured at the wrong time, after swap buffers has finished, instead of when vsync wait is finished.

It's possible that this is too early and causes a race condition though. At least it introduces one frame of extra latency. I think we could try switching the `swap_buffers` and `wait_for_vsync` around here https://github.com/neovide/neovide/blob/c2c2179060f89766939b036f72b8404f7aeb7536/src/window/window_wrapper.rs#L327-L334.

9mm:
Ill check again, its possible but i tried to be very careful that i ran the commands right to not waste time. i did use the VSYNC env variable before the command, so ideally that should have worked, ill check again a bit later

fredizzimo:
BTW, you can open the trace yourself to check the frame rate. Just open the GUI tracy profiler, and "open saved trace". And if you have already opened a trace, the red "power" button.

fredizzimo:
Actually, I think you can ignore this branch, at least for now. The parameter `--no-idle` should have the same effect. 

So, I think captures from https://github.com/neovide/neovide/pull/2149 is enough
1. `--no-idle` default (vsync enabled)
2. `--no-idle --no-vsync` refresh rate 960

And then maybe 1. with the wait and swap changed the other way as explained a few messages up.

fredizzimo:
BTW, I'm not surprised that it's choppy with a zero delay animation, but I am surprised that you see an improvement with vsync disabled and a high refresh rate. For me it's just as choppy with both. The longer the animation is, the smoother it gets. 

And the reason for that is clear, it's the uneven pacing of the keyrepeats, or rather updates triggered by the presses from Neovim, and a longer animation masks that.

9mm:
Ive had a lot of urgent fires today (primary DB got 20x the amount of traffic etc), but i will try to make a quicktime video so hopefully there are no dropped frames in the animation to see if it shows up in the video

fredizzimo:
No, problems, I'm not expecting you to answer immediately, even if I write many messages :smile:

Feel free to do it whenever you have more time, even if it takes days.

9mm:
haha right on, thanks 🙏🏼

9mm:
I didnt bother re-capturing the 960 one because even if it was wrong, both "versions" that I tested were both quite laggy. It's almost not even worth doing a comparison there because my original thought was on that branch I could show you how its smooth, but i think something else changed which affects that FPS method, so instead... I moved your tracy update on to main and just did a real test there to have something to compare to.

This is what it looks like when its perfectly smooth

9mm:
Ugh 1 moment i hit enter too soon

9mm:
It turns out the Env variable didnt work and thats probably why it uploaded the same thing twice. I had to explicitly use `--no-vsync` option

Config:

```lua
  vim.g.neovide_refresh_rate = 960 -- fps peaks over 1000+ (requires NEOVIDE_VSYNC=0)
  vim.g.neovide_refresh_rate_idle = 60
  vim.g.neovide_hide_mouse_when_typing = true
  vim.g.neovide_underline_stroke_scale = 1.2

  vim.g.neovide_cursor_animation_length = 0
  vim.g.neovide_cursor_trail_size = 0
  vim.g.neovide_scroll_animation_length = 0
  vim.g.neovide_cursor_animate_command_line = false
```

Command line:

```
NEOVIDE_VSYNC=0 tracy-capture -o vsync.tracy & cargo run --release --features profiling -- --no-fork --no-idle
```

Branch: Main, with tracy

Here it is extremely smooth

[vsync.tracy.zip](https://github.com/neovide/neovide/files/13489165/vsync.tracy.zip)

9mm:
Here are raw quicktime files.

Note that the difference HAS improved since the first time, it used to be very bad, now its noticeably improved and the gap is less between these 2 videos. its mostly a lot of "stuttering" which is noticeable

Please tell me when you download these so i can delete them

**note** also download them and watch, dropbox will try to compress them if you preview them there

960FPS:

https://www.dropbox.com/scl/fi/swlwrs0orkku61qnvol2w/vsync_0_960.mov?rlkey=2hg5tt65o6n3xkgcnyz0yijcc&dl=1

this branch:

https://www.dropbox.com/scl/fi/ac1tvmqz11po3jc7w35ve/vsync_1.mov?rlkey=6rioclnaabrbue5tbtn34fpea&dl=1

fredizzimo:
@9mm, I'm still not sure what's wrong. But I think it has to do with the update frequency of the inputs. We can only read inputs and process the results back from neovim between the frames, so with a higher frame rate the inputs are read more often and the timing of them might also be more accurate.

I'm not really sure if there's anything thing that can easily be done about it, but I think that's ok for now, we can add a faq entry for this special case. The default settings are smooth so that should be enough for the majority.

fredizzimo:
I'm closing this, since I don't think it will help.

PR #2152, opened 2023-11-29
fix: Change the order of vsync wait and swap buffers to reduce latency

fredizzimo:
This changes the order of waiting for vsync and swapping the buffers. This reduces the input latency since reading the inputs and the swap happens during the same vsync interval, rather than different ones.

- No


fredizzimo:
@9mm. I don't think this fixes your problem. But if you want you can try if it has any effect.

Kethku:
LGTM

Issue #2140, opened 2023-11-20
Crash using multigrid when loading certain Session.vim files


Ontonator:
**Describe the bug**
Neovide crashes when certain commands are run. (Sorry I can't provide a better description, I have no idea why this happens.)

**To Reproduce**
1. Create a .vim file containing the following minimised script:
```vim
set shortmess=A
tabnew +setlocal\ bufhidden=wipe
resize 1
tabnext
edit newfile.txt
```
2. Run Neovide with multigrid enabled
3. `:so` the script

**Expected behavior**
Neovide should not crash and should exhibit the same behaviour as without multigrid.

**Screenshots**

**Desktop (please complete the following information):**
- OS: Windows 11 22H2 with WSL Ubuntu 22.04
- Neovide Version: 0.11.2 (59e4ed47e72076bc8cec09f11d73c389624b19fc)
- Neovim Version: NVIM v0.10.0-dev

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
```log
2023-11-20 04:18:23 - Neovide panicked with the message 'Could not parse event from neovim: invalid event format for event 'grid_cursor_goto' - [Integer(PosInt(3)), Integer(NegInt(-57)), Integer(PosInt(19))] - invalid u64 format -57'. (File: src\error_handling.rs; Line: 18, Column: 5)
   0: backtrace::backtrace::dbghelp::trace
             at C:\Users\[redacted]\.cargo\registry\src\index.crates.io-6f17d22bba15001f\backtrace-0.3.69\src\backtrace\dbghelp.rs:98
   1: backtrace::backtrace::trace_unsynchronized<backtrace::capture::impl$1::create::closure_env$0>
             at C:\Users\[redacted]\.cargo\registry\src\index.crates.io-6f17d22bba15001f\backtrace-0.3.69\src\backtrace\mod.rs:66
   2: backtrace::backtrace::trace<backtrace::capture::impl$1::create::closure_env$0>
             at C:\Users\[redacted]\.cargo\registry\src\index.crates.io-6f17d22bba15001f\backtrace-0.3.69\src\backtrace\mod.rs:53
   3: backtrace::capture::Backtrace::create
             at C:\Users\[redacted]\.cargo\registry\src\index.crates.io-6f17d22bba15001f\backtrace-0.3.69\src\capture.rs:176
   4: backtrace::capture::Backtrace::new
[...]

Issue #1949, opened 2023-07-27
Neovide crashes when searching, with two line config


mawkler:
**Describe the bug**
Neovide started crashing for me when when starting a search, and I've boiled it down to the two following lines:

```lua
vim.o.cmdheight = 0
vim.keymap.set('n', '/', 'ms/', { silent = true })
```

**To reproduce**

1. `neovide -- -u NONE -c "lua vim.o.cmdheight = 0" -c "lua vim.keymap.set('n', '/', 'ms/', { silent = true })"`
2. Type `/f` (i.e. search for any character)
3. Neovide crashes

**Expected behavior**

Neovide doesn't crash

**Desktop:**
 - OS: Manjaro Linux x86_64
 - Neovide Version 0.10.4
 - Neovim Version v0.10.0-dev-735+g8fe9f41f7

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

GitHub complained that my post was too long so I cropped out a couple of hundred lines in the middle of the log that hopefully aren't important.

<details>
<summary><b>Click to expand</b></summary>

```
TRACE [neovide] Neovide version: 0.10.4
DEBUG [neovide::bridge::command] Starting neovim with: Command { std: "/usr/bin/nvim" "--embed" "-p" "-u" "NONE" "-c" "lua vim.o.cmdheight = 0" "-c" "lua vim.keymap.set(\'n\', \'/\', \'ms/\', { silent = true })", kill_on_drop: false }
INFO [neovide::bridge::setup] Neovide registered to nvim with channel id 1
INFO [neovide::bridge] Neovim process attached
TRACE [neovide::bridge::handler] Neovim notification: "redraw"

[...]

TRACE [neovide::redraw_scheduler] Next frame queued
TRACE [neovide::channel_utils] neovide::bridge::ui_commands::UiCommand Serial(Keyboard("/"))
TRACE [neovide::bridge::ui_commands] Keyboard Input Sent: /
TRACE [neovide::bridge::handler] Neovim notification: "redraw"
TRACE [neovide::channel_utils] neovide::editor::EditorCommand NeovimRedrawEvent(ModeChange { mode: CmdLine, mode_index: 4 })
TRACE [neovide::channel_utils] neovide::editor::EditorCommand NeovimRedrawEvent(MouseOff)
TRACE [neovide::channel_utils] neovide::editor::EditorCommand NeovimRedrawEvent(Flush)
[...]

PR #2157, opened 2023-12-01
Ignore negative cursor positions

fredizzimo:
There are a few different Neovim bugs that causes invalid cursor positions to be sent. So instead of crashing Neovide, just give a warning and set the values to zero. In many cases the correct cursor position is sent right after that anyway. And in the worst case it just causes a visual glitch.

The drawback is that we don't get to know about the problems so they can be reported upstream as easily. But all the current known bugs are reported, and if there are no visual glitches then perhaps it does not need to be fixed at all.

* Fixes https://github.com/neovide/neovide/issues/2141
* Fixes https://github.com/neovide/neovide/issues/2140
* Fixes https://github.com/neovide/neovide/issues/1949

- No

### DIFF
--- a/src/bridge/events.rs
+++ b/src/bridge/events.rs
@@ -4,7 +4,7 @@ use std::{
     fmt::{self, Debug},
 };
 
-use log::debug;
+use log::{debug, warn};
 use rmpv::Value;
 use skia_safe::Color4f;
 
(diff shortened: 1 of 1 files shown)


MultisampledNight (approved):

MultisampledNight:
Thank you! :steam_locomotive: Choooooooooo chooooooooooo

PR #2161, opened 2023-12-02
feat: X1 and X2 mouse button support

### LINKED ISSUE #2148: Support for X1Mouse and X2Mouse

I have two mouse side-buttons, which in old Vim keycodes are typically `<X1Mouse>` and `<X2Mouse>`. I would like to be able to use them in Neovide. I use them all the time in other code environments to navigate forward and backward, and would do the same in Neovide + Neovim with binding them to `<C-O>` and `<C-I>`.

I would like the events to be sent to Neovim such that they map to X1Mouse and X2Mouse events.

I've checked if they're supported already, and don't seem to be. `C-V` followed by pressing them does nothing. As I can't capture an event myself, the only other thing I could do would be to remap the mouse buttons themselves to keybinds, but that would break them in, say, my browser instead.

**Additional context**

They appear to be available as events in winit: https://docs.rs/winit/latest/winit/event/enum.MouseButton.html

agraven:
Closes #2148 

Implement support for X1 and X2 mouse buttons (commonly known as "back" and "forward")

- Feature

- No

### DIFF
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -5,7 +5,10 @@ use log::trace;
 use anyhow::{Context, Result};
 use nvim_rs::{call_args, error::CallError, rpc::model::IntoVal, Neovim, Value};
 use strum::AsRefStr;
-use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver};
+use tokio::sync::{
+    mpsc::{unbounded_channel, UnboundedReceiver},
+    OnceCell,
+};
 
 use super::{show_error_message, show_intro_message};
 use crate::{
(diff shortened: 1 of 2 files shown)


fredizzimo:
Thank you, but I think we rather wait for your Nvim PR to get merged:
* https://github.com/neovim/neovim/pull/26279

We can add a api version check to enable the buttons when running against the nightly version. Not that many people have requested this support, so I think it's fine supporting it only together with the nightly version.

I'm leaving this open still, in case other maintainers have different opinions.

agraven:
@fredizzimo I'm giving a proper implementation a go and there's one thing I would appreciate input on, where should the version check be performed and stored?

fredizzimo:
Hi @agraven,

We already query the api information here https://github.com/neovide/neovide/blob/ce0174a17674b1984c411e9f81b752d81b1b83d4/src/bridge/setup.rs#L48-L51
So at first, I thought that it would make sense to do it there, and maybe query for more information if needed. But then we need to store it in a lazy static or something, and because the code is asynchronous the rest of the systems also need to be prepared to deal with the situation where it's not yet available.

So,  I think another approach is better. Don't do any checks in the MouseManager itself, just pretend that it works, and make the actual RPC call a NOP in https://github.com/neovide/neovide/blob/main/src/bridge/ui_commands.rs. 

That means that you can query it at the startup of the SerialCommand thread here 
https://github.com/neovide/neovide/blob/ce0174a17674b1984c411e9f81b752d81b1b83d4/src/bridge/ui_commands.rs#L310-L321

And then just pass it on to the execute function, keeping it totally local to the task.

agraven:
I ended up implementing it with `OnceCell` instead since that felt like it had better locality when the check is going to be relevant to the serial command handling anyway. I'm happy to change it if you prefer though.

MultisampledNight:
Thank you! Choooooooooo choooooooooooo :steam_locomotive:!

Issue #1550, opened 2022-09-22
winbar moves with the text in multigrid mode


mrshmllow:
Using the winbar in neovim moves around when scrolling, despite all other UI elements staying in place
![out](https://user-images.githubusercontent.com/40532058/191671025-d495508c-1815-42ad-9187-59e477fc614b.gif)

Im guessing is likely because the winbar is kinda tacked onto the window contents and not _really_ a UI element?

ronisbr:
Is there any work around for this issue? Like disable smooth scrolling with `multigrid`?

fredizzimo:
We have this issue open on the Neovim side https://github.com/neovim/neovim/issues/23610.

Unfortunately, it has since then turned out that the built-in borders of floating windows also suffers from the same problem, so we might have to think about some better option to for solving both. I'm not sure if we can do it in Neovide alone or not.

fredizzimo:
There's a fix for this in https://github.com/neovide/neovide/pull/2165

mikaraunio:
Hmm, I'm still seeing this with lualine winbars in 0.12.2. Scrolling in windows with borders is fine, though.

thetic:
i am also seeing this issue in 0.12.2

fredizzimo:
The remaining with the winbar should be fixed on main together with Neovim nightly with:
* https://github.com/neovide/neovide/pull/2455
But there was a regression for the floating window borders, which will be fixed here
* https://github.com/neovide/neovide/pull/2487

Issue #1446, opened 2022-08-03
Floating windows with borders doesn't work well with smooth scrolling


archiif:
**Describe the bug**
When smooth scrolling is activated with `--multigrid`, scrolling in a floating window with borders causes the border to shake, making it really dizzying and hard to follow.

**To Reproduce**
1. Launch neovide with `--multigrid`
2. Spawn a floating window with borders
3. Scroll in it

**Expected behavior**
The border in the floating window should not shake when scrolling

**Screenshots**
![2022-08-03_08-25-21](https://user-images.githubusercontent.com/60074051/182504416-6b7193af-cbe7-4bab-b2de-16e61c729d19.gif)

**Desktop (please complete the following information):**
 - OS: Microsoft Windows [Version 10.0.19043.1826]
 - Neovide Version 0.9.0
 - Neovim Version v0.8.0-dev-739-ge59bc078d

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/9247226/neovide_rCURRENT.log)

**Additional context**

AlexvZyl:
Yeah, I would love to try and fix it.  If one of the maintainers can point me in the right direction I would love to give it a shot (assuming that it is not super difficult...).

PR #2165, opened 2023-12-03
fix: Scrolling of windows with winbar or native borders

fredizzimo:
This fixes the scrolling of windows with winbar or native borders by only scrolling the inner region. 

The solution is quite hacky, see the new `neovide_unlink_border_highlights` option for example. But I think this is the best we can do with the current version of Neovim.

I also changed `Arc<Mutex<Line>>`  `<RC<RefCell<Line>>` because it's never accessed concurrently.

* Fixes https://github.com/neovide/neovide/issues/1550
* Fixes https://github.com/neovide/neovide/issues/1446

**NOTE:** This probably needs a bit more testing before merging.

- No

### DIFF
--- a/src/bridge/events.rs
+++ b/src/bridge/events.rs
@@ -9,7 +9,9 @@ use rmpv::Value;
 use skia_safe::Color4f;
 use strum::AsRefStr;
 
-use crate::editor::{Colors, CursorMode, CursorShape, Style, UnderlineStyle};
+use crate::editor::{
+    Colors, CursorMode, CursorShape, HighlightInfo, HighlightKind, Style, UnderlineStyle,
+};
 
 #[derive(Clone, Debug)]
 pub enum ParseError {
@@ -476,7 +478,7 @@ fn parse_default_colors(default_colors_arguments: Vec<Value>) -> Result<RedrawEv
     })
 }
 
-fn parse_style(style_map: Value) -> Result<Style> {
+fn parse_style(style_map: Value, info_array: Value) -> Result<Style> {
     let attributes = parse_map(style_map)?;
 
     let mut style = Style::new(Colors::new(None, None, None));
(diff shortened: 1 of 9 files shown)


src/bridge/events.rs
```diff
@@ -523,13 +525,73 @@ fn parse_style(style_map: Value) -> Result<Style> {
...
+
+    for attribute in attributes {
+        if let (Value::String(name), value) = attribute {
+            match (name.as_str().unwrap(), value) {
+                ("kind", value) => {
+                    let kind_str = parse_string(value)?;
+                    match kind_str.as_str() {
+                        "ui" => kind = Some(HighlightKind::Ui),
+                        "syntax" => kind = Some(HighlightKind::Syntax),
+                        // The documentation says terminal but Neovim 0.9.4 sends term...
```
Kethku, line 552: Is there an issue to link here?

src/bridge/events.rs
```diff
@@ -523,13 +525,73 @@ fn parse_style(style_map: Value) -> Result<Style> {
...
+        "kind field not found in highlight info".to_string(),
+    ))?;
+    let ui_name = if kind == HighlightKind::Ui {
+        ui_name.ok_or(ParseError::Format(
+            "ui_name field not found in highlight info".to_string(),
+        ))?
+    } else {
+        String::default()
+    };
+    // hi_name can actually be absent for terminal, even though the documentation indicates otherwise
```
Kethku, line 578: same here

Kethku:
Looks good to me. Ship it ✅

cshuaimin:
I tested this and it works for simple case like `set winbar=%F`, but not works with dropbar.nvim. When I change the highlight group, winbar scrolls with the content too, try `set winbar=%#Visual#%F`.

MultisampledNight (approved):

cshuaimin:
That makes sense, thanks for the explanation. Float window border works well, great!

ttytm:
I greatly appreciate the recent work of @fredizzimo. I want to say thanks at this point.

For this PR, I don't think it is ready for a merge. With the change, when using a winbar content that is more complex than something like `set winbar=%F`, the winbar content changes when e.g. opening a telescope picker. A change like this should not come with a downside for cases like @cshuaimin is mentioning or for users using things like dropbar. Here fix is not working anyway atm. But yeah, unfortunately, this again requires that upstream neovim allows for a proper fix.

E.g. current master opening a telescope file picker
Screenshot_20231213_151804

With the changes:
Screenshot_20231213_152130

fredizzimo:
Hm.. this should be strictly better than main in all cases, winbars with custom highlight groups the bug should be the same as on main, and the content scrolls with the content like @cshuaimin explains,

I can't quite explain the different highlights in the screenshot above. But maybe the highlight groups are dynamically updated, and in that case the `neovide_unlink_border_highlights` hack does not work. @ttytm, can you test if setting that neovide option to false fixes it? 

If it does, there's no more hacks I can think of to get this work, and we will have to get a fix Neovim itself. Maybe we could merge this with the option set to false, to allow people to still have proper border and winbar scrolling with simple winbars and borders.

ttytm:
> [...] If it does, there's no more hacks I can think of to get this work, and we will have to get a fix Neovim itself. Maybe we could merge this with the option set to false, to allow people to still have proper border and winbar scrolling with simple winbars and borders.

Good suggestion, I unfortunately cannot test it today anymore. I'll leave an update tomorrow.

ttytm:
@fredizzimo the option has no effect on the mentioned behavior.

fredizzimo:
@ttytm, I think I need a Neovim configuration, so that I can re-produce the problem myself.

Ideally following this format https://github.com/neovim/neovim/blob/master/contrib/minimal.lua

But if that is not possible, a complete `.config/nvim` directory with lazy as the plugin manager is also OK, since I can use `NVIM_APPNAME` to run the config without messing with my own.

ttytm:
Hey @fredizzimo sorry bout the delay. I don't have the capacity atm to create a minimal repro. It's very unfortunate as I don't like to be a blocking factor. 

I can share a full nvim config where this happens with: https://github.com/tenxsoydev/nxvim.
Also, please feel free to ignore the mentioned objections.

fredizzimo:
@ttytm, I tested your configuration, and I can repeat the problems with the winbar highlights getting brighter when telescope is opened. But I can also repeat that on the current master, so there must be something else going on here, mabye some earlier PR that broke it.

In any case, I will try to make this more robust by comparing the content instead of using highlight groups. We know how many lines are scrolled, and from that we see which lines stay in place. If it works, it should not take long to implement, so I'm putting this as draft until then.

ttytm:
I'm discovering that it might not be a problem with Neovide but a problem with the config.
As I can reproduce the winbar problem when using nvim in Kitty on a mac. Somehow no the issue only appears in neovide with the PR but not in kitty at current master on Linux.

fredizzimo:
I'm moving this out of draft again. The content based approach has too many special cases, and I'm not confident that it can be reliable enough. 

My idea was basically to check how many lines have stayed in place at the top and bottom, every time the screen is updated, and consider those borders or winbars, which are treated the same way in this PR. The biggest issue with that approach is duplicate lines, but also lines that changes due to highlight updates and stuff like that, or winbars with dynamic content. 

So instead I think we can go with what we have, perhaps even with `neovide_unlink_border_highlights` set to false by default, to make sure that it has no negative effects at all. And people that want the support can set that to true. Note that it needs to be set before the colorscheme is loaded.

But I'm leaving it enabled by default for now, until I get more feedback.

src/renderer/rendered_window.rs
```diff
@@ -453,12 +507,37 @@ impl RenderedWindow {
...
+                                // The specification seems to indicate that kind should be UI and
+                                // then we only need to test ui_name. But at least for FloatTitle,
+                                // that is not the case, the kind is set to syntax and hi_name is
+                                // set.
+                                .map(|v| info.ui_name == *v || info.hi_name == *v)
+                                .any(|v| v)
+                            })
+                        })
+                    })
+                    .all(|v| v);
```
cshuaimin, line 534:
```suggestion
                    .any(|v| v);
```

> lines with purly border highlight groups are considered borders.

I believe if we change this to "lines **contain** border highlight grouops are considered borders", then dropbar.nvim will work! Don't know if this has false positives. @fredizzimo what do you think?

cshuaimin: However with this change, it seems that we need another way to ignore left and right borders.

fredizzimo: That's a great idea!

We could use all to detect the borders and any to detect winbars. I will try that later today.

fredizzimo: I implemented this, and it does seem to work. 

But I have only done a very quick test yet and it's a bit late for me to do more testing now.

MultisampledNight:
Thank you! Choooooooooooo choooooooooooo :steam_locomotive:

Issue #2164, opened 2023-12-02
"blinkon" doesn't blink


9mm:
**Describe the bug**

If using a vim setting like this:

```
vim.opt.guicursor = "a:blinkon100"
```

The square cursor doesn't blink

**Expected behavior**

It should blink 👁️

**Desktop (please complete the following information):**
 - OS: 14.1.2 (23B92)
 - Neovide Version Latest `main` branch
 - Neovim Version NVIM v0.9.4

fredizzimo:
I think it might need the full format `set guicursor=a:blinkwait700-blinkoff400-blinkon175` but that does not work either. The actual option works, since the shape can be changed.

9mm:
Yes sorry I should have added that.

I actually had `guicursor = ""` for the longest time, and only realized later hmm......... maybe I want to try blink.

fredizzimo:
This should be fixed by
* https://github.com/neovide/neovide/pull/2167

9mm:
Ah, I just noticed you already said that about the full format. I must have missed that the first time.

9mm:
thanks!

ddevaal:
Is this fixed in Neovide 0.12.0? I'm on 0.12.2 (macOS) and I can't get blinking to work. `vim.opt.guicursor = "a:blinkon100"` doesn't do anything for me. It would be so cool if we can get this animation too: https://github.com/neovide/neovide/pull/2167#issuecomment-1859037271 ❤️‍🔥❤️‍🔥❤️‍🔥

Issue #2131, opened 2023-11-15
Neovide transparency wont update until scrolling


unplugred:
Ticket describing the comment from https://github.com/neovide/neovide/issues/2095#issuecomment-1807059512

**Describe the bug**
I've got the following script to change the transparency when doing CTRL+SHIFT+Scroll
While it did operate correctly in 0.11.2, it broke when compiling from main.
The transparency does not update, until scrolling up and down the currently opened file.

**To Reproduce**
1. Write the following code in vimrc:
```
let g:neovide_transparency=0.9
highlight Normal ctermbg=none
highlight NonText ctermbg=none
function! ChangeTransparency(delta)
  let g:neovide_transparency = g:neovide_transparency + a:delta
endfunction
noremap <expr><C-S-ScrollWheelUp> ChangeTransparency(0.02)
noremap <expr><C-S-ScrollWheelDown> ChangeTransparency(-0.02)
```
2. Open a new instance of vim, CTRL+SHIFT+Scroll to change the transparency. The transparency won't change.
3. Scroll up and down the file. The transparency will update

UPDATE: adding `redraw!` after changing the transparency fixes it.

**Expected behavior**
Transparency should update when updating g:neovide_transparency

**Desktop:**
 - OS: Windows 11
 - Neovide Version: Commit b79b7cc0be18fd36b469713e55c558835178dad5
 - Neovim Version: v0.9.2

fredizzimo:
This should be fixed by:
* https://github.com/neovide/neovide/pull/2167

Issue #2162, opened 2023-12-02
Performance regression on Wayland with 6c9b9219dbbd43e3eb6be9fe725d68e8e9ee2561


wroyca:
**Describe the bug**
Starting with https://github.com/neovide/neovide/commit/6c9b9219dbbd43e3eb6be9fe725d68e8e9ee2561, there's a performance regression on Wayland, with noticeable latency and lag input while typing and scrolling.

**To Reproduce**
1. Checkout 6c9b9219dbbd43e3eb6be9fe725d68e8e9ee2561 (or master)
2. Install with Cargo.
3. Start using Neovim as you normally would.
4. See noticeable performance issues.

**Expected behavior**
This is not an issue with the preceding commit (c2c2179060f89766939b036f72b8404f7aeb7536), in which performance maintains proper smoothness with no latency.

**Desktop (please complete the following information):**
 - OS: Fedora 39 (GNOME 45)
 - Neovide Version: 0787fd1828acb2b87ae2b73c9482adc1dcd0976f
 - Neovim Version: v0.10.0-dev+1741-gc45d1e32a

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/13537059/neovide_rCURRENT.log)

**Additional context**
I run the kernel with ```preempt=full``` which may (or may not) impact said performance.

fredizzimo:
It should be no worse than other platforms, and it does not really bother me, so I haven't made a fix yet. But there's indeed a pointless latency, due to the event loop only waking up with a timer or when there are window events, and not when there are new updates from Neovim. We should probably deliver those as winit user events instead to properly wake up the loop.

A workaround is to set `neovide_refresh_rate_idle` to something smaller like 60 or 120.

wroyca:
> But there's indeed a pointless latency, due to the event loop only waking up with a timer or when there are window events, and not when there are new updates from Neovim. 

Thank for the insight!

> A workaround is to set `neovide_refresh_rate_idle` to something smaller like 60 or 120.

Unfortunately, that doesn't help on my side. I suspect this is very hardware-specific and that I'm perhaps I'm just one isolated case

fredizzimo:
Could you install tracy https://github.com/wolfpld/tracy and run `tracy-capture -o latency.tracy & cargo run --release --features profiling -- --no-fork`?  And attach `latency.tacy` here.

[...]

PR #2167, opened 2023-12-04
fix: Completely event driven update loop

fredizzimo:
* This changes all update loop events to be sent through the winit event queue. This reduces the latency, since the rendering starts immediately when the event is received, rather than waiting for the `neovide_refresh_rate_idle` wakeup.
* The way the cursor blink is updated has also changed, there was a regression due to some merge conflicts earlier. Now the cursor blink controls how long the event loop is waiting. 
* The test `window_separator_modifies_grid_and_sends_draw_command` had to be removed since supporting it would have been too complex requiring an event loop and everything.
* The `DrawCommandBatcher` has been greatly simplified
* The  `EventAggregator` has been eliminated, and replaced by either raw channels or the winit `event_loop_proxy` for winit events. `UICommand`s are now sent with `send_ui`
* The overall communication architecture has been improved, so that there's a clear flow between the components, which much less risk of mistakes, since there's no global `EventAggregator` that allows almost any kind of communication any longer. These changes have been documented in the architecture comment in `main.rs`

**NOTE:** The changes are rather big and a bit more testing would be good. So far I have only done basic testing on Wayland and Windows.

* Fixes https://github.com/neovide/neovide/issues/2164
* Fixes https://github.com/neovide/neovide/issues/2131
* Not completely sure about https://github.com/neovide/neovide/issues/2162

- No


9mm:
I'm not sure if im doing something wring but I tried this:

```
vim.opt.guicursor = "i:blinkon500"
```

No blinking

fredizzimo:
You need all three values: https://neovim.io/doc/user/options.html#'guicursor'
> blinkoff{N} blink times for cursor: blinkwait is the delay before the cursor starts blinking, blinkon is the time that the cursor is shown and blinkoff is the time that the cursor is not shown. Times are in msec. **When one of the numbers is zero, there is no blinking**

I implemented it according to that specification

fredizzimo:
@9mm, since this also changes how the event loop is run, could you also test if it makes any difference when you run with vsync?

AThePeanut4:
I'm getting what I believe is the same issue as #2162, so I've just tried this branch to see if it fixes it.

The easiest symptom to spot is that animations are "delayed" after being idle, sometimes to the point of being cut off entirely, meaning that the cursor "teleports" to its new location. It seems like Neovide is realising it should be animating the movement way too late - which fits perfectly with "rendering having to wait for the neovide_refresh_rate_idle wakeup".

As suggested in https://github.com/neovide/neovide/issues/2162#issuecomment-1837244375, I set `neovide_refresh_rate_idle` to `144` (my monitor's refresh rate), and that seems to fix the issue entirely for me. But I assume that isn't ideal, as then there's no "idling" at all.

On this branch, the issue is almost entirely gone, but not entirely. When using the keyboard to navigate, everything is good. But when using a mouse to click, the animations are still sometimes a bit delayed. Not always, but it is still happening.

I've included three screen recordings below which I hope demonstrate the issue: on the latest `main` commit, on this PR branch, and on the commit before 6c9b9219dbbd43e3eb6be9fe725d68e8e9ee2561, which is where this issue first started occurring.

### On the latest `main` (b6374265)

https://github.com/neovide/neovide/assets/49614525/6bb21dd3-b28a-404b-b82e-c5d55af5fc52

### On this branch

https://github.com/neovide/neovide/assets/49614525/e6bae652-75bb-4823-85e4-45f7ea805960

### On c2c2179060f89766939b036f72b8404f7aeb7536 (the commit before 6c9b9219dbbd43e3eb6be9fe725d68e8e9ee2561)

https://github.com/neovide/neovide/assets/49614525/bd553a2d-959e-4d55-b788-c0066e8883e9

AThePeanut4:
I think I've discovered why the mouse seems to trigger it more often. When using the keyboard, the `showcmd` info box thing seems to trigger a "wakeup" just before the cursor is actually moved, and Neovide is "ready" to start the animation on time. If I set `noshowcmd`, which I have done in the screen recording below, then I can trigger the delayed animations with the keyboard as well. It still doesn't happen every time, but it does happen.

https://github.com/neovide/neovide/assets/49614525/7e4aefbd-1439-457e-b380-717d71b4afc7

9mm:
> @9mm, since this also changes how the event loop is run, could you also test if it makes any difference when you run with vsync?

```
cargo run --release -- --vsync
```

On OSX on this PR i get the same performance as before, so I'd say that's good. It may be even slightly better but it's hard to tell, it is definitely not worse though.

fredizzimo:
@AThePeanut4, I don’t have those problems on my system with wayland. And it’s hard to tell without actual profiling data.
But it’s possible that wayland does not give us ”permission” to render when we request it. I will try to change it tomorrow to always render the first frame after being idle.

fredizzimo:
That is, don’t take this branch when `num_consecutive_rendered` is 0, if you want to try to modify it yourself. I’m not at the computer myself.

https://github.com/neovide/neovide/blob/aeaab024ce7b37785b73717beb7b3c341ff8bb38/src/window/update_loop.rs#L175

AThePeanut4:
> That is, don’t take this branch when `num_consecutive_rendered` is 0, if you want to try to modify it yourself. I’m not at the computer myself.
> 
> https://github.com/neovide/neovide/blob/aeaab024ce7b37785b73717beb7b3c341ff8bb38/src/window/update_loop.rs#L175

I made this change:
```diff
diff --git a/src/window/update_loop.rs b/src/window/update_loop.rs
index b67147a..cab0449 100644
--- a/src/window/update_loop.rs
+++ b/src/window/update_loop.rs
@@ -172,7 +172,7 @@ impl UpdateLoop {
             Ok(Event::AboutToWait) | Err(false) => {
                 self.should_render.update(window_wrapper.prepare_frame());
                 if self.should_render == ShouldRender::Immediately || !self.idle {
-                    if window_wrapper.vsync.uses_winit_throttling() {
+                    if self.num_consecutive_rendered > 0 && window_wrapper.vsync.uses_winit_throttling() {
                         window_wrapper.windowed_context.window().request_redraw();
                     } else {
                         self.render(window_wrapper);
```

but it didn't appear to make a difference unfortunately.

FYI, my system info:
- Compositor: Hyprland (latest commit https://github.com/hyprwm/Hyprland/commit/3bb9c7c5cf4f2ee30bf821501499f2308d616f94)
- OS: Arch Linux, with the `linux-zen` kernel (perhaps this is relevant?).
- CPU: Ryzen 7 5700X
- GPU: RX 6700
- And a 144Hz monitor, as already mentioned.

I've attached tracy captures, one with the change above ([latency_change.tracy.gz](https://github.com/neovide/neovide/files/13561070/latency_change.tracy.gz)) and one without ([latency.tracy.gz](https://github.com/neovide/neovide/files/13561068/latency.tracy.gz)), following these instructions from https://github.com/neovide/neovide/issues/2162#issuecomment-1837251296:
 > Could you install tracy https://github.com/wolfpld/tracy and run `tracy-capture -o latency.tracy & cargo run --release --features profiling -- --no-fork`? And attach `latency.tacy` here.

I hope they're useful - please let me know if you need me to do anything else.

fredizzimo:
@AThePeanut4, thanks for the traces. I can definitely see a change between those two captures. I reduces the latency, but it's probably not the latency you are seeing, since it's at most one frame at 120 FPS, so 8ms.

The changed versions renders as soon as the data is received from Neovim, which makes me think that the the latency is something else. 

I added some more profiling that should allow us to determine the exact delay from the input until it's rendered. But there's some delay until it's shown on the screen out of our control (the compositor and graphics pipeline).

BTW, intrestingly hyprland seem to reduce the frame rate to 60 FPS when the window is not active. Which could mess a bit with the frame dt detection we currently have, but I don't think that's the main problem here.

fredizzimo:
I changed this to draft, while figuring out the problems and cleaning up the code a bit.

AThePeanut4:
@fredizzimo Here's a new capture, with `noshowcmd` enabled, using the latest changes (e60ecf0). The animation lag is still there.

[latency.tracy.gz](https://github.com/neovide/neovide/files/13587923/latency.tracy.gz)

fredizzimo:
@AThePeanut4 don't worry, that's just the change you already tested, with the `num_consecutive_rendered` test.

fredizzimo:
Unfortunately I don't see any latency, in the traces, so there must be some other effect, for example

image

The first green is when we get the key press from winit, and the end of the swap buffer is when the rendered view goes out, and between those there's just 1.5ms.

And if you zoom out:
image

The animation is shown for 270ms with a constant 120 FPS. What's your `neovide_scroll_animation_lenght`, that should be roughly the same.

AThePeanut4:
@fredizzimo I set 
```lua
vim.g.neovide_scroll_animation_length = 0.1
vim.g.neovide_cursor_animation_length = 0.06
vim.g.neovide_cursor_trail_size = 0.3
vim.g.neovide_underline_stroke_scale = 0.5
```

If I comment those lines out (i.e. use the defaults), I still get the issue (the cursor "teleports").

Should I make a screen recording while doing a capture? Would that help?

fredizzimo:
I think there might be some small hitches after you unfocus the window or perhaps go to another workspace and come back though. 

@AThePeanut4, can you explain exactly what I'm supposed to see in those videos. I'm not sure if I can see any problems, but that's perhaps because I don't know when you actually physically click something. Or maybe I'm looking at the completely wrong thing.

fredizzimo:
> If I comment those lines out (i.e. use the defaults), I still get the issue (the cursor "teleports").

Ah I see now. I think that must have something to do with the delta time calculation then. If that goes wrong the speed will be wrong, even if the frame rate is correct.

fredizzimo:
Indeed zooming in on the first frame shows this
image

The animation dt is 120 FPS, and there's 5 frames there, which means that the first 40 ms of the cursor and scroll animation is skipped. The looks good, at least in the portion I'm looking at.

AThePeanut4:
@fredizzimo Does that explain the issue then?

It doesn't always happen - sometimes it animates fine (or only a tiny bit at the beginning is cut off), but sometimes it's as if the entire animation is skipped.

Out of interest I picked a random one in `latency2.tracy`, and here it seems like there are 8 `cursor_animate`s (so I'm assuming 64 ms).
image

fredizzimo:
Yes. it explains the issue. I need to think about another way of calculating the dt. Preferably one that is cross platform, works with both vsync and without, but without a reliable way of knowing which is used. It should also take into account that timing between two frames on the cpu is unreliable due to modern pipelined graphics display.

fredizzimo:
That said, I don’t understand how we get a high average here, or why last dt would be used instead. https://github.com/neovide/neovide/blob/b6374265be52d4171b2348bd9ebfbaa9434d060e/src/window/update_loop.rs#L83C24-L83C24

Could you try to add some logging to try to find out?

fredizzimo:
No it’s clear, the num consecutive check should not be there. That should fix the issue, could you try to remove it @AThePeanut4 

This line https://github.com/neovide/neovide/blob/b6374265be52d4171b2348bd9ebfbaa9434d060e/src/window/update_loop.rs#L82C1-L82C1

AThePeanut4:
@fredizzimo Erm okay I've found another issue, I think it's related...

When I press `:` to go into command mode, the cursor animation appears to hitch a bit at the beginning (in the first frame). Setting `neovide_refresh_rate_idle` to 144 fixes it, as before.

I've set the recording FPS to 120 (which is the max OBS is letting me set it to) to hopefully help see better, but it's still not great. In person it's more obvious.

This is on a646304, with `noshowcmd` again, and with my usual animation config from above:

> ```lua
> vim.g.neovide_scroll_animation_length = 0.1
> vim.g.neovide_cursor_animation_length = 0.06
> vim.g.neovide_cursor_trail_size = 0.3
> vim.g.neovide_underline_stroke_scale = 0.5
> ```

[latency.tracy.gz](https://github.com/neovide/neovide/files/13589161/latency.tracy.gz)

https://github.com/neovide/neovide/assets/49614525/1c91a144-7b7a-417d-a035-83bc26564d3c

fredizzimo:
I think that might be another issue. It's even more obvious if you set `vim.g.neovide_scroll_animation_length = 1`. The cursor moves a long distance the very first frame. 

I think that's caused by the animation ramp we are using for the cursor, which is an exponential ease out https://easings.net/

fredizzimo:
It's not possible to tune at the moment, but you can modify the code here
https://github.com/neovide/neovide/blob/b6374265be52d4171b2348bd9ebfbaa9434d060e/src/renderer/cursor_renderer/mod.rs#L163

With functions from this file https://github.com/neovide/neovide/blob/main/src/renderer/animation_utils.rs

AThePeanut4:
@fredizzimo It seems like it's triggered by holding in `:` - it stutters really badly if you don't release the `:` key. If you quickly tap `Shift+:` and then immediately release the keys, then the animation seems fine.

fredizzimo:
> @fredizzimo It seems like it's triggered by holding in `:` - it stutters really badly if you don't release the `:` key. If you quickly tap `Shift+:` and then immediately release the keys, then the animation seems fine.

Is it possible that you have some autocomplete that's lagging in that case? At least for me `nvim-cmp` kicks in when a second `:` is typed.

AThePeanut4:
> Is it possible that you have some autocomplete that's lagging in that case? At least for me `nvim-cmp` kicks in when a second `:` is typed.

Doesn't seem so no, I disabled `nvim-cmp` and it is still happening. The issue also only happens when `noshowcmd` is set.

Here's a capture and screen recording of the issue, with `noshowcmd` set and `nvim-cmp` disabled. I'm alternating between quickly pressing and releasing `Shift+:`, and holding the keys in until after the animation completes, starting with the former. I am able to very consistently trigger it by just holding the keys down.

[latency.tracy.gz](https://github.com/neovide/neovide/files/13591306/latency.tracy.gz)

https://github.com/neovide/neovide/assets/49614525/10e6f8de-7609-41fd-8c24-6d94af02f398

fredizzimo:
@AThePeanut4. I actually force pushed a slightly different fix now. The previous fix could render an extra frame in some cases.

But I think the result should be the same.

fredizzimo:
Hm.. I think something else is broken now.. the scrolling is not smooth

fredizzimo:
@AThePeanut4, that was actually your configuration that I still had enabled by mistake.
`vim.g.neovide_scroll_animation_length = 0.1` is too short, at least on my system, and is unable to smooth out the uneven repeat rate of the keys. 0.2 is still not completely smooth, only 0.3 feels completely smooth.

I'm not sure how to solve that, I think we need some kind of filtering, but that also adds some latency. And predictive solutions can be wrong.

AThePeanut4:
@fredizzimo Can confirm that on e34d9d9 the command mode animation issue and the initial issue are both still fixed.

As for the scroll animation length, I think I understand what you mean (I'm holding down `<C-D>` on a really long file, and yeah the scrolling isn't perfectly smooth). The reason I set it to `0.1` is because I don't like how long the scrolling animation is when only pressing `<C-D>` _once_, and that is what I find myself doing way more often than holding it down. If I need to find something that I can't search for (e.g. I'd forgotten its name) and thus need to scroll through an entire file, I'd probably tap `<C-D>` successively rather than hold it down anyway, so that I can actually see what I'm scrolling over.

Perhaps animations could be different when holding keys down, if that's something that would even be possible. Maybe a different animation type (something more linear, perhaps?) or just a different animation length. But that certainly belongs in a separate feature request.

src/main.rs
```diff
@@ -146,23 +144,23 @@ fn setup() -> Result<(WindowSize, NeovimRuntime)> {
     // Neovide also includes some other systems which are globally available via lazy static
```
Kethku, line 144: nit: I believe the only remaining lazy static is settings? Maybe this can be reworded to not imply that there are multiple

MultisampledNight (approved):

9mm:
[neovide_backtraces.log](https://github.com/neovide/neovide/files/13630112/neovide_backtraces.log)

I got this backtrace, im 90% sure it was when using this branch. Not sure if this is directly related though, I dont remember what I did i just know neovide froze for a long time and i had to force quit it, then i noticed it was there

like i said though, its never happened any other time, and neovide never crashes, so might not be worth trying to chase down

fredizzimo:
@9mm. I don't think it's related to this, but by studying the code there appears to be a bug where it will crash like that if you drag the window on top of a window that is closed while dragging. It would probably be better to report a new issue, and also check if you can repeat it like that.

You probably need some window that autocloses with mouse actions.

MOldtime:
One suggestion is to animate the cursor blink, such as vsocde

https://github.com/neovide/neovide/assets/38435967/3bdfb4fc-8316-4e60-8c22-1438d8df0b6e

src/main.rs
fredizzimo: I added documentation for the running tracker, which is still remaining.

Kethku:
> One suggestion is to animate the cursor blink, such as vsocde
> 
>  2023-12-17.12-34-21.mp4

This is a reasonable feature request but likely out of scope for this PR

Kethku:
@MOldtime if you were interested in contributing such a change, I'd be happy to give pointers for how to do it. I dont think itd be too hard. First step would be to make an issue with the feature request to discuss further

MOldtime:
> @MOldtime if you were interested in contributing such a change, I'd be happy to give pointers for how to do it. I dont think itd be too hard. First step would be to make an issue with the feature request to discuss further

Thank you for your kindness. This is a great project, but I don't have the experience of developing Rust. I decided to learn. Rust looks like many advantages. This takes a while

mati865:
This PR has almost solved my issue with Neovide fully utilizing single core after about a minute of being unfocused on Ubuntu 22.04 running in VirtualBox and with Wayland enabled. Previously this has happened every time, now it's just "sometimes".
I haven't had time to debug it but when I do I'll create an issue (unless some other PR solves it fully).

fredizzimo:
@mati865, there's a small chance that https://github.com/neovide/neovide/pull/2188 improves or fixes it, but if not, then please open an issue. I think we will need a log preferably from that PR, which has more diagnostics.

I gave instructions for the tracy usage in this message https://github.com/neovide/neovide/issues/2162#issuecomment-1837251296

mati865:
Yeah, I've been watching https://github.com/neovide/neovide/pull/2188 already but planned to wait until it's merged.
However I have tested it today and haven't encountered high CPU usage when unfocused even once for the whole day.

Issue #2087, opened 2023-10-21
macOS cannot close window without frame


hahuang65:
**Describe the bug**
Does not respond to cmd+w on macOS (with `--frame none`)

**To Reproduce**
1. Open `neovide --frame none`
2. Press cmd+w

**Expected behavior**
Window should close (and application if it's the final window)
This is a particular problem for me because I use a tiling window manager ([yabai](https://github.com/koekeishiya/yabai)) and I have muscle memory that presses alt-q (which maps to cmd+w) to close all my windows/applications.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOS 13.3.1 (22E261)
 - Neovide Version 0.11.2
 - Neovim Version 0.10.0-dev-1363+g2eecb1b85-Homebrew

MultisampledNight:
Hi, thanks for reporting!

At least as a quick fix, you can remap <kbd>Cmd</kbd> + <kbd>W</kbd> to `:q` in Neovim, for example in VimScript:

```vim
noremap <D-w> <Cmd>quit<CR>
```

crupest:
I can stably reproduce this bug but in another way. I don't have any special map. In no-frame mode, `:q` makes the window freeze. Command + Q quits Neovide as expected like other apps. I think I can give it a try on fixing this bug.

hahuang65:
For what it's worth, `neovide --frame buttonless` does exactly what I want... it actually does NOT have a frame in macOS (and has rounded window corners, where as `--frame none` has squared window corners), and allows for closing the window.

fredizzimo:
@crupest, regarding that hang, it was encountered again on discord, and we decided to see if refactoring the Window creation to inside the event loop helps. Otherwise, we will report it to Winit.

crupest:
That's what I think. I guess the cause is either that we are doing something wrong with event loop and window system or an internal bug in winit.

fredizzimo:
For a more detailed description of moving the window creation, see this https://github.com/neovide/neovide/issues/2182#issuecomment-1886098109

fredizzimo:
I have a potential fix for this here:
* https://github.com/neovide/neovide/pull/2562

It would be good if someone could test it.

PR #2168, opened 2023-12-05
feat!: fix appearance of `--frame transparent` and deprecate `neovide_background_color`.

### LINKED ISSUE #2094: Transparency issues on macOS
The rendering of transparent windows have changed on macOS, and the old recommendations about using the [background color](https://neovide.dev/configuration.html#background-color-currently-macos-only) to control the dependency does no longer apply. The reason for that is that the background color transparency and the regular transparency  (blend/pumblend/neovide_transparency) conflicts with each other.

The workaround is a bit unclear to me since I don't own a mac to test it out. But I think that you should start by  disabling the background by not setting it in your settings. After that you can control the global transparency (affects all background colors) with `g:neovide_transparency`, and the native Neovim `pumblend` and `blend` options.

The background color was originally made to control the color of the title bar, see:
* https://github.com/neovide/neovide/issues/1298
* https://github.com/neovide/neovide/pull/1459

So we need an alternative to that. Some users have reported that setting [`--frame`](https://neovide.dev/command-line-reference.html#frame) to `none` works reasonably well for certain use cases. But I think we should remove `g:neovide_background_color` completely, and implement the title  bar color some other way. And even in the case where `--frame` is the final workaround, I don't see any use for the background color, since it just cause problems and makes macOS work differently than other platforms.

We need help from mac users to find the final solution for this. Preferably a complete pull request with the required code and documentation changes.

Some discussion about the transparency issues
* https://github.com/neovide/neovide/pull/1870#pullrequestreview-1508224104
* https://github.com/neovide/neovide/pull/1870#issuecomment-1649503004 (forward from that message)

crupest:
Fix #2094

This pr does following changes:

1. fix the bug in `--frame transparent`. Before this pr, the titlebar floats on the head of content. This pr adds a top padding of titlebar height, so the appearance looks better.
2. deprecated the `neovide_background_color` option because the color of titlebar now honors `neovide_transparency`. An error message will be printed if user uses it.
3. code refactor: `draw_background` code merges into the feature code of this pr. And docs are updated.

And here are some other related changes:

1. The light/dark theme of the window will be set based on the background color. So we can get the titlebar text in the right color (white/black).

This pr is almost done but still has some details to discuss. If you want to experience this right now, you can download it from the CI at https://github.com/crupest/neovide/actions/runs/7626642491 (currently based on `3e575fc` on main branch). Or just checkout the branch of this pr and `cargo run --release -- --frame transparent`.


crupest:
@fredizzimo I have to say, things are much simpler than I think. We don't have to draw the title bar ourselves.

Here is a style mask on macos [`NSWindowStyleMaskFullSizeContentView`](https://developer.apple.com/documentation/appkit/nswindowstylemask/nswindowstylemaskfullsizecontentview?language=objc), which makes a effect like content view spreads to the whole window and the title bar floats on the top.

So things get simple. We always specify `NSWindowStyleMaskFullSizeContentView` to get full control of the window and use `setTitlebarAppearsTransparent_` to make the title bar have no background if there is one. When we don't want a title bar, just don't specify `NSWindowStyleMaskTitled`. When we want a title bar, specify `NSWindowStyleMaskTitled` and related masks, make a top padding and it's done.

Here is the result.
Screenshot 2023-12-08 at 21 42 24

Then the problem becomes how do we adjust old configs. I think we can deprecate all related macos specific configs in the doc and try to leave them there if possible so people won't get surprised that old configs not work. I want to deprecate the two macos specific `transparent` and `buttonless` options of `--frame` and silently make them work like `borderless` since I think these two ones are meaningless. And I also want to deprecate `neovide_background_color` but leave it still work since it does not complicates many things. How do you think? Is there anything I didn't notice?

fredizzimo:
@crupest, that's great news!

We actually discussed the deprecation of `neovide_background_color` yesterday on Discord. 

I think we should just show a message with the link to the appropriate issue and documentation here, for people that tries to use it. The `error_msg!` macro in `error_handling.rs` shows a mesage using the standard Neovim message system, so that could be used.

crupest:
@fredizzimo My network environment does not allow me to access discord. I'm sorry I can't synchronize with you guys there.

I have merged `draw_background` into my `macos_feature`. I'll follow your suggestion to use neovim message to show the error. What about the `--frame` options? Maybe you can discuss on this topic to get what most people would like.

I think I'll have plenty of time to deal with this this weekend. During this period, I'll also complete the documentation change in the pr.

fredizzimo:
I don't know if it's possible, but maybe it's possible to make the text field as wide as the titlebar? Even if the actual visible text doesn't take up all the space. 

Or some other invisible title bar elements? 

It would be rather easy to ignore all events in a certain area in the `mouse_manager`, but I think there will be problems when the mouse up and mouse down are in different areas.

crupest:
That's a good idea. I'll try to put a native invisible control there and handle the zoom.

crupest:
@fredizzimo It's a little awkward for me to misunderstand a lot of things. I should have tried the `transparent` and `buttonless` options before I did all these. But luckily I have learned a lot of things. I want to share them here before I go to rest today.

Here are the our implementations,
1. `--frame none` = `Borderless`. Borderless window has **no round corner** and suffers bug #2087. This is useful for tiling window manager as the issue author said.
2. `--frame buttonless` = MY `--frame none` = `Titled | FullSizeContentView` with **invisible title bar**. Different to borderless, this **has round corner** and does not have above bug.
3. `--frame transparent` = MY `--frame full` without top padding =  `Titled | FullSizeContentView | [window buttons flags]` . So the window buttons are float on the content, which makes it ugly. So I guess nobody use this.

So everything is just a composition of complicated things. I think we have to sort these again. I think we can leave 3 final choices:
1. borderless (no round corner) and fix #2087 
2. borderless (with round corner), already works now
3. decorated, and fix the title bar issue mentioned above

I will revise my codes to reflect these new options tomorrow. As for the title bar problem, I have investigated a long time. The way of putting a native control is infeasible. We have so little ability to interact with native platform, especially for macos, which use obj-c but not pure-c api. And it requires to use obj-c api to do the event handling and other related things. We can use the rust library objc but it only makes the code ugly and unmaintainable. (Maybe there is other library?) Logically, in this case, title bar is under our control and we then have the duty to manage that area. But it in turn breaks our current codes a lot.

I'll keep searching for solutions tomorrow. Just to leave a note here. I'm really new to macos developing. And if somebody has any experience on this, I'll be very glad to learn from you!

fredizzimo:
No, problems, this is complicated, and I don't think anyone actually understands it, that's why we have these issues opened for months without any comments about implementation ideas. I have problems even imaging the concepts since I don't use macOS.

As for the objc, winit uses https://github.com/madsmtm/objc2, so I guess that's about as good as it gets. And talking about winit, it's possible that some of these things might make more sense to implement there, or at least port it there later.

crupest:
@fredizzimo I have a good news to share with you. The method of using a stub view at titlebar works perfectly. I have already pushed the working codes. I'll do code clean tomorrow and then I think we can merge. Cheers🍺.

crupest:
And I will also try to improve macos specific code, use better apis and make code as clean as I can do.

crupest:
> Edit: I forgot to say, from what I have seen, the macOS documentation itself is some of the worst I have seen...

Exactly. Far worse than M$. I develop totally via trials.

crupest:
@fredizzimo Coming again. After struggling for several days to try, I failed and gave up to find a way to layout title bar stub completely in macos codes. I still need to use winit resize event to update title bar stub's position and size. It makes the code kind of unclean but I can't find a better way. The biggest problem is winit always pacics here in whatever way I try:

```
eovide panicked with the message 'called `Result::unwrap()` on an `Err` value: PoisonError { .. }'. (File: /Users/crupest/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.29.4/src/platform_impl/macos/app_state.rs; Line: 383, Column: 33)
```

I'm not sure whether there is a more gentle way to handle this. However, current code works well in function. Do you think we can try more about making the code neater or we just stop here.

Here is the final effects. Just ignore the lag of resizing. It's an old problem.

https://github.com/neovide/neovide/assets/15831781/17c1190a-b279-4ad5-8815-e258136b5492

fredizzimo:
@crupest, I did not take a too detailed look at the code, but it looks OK to me and we can always improve it later.  Even small glitches are OK at this point, since it should be strictly better than what we had before, as long as it does not crash. 

@madsmtm, @kchibisov, do you have some advice from the Winit side? The idea here is to support transparent backgrounds and titlebars, by drawing them ourselves. Previously we used a custom background color for the whole window, but that was interfering with the transparency of the Neovim background.

kchibisov:
@fredizzimo winit has option to do transparent decorations or borderless, so you can offset the title bar, and you still draw into it (alacritty has transparent setting on macOS).

It's not like you need to draw them on your own entirely, you can just clear and then update viewport?

fredizzimo:
@crupest, I guess you tried some approaches using those options before starting to work on this approach? 

Personally, I'm quite blind on this, since I don't have a macOS computer to test on, and I'm just going on what's been reported in different issues here. 

But yes, now I saw this old issue, especially this image https://github.com/neovide/neovide/issues/1298#issuecomment-1165877453, and that does indeed look like exactly what we want. We just need to offset the content, which is already possible through a user option `g:neovide_padding_top`, but that should probably be automatic in this case.

crupest:
You are right. A top padding is enough to make the appearance good. But there are several extra points of my codes.

1. I removed transparent frame option because it has no top padding and makes no actual sense then.

2. I write a way to correctly, maybe, get the title bar height.

3. The appearance looks good. But the double click on title bar to zoom doesn't work. I don't know if winit handles this well if just to use winit api to set the window style. Maybe I'll try it tomorrow.

Besides, there are some clean work. Like move the codes of previous draw background and update docs.

crupest:
> It's not like you need to draw them on your own entirely, you can just clear and then update viewport? 

Actually I don't draw the title bar by ourselves now. The solution is dropped after I find the way to add a top padding. I lined out the original solution and point out the comment of the new solution.

kchibisov:
Yeah, but it's pretty much the same with transparent style, no? The only issue is that you need to add handling for clicks and padding yourself, but winit has all the functions available for stuff like that, including moving.

crupest:
To be clear, actually the core solution of my codes is exactly adding a top padding. And there are the other points mentioned above.

crupest:
> Yeah, but it's pretty much the same with transparent style, no? The only issue is that you need to add handling for clicks and padding yourself, but winit has all the functions available for stuff like that, including moving.

That's great. I'm to blame for not even trying the apis. I'll do the tests tomorrow. If everything works well, it is absolutely better to use winit apis.

crupest:
I can't wait and get up from the bed and did some tests. 🤣

With `.with_transparent(true).with_titlebar_transparent(true)`, it looks like this. The title bar is transparent. But we can't draw on it and we need to set background color to color it, which goes back to the situation before this pr.
Screenshot 2023-12-16 at 03 05 45

Then we need `.with_fullsize_content_view(true)`. This goes back to the problem of no top padding and titlebar click issue.
Screenshot 2023-12-16 at 03 07 31

I don't know whether winit did some special handling in this case. Or I missed some other winit options?

kchibisov:
You need to offset the viewport yourself.

crupest:
Emmm. alacritty does not handle it as we want, either. And it also suffers from the titlebar click issue.

Screenshot 2023-12-16 at 03 34 36

For `Buttonless`,
Screenshot 2023-12-16 at 03 35 53

So I guess my workaround is still needed. 😵‍💫

kchibisov:
>Emmm. alacritty does not handle it as we want, either. And it also suffers from the titlebar click issue.

That's because no-one cares to fix it in alacritty.

crupest:
Hey guys! I'm so sorry if I offend you or someone else by chance. I'm not a native English speaker and can't read your emotions behind your words easily. I might even misunderstand you. Or I didn't behave correctly in this community. It looks like the situation is a little awkward here. I'm not finding an excuse for me. I'm so sorry if I did something wrong, sincerely.

I think we should still push this pr forward whatever happens. Either we are ok with current solution or we have to come up with another one. Or whatever I think wrong. Please point it out and I'm very willing to listen and fix it!

Best regards!

fredizzimo:
@crupest, I don't think you offended anyone, I think this is something that both Alacritty and Neovide need to resolve, perhaps eventually through something common in Winit. I understand the comment as the reason why it's not done yet for Alacrity, is because it has had quite low priority there, so no-one has got around to do it yet.

I'm quite happy with this solution, since everything else seem to have drawbacks. But what I stil would like to know, is if you can react to the resizes without panics inside winit. That was the advice I was mostly hoping from the winit team initially, although it was also good to get a second validation of the solution you picked.

crupest:
Yes. I agree that current code is really a little ugly although I have put a lot of efforts on it. It is very tedious, unnatural and error-prone. This solution is absolutely the last sort and it could be used only If we really want this feature so quickly and no other ways are feasible. It's quite ok to leave it here currently.

I'll continue searching for a way to make it better. My current idea is,

1. Firstly limit all these things in macos specific code and not pollute the main codes.
2. No panic in macos codes at least I can't run into panic on my machine.
3. Use better apis to get rid of the titlebar stub if possible.

And according to what we have discussed, further goal is to let winit does all this stuff and alacritty can make use of it as well. I might finally have to deal with winit directly. It will be a long time.😂

Again, if you or someone else gets new ideas on this, whatever it is, please leave a message here and I'm very willing to learn from you!

crupest:
Emmm, looks like I always can't get the key points and make things formalized...

I think the key point you want to say is to deal with the strange panic when handling macos code like ui events. I'll look into it and winit-related things.

crupest:
@fredizzimo I finally find a way to do the layout of titlebar stub in pure macos-specific code. It no longer pollutes our event handling code. And the panic goes away when my layout can work correctly. I guess it might relate to the resizing handling. So when my layout is ok, the bug is away.

I have pushed the new code. I don't put a screen record here because the effect is the same as above.

There is a small thing. I have to double click to activate visual mode. But I have tested on current main branch. It is there, too. I don't know if this is intentional or it is another small bug.

Is there any other problems? I think the `MacosTitlebarStub` might be not very accurate. I'm not a native English speaker and can't use a good work to describe it. Do you think we need to rename it?

fredizzimo:
@crupest Great! I will do a full code review a bit a later, and some of the other maintainers will probably also have a look.

There is one clippy error, that you should fix though, you can see it in the "files changed" tab.

>  I think the MacosTitlebarStub might be not very accurate. I'm not a native English speaker and can't use a good work to describe it. Do you think we need to rename it?

I'm not a native English speaker either, but I would probably just remove the `Stub` part, to make it `MacosTitlebar`. A stub usually means a dummy implementation to get things compiling, or when unit testing, something that returns canned responses for testing purposes.

Or more desciptive `MacosTitlebarClickHandler`, but since the file already implies that it is macOS specific, it can be shortened to `TitlebarClickHandler`

> There is a small thing. I have to double click to activate visual mode. But I have tested on current main branch. It is there, too. I don't know if this is intentional or it is another small bug.

I think that's how it's supposed to work, a single click moves the cursor, a double click selects the current word and enters visual mode. You can also click and drag to select and enter visual mode (or select mode, if neovim is configured that way)

crupest:
@fredizzimo I have made some revise.

1. `MacosTitlebarStub` => `TitlebarClickHandler`, I think the new name is great since the view just handles click.
2. `title_bar` => `titlebar` and fix the clippy problem.
3. Cache titlebar height in dpi-independent length, since it will never change.
4. Add docs at some places.

fredizzimo:
@falcucci, would you like to test this? Since you were the one who originally reported the transparency problems on macOS.

@9mm, since you have been testing a lot of macOS stuff lately, and if you have time and want to, it would be good to test this as well.

9mm:
Sure thing will do

falcucci:
@fredizzimo thank you, gonna try it now and share the results soon. Sorry beein away for so long, busy on personal stuff atm.

falcucci:
the first feedback so far:

- The transparency seems to work correctly without the use of --frame transparency;
- The deprecated messages is gonna be really useful, good catch!;
- I am not convinced about the padding values, I think we should update to have a more simetrical feeling. Right now it looks like it's almost beeing cutted by the borders.

Try to notice the difference (without transparency to be more visible):

CleanShot 2023-12-22 at 02 15 46@2x

now using the following padding values:

```rust
let g:neovide_padding_top = 60
let g:neovide_padding_bottom = 0
let g:neovide_padding_right = 40
let g:neovide_padding_left = 40
```

CleanShot 2023-12-22 at 02 10 13@2x

those are my 2 cents for now.

crupest:
@falcucci Really thanks for what you did.

> The transparency seems to work correctly without the use of --frame transparency

By what you said, I suddenly notice that we needed to specify `--frame transparency` to make titlebar transparent before. And if we don't specify it, the titlebar is solid-colored. So current behavior of thischange is kind of differed from before. And we are no longer able to get a solid-colored titlebar. Maybe we should make the titlebar solid by default and only make it transparent when specifying `--frame transparency`. How do you think of this change?

> Maybe the padding_top is currently low and could be increased by 5.

The default padding_top is purely equal to the titlebar's height with no extra padding. But I'm happy to make some additional padding. How do you think about this?

Really thanks for your feedback!

9mm:
Hey, here are some comments:

Title is in black for me:

image

I always use buttonless (definitely keep this! I saw talk above about removing options) but ^ that is the default.

I also noticed you can't resize the window now but that only occurs on frameless. I didn't know that was desired for certain windows on mac, can you tell me a scenario where that would be the case? I don't know if thats difficult to make that resizeable, but that might be my preferred layout given it makes it very crisp/square, but is unusuable if I cant resize it. but if this is a common mac thing i dont know about Id love to hear.

This next one is minor and may not be fixable, but the cmd line is the only thing with transparency 0%. I realize it might be contextually some kind of "popup" though or something, and should be set to 100% like other popups. This issue is minor, and may not be an issue.

image

crupest:
@9mm Really thanks for your testing. I'm off my computer but I can answer some of your questions.

The titlebar text color is based on the dark/light theme of your system. It is controlled by macOS. For me, in dark mode it is white and in light mode it is black. This can be annoying. But it is a little complex to make this automatically change. The background color is now controlled by the theme of neovim. And we have to detect the color set by neovim to change the text color automatically. One idea of me is to create an option to override the light/dark theme of Neovide. I think there should be an api to force the light/dark theme in one app. I don't know how this is handled on other platforms.

For the frameless window issue, the window is not movable and resizable by intention. I didn't test it on other platforms. But according to current documentation, it should be not movable and resizable on other platforms, either. The usage of it might be like #2087, used in tiling window manager.

As for the cmd line, I guess it is controlled by neovim, I'm not sure. I didn't notice it before. I'll have a test on it tomorrow by trying some other color themes.

Just post your opinions about everything! I'm very glad to talk about this issue with you!

crupest:
I think I'll make a change of the `--frame` option. No longer deprecate any options.
The default one, aka `full`, is to get a solid color titlebar. And `transparency` is to get a transparent titlebar. The other two do not change. This will match the old behavior better and give more options for user.

crupest:
@fredizzimo In the discussion above, I want to learn something from you.

1. How is the titlebar text color handled on other platforms when the titlebar is transparent? Is the color adapted for the background automatically?

2. Is the window not movable and resizable when `--frame none` is specified on other platforms?

3. Is the transparency applied to everything in neovide? Or it is just applied for some parts according to the circumstances. For explaining the cmd line transparency issue. I don't use transparency quite much and don't notice it carefully.

crupest:
Sorry to bother again.
For the square corner problem, if we do want it, I'm very willing to find a way to implement it! But I'm not familiar with coding on macOS, so I can't tell you a solution immediately. Another problem raised for this is we might have to find an elegant way to organize these options.
I think this is kind of away from this pr. So maybe we can talk about it in another issue or pr.

9mm:
> I think I'll make a change of the `--frame` option. No longer deprecate any options. The default one, aka `full`, is to get a solid color titlebar. And `transparency` is to get a transparent titlebar. The other two do not change. This will match the old behavior better and give more options for user.

Regarding this, can you say what exactly you're changing again? I think the way you have it is great, I don't think i'd change it, I would just simply find a way to read the light/dark mode of the persons system?

Or, default it to light mode as thats how macs are by default I think?

Or as you said make an option... but as far as changing all of it back to default I actually thought it looked good as the default as you have it.

9mm:
I do think having a title bar change along with the body is a much much better user experience, so IMO that should be the new default. just my .02

crupest:
Sorry for being unclear. That's two different problems.

The `--frame` one is, in current code, the default is to have a transparent titlebar but it is solid color before. I have deprecated `transparency` because it is the same as default in current code. The new change I'll make is to get `transparency` back and using it makes the titlebar transparent while the default frame is to have a solid titlebar. This will match the old behavior.

The titlebar issue I think we do need to solve it. The solution I can come up with is to manage the titlebar text by ourselves. We detect the background color and choose the inverse one from white/black.

9mm:
Hmm ok, I think I follow. When you say "transparent title bar" do you mean a title which can be controlled with neovide_transparency?

All I mean is that I think the default should be where if someone sets neovide_transparency to say.. 0.9, that both titlebar and body change, vs the title bar staying opaque which I think is unexpected behavior

crupest:
I don't know how macOS controls the titlebar text color exactly. I can only tell it relates to the dark/light theme. If we really want to solve it, we can't assume any fixed default color. Although macOS default is light theme, someone may change it, like I'm using auto theme. (That's how I know the cause of the bug.)

crupest:
Yeah, you get the point. Before this pr, the default is to have an opaque titlebar. If you want a transparent titlebar, aka, the color is the same as body, you have to use the buggy `--frame transparency`. But now I make transparent titlebar default. This  does not match the old behavior. I don't use transparency much. But I think it should be better to match the old behavior.

Emmm, it looks like this is not related to what you feedback. But if you can confirm the transparent titlebar works correctly, it is just enough for me to ensure this feature works correctly also on others' computers.

9mm:
Yep transparency seems to work good on my end, but maybe @fredizzimo can weigh in on changing default behavior. To me it seems much better even if it breaks it, simply because it actually works correctly, but I don't know much about how bad it is to break previous behaviors, etc.

crupest:
You are right. That's why I'm very happy to hear from your feedbacks, so I can make it match our demands better, like making the default options as what we expect. Really thank you for contributing time to test on it!

I'll work on the titlebar text color issue this weekend.

src/window/macos.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// A trait for winit Window on macos, by which we can easily manipulate the window.
+pub struct MacosWindowFeature {
+    ns_window: Id<NSWindow>,
+    has_titlebar: bool,
+}
+
+impl MacosWindowFeature {
+    pub fn from_winit_window(window: &Window) -> MacosWindowFeature {
```
madsmtm, line 57: This is only safe to call from the main thread, since it accesses `NSWindow` and `NSView`.

I'd suggest using `icrate::Foundation::MainThreadMarker` to help with that.

src/window/macos.rs
```diff
@@ -0,0 +1,152 @@
...
+        let ns_window = if let RawWindowHandle::AppKit(handle) = window.raw_window_handle() {
+            unsafe { Id::new(handle.ns_window as *mut NSWindow).unwrap() }
```
madsmtm, line 59: The pointer does not have `+1` retain count, so this ends up doing a double-free.

```suggestion
            unsafe { Id::retain(handle.ns_window as *mut NSWindow).unwrap() }
```

See also https://github.com/madsmtm/objc2/issues/543.

src/window/macos.rs
```diff
@@ -0,0 +1,152 @@
...
+                Frame::Transparent => full_frame_style_mask,
+                Frame::Full => full_frame_style_mask,
+            };
+
+            ns_window.setStyleMask(style_mask);
+            ns_window.setTitlebarAppearsTransparent(true);
+
+            if frame == Frame::Buttonless {
+                ns_window.setTitleVisibility(NSWindowTitleHidden);
+            }
```
madsmtm, line 89: Just FYI, `winit` doesn't guarantee that these aren't overwritten by some other operation done on the window (e.g. when entering or exiting full screen), or that the behaviour here doesn't change subtly in the future (while the previous approach of setting window builder attributes is generally better).

Basically: Use with care.

src/window/macos.rs
```diff
@@ -0,0 +1,152 @@
...
+        unsafe fn mouse_down(&self, event: &NSEvent) {
+            if event.clickCount() == 2 {
+                self.window().unwrap().zoom(Some(&self));
+            }
+        }
+    }
+);
+
+impl TitlebarClickHandler {
+    pub fn new() -> Id<TitlebarClickHandler> {
```
madsmtm, line 41: Requires `MainThreadMarker` too.

madsmtm (commented):
I haven't followed the implementation history here, but the approach in the current diff seems okay-ish (probably better ways to do this if integrating more fully with AppKit, but since you're using `winit`, that can be kinda difficult).

Reviewed with a focus on Objective-C safety, not entirely sure if all the AppKit APIs are used correctly.

falcucci:
@9mm 
> This next one is minor and may not be fixable, but the cmd line is the only thing with transparency 0%. I realize it might be contextually some kind of "popup" though or something, and should be set to 100% like other popups. This issue is minor, and may not be an issue.

this is a well-known issue which has an old fix for that https://github.com/neovide/neovide/pull/1914

AFAIK for compatibility problems it wasn't merged.

@crupest is up to you to pick this up in this pr but I would not mix both to avoid a bigger chance.

For the future this fix could be maybe enabled only on macOS if we use --frame transparency to avoid this compatibility behavior.

@crupest also just a little nip: it would be awesome to have any settings changes updated in the description of the pull requests so we have an eye on what has changed to test it out :)

fredizzimo:
@falcucci, the reason why that was not merged, was that it was completely broken on other platforms
https://github.com/neovide/neovide/pull/1870#discussion_r1264691807 (wait a while until the correct message opens, it loads for many seconds)

It was just a hack to workaround the macOS special problems due to the background color being used for transparency, which is what this PR removes, so it should be the same as other platforms now.

I can’t tell why the commandline is not transparent for @9mm though, since as you can see from my image in the link, the transparency applies to the commandline for me.

fredizzimo:
It’s not as transparent as the background, but not opaque either. The reason for that is that it’s technically a floating window, so its like two semi transparent floating windows on top of each other. I think that’s something that needs to be fixed on the Neovim side. Unless, we make some serious hacks and detect the cmdline and disable all rendering beneath it.

fredizzimo:
@crupest 

> @fredizzimo In the discussion above, I want to learn something from you.
> 
> 1. How is the titlebar text color handled on other platforms when the titlebar is transparent? Is the color adapted for the background automatically?
The other platforms don’t support transparent titlebar. But if they did, we would propbably use the neovide default background color.

> 2. Is the window not movable and resizable when `--frame none` is specified on other platforms?
No
> 3. Is the transparency applied to everything in neovide? Or it is just applied for some parts according to the circumstances. For explaining the cmd line transparency issue. I don't use transparency quite much and don't notice it carefully.

The transparency applies to all background colors, and overlapping transparencies use the standard ”over” formula https://en.wikipedia.org/wiki/Alpha_compositing#Description. So two half transparent window on top of each other becomes 25% transparent.

fredizzimo:
For the frame option I would use this to match the other platforms, and leave room to implement transparent on those as well
`full` - system titlebar (default)
`transparen` - this new implementation, the neovim background color
`none` - no titlebar, not movable or resizable

fredizzimo:
You can query the theme color from winit https://docs.rs/winit/latest/winit/window/struct.Window.html#method.theme, but perhaps that’s not needed if you let winit deal with the `full` option.

NOTE: At some point we will probably remove this https://neovide.dev/configuration.html#theme, and pass dark/light forward to neovim as the default theme

fredizzimo:
@crupest, I don’t know if the scale factor can change dynamically on macOS. But if, not, perhaps you need to hook into this to make the height update. https://github.com/neovide/neovide/blob/ce0174a17674b1984c411e9f81b752d81b1b83d4/src/window/window_wrapper.rs#L269

crupest:
Thanks for all involvement of everybody!

## Replies

@madsmtm Thanks for your time on review this pr. Your suggestion on the code is very valuable and I have followed it. The obj2 library made by you is a really great work.

> @crupest also just a little nip: it would be awesome to have any settings changes updated in the description of this pull request so we have an eye on what has changed to test it out :)

@falcucci You are right. The communication lasts very long here. I've updated the original pr description to reflect what has changed.

## New changes in codes

`--frame transparency` is back now. Every unnecessary changes are all reverted. Bugs in objc are fixed. Docs are updated. The final changes are following, and also described in the pr description at the head of this page,

1. fix the bug in `--frame transparent`. Before this pr, the titlebar floats on the top of content. This pr adds a top padding of titlebar height, so the appearance looks better.
2. deprecated the `neovide_background_color` option because the color of titlebar now honors `neovide_transparency`. An error message will be printed if user uses it.
3. code refactor: `draw_background` code merges into the feature code of this pr. And docs are updated.

## What we have done already

1. the behavior changes now get an agreement. Nothing in `--frame` is changed except `transparency` appearance is fixed. `neovide_background_color` is definitely deprecated.
2. The `TitlebarClickHandler` workaround works correctly. Thanks for @falcucci @9mm 's tests. So we can no longer worry about the function.
3. Objc codes have been reviewed by @madsmtm. So there should be no apparent bugs.

crupest:
> @crupest, I don’t know if the scale factor can change dynamically on macOS. But if, not, perhaps you need to hook into this to make the height update.
> 
> https://github.com/neovide/neovide/blob/ce0174a17674b1984c411e9f81b752d81b1b83d4/src/window/window_wrapper.rs#L269

I think the dpi can change dynamically on every platforms. Either by user explicitly setting, or by screen change... Ok, I'll hook into that function.

I'm currently working on titlebar text color.

crupest:
@fredizzimo I have some problems about the titlebar text color.

The solution I can come up with, is to calculate white/black from neovim background color, and set the light/dark mode of app to get the desired color of the titlebar text. But this will make `neovide_theme` and `WindowEvent::ThemeChanged` totally broken.

Do you think there is a correct to handle this?

> You can query the theme color from winit https://docs.rs/winit/latest/winit/window/struct.Window.html#method.theme, but perhaps that’s not needed if you let winit deal with the `full` option.
> 
> NOTE: At some point we will probably remove this https://neovide.dev/configuration.html#theme, and pass dark/light forward to neovim as the default theme. The way we were going to do it has been deprecated though [neovim/neovim#25918](https://github.com/neovim/neovim/pull/25918). But I believe that’s a mistake, since their new autodetection only works for terminals.

The light/dark mode behavior is a little messy in neovim. It can be

1. calculated from background color
2. set by user or ui like neovide
3. set by color scheme

Actually I don't think it's really a bug because ideally the color scheme, neovim's light/dark mode and neovide app's one should be synced perfectly. But here, they are totally out of sync and may even have incorrect value. This is a specific bug when the titlebar is transparent. I wonder if we should really "fix" this.

crupest:
> @crupest, I don’t know if the scale factor can change dynamically on macOS. But if, not, perhaps you need to hook into this to make the height update. https://github.com/neovide/neovide/blob/ce0174a17674b1984c411e9f81b752d81b1b83d4/src/window/window_wrapper.rs#L269

Oh, I might misunderstand you. The old code already calculated the height with the scale factor. But hook this event may have less issue and improve the performance since we don't have to run into objc to do the calculation.

fredizzimo:
> Oh, I might misunderstand you. The old code already calculated the height with the scale factor. But hook this event may have less issue and improve the performance since we don't have to run into objc to do the calculation.

I might have misunderstood the code, but I thought it was run only once at the startup.

src/window/macos.rs
```diff
@@ -0,0 +1,138 @@
...
+    // Extra titlebar height in --frame transparecy. 0 in other cases.
+    extra_titlebar_height_in_pixel: u32,
+}
+
+impl MacosWindowFeature {
+    pub fn from_winit_window(window: &Window, mtm: MainThreadMarker) -> MacosWindowFeature {
+        let ns_window = if let RawWindowHandle::AppKit(handle) = window.raw_window_handle() {
+            unsafe { Id::retain(handle.ns_window as *mut NSWindow).unwrap() }
+        } else {
+            panic!("Not a appkit window.");
```
MultisampledNight, line 59: Typo

```suggestion
            panic!("Not an appkit window.");
```

src/window/macos.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+}
+
+lazy_static! {
+    // This height is in dpi-independent length, convert it to pixel length by multiply with scale factor.
+    static ref TITLEBAR_HEIGHT: f64 = MacosWindowFeature::calculate_titlebar_height();
+}
+
+pub struct MacosWindowFeature {
+    // Extra titlebar height in --frame transparecy. 0 in other cases.
```
MultisampledNight, line 50: Typo

```suggestion
    // Extra titlebar height in --frame transparency. 0 in other cases.
```

src/window/macos.rs
```diff
@@ -0,0 +1,138 @@
...
+);
+
+impl TitlebarClickHandler {
+    pub fn new(_mtm: MainThreadMarker) -> Id<TitlebarClickHandler> {
+        unsafe { msg_send_id![Self::alloc(), init] }
+    }
+}
+
+lazy_static! {
+    // This height is in dpi-independent length, convert it to pixel length by multiply with scale factor.
```
MultisampledNight, line 45: Slightly confusing wording imo.

```suggestion
    // This height is in dpi-independent length, convert it to pixel length by multiplying it with scale factor.
```

src/window/macos.rs
```diff
@@ -0,0 +1,138 @@
...
+        } else {
+            panic!("Not a appkit window.");
+        };
+
+        if let Ok(color) = &SETTINGS
+            .get::<WindowSettings>()
+            .background_color
+            .parse::<Color>()
+        {
+            error_msg!("neovide_background_color has now been deprecated. Use neovide_transparency instead if you want to get a transparent window titlebar. Please check https://neovide.dev/configuration.html#background-color-deprecated-currently-macos-only for more information.");
```
MultisampledNight, line 67: Might make sense to split this over multiple lines? (idk if that exact suggestion works, might need to extract it into a variable)

```suggestion
            error_msg!(concat!(
                "neovide_background_color has now been deprecated. ",
                "Use neovide_transparency instead if you want to get a transparent window titlebar. ",
                "Please check https://neovide.dev/configuration.html#background-color-deprecated-currently-macos-only for more information.",
            ));
```

src/window/macos.rs
```diff
@@ -0,0 +1,138 @@
...
+            }
+        }
+
+        MacosWindowFeature {
+            extra_titlebar_height_in_pixel,
+        }
+    }
+
+    // Used to calculate the value of TITLEBAR_HEIGHT, aka, titlebar height in dpi-independent length.
+    fn calculate_titlebar_height() -> f64 {
```
MultisampledNight, line 110: Also about `calculate_titlebar_height_in_pixel`: I believe the `calculate_` prefix is superflous, as it's likely not to be just a getter anyway.

src/window/macos.rs
```diff
@@ -0,0 +1,138 @@
...
+                    Self::calculate_titlebar_height_in_pixel(window.scale_factor());
+            }
+        }
+
+        MacosWindowFeature {
+            extra_titlebar_height_in_pixel,
+        }
+    }
+
+    // Used to calculate the value of TITLEBAR_HEIGHT, aka, titlebar height in dpi-independent length.
```
MultisampledNight, line 109: could be a doccomment (see [the reference](https://doc.rust-lang.org/stable/reference/comments.html#doc-comments) and/or [the book](https://doc.rust-lang.org/stable/book/ch14-02-publishing-to-crates-io.html#making-useful-documentation-comments)), so that it appears in the documentation generated by `cargo doc` as well:

```suggestion
    /// Used to calculate the value of TITLEBAR_HEIGHT, aka, titlebar height in dpi-independent length.
```

MultisampledNight (commented):
First of all: Thank you for the PR!

You might want to run [typos](https://github.com/crate-ci/typos) from time to time.

Personally, I'm not on macOS, so I can't really say much about the actual changes and how effective they are. The code looks good to me though.

crupest:
@MultisampledNight 

> You might want to run [typos](https://github.com/crate-ci/typos) from time to time.

Thanks for your careful review. It is really careless of me to not pay attention to these small things. I'll do it better in the future.

> Personally, I'm not on macOS, so I can't really say much about the actual changes and how effective they are. The code looks good to me though.

It's ok. @9mm and @falcucci have already tested the function on their computers. (Thank them very much.) So I think there is no apparent issue.

By the way, this pr has not ended. I have to fix the titlebar text color problem before this pr is ready to merge. However, existing code you have reviewed will not change.

Best regards!

crupest:
@fredizzimo I have finished fixing the titlebar text color problem.

The idea of implementing this is,

1. Add a new `WindowCommand::ThemeChanged`.
2. Hook into `RedrawEvent::DefaultColorsSet` and set the light/dark theme there using above command. I don't follow your instrcutions, because I find `RedrawEvent::SetTitle` there. And I think there should be no difference, because it looks like `RedrawEvent::DefaultColorsSet` finally leads to `DrawCommand::DefaultStyleChanged`. I'm not sure if this is right. Please point it out if I'm doing it wrongly.
3. The light/dark theme is calculated base on the formula [here](https://graphicdesign.stackexchange.com/questions/62368/automatically-select-a-foreground-color-based-on-a-background-color). It has been documented in code as well.

This works well on my computer. But maybe it has to be tested on others'. @9mm Would you mind testing this?

MultisampledNight:
> Thanks for your careful review. It is really careless of me to not pay attention to these small things. I'll do it better in the future.

Eh, it's fine! We're all not perfect knights who do everything right at the first time, and things like typos are so small and irrelevant that they barely matter! Toning yourself down in face of so small stuff like this might unnecessarily take motivation, but your perspective might differ.

Primarily it's about the function, how it's packaged is secondary and easily changeable!

crupest:
> > Thanks for your careful review. It is really careless of me to not pay attention to these small things. I'll do it better in the future.
> 
> 
> 
> Eh, it's fine! We're all not perfect knights who do everything right at the first time, and things like typos are so small and irrelevant that they barely matter! Toning yourself down in face of so small stuff like this might unnecessarily take motivation, but your perspective might differ.
> 
> 
> 
> Primarily it's about the functionality, how it's packaged is secondary and easily changeable!

Though it is a little off topic, I do find myself too sensitive in emotion. This is the first time that I cooperate with others in such a big community, so I do feel kind of, maybe, afraid to behave badly. This might be caused by my personality, and actually I don't feel that bad to say those. But now I do think it might make others feel kind of uncomfortable and, maybe, tired to maintain the emotion. I keep inspecting myself in this process. I agree what you point out and I will concentrate more on content. Our final goal is build Neovide better, which already makes us feel happy. Whatever, I really appreciate everybody here. You are really nice guys.

src/window/window_wrapper.rs
```diff
@@ -211,6 +232,11 @@ impl WinitWindowWrapper {
         self.windowed_context.window().set_title(&self.title);
     }
 
+    pub fn handle_theme_changed(&mut self, new_theme: Option<Theme>) {
+        self.theme = new_theme;
+        self.windowed_context.window().set_theme(self.theme);
```
fredizzimo, line 237: I think this might call an infinite recursion, at least on some platforms, since `WindowEvent` is windows specific, and should be fired when the theme of a window is changed.

And by quickly looking at the code it looks like the macOS implementation checks the theme from the window https://github.com/rust-windowing/winit/blob/4f6fd44c6ce9d5020360257140aee351813664b4/src/platform_impl/macos/window_delegate.rs#L400-L404

So, I think this should be removed for now, and that winit handles the theme change by itself already internally.

At some point, we might want to notify Neovim about this change though, and let it automatically switch `background`, unless it's set to something in the user configuration.

fredizzimo: I just realized that I'm wrong, this is our own theme changed event, and not the winit one. I think we need to rename `WindowCommand` to something else at some point, it's too similar to `WindowEvent`

crupest: > I just realized that I'm wrong, this is our own theme changed event, and not the winit one. I think we need to rename `WindowCommand` to something else at some point, it's too similar to `WindowEvent`

Yeah, I was just about to ask you what the dead loop is. I put it at `WindowCommand` because, I read the code carefully and I find `WindowCommand::TitleChanged` there. Maybe you can first follow `WindowCommand::TitleChanged` to get the logic clearly.

As I can understand, the `WindowCommand` is a way to set some window properties from other components, like window title is asked to change, since not every components hold a ref to window object.

fredizzimo:
@crupest

> @fredizzimo I have finished fixing the titlebar text color problem.
> 
> The idea of implementing this is,
> 
> 1. Add a new `WindowCommand::ThemeChanged`.

> 2. Hook into `RedrawEvent::DefaultColorsSet` and set the light/dark theme there using above command. I don't follow your instrcutions, because I find `RedrawEvent::SetTitle` there. And I think there should be no difference, because it looks like `RedrawEvent::DefaultColorsSet` finally leads to `DrawCommand::DefaultStyleChanged`. I'm not sure if this is right. Please point it out if I'm doing it wrongly.

> 3. The light/dark theme is calculated base on the formula [here](https://graphicdesign.stackexchange.com/questions/62368/automatically-select-a-foreground-color-based-on-a-background-color). It has been documented in code as well.
> 
> This works well on my computer. But maybe it has to be tested on others'. @9mm Would you mind testing this?

I'm not sure if these changes are desired or not, but it's more opinion based than anything else. IMO, the window frame should follow the OS theme, independently of the Neovim theme. But I can see the arguments for the other way around too. I think the actual implementation is good though, if this is the way we decide to go.

My previous comment was about the `transparent` frame. Where I think the titlebar font should follow the one of the Neovide theme. And perhpas `transparent` we could also rename the option to something else, since you don't really need to use transparency for this option. Perhaps `frame=neovim`, to indicate a neovim themed frame?

src/window/window_wrapper.rs
fredizzimo: I think the name `WindowCommand` comes from the fact that those events target our `WinitWindowWrapper`. And is usually something that needs to be further sent to winit. While our draw commands are more about the Neovide grid. 

So you did correctly here, and I was not not reading the code carefully enough before writing the message.

fredizzimo:
> > > Thanks for your careful review. It is really careless of me to not pay attention to these small things. I'll do it better in the future.
> > 
> > 
> > Eh, it's fine! We're all not perfect knights who do everything right at the first time, and things like typos are so small and irrelevant that they barely matter! Toning yourself down in face of so small stuff like this might unnecessarily take motivation, but your perspective might differ.
> > Primarily it's about the functionality, how it's packaged is secondary and easily changeable!
> 
> Though it is a little off topic, I do find myself too sensitive in emotion. This is the first time that I cooperate with others in such a big community, so I do feel kind of, maybe, afraid to behave badly. This might be caused by my personality, and actually I don't feel that bad to say those. But now I do think it might make others feel kind of uncomfortable and, maybe, tired to maintain the emotion. I keep inspecting myself in this process. I agree what you point out and I will concentrate more on content. Our final goal is build Neovide better, which already makes us feel happy. Whatever, I really appreciate everybody here. You are really nice guys.

You are doing a really good job here and we really appreciate your work. It's very typical that pull requests are not perfect the first time, and needs a lot of iteration. That not only applies to big open source projects, but also to almost any project that has good code reviewing practices. It's very rare that a single person can think of everything.

One thing to note, that there's no longer any hurry to get this fixed, we are in the process of releasing Neovide 12.0 right now, with the following release notes
> The transparency rendering on macOS has been changed, and a full fix is work in progress here https://github.com/neovide/neovide/pull/2168. Meanwhile, it's recommended to unset `neovide_background_color` and set `neovide_transparency according to your preference. Note that the the title bar will remain solid with the default OS colors.

So I believe that's enough until we get confident that these fixes are doing what we want and works properly.

crupest:
@fredizzimo

One thing you got correctly is this change is a little far from what this pr is about and the logic might even be wrong.

The goal of this change is to **correct the titlebar text color, aka, white on dark background and black on light ground**. But this change seems too big for achieving this small thing.

Why it is implemented by **changing the window theme** is, on macOS the titlebar text color is determined by the window theme. In dark mode it is white, and in light mode it is black. Light/dark background == Light/dark theme.

The logic here is, dark/light theme is determined by neovim's background color. But now I believe this is not correct. Ideally,

1. the light/dark type of background color
2. `bg` option
3. the light/dark flavor of color scheme
4. the theme of the window

should be synced.

But in this case, the background color is not synced with the theme of window. And I didn't pay attention to `bg` and color scheme. I don't know whether they are synced. Ideally the color scheme changes the colors and set a light/dark background based on the value of `bg` and we can choose the value of `bg` to get the flavor we want. If this is true, then we only need to monitor `bg` value and sync the window theme and titlebar text color will be correct then.

Do you have any idea on this? Actually I was thinking this bug should not be fixed in this pr and open another issue for this problem. But it looks really ugly if the titlebar text color is black on a dark background.

crupest:
> One thing to note, that there's no longer any hurry to get this fixed, we are in the process of releasing Neovide 0.12.0 right now, with the following release notes
> 
> > The transparency rendering on macOS has been changed, and a full fix is work in progress here #2168. Meanwhile, it's recommended to unset `neovide_background_color` and set `neovide_transparency according to your preference. Note that the the title bar will remain solid with the default OS colors.
> 
> So I believe that's enough until we get confident that these fixes are doing what we want and works properly.

That's great! I was really trying to make this pr good enough quickly because you said this is a barrier of releasing a new version. Actually even if we are in hurry, I don't want this gets merged hastily, because it only makes things worse.

fredizzimo:
Yes, we decided to make the release anyway, since we didn't want to rush this through anymore. For a long time it looked like this was close to ready, but there are so many small details that still need proper attention and we had everything else ready. It's not your fault, things were just more complex than what they appeared on the surface.

I also got enough understanding of the problem now, so I think I was able to come up with the release notes with a workaround that is good enough for most people. That was what I was most worried about, getting tons of people with a completely broken configuration.

So. let's think about this properly and enjoy the holidays before taking the next steps. I think this PR is in a good enough state now, that people who desperately needs this can use it.

crupest:
That's great. The count of comment is nearly 100. Really thank everyone who have involved in these. This would be nothing without your generous time and devotion. I have learned so much from you. Let's have a rest here. (I suddenly found it is Christmas now. Sorry we don't celebrate this festival quite much in our cultural.😂) Enjoy holidays! Merry Christmas! ♥️♥️♥️

9mm:
It would appear that worked, nice...

If were not going to put this as the default I'd add it to the docs then near the top where people will see it that they need to try the different frame options. This is 10000x better than the default on OSX i think (mine looks a little akward because i use a tab plugin, but i dont think most people do)

```
cargo run --release -- --frame transparent
```

image


Default:

image

crupest:
I have put a note in the pr description at the top of this page. Actually you don't have to build this by yourself. You can just download it from CI. In fact, for other branches and prs, you can download from CI and experience it without building by yourself, too.

> This pr is almost done but still has some details to discuss. If you want to experience this right now, you can download it from the CI at [here](https://github.com/crupest/neovide/actions/runs/7314260190). Or just checkout the branch of this pr and `cargo run --release -- --frame transparent`.

falcucci:
@crupest I must say that the PR title is too generic for me.

It sounds like these changes does a big transparency refactor but does macOS window features updates like the title bar. Not the way neovide handles transparency itself.

crupest:
@falcucci You are right. The title was made at the very beginning of this pr. At that time, I thought there would be a big change. But now, the problem becomes much smaller from begin than I thought. The title has updated to better reflects the change!
Thanks for your advice!

fredizzimo:
I think the only unresolved issue that is left, is the handling of the theme. I think we should change it to only follow the neovim theme when the frame is transparent and on macOS. In all other cases the os theme should be used.

There's one additional change that we are planning to do separately, and that is to set the default neovim background color based on the theme, the terminal also does something similar.

polachok:
I think there's a problem here, when window is maximized the title is hidden, but padding remains, so everythin is offset.

src/editor/mod.rs
```diff
@@ -558,3 +577,20 @@ pub fn start_editor(event_loop_proxy: EventLoopProxy<UserEvent>) -> NeovimHandle
...
+            > 0.5
+        {
+            Some(Theme::Light)
+        } else {
+            Some(Theme::Dark)
+        }
+    } else {
+        None
+    }
+}
```
falcucci, line 596: instead of nested if/else, this would be more legible. At least creating a function that self describes what it is doing by their annotation would be great. for example:

```rust
/// Based on formula in https://graphicdesign.stackexchange.com/questions/62368/automatically-select-a-foreground-color-based-on-a-background-color
/// Checks if the color is light.
#[cfg(target_os = "macos")]
fn is_light_theme(color: &Color4f) -> bool {
    0.2126 * color.r + 0.7152 * color.g + 0.0722 * color.b > 0.5
}

/// Get proper dark(white text)/light(black text) theme of window for background color.
#[cfg(target_os = "macos")]
fn window_theme_for_background(background_color: Option<Color4f>) -> Option<Theme> {
    background_color?;

    match background_color.unwrap() {
        color if is_light_theme(&color) => Some(Theme::Light),
        _ => Some(Theme::Dark),
    }
}
```

falcucci (changes requested):

website/docs/configuration.md
```diff
@@ -144,7 +144,13 @@ vim.g.neovide_padding_left = 0
...
 background color instead.
 
-#### Background Color (Currently macOS only)
+#### Background Color (**Deprecated**, Currently macOS only)
+
+This configuration is deprecated now and might be removed in the future. In
+[#2168](https://github.com/neovide/neovide/issues/2168), we have made Neovide control the title bar
+color itself. The color of title bar now honors [`neovide_transparency`](#transparency). If you want
+a transparent title bar, setting `neovide_transparency` is sufficient. If you think you still need
+some function like this, please open an issue and tell us.
```
falcucci, line 153:
```suggestion
a transparent title bar, setting `neovide_transparency` is sufficient.
```

falcucci (changes requested):

crupest:
@polachok Thanks for your bug report! I have pushed a fix. You can give it another try! Download it at https://github.com/crupest/neovide/actions/runs/7626642491 .

@falcucci Thanks for your code review. I have applied most changes except changing name of `MacosWindowFeature`. Actually I did have some difficulty making a name for this before. But I am not sure removing `Macos` is good or not. Also `Feature` is not very proper, I think. But I can't think up a better name for it. It is not a real feature, but is some kind of hack.

fredizzimo (approved):
From my point of view this looks good now! But I will leave it open a while before merging so that people have time to test it.

falcucci:
@crupest i'll test it for a last look, thank you

D00mch:
I downloaded the version from the https://github.com/crupest/neovide/actions/runs/7626642491 , works pretty good! Thank you!

fredizzimo:
I will merge this just before making the release, probably tomorrow, so you will have some last chance of testing and reviewing before that.

polachok:
Works fine for me. Not a fan of the animation while entering/exiting fullscreen, but it can be fixed later.

crupest:
> Works fine for me. Not a fan of the animation while entering/exiting fullscreen, but it can be fixed later.

@polachok That's kind of annoying indeed. But it may take some time for me to find a way of handling it better.

MacOS automatically zooms the window content when toggling fullscreen, and then the new padding is painted. So the visual effect is a little bad. Maybe we have to find a way to update the padding and draw the surface before the zoom happens. Or we have to make a smooth animation so it looks better.

> my suggestion is to create a setting to toggle it as the user wishes.

@falcucci I'm not sure about this problem. I was assuming that `--frame transparent` means that the titlebar has a transparent background and everything is the same as not specifying it. And the visibility of titlebar text is another feature. I can implement it in this pr. But I'm afraid that it will make this pr too huge and hard to maintain.

falcucci:
@crupest I am sure about this problem: a default feature has been changed and we are deprecating it in this pr.

I see two ways to handle: 
- we add a toggle since we shouldn't deprecate the hidden title instead of making it visible as default. (this should be the starting change and then proceed with everything else related to title bar);
- we add a setting to toggle it in another PR before the release of possible (this should be a minor task)

crupest:
@falcucci Well, I think the former way is better. Since we are going to make it anyway, let's just make them together instead of creating another one immediately after this. But it may take some time for me to hack into the settings part.

fredizzimo:
Could you take screenshots of all four frame variations, so it's easer for us that are not on Mac to judge?

I think the title can be hidden by just setting it to empty inside Neovim, so we don't necessarily need to do anything

falcucci:
@crupest sure, sounds good. I can help creating the toggle PR, but I am not sure if today I can do that (far from computer currently).

@fredizzimo you still plan to release today?

crupest:
Default:

Screenshot 2024-01-27 at 23 09 39

With `--frame none`:

Screenshot 2024-01-27 at 23 09 59

With `--frame buttonless`:

Screenshot 2024-01-27 at 23 10 22

With `--frame transparent`:

Screenshot 2024-01-27 at 23 10 40


> I think the title can be hidden by just setting it to empty inside Neovim, so we don't necessarily need to do anything

That's an interesting idea. But I can't find the option right now. Would you mind telling me?

falcucci:
@fredizzimo it's just the title being displayed or not.

I'm not sure if a neovim config would do that but I am curious to try since this is a winit platform specific property.

crupest:
> I'm not sure if a neovim config would do that but I am curious to try since this is a winit platform specific property.

Looks like the options are `title` and `titlestring`. But I failed to make it empty because neovim seems to not allow an empty value. If you set it to empty string, neovim will use a default one.

falcucci (approved):

falcucci:
@fredizzimo @crupest let's merge this and i'll handle to toggle the title hidden option in another PR, I'll try to do it before going to the theater today.

fredizzimo:
I realized, I made a mistake when reviewing this PR
* https://github.com/neovide/neovide/pull/2126

I thought the override only applied when the user don't set it to empty himself. At least Ithat's the way I think it should work, "Neovide" should only be the default

crupest:
> let's merge this and i'll handle to toggle the title hidden option in another PR, I'll try to do it before going to the theater today.

That would be great!

falcucci:
https://github.com/neovide/neovide/pull/2319 the draft has been created, i'll add the docs and needs review

fredizzimo:
Thank you! Let's fix further problems separately if they come up.

falcucci:
🎉💯

PR #2176, opened 2023-12-13
Clarity on the 'frame' attribute

dastrus:
It states 'full' but must be 'Full'. All values are case sensitive and must start with a capital letter.

- Fix
- Documentation

- No

### DIFF
--- a/website/docs/config-file.md
+++ b/website/docs/config-file.md
@@ -29,7 +29,7 @@ maximized = false
 srgb = false
 idle = true
 neovim-bin = "/usr/bin/nvim" # in reality found dynamically on $PATH if unset
-frame = "full"
+frame = "Full" # Full, Transparent, Buttonless, None -- case sensitive
 ```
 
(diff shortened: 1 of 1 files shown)


fredizzimo:
Thanks, but I think it should be fixed on master to take lowercase values https://github.com/neovide/neovide/blob/main/src/frame.rs#L9

https://github.com/neovide/neovide/commit/398353677fa326e863aa3a8e3935e175e2dc6335

I think the problem is that our documentation shows the state of master, not the release. That should probably be fixed somehow.

dastrus:
Case-insensitive would probably be preferable, but lowercase is definitely sufficient ;) 

Thanks for a great Neovim Front End! I'm almost 24 hours in and it's working pretty damn well.

fredizzimo:
I think that would be possible if we use [serde-aux](https://docs.rs/serde-aux/0.5.1/serde_aux/container_attributes/fn.deserialize_struct_case_insensitive.html)

The main serde library does not seem to support it https://github.com/serde-rs/serde/issues/586

PR #2186, opened 2023-12-18
chore: Improved profiling

fredizzimo:
* Add support for profiling async tasks
* Add support for plotting graphs when profiling
* Add support for named frames (shows the duration of a re-occuring task) with a name
* Add more profiling events that helps especially with viewing the event flow

- No

### DIFF
--- a/src/renderer/cursor_renderer/mod.rs
+++ b/src/renderer/cursor_renderer/mod.rs
@@ -9,7 +9,7 @@ use winit::event::{Event, WindowEvent};
 use crate::{
     bridge::EditorMode,
     editor::{Cursor, CursorShape},
-    profiling::tracy_zone,
+    profiling::{tracy_plot, tracy_zone},
     renderer::animation_utils::*,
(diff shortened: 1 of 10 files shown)


src/renderer/cursor_renderer/mod.rs
```diff
@@ -422,6 +422,7 @@ impl CursorRenderer {
         if !animating {
             self.previous_editor_mode = current_mode.clone();
         }
+        tracy_plot!("cursor animating", animating as u8 as f64);
```
MultisampledNight, line 425: What is the thought behind the double cast?

fredizzimo: The plots only takes f64, and animating as a boolean, which is not directly convertible to floats

MultisampledNight: Oh. That makes sense, thanks!

MultisampledNight:
Thank you! :steam_locomotive: Chooooooooo choooooo!

Issue #2137, opened 2023-11-17
neovide maximized does not work in wayland 


busaster:
Linux Ubuntu 22.04.3 LTS with wayland

Neoide version 0.10.0-dev

Launch command "neoide --maximized" from command line or via a .desktop file
The windows pops up but it is not maximized.

![image](https://github.com/neovide/neovide/assets/1733338/5bb6a315-ccca-40c1-a1b3-d64d39a2e2e1)

fredizzimo:
On my system this happens maybe once in ten tries. Most of the time it's maximized correctly.

In any case it looks like a winit, wayland or gnome bug, but probably need a bit more investigation before reporting. I'm currently on gnome and haven't tested with any other DE/window manger yet.

I can't repeat it with the `maximized` winit example, so I don't think changing the `maximized` state is the trigger, it's probably the fact that we call `show` and `maximized` right after each other.

mati865:
FYI Alacritty has the same issue, it supposedly has been fixed on the master but hasn't made it to the release yet: https://github.com/alacritty/alacritty/issues/6184

fredizzimo:
Can you test if this is fixed on main? I think it might have been fixed by
* https://github.com/neovide/neovide/pull/2205

But I don't have a computer to test on for one week more.

mati865:
I haven't observed it since that PR was merged to main branch.

fredizzimo:
Nice, I'm going to close this then.

PR #2188, opened 2023-12-18
feat: Smoother render loop

fredizzimo:
This makes rendering a little bit smoother, at least on some platforms, and also cleans up and unifies some code.

* Remove some redundant polling for new events after the first frame is rendered
* Do the animation and shaping as early as possible in the frame, which leaves just rendering to the end. This makes the rendering part much more predictable, which plays nicer with how some Wayland compositors are made. For example on Gnome many frames were too late because of the unpredictability. 
* Got rid of the manual dt calculation, and tries to use the monitor refresh rate instead, with a fallback to `neovide_refresh_rate`. 
* The frames are generated using a real-time clock and the frame rate calculated above, duplicating frames when rendering too fast. But that duplication does not matter since the frames would not be visible without enabling tearing anyway. And when the vsync is disabled, the configured refresh rate is used as the base, so it does not matter in that case either.
* If frames are skipped by the compositor, the animation step is still done, which reduces the visual artefacts. This actually happens maybe once every 10 second on the gnome system I was mostly profiling on.
* The Windows and macOS vsync implementation has been simplified and emulates the the native winit vsync support, which also unifies the rest of the code. But the timer based fallback and X11 opengl based vsync still uses the old system, with the latter almost impossible to port.
* The render thread is removed on Windows. Winit has done some optimisations and the event loop is just a tiny bit on the slower side now. Furthermore, both the vsync and the event driven update loop relies on it being fast. This also simplifies the code, and removes one of the platform specific code paths.
* On Windows the frames start rendering in the middle of the Vsync interval. According to my tests this is as smooth as I think we can get on OpenGL, we need to switch to D3D and waitable swap chains to do better. Maybe it's placebo, but it's still clearly better IMO, but not as smooth as on Wayland for example.

NOTE: This is functionality complete and can be tested, but depends on, so it's best to not review before that is merged. That's also why it's marked as a draft.
* https://github.com/neovide/neovide/pull/2186

I have also not tested this properly on all platforms, I have only really done testing on Wayland and Windows. So we need testing on macOS and X11 before merging this. I have also only tested the very latest commit on a remote desktop connection to Windows.

- No

### DIFF
--- a/src/renderer/vsync/vsync_macos.rs
+++ b/src/renderer/vsync/vsync_macos.rs
@@ -1,60 +1,52 @@
-use std::sync::{Arc, Condvar, Mutex};
-
 use log::{error, trace, warn};
 
-use crate::renderer::WindowedContext;
+use winit::event_loop::EventLoopProxy;
+
+use crate::{renderer::WindowedContext, window::UserEvent};
 
 use super::macos_display_link::{
     core_video, get_display_id_of_window, MacosDisplayLink, MacosDisplayLinkCallbackArgs,
 };
 
 struct VSyncMacosDisplayLinkUserData {
-    vsync_count: Arc<(Mutex<usize>, Condvar)>,
+    proxy: EventLoopProxy<UserEvent>,
 }
 
 fn vsync_macos_display_link_callback(
     _args: &mut MacosDisplayLinkCallbackArgs,
     user_data: &mut VSyncMacosDisplayLinkUserData,
 ) {
-    let (lock, cvar) = &*user_data.vsync_count;
-    let mut count = lock.lock().unwrap();
-    *count += 1;
-    cvar.notify_one();
+    let _ = user_data.proxy.send_event(UserEvent::RedrawRequested);
 }
 
 pub struct VSyncMacos {
     old_display: core_video::CGDirectDisplayID,
     display_link: Option<MacosDisplayLink<VSyncMacosDisplayLinkUserData>>,
-    vsync_count: Arc<(Mutex<usize>, Condvar)>,
-    last_vsync: usize,
+    proxy: EventLoopProxy<UserEvent>,
 }
 
 impl VSyncMacos {
-    pub fn new(context: &WindowedContext) -> Self {
+    pub fn new(context: &WindowedContext, proxy: EventLoopProxy<UserEvent>) -> VSyncMacos {
         let mut vsync = VSyncMacos {
             old_display: 0,
             display_link: None,
-            vsync_count: Arc::new((Mutex::new(0), Condvar::new())),
-            last_vsync: 0,
+            proxy,
         };
 
-        vsync.display_link = vsync.create_display_link(context);
+        vsync.create_display_link(context);
 
         vsync
     }
 
-    fn create_display_link(
-        self: &mut Self,
-        context: &WindowedContext,
-    ) -> Option<MacosDisplayLink<VSyncMacosDisplayLinkUserData>> {
+    fn create_display_link(&mut self, context: &WindowedContext) {
         self.old_display = get_display_id_of_window(context.window());
 
-        let vsync_count = self.vsync_count.clone();
-
-        match MacosDisplayLink::new_from_display(
+        let display_link = match MacosDisplayLink::new_from_display(
             self.old_display,
             vsync_macos_display_link_callback,
-            VSyncMacosDisplayLinkUserData { vsync_count },
+            VSyncMacosDisplayLinkUserData {
+                proxy: self.proxy.clone(),
+            },
         ) {
             Ok(display_link) => {
                 trace!("Succeeded to create display link.");
@@ -77,22 +69,17 @@ impl VSyncMacos {
                 error!("Failed to create display link, CVReturn code: {}.", code);
                 None
             }
-        }
+        };
+        self.display_link = display_link;
     }
 
-    pub fn wait_for_vsync(&mut self) {
-        let (lock, cvar) = &*self.vsync_count;
-        let count = cvar
-            .wait_while(lock.lock().unwrap(), |count| *count < self.last_vsync + 1)
-            .unwrap();
-        self.last_vsync = *count;
-    }
+    pub fn wait_for_vsync(&mut self) {}
 
     pub fn update(&mut self, context: &WindowedContext) {
         let new_display = get_display_id_of_window(context.window());
         if new_display != self.old_display {
             trace!("Window moved to a new screen, try to re-create the display link.");
-            self.display_link = self.create_display_link(context);
+            self.create_display_link(context);
         }
     }
 }
--- a/src/renderer/vsync/vsync_macos.rs
+++ b/src/renderer/vsync/vsync_macos.rs
@@ -1,4 +1,8 @@
 use log::{error, trace, warn};
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc,
+};
 
 use winit::event_loop::EventLoopProxy;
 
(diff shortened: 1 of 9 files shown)


fredizzimo:
@9mm, could you test this at some point? There's no hurry, but I wonder if this makes the it possible for you to run with the default settings and vsync enabled, without the need for disabling it and specifying a high refresh rate.

9mm:
Absolutely, will check tomorrow morning!

9mm:
Ooooo, this definitely is much smoother, if not AS smooth as the other way I was doing it, and this even with animation set to `0.0` which makes it much more obvious when something lags.

If I use the profiler it looks like it hovers around 65-70 fps, which is much lower than the 120 hz of my screen, but I can say even despite that (somehow?). Also the max hovers around 13-15ms. That is a bit confusing to me though given how smooth everything is, I also checked scrolling.

Edit: Actually it looks like scrolling is closer to 120 but scrolling through a list is less than 120, so maybe just item scrolling through a telescope list is maxed out already. When scrolling through a long file it hovers around 8.3ms max

fredizzimo:
@9mm, that's very good news! 

On Windows, the news is not as good though. On my work computer it still stutters a lot, even if it was almost smooth on my home computer. I don't think that can be solvable without moving to D3D.

mati865:
On Wayland Gnome under VirtualBox VM it's even smoother than current main but more importantly for me this PR solves high CPU usage when unfocused like I already said in https://github.com/neovide/neovide/pull/2167#issuecomment-1863544151

I have noticed weird issue however, very rarely when I bring back Neovide to the front the window will shrink from maximized state. Profiling would be though as it has happened like three times during the whole workday.

fredizzimo:
@mati865, I think this issue might be related to the maximised state bug you are having 
* https://github.com/neovide/neovide/issues/2137. 

I have not properly checked what's going on, but I think it's a bug in winit rather than Neovide, since we are not really doing much special. It might also be that winit sends us some window size before the maximized one, and then we send that window size to Neovim, which in turn resizes the window back. But it needs more investigation to fully understand what's going on.

mati865:
I do hit #2137 when launching Neovide as the window typically (but not always) starts as a small square-ish box.
The issue I described in the previous comment is a bit different: very rarely Neovide would shrink vertically (horizontal size doesn't seem to change) when focusing it.
I haven't observed it before switching to this branch but there are too many variables (and too rare reproduction) to pinpoint the problem, so it might be something entirely unrelated to this branch.

src/renderer/vsync/vsync_macos.rs
```diff
@@ -1,60 +1,63 @@
...
+use crate::{renderer::WindowedContext, window::UserEvent};
 
 use super::macos_display_link::{
     core_video, get_display_id_of_window, MacosDisplayLink, MacosDisplayLinkCallbackArgs,
 };
 
 struct VSyncMacosDisplayLinkUserData {
-    vsync_count: Arc<(Mutex<usize>, Condvar)>,
+    proxy: EventLoopProxy<UserEvent>,
+    redraw_requested: Arc<AtomicBool>,
```
MultisampledNight, line 17: Is the `Arc` needed? 

- `Arc` also just uses an `AtomicUsize` under the hood to hold referrers iirc
- `AtomicUsize::swap` and `AtomicUsize::store` take `&self` immutably

so anything shared to could just hold references.

src/renderer/vsync/vsync_macos.rs
```diff
@@ -77,22 +80,21 @@ impl VSyncMacos {
...
-    pub fn wait_for_vsync(&mut self) {
-        let (lock, cvar) = &*self.vsync_count;
-        let count = cvar
-            .wait_while(lock.lock().unwrap(), |count| *count < self.last_vsync + 1)
-            .unwrap();
-        self.last_vsync = *count;
+    pub fn wait_for_vsync(&mut self) {}
+
+    pub fn request_redraw(&mut self) {
+        self.redraw_requested.store(true, Ordering::Relaxed);
```
MultisampledNight, line 90: Shouldn't this be `Ordering::Acquire` and the `swap` above `Ordering::AcqRel`, as this is relying on the `redraw_requested` to be propagated before it's seen above?

src/renderer/vsync/vsync_win.rs
```diff
@@ -1,64 +1,115 @@
...
+    redraw_requested: Arc<AtomicBool>,
+}
+
+/// Calculates the time until the vblank, taking into account that the vblank is cyclic, so this
+/// always finds the next vblank forward
+fn time_until_vblank_forward(delay: f64, period: f64) -> f64 {
+    delay.rem_euclid(period)
+}
+
+/// Calculates the time to wait to hit the vblank + offset given a period Note always tries to wait
```
MultisampledNight, line 36: `tq` probably had fun here.

```suggestion
/// Calculates the time to wait to hit the vblank + offset given a period
///
/// Note: always tries to wait
```

src/renderer/vsync/vsync_win.rs
```diff
@@ -1,64 +1,115 @@
...
 
-use crate::profiling::tracy_zone;
+use crate::{
+    profiling::{tracy_plot, tracy_zone},
+    window::UserEvent,
+};
 
 pub struct VSyncWin {
     should_exit: Arc<AtomicBool>,
     vsync_thread: Option<JoinHandle<()>>,
```
MultisampledNight, line 26: I guess you left that since it's not really affected by this PR, but `vsync_thread` seems to be never `None`, so the `Option` wrapper could be removed.

src/window/update_loop.rs
```diff
@@ -175,38 +207,77 @@ impl UpdateLoop {
                 return Err(());
             }
             Ok(Event::AboutToWait) | Err(false) => {
-                self.should_render.update(window_wrapper.prepare_frame());
-                if self.should_render == ShouldRender::Immediately || !self.idle {
-                    // Always draw immediately for reduced latency if we have been idling
-                    if self.num_consecutive_rendered > 0
-                        && window_wrapper.vsync.uses_winit_throttling()
+                // We will also animate, but not render when frames are skipped(or very late) to reduce visual artifacts
```
MultisampledNight, line 210: Style (and pretty unimportant): This is quite nested, so much that it becomes hard to follow what brace closes what. It might make sense to extract this into its own function `maybe_render` or the like?

MultisampledNight (commented):
First: Thank you for the PR!

src/renderer/vsync/vsync_win.rs
fredizzimo: I'm not sure if there's a better way, but the problem is that the join handle is not valid after joining, so there's problems with the ownership.

Using an option seem to be a quite common recommendation, here https://stackoverflow.com/questions/69189818/why-can-t-i-use-a-thread-handle-which-is-stored-in-a-struct and here https://users.rust-lang.org/t/joinhandle-join/81874/3

MultisampledNight: Sounds good!

src/renderer/vsync/vsync_macos.rs
fredizzimo: Acquire and Release are only really needed if there are some other memory involved, but there we are just comparing a boolean value.

src/renderer/vsync/vsync_macos.rs
fredizzimo: No, I think it was a copy paste issue, because I copied the code from the windows implementation. We should be able to remove it.

src/window/update_loop.rs
fredizzimo: Yes, I think there are a few things that could be refactored here. I don't want to risk breaking any functionality right now though. So I think it can wait a bit and be done in a separate PR.

MultisampledNight: Sounds fair!

src/renderer/vsync/vsync_macos.rs
MultisampledNight: Alright!

src/renderer/vsync/vsync_macos.rs
MultisampledNight: Turns out we can't without doing self-referential magic since `VSyncMacosDisplayLinkUserData` is referenced by `VSyncMacos` itself. I'll drop this then.

MultisampledNight:
Thank you! Chooooooooooooooooooooooooooooooooooooo chooooooooo! :steam_locomotive:

9mm:
On a roll today 😎

MultisampledNight:
Yep, we're getting ready for release!

Issue #1796, opened 2023-03-11
No externalfile drag possible on self compiled binary but works on released binary (windows)


mroxh:
**Describe the bug**
I have followed the [windows source installation](https://neovide.dev/installation.html#windows-source) and I am able to compile the windows neovide binary and run it.

With either the "debug" or "release" version, 
- start neovide binary
- I try to drop a file from the explorer and place this in neovide, the mouse arrow changes to a stop sign.
- release the mouse button, nothing happens.

However, if use the windows neovide binary from scoop or from the [release page](https://github.com/neovide/neovide/releases/tag/0.10.3)
- start neovide binary
- I try to drop a file from the explorer and place this in neovide, the mouse arrow changes to an arrow with dotted rectangular and + sign
- release the mouse button and the file name/path is dropped in neovide/neovim

**To Reproduce**
1 -  follow the [windows source installation](https://neovide.dev/installation.html#windows-source)
2 - cargo build --release
3 - Start the binary from the target\release\neovide.exe
4 - I try to drop a file from the explorer and place this in neovide, the mouse arrow changes to a stop sign icon.
5 - release the mouse button, nothing happens.

**Expected behavior**
The self compiled version should behave the same as the released version.
Is there something missing in the build instructions to have drag and drop enabled? I guess I am doing something wrong to build it properly? I can't find any documentation hence this bug. Please refer to other documentation that I would have missed. I have read the maintainer cookbook, but I see no specific instructions related to this problem. This is my first time compiling neovide.

**Screenshots**
Difficult because the stop sign icon is lost the moment I release my mouse button, to make a screenshot

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Neovide Version 0.10.3
 - Neovim Version 0.8.1

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
I tried to for both cases with --log file, but I see nothing specific. If needed I can still attach.

**Additional context**
When running neovide in wsl mode (neovide.exe --wsl), dragging a file will pass the windows os path (e.g d:\\textfile.log) to neovim running in wsl which is incorrect (dtextfile.log). I would like to convert this to WSL1 path (/mnt/d/textfile.log), by adjusting the code, but I would need to be able to drag and drop from the self build version to reproduce the situation.
My development environment is in WSL1, but I sometimes get log or other text files in windows.
Currently I open this in a direct windows neovide, but I would like to open these files in a  windows neovide in wsl mode.

Kethku:
[...]

Issue #696, opened 2021-05-29
Feature request: Suppport app argument to open files or folders when we drag drop them


h4kbas:
As far as I observed there is no way to open my files by simply dragging them into the app like Vs code. Can you handle the incoming app  arguments to open folders or files if we send a parameter on cmd or drag them into the app?

PyGamer0:
Does it already do that?

h4kbas:
Nope it does not care the files, i have to open the app and use :e to edit a new file

h4kbas:
Version 0.6.0 (2021041315)
MacOS BIg Sur 11.0.1 

Oh, I don't know but it doesn't work for me, interesing

PyGamer0:
@h4kbas try building neovide or get an executable from [here(https://github.com/Kethku/neovide/actions/runs/984701889)

fredizzimo:
This was fixed for macOS in
* https://github.com/neovide/neovide/pull/2395

Issue #330, opened 2020-08-04
Drag-and-drop files no longer works


alexozer:
Previously it was possible to drag-and-drop a file into the neovide window to open it, but it seems like this is no longer possible on 0.5.0. Is this due to SDL2? Will it be re-implemented?

Kethku:
yes this is due to winit. Drag and drop is only implemented for some platforms in sdl2. Eventually we may bring back support for winit in which case it should again work.

xarthurx:
Is there any time line for this drag-n-drop feature? Seems 1.5 yrs has passed.
I'm on Windows platform.

MultisampledNight:
This should be fairly easy to implement, looking at [`winit::event::WindowEvent::DroppedFile`](https://docs.rs/winit/0.24.0/winit/event/enum.WindowEvent.html#variant.DroppedFile). Neovide seems to still have the skeleton of such a drop event:

https://github.com/neovide/neovide/blob/9c1481e1c5b00b4a1f6233665445dcb28cd740be/src/bridge/ui_commands.rs#L144-L145

This could be triggered by matching for ~~winit's~~glutin's event here:

https://github.com/neovide/neovide/blob/9c1481e1c5b00b4a1f6233665445dcb28cd740be/src/window/mod.rs#L142

Is anyone down to writing a few lines? If not, I can do it tomorrow or the day after tomorrow.

MultisampledNight:
Wait, lol. This is already handled:

https://github.com/neovide/neovide/blob/9c1481e1c5b00b4a1f6233665445dcb28cd740be/src/window/mod.rs#L161-L167

xarthurx:
> Wait, lol. This is already handled:
> 
> 
> 
> https://github.com/neovide/neovide/blob/9c1481e1c5b00b4a1f6233665445dcb28cd740be/src/window/mod.rs#L161-L167

Thanks for looking into this.
I don't understand the mechanism, but it seems the drag-n-drop feature is not working in windows...

9mm:
@fredizzimo  Ok after tons of looking, it seems on OSX the problem is... the plist doesn't contain `CFBundleTypeExtensions`.

I have looked at MacVim's and they have literally about 1000 entries in there, probably for right click context menus (ie, "open with")

I started to research a much simpler list that would work with any file extension for instance, but that's when I realized the second problem which is...

It looks like Cargo bundle doesn't seem to support anything other than the most rudimentary basic list of items :(

[...]

Issue #1682, opened 2022-12-19
Macos open with doesnt work


ta3pks:
**Describe the bug**
trying to open a file with neovide using open with opens an empty neovide instance on mac

**To Reproduce**
1. right click a text file in finder
2. hover "open with"
3. select neovide to open with
4. an empty neovide opens instead of the selected file

**Expected behavior**
selected file to open in neovide

**Desktop (please complete the following information):**
 - OS: macOS 13.1 ventura
 - Neovide Version 0.10.3
 - Neovim Version 0.8.1

owittek:
Does opening via Spotlight work? Might be related to my issue (https://github.com/neovide/neovide/issues/1679#issue-1500989320)

girishji:
Same issue here. 
1) You can right click on a file in Finder,  click "get info", change default application to Neovide. Then double click on the file, it opens empty buffer instead of said file.
2) "open -a Neovide filename" should open filename, but it opens empty buffer. however "open -a Neovide --args filename" opens filename correctly.

kabouzeid:
fyi this problem has been reported multiple times. +1 i would also like to have this. hopefully someday someone will make a PR for this

NikosEfthias:
tried messing around a little creating a dummy project which is just 
```
fn main() {
    let args: Vec<String> = std::env::args().collect();
    std::fs::write("/tmp/out.txt", args.join(" ")).unwrap();
}
```
then I bundled this using cargo bundle when I set it as default for any file and double click MacOS does not pass any extra arguments.\
I have been researching about this whole night but every bit of info i could find on internet says macos does pass arguments.

If anybody can direct me to some piece of documentation i would be glad to open a pr about this

girishji:
Maybe the way neovide is bundled, it is ignoring the filename argument passed by macos when you set neovide as default application for opening a file. Instead it is looking for explicit arguments passed through 'open --args'.

kabouzeid:
I have no idea. I wanted to fix this myself, then finally found out about Apple Events, and then lost interest :D

NikosEfthias:
okay i will try to test thanks for the headers anyway :D

NikosEfthias:
I figured getting filepath out before i open a pr wanted to ask which would be a better way to handle openFile events on mac
[...]

PR #2191, opened 2023-12-19
fix: Dragging and dropping files in to OSX dock icon

9mm:

Currently when a file is dragged over the Neovide dock icon, it doesn't change appearance (indicating it doesn't accept drag and drop). Furthermore, when the file is dragged over (after allowing extensions) it still didn't accept the drag event so is intending to be fixed too.

* fixes https://github.com/neovide/neovide/issues/696
* partially fixes https://github.com/neovide/neovide/issues/330

There are some other issues that could potentially be fixed as well if someone with Windows wants to help. It's possible that drag and dropping on windows doesn't work for a similar reason?

* potentially fixes https://github.com/neovide/neovide/issues/1796

I actually think maybe this is related too, as thats the purpose of the massive CF type bundle is to say every type thats supported as an "Editor" role: https://github.com/neovide/neovide/issues/1682

- No


9mm:
I made this PR as a draft before getting it working, so I can save all my notes here and get feedback on best approach, ie: how to handle the cargo bundle not supporting it. Can we just fork that or something, and merge one of those PR's? or is there a better approach

9mm:
Another maybe better option is in build.rs modify the plist file manually with https://crates.io/crates/plist ? The other package is no longer maintained btw it seems

fredizzimo:
I think we are missing some information here. What does cargo bundle not support? And which PRs are you talking about?

9mm:
Basically in order to have drag and drop support on OSX, it looks like in the `Info.plist` you need to have a key called `CFBundleDocumentTypes`.

https://stackoverflow.com/a/14527463/794481

I checked the bundled `Neovide.app` and it's missing that item from `Info.plist`

Given that this file isn't anywhere in the repo, I am assuming it's coming from here:

https://github.com/neovide/neovide/blob/main/Cargo.toml#L106-L118

I am assuming `cargo-bundle` is using that to generate the `Info.plist.`

https://github.com/burtonageo/cargo-bundle

^ In their docs, they only support a small variety of keys that can be added. There's several complaints in all the Github issues about how he isn't maintaining it with new types, and theres months old open PR's where he's also not adding new types.

Heres an example of a PR adding a key: https://github.com/burtonageo/cargo-bundle/pull/123

That one doesn't help us just showing that there's no actual way to just write your own keys/values which seems crazy to me.

I also could be missing something obvious here... but otherwise I don't know how we can write the proper keys to the Info.plist, as full MacOS support requires a lot of identifiers which are missing, for example all the right click "open with" context menus, setting a program as a default when opening, etc, are all controlled via that.

9mm:
https://github.com/burtonageo/cargo-bundle/issues/33

fredizzimo:
Ah, I see. I will search around if there are some alternatives to cargo bundle. The status of the projects says "very early alpha", and if the maintainer doesn't have time to work on it, then we have to look for alternatives.

9mm:
I only did a rudimentary 5 mins of googling so this may not end up being any good, but maybe "tauri-bundler" which looks to be [based on cargo-bundler](https://internals.rust-lang.org/t/cross-platform-bundling/16773)

It looks like tauri is a full build system which wouldn't be desired, but maybe tauri-bundler can just be used separately

Im trying to see though if it supports arbitrary keys... still searching

fredizzimo:
I checked Alacritty  and they seem to be doing it manually with a makefile and `hdiutil` https://github.com/alacritty/alacritty/blob/master/Makefile

And with the icons and plist included in the repository https://github.com/alacritty/alacritty/tree/master/extra/osx/Alacritty.app/Contents

So, I guess that's one way to go. But not sure if it's the best. 

I will do more searching later.

9mm:
But yes - whatever you think. This bit is outside my wheelhouse haha

fredizzimo:
Maybe we could use the tool from MacVim https://github.com/macvim-dev/macvim/tree/master/src/MacVim/create-dmg

9mm:
That might be a great idea, I can certainly say MacVim has dialed in the native Mac experience, that's for sure. It's almost too much how they're dialed in 🤣 For example, they're basically in every single right click menu, context menu, top-bar menu. Not that we need all that but good to know it would be supported.

shaunsingh:
> [burtonageo/cargo-bundle#33](https://github.com/burtonageo/cargo-bundle/issues/33)
> 
> ^ Other people wanting keys added, and him saying he doesn't have time to work on it anymore

I think it's in our best interest to move away from cargo-bundle for macOS at the least. Tauri builder (https://github.com/tauri-apps/tauri/tree/dev/tooling/bundler) at the least has not just support for `Info.plist`, but shipping a `dmg` instead of our current `.app` natively.  I can't speak for other platforms, but I imagine there are additional bug fixes.

If everything looks good, I'll whip up a PR to move the build process over to tauri-builder. Hopefully that should make this PR a bit easier to implement.

9mm:
I'm good wil tauri bundler but I'll defer to @fredizzimo - it looks like there's slight improvements on Linux + Windows as well, maybe some other issues would be fixed by that (ie, drag and drop on other OS's, etc)

fredizzimo:
I would personally go with the MacVim's tool, since it seems more generic, while the tauri bundler seems specialized for their usecase of packaging tauri applications. 

I'm also not sure if the cargo.toml support is there in the tauri bundler anymore, they mentioned that they refactored into a library, and I can only see the library documentation. So if the support is there, it seems to be undocumented. And their CLI tool only seem to be able to build tauri applications.

But if @shaunsingh, can get that to work, and you figure out how to enable the plist properies you need, then it should be fine. 

But for example it looks like you have to pass the file associations in a settings struct to enable what you want https://github.com/tauri-apps/tauri/blob/6892a8cbc1df39c48c3b83f3700b873cbad086ed/tooling/bundler/src/bundle/macos/app.rs#L221

9mm:
Ah very good points, well I'm good with Macvim's obviously as well too, it does what we need for sure

fredizzimo:
For that one, I think you could try it yourself, you just need to build a release build, and give the right parameters the path to the executable, the path to the icons and so on.

9mm:
Awesome, I might have some questions but I'll give it a go

fredizzimo:
Yes the macOS CI runs on this https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md

fredizzimo:
I also just noticed that the official repository for create-dmg is https://github.com/create-dmg/create-dmg

9mm:
Awesome, I'm workin on it already I'll let you know what i come up with.

fredizzimo:
Btw, if I understand it correctly,this is the source folder macvim use, but you only need a few things from there, for example the plist with our own modifications

https://github.com/macvim-dev/macvim/tree/master/src/MacVim

fredizzimo:
Actually here's the command they use https://github.com/macvim-dev/macvim/blob/1a8aef0b01ce606f17115b3e71d4864f0640d1b2/src/Makefile#L3679-L3694

fredizzimo:
Sorry, after studying more, it seems to be more complex than that :cry: 

The plist needs to be compiled into the executable https://github.com/macvim-dev/macvim/blob/1a8aef0b01ce606f17115b3e71d4864f0640d1b2/src/MacVim/MacVim_xcode8.xcodeproj/project.pbxproj#L261 And that's probably partly what those bundle tools does

9mm:
Yes I was kind of coming to that conclusion at the same time, it also seems like it's just for dmg's after you already have the .app built

fredizzimo:
Hm. based on the Alacritty makefile, I don't see anything special, it copies the executable to the same folder as the plist and renames it to Alacritty.app. So the Alacritty.app goes into a folder with the following structure https://github.com/alacritty/alacritty/tree/3d7d81c8482eb9465763020a397290765b70b541/extra/osx/Alacritty.app/Contents

Then it creates the dmg from that https://github.com/alacritty/alacritty/blob/3d7d81c8482eb9465763020a397290765b70b541/Makefile

9mm:
Interesting, well I can make a rough version in a single file, and I will try to cobble it together and then someone perhaps could refine it, but ill get the core parts of it working first, if you think thats good approach

fredizzimo:
Yes, I think that's a good approach, if it works, then we at least have full control of the plist file and everything else ourselves.

9mm:
@fredizzimo im getting close to a good working version, do you have a recommendation for where this file should live?

Id rather its in the proper spot it will eventually be, but if its just going to be moved entirely to build.yml or some github process I suppose it doesn't matter

fredizzimo:
We have the folders `wix` and `snap`. So perhaps `macos-app`? We also have common assets in the `assets' folder, where the Linux deskop file is. So if it's just a single file, everything could also go in asserts. 

The actual commands will go to build.yml, but ideally there would be a shell script that it calls, which can also be manually run. So perhaps putting that and the plist and other macOS specific stuff in macos-app folder is the best. 

At some point we might want to clean this up and move all of these under the same main folder, with subfolders for the different platforms. So I don't think the actual location matters too much at this point.

9mm:
yep makes sense, thats pretty much how i had it anyway. Thanks! will have something soon

9mm:
Ok here we go 😎

I made 2 scripts:

1. `macos-builder/make-icns`
  * this generates a `tiff` from 2 input pngs for the purpose of supporting retina in the dmg installer background. If you just use a standard png it will be blurry, so a properly composed tiff must be used
  * this generates an `.icns` file from the input 1024 sized png that I added. This only needs to be run when the icon is changed
3. `macos-builder/run` this generates both an `app` and a `dmg`

Here's a preview of the `dmg` installer I designed:

image

I moved over all the old Info.plist keys. Now im going to work on the actual issue/problem with file types, but wanted to get this up for review first

Current "to solve" list

- this does not auto-clear out bundled items (and it wont overwrite). I didnt want to add `rm` stuff to script as im sure that will change. both the `app` and `dmg` should be cleared out before running
- potentially using the values from `Cargo.toml` instead of redefining them
- interpolating the [CFBundleShortVersionString](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring) and [CFBundleVersion](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion) version numbers into `Info.plist` (I'm not sure the difference but `alacritty` seems to just put a `1` for `CFBundleVersion`, potentially hardcoded

9mm:
Regarding all the copyright stuff, I made everything by hand using existing resources. The font is `JetBrainsMono Nerd Font` Bold Italic so I think that's safe to use

9mm:
Just as a note, when you run it, and the dmg temporarily opens... don't click it. Its part of the compile process where it mounts the dmg for some reason

```
➜  neovide git:(drag-file-event) macos-builder/run
Created 'Neovide.app' in 'target/release/bundle/osx'
Creating disk image...
...................................................................................................................................
created: /Users/zesty/Code/neovide/target/release/bundle/osx/rw.51273.Neovide.dmg
Mounting disk image...
Device name:     /dev/disk4
Searching for mounted interstitial disk image using /dev/disk4s...
Mount dir:       /Volumes/dmg.NgpJlq
Copying background file 'assets/neovide-dmg-background.tiff'...
Making link to Applications dir...
/Volumes/dmg.NgpJlq
Copying volume icon file 'target/release/bundle/osx/Neovide.app/Contents/Resources/Neovide.icns'...
Running AppleScript to make Finder stuff pretty: /usr/bin/osascript "/var/folders/t8/nccp4pks5xz93n7m_h3mj6mr0000gn/T/createdmg.tmp.XXXXXXXXXX.T45dzPQegQ" "dmg.NgpJlq"
waited 1 seconds for .DS_STORE to be created.
Done running the AppleScript...
Fixing permissions...
Done fixing permissions
Skipping blessing on sandbox
Deleting .fseventsd
Unmounting disk image...
"disk4" ejected.
Compressing disk image...
Preparing imaging engine…
Reading Protective Master Boot Record (MBR : 0)…
   (CRC32 $F76F4199: Protective Master Boot Record (MBR : 0))
Reading GPT Header (Primary GPT Header : 1)…
   (CRC32 $1B77B0BA: GPT Header (Primary GPT Header : 1))
Reading GPT Partition Data (Primary GPT Table : 2)…
   (CRC32 $A0D8ACC9: GPT Partition Data (Primary GPT Table : 2))
Reading  (Apple_Free : 3)…
   (CRC32 $00000000:  (Apple_Free : 3))
Reading disk image (Apple_HFS : 4)…
.............................................................................................................................................................................................
   (CRC32 $58985474: disk image (Apple_HFS : 4))
Reading  (Apple_Free : 5)…
.............................................................................................................................................................................................
   (CRC32 $00000000:  (Apple_Free : 5))
Reading GPT Partition Data (Backup GPT Table : 6)…
.............................................................................................................................................................................................
   (CRC32 $A0D8ACC9: GPT Partition Data (Backup GPT Table : 6))
Reading GPT Header (Backup GPT Header : 7)…
.............................................................................................................................................................................................
   (CRC32 $74C20A2D: GPT Header (Backup GPT Header : 7))
Adding resources…
.............................................................................................................................................................................................
Elapsed Time:  3.943s
File size: 11259856 bytes, Checksum: CRC32 $6DCF9612
Sectors processed: 108621, 60644 compressed
Speed: 7.5MB/s
Savings: 79.8%
created: /Users/zesty/Code/neovide/target/release/bundle/osx/Neovide.dmg
hdiutil does not support internet-enable. Note it was removed in macOS 10.15.
Disk image done
```

9mm:
File types have been added! So everything now is passing in. I think maybe the problem might be the fact its forking now, perhaps its not properly passing the item to the 2nd process, so im going to find a way to test the fork code with this code, and maybe both should be merged together.

If its not forking, then ill figure out what it is

9mm:
Ok, that was the problem... when I use the `NEOVIDE_FORK=0` env variable, it now gives an error message instead of just disappearing.

Now the problem seems to be this: https://github.com/neovide/neovide/issues/1682

ill look more into that later but i feel very good with progress so far

image

9mm:
Hey @fredizzimo I have hit a temporary roadblock. From what I gather it seems I need to implement `NSapplication` using `cocoa` crate.

https://stackoverflow.com/a/2490050/794481

I see some talk about the exact error message here (in first issue comment how he explains 'the application cannot handle this file')

https://github.com/rust-windowing/winit/issues/1751

Currently this looks unresolved... I'm not sure if there's a way to do this manually outside?

If you have any code examples from other open source projects I could potentially use as an example i can just try to get something working

9mm:
I did see this maybe? https://github.com/rust-windowing/winit/issues/1855#issuecomment-786403931

fredizzimo:
@9mm. This is probably out of scope for winit, since it only deals with windows. But based on some quick searches, it looks like we need to design a document based application. https://developer.apple.com/documentation/appkit/documents_data_and_pasteboard/developing_a_document-based_app/

But since, I just did a very quick search, it's possible that I'm completely wrong.

fredizzimo:
@9mm, the thread you linked mentions https://mrmekon.github.io/fruitbasket/fruitbasket/. Which after a short investigation, I think should allow us to get access to the application and listen to events.

9mm:
Haha, I just saw that in the comment above yours :)

I did start to attempt to use that but gave up after not having any idea what I'm doing. I'll try again

fredizzimo:
I think you just need to create the instance, and use https://mrmekon.github.io/fruitbasket/fruitbasket/struct.FruitApp.html#method.register_callback and https://mrmekon.github.io/fruitbasket/fruitbasket/struct.FruitApp.html#method.register_apple_event

The loop should already be run by winit

9mm:
[this](https://stackoverflow.com/questions/50996326/why-isnt-applicationopenfile-called-when-the-application-is-already-running) looks helpful even though I have no idea what it means... saving

9mm:
So I have tried 4 or 5 times throughout the day w/ different attempts. For starters I don't do any ios/osx dev, or really even any native app work with loops/threads, only web architecture with threads which is very different than dealing with OS level resources, windows, etc

I would say that my impression is that Fruitbasket has to try and boot itself as a separate app or something. I didn't really get their examples working.

I did also just try a regular cocoa based approach based on some example I found, and when I placed it before the loop it said I can't do that before initializing the loop (winit error), and when i placed it after event loop it would just not show the window.

I'll try again but Im' not sure how much further I will get, I have spent a good 7 or 8 hours trying different things or even make some small progress but have not achieved that yet.

9mm:
Also more info about using fruitbasket as a document type handler

https://github.com/mrmekon/fruitbasket/issues/31

https://github.com/woelper/oculante/issues/33

This better explains what I have experienced despite being out of my league, all my attempts eluded to something like this, and I just don't know enough how to handle multiple loops, etc

> My main problem is that fruitbasket brings its own event loop, and the library I am using (notan) also has its own even loop (receiving input, drawing stuff on screen). I believe that I can't have both. I tried putting fruitbasket's loop in a background thread to listen for file events and then send something to oculante, but I believe fruitbasket only works on the main thread. This is why only on Mac, I start fruitbasket's loop, wait as long as I get a file sent to open, then start the app again with the file as an argument. This works "well" for initial opening, but not for opening files after one was already openend, as fruitbasket is not active any more.

fredizzimo:
@9mm, don't worry, the official Apple documentation is very bad, and you are dealing with layers of abstractions on top of it (winit/fruitbasket), which makes everything very hard.

It might not even be possible to do without modifications to winit. So, I think that we need to open an issue there if we can't find the solution. 

I will also read up a bit about this, but I don't know when, probably not in the next few days at least. But maybe someone else is reading this and can help you.

9mm:
Thanks, I appreciate that. It definitely is painful, even though a bit fun trying to figure things out from just sheer raw intuition and reading.

No rush or worries! That would be cool, if you feel like it - although Im not trying to just open issues then make people work on them.I I think I will feel inspired again soon and try again.

abhillman:
Just an FYI -- this compiles, but does not deliver said functionality on my box as currently authored.

```
% uname -a
Darwin MBP.local 23.2.0 Darwin Kernel Version 23.2.0: Wed Nov 15 21:54:05 PST 2023; root:xnu-10002.61.3~2/RELEASE_ARM64_T6031 arm64
```

abhillman:
Ideally, a change that would enable document handling for macOS might be upstreamed in cargo-bundle (e.g. https://github.com/burtonageo/cargo-bundle/issues/174).

9mm:
@abhillman 
it doesnt offer said functionality because it's a draft. namely there is no good way using winit to access `application:openFile:`

I'm not an OSX developer so I dont have the skills to find a way for Neovide to hook into the event loop.

If you read the comments you will see theres multiple event loops that need to be handled, there were some possible ways with fruitbasket, etc, but I couldn't get them to work.

I have already done a lot of the heavy lifting which was a PRECURSOR to this which was adding an installer DMG and supporting the necessary `Info.plist` attributes.

But as far as hooking into that event, I'm at a loss.

Which is a bummer because I'd really love this functionality. I work with dozens of CSV's every day on my desktop and i have no easy way to open them. I have to keep macvim on there to use it for that use case... which sux

fredizzimo:
Also,if you read the first messages, you will find the reason why we decided to move away from `cargo bundle`

The [project status](https://github.com/burtonageo/cargo-bundle#status) is
> Very early alpha. Expect the format of the [package.metadata.bundle] section to change, and there is no guarantee of stabili

And that combined with very sporadic development and, and useful pull requests getting left open without any comments.

9mm:
^ yes that too. TBH it would be really cool to ship this issue eventually as I did spend a lot of time making a nice installer etc, which separates the good OSX from the chaff

heres also what it is stuck on it seems:

https://github.com/rust-windowing/winit/issues/1751

Would definitely be cool to find another way though, as who knows how far away that is

image

abhillman:
> Also,if you read the first messages, you will find the reason why we decided to move away from `cargo bundle`
> 
> The [project status](https://github.com/burtonageo/cargo-bundle#status) is
> 
> > Very early alpha. Expect the format of the [package.metadata.bundle] section to change, and there is no guarantee of stabili
> 
> And that combined with very sporadic development and useful pull requests getting left open without any comments.

Although `cargo-bundle` is in alpha stage, it has thus-far served the needs of `neovide`. It would be a real gift to the rust community and to rust's cross-platform story to add this functionality to `cargo-bundle`.

9mm:
@abhillman i doubt cargo-bundle will work any time soon given the sheer amount of new keys required, unless they specify some way to allow arbitrary key entry. You can see there are a huge amount of new keys here that added:

https://github.com/neovide/neovide/pull/2191/files#diff-971b2ac9922c3c9ca5b99bb2125e7bfdd78cd3175e349608c4d723f7aba023d8

abhillman:
> @abhillman i doubt cargo-bundle will work any time soon given the sheer amount of new keys required, unless they specify some way to allow arbitrary key entry. You can see there are a huge amount of new keys here that added:
> 
> https://github.com/neovide/neovide/pull/2191/files#diff-971b2ac9922c3c9ca5b99bb2125e7bfdd78cd3175e349608c4d723f7aba023d8

Let's put aside the `cargo-bundle` thing for now! I think however you and other folks think is best is ideal. That said, I played around with this for a bit. The minimal required addition to a `plist` file to allow for a document to be dropped on top of a app bundle is the following:

```plist
<key>CFBundleDocumentTypes</key>
<array>
  <dict>
    <key>LSItemContentTypes</key>
    <array>
      <string>public.data</string>
    </array>
  </dict>
</array>
```

In addition to this piece and other logistics, what remains is registering a message handler for the OS.

abhillman:
Here's a sketch of how to bring this much of the rest of the way.

**Background**

* [`winit`](https://github.com/rust-windowing/winit), the library responsible for creating a window for `neovide`, produces an [`AppDelegate`](https://github.com/rust-windowing/winit/blob/f8b7c4b78f41beb30ca19fe357abdec0d98a6434/src/platform_impl/macos/app_delegate.rs#L50)

* the way that you would get this to work (i.e. dropping documents on an app bundle's icon and calling a method) using Objective-C is by adding the instance method [`application:openFiles:`](https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428742-application?language=objc) to a project's `AppDelegate.m`:

```
- (void)application:(NSApplication *)sender openFiles:(NSArray *) fileNames {
    NSLog(@"Files dragged on: %@", fileNames);
}
```

* Objective-C has a runtime that has advanced reflection and related runtime features (i.e. classes can be fetched by name, methods can be added, etc.)

**Sketch of the solution**

```
use objc::runtime::{
    objc_getClass
};

unsafe fn add_url_handler() {
    let class_name = CString::new("WinitApplicationDelegate").expect("CString::new failed");
    let res = objc_getClass(class_name.as_ptr());

    // TODO: use class_addMethod http://sasheldon.com/rust-objc/objc/runtime/fn.class_addMethod.html to add a `application:openFiles:` method to the application delegate

    // TODO: tell neovim to open the file
}
```

abhillman:
For a simple example demonstrating fetching the `ApplicationDelegate` that can be expanded upon as a kind of harness, if desired, which could then be ported to this repo:

```
use std::ffi::CString;
use winit::{
    event_loop::EventLoop,
    window::WindowBuilder
};

use objc::runtime::{
    objc_getClass
};
use winit::event::Event;

unsafe fn add_url_handler() {
    let class_name = CString::new("WinitApplicationDelegate").expect("CString::new failed");
    let res = objc_getClass(class_name.as_ptr());

    // Prints a memory address like 0x6000019d85a0
    println!("{:?}", res);
}

fn main() {
    let args: Vec<String> = std::env::args().collect();
    std::fs::write("/tmp/argv.txt", args.join(" ")).unwrap();

    let event_loop = EventLoop::new().unwrap();
    let _window = WindowBuilder::new().build(&event_loop).unwrap();

    let loop_proxy = event_loop.create_proxy();
    loop_proxy.send_event(()).unwrap();

    event_loop.run(|event, target| {
        match event {
            Event::UserEvent(_) => {
                // As-written, this will only be called once. It is called
                // due to the `loop_proxy.send_event(())` call above.
                unsafe { add_url_handler() }
            }
            _ => {}
        }
    }).expect("TODO: panic message");
}
```

In order to use it, one would create a rust project with the proper dependencies, run `cargo bundle`, and modify the resulting app bundle's `Info.plist` per https://github.com/neovide/neovide/pull/2191#issuecomment-1913460232. From there, dropping a document on the binary will result in calling the application delegate's `application:openFiles:` method, if defined.

abhillman:
I'm a novice at Rust and macOS development, so please forgive me in advance if I am speaking in a way that is it all overly basic.

9mm:
@abhillman awesome! regarding your Info.plist, the only reason all the other keys are required if I remember right is that OSX requires that you explicitly specify every single filetype. Wildcard `*` does not work anymore (it's been awhile since I checked). I have confirmed in a number of other OSX apps and they all do this (specify every single file type), so I followed their lead.

@fredizzimo what do you think of that above? I don't really know rust well enough. That would be awesome if this works tho

Is there a way 2 people can collaborate on a PR so he can actually commit that to my PR?

9mm:
I can also give this a try momentarily ,my brain is cobwebs on all this though, I'll have a look again today

fredizzimo:
I don't really know what a good approach is, perhaps it would be worth asking the Winit team what their long-term vision is, and if this could be a good short-term workaround?

@9mm, regarding collaboration, you can either give the other person write permissions to your branch, or they can send a PR to your fork targeting your branch, which you can then merge. But you have to be careful with re-basing if you are co-operating on the same branch, it's the safest to always use merges, they will be cleaned up anyway when we finally squash the PR.

polachok:
I sort of implemented this in https://github.com/polachok/neovide/commit/9802ab31cd91773b07bb0e7b38f113355ac1ef2a with some limitations:
* it doesn't work when neovide isn't already open (not sure how to fix this)
* if multiple files are dragged onto icon, only one file is visible (macvim opens them in tabs)
* uses the `WinitApplicationDelegate` hack above

9mm:
Awesome! I will check it out soon 💯

polachok:
>it doesn't work when neovide isn't already open (not sure how to fix this)

It actually works if I disable forking. Also, this makes "open with>Neovide" menu in finder work. I guess the "openFiles:" request is sent to the original process.

9mm:
Interesting, thats def a question for @fredizzimo 

Nice work this is super exciting. This is the main feature left thats painful, so Im really stoked!

fredizzimo:
For the forking, we have this issue:
* https://github.com/neovide/neovide/issues/2147

TLDR; 
I propose that we change the defaults to not fork, neither from the GUI nor from the terminal. That aligns better with how most applications behave when launched from the terminal. But it's not a decision I can take by my own.

polachok:
Just opened a draft PR for this: https://github.com/neovide/neovide/pull/2395

fredizzimo:
I guess this can be closed now?

9mm:
I believe so!

PR #2198, opened 2023-12-24
add IME preedit support

OliverChao:
- Feature

IME preedit is something lost in Neovide. 
This pr support ime preedit.

https://github.com/neovide/neovide/assets/39361033/332d09a9-edae-456b-aa1c-1edd675da74b

- No


src/window/keyboard_manager.rs
```diff
@@ -56,10 +56,13 @@ impl KeyboardManager {
...
-                event: WindowEvent::Ime(Ime::Preedit(text, cursor_offset)),
-                ..
-            } => self.ime_preedit = (text.to_string(), *cursor_offset),
+            // Event::WindowEvent {
+            //     event: WindowEvent::Ime(Ime::Preedit(text, cursor_offset)),
+            //     ..
+            // } => {
+            //     self.ime_preedit = (text.to_string(), *cursor_offset);
+            //     log::trace!("Ime preedit {text}");
+            // }
```
fredizzimo, line 65: I would prefer to keep the ime state in the keyboard manager. And I also think that this breaks something, since we used to block keyboard inputs when the pre-edit was active, and that code is no longer effective.

src/window/window_wrapper.rs
```diff
@@ -559,8 +596,9 @@ impl WinitWindowWrapper {
             cursor_position.x.round() as i32,
             cursor_position.y.round() as i32 + font_dimensions.height as i32,
         );
-        if position != self.ime_position {
-            self.ime_position = position;
+        if position != self.ime.position() {
+            // self.ime.position = position;
```
fredizzimo, line 600: This commented code should be removed

src/window/window_wrapper.rs
```diff
@@ -317,6 +321,32 @@ impl WinitWindowWrapper {
...
+                ..
+            } => {
+                if size.width != 0 && size.height != 0 {
+                    self.window_width = size.width as f32;
+                }
+            }
+            Event::WindowEvent {
+                event: WindowEvent::Ime(ime),
+                ..
+            } => match ime {
```
fredizzimo, line 336: The clippy warnings should be fixed

src/renderer/ime.rs
```diff
@@ -0,0 +1,82 @@
...
+        Self {
+            text,
+            cursor_offset,
+        }
+    }
+    pub fn preedit_text(&self) -> &String {
+        &self.text
+    }
+    pub fn cursor_offset(&self) -> Option<usize> {
+        if let Some(offset) = self.cursor_offset {
```
fredizzimo, line 76: This clippy warning should be fixed

src/renderer/ime.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    #[inline]
+    pub fn set_position(&mut self, p: PhysicalPosition<i32>) {
+        self.position = p;
+    }
+
+    #[inline]
+    pub fn position(&self) -> PhysicalPosition<i32> {
+        return self.position;
```
fredizzimo, line 51: This clippy warning should be fixed

src/renderer/cursor_renderer/mod.rs
```diff
@@ -341,6 +347,54 @@ impl CursorRenderer {
...
+            let off = if let Some(off) = preedit.cursor_offset() {
+                off as f32 * factor
+            } else {
+                0f32
+            };
+
+            let mut start = self.destination.x;
+            let mut end = self.destination.x + off;
+            if end + factor > w {
+                start = start - (end + factor - w);
```
fredizzimo, line 364: This clippy warning should be fixed

src/renderer/cursor_renderer/mod.rs
```diff
@@ -341,6 +347,54 @@ impl CursorRenderer {
 
         canvas.restore();
 
+        // draw ime if possible under current cursor
+        if let Some(preedit) = preedit {
+            let factor = grid_renderer.font_dimensions.width as f32;
+            let y_adjustment = y_adjustment as f32 + factor / 5.0;
```
fredizzimo, line 353: What does this factor of 5.0 do?

src/renderer/cursor_renderer/mod.rs
```diff
@@ -341,6 +347,54 @@ impl CursorRenderer {
...
+            for blob in blobs.iter() {
+                canvas.draw_text_blob(blob, (start, self.destination.y + y_adjustment), &paint);
+            }
+
+            paint.set_color(skia_safe::colors::GREY.to_color());
+            let r = Rect::new(
+                start,
+                self.destination.y + y_adjustment,
+                end,
+                self.destination.y + y_adjustment + factor / 4.0,
```
fredizzimo, line 393: What does this factor 4.0 do?

src/renderer/cursor_renderer/mod.rs
```diff
@@ -341,6 +347,54 @@ impl CursorRenderer {
...
+            let blobs =
+                &grid_renderer
+                    .shaper
+                    .shape_cached(preedit.preedit_text().clone(), bold, italic);
+            paint.set_color(foreground_color);
+            for blob in blobs.iter() {
+                canvas.draw_text_blob(blob, (start, self.destination.y + y_adjustment), &paint);
+            }
+
+            paint.set_color(skia_safe::colors::GREY.to_color());
```
fredizzimo, line 388: This should use the correct highlight group color, otherwise it might be invisible. I think this is the selection right?

Also, shouldn’t it be drawn before the text? So that the text remain invisible, with the text color also set apporopriately through the highlight groups.

I also think that the floating window transparency need to be taken into account.

src/renderer/mod.rs
```diff
@@ -145,7 +147,7 @@ impl Renderer {
         self.cursor_renderer.prepare_frame()
     }
 
-    pub fn draw_frame(&mut self, root_canvas: &Canvas, dt: f32) {
+    pub fn draw_frame(&mut self, root_canvas: &Canvas, dt: f32, ime: &Ime, w: f32) {
```
fredizzimo, line 150: Can you give the  the `w` parameter a more descriptive name? We don’t use the shorthand for width anywhere else?

src/renderer/cursor_renderer/mod.rs
```diff
@@ -341,6 +347,54 @@ impl CursorRenderer {
...
+
+            let off = if let Some(off) = preedit.cursor_offset() {
+                off as f32 * factor
+            } else {
+                0f32
+            };
+
+            let mut start = self.destination.x;
+            let mut end = self.destination.x + off;
+            if end + factor > w {
```
fredizzimo, line 363: I think ideally we should use the neovim grid window width here, not the global window width, otherwise it will render outside floating window and splits for example.

This als means that the code better belongs in `rendered_window`, where the width and tranaparency needed is available.

fredizzimo (changes requested):
Once the issues are fixed, I think we can go with this, since it’s better than nothing.

But there will still be problems, like word wrapping, especially on the last line. The colors might also not be as expected in all places. So I think we eventually should change the implementation to something else.

When I was thinking about this previously, I had the idea using the regula neovim insert mode for displaying the text, with a lua side plugin helping to deal with that. For example to handle aborts, or selecting another entry. The cursor would also be native, and the selection could use a custom highlight group or perhaps a built in, again, using the lua plugin to mark it.

That way we would not need any special rendering in Neovide, and word wrapping/scrolling of the buffer would happen automatically. We could even integrate proper code completion that way.

OliverChao:
very sorry for the late reply. I am a lot bit busy currently. 
This pr seems far from completed especially blob string drawing. 
I am finding one easy way to handle it.

fredizzimo:
Don't worry, I think it's quite close, if we just want an initial version that is usable, but indeed there's a lot to think about and fix if we want it to work perfectly. 

You don't need to hurry though. And if you want to attempt the other way, using Neovim insert mode, I can give you a more detailed design of what I had planned. I think that's the best way of doing it, but the implementation is more complex, at least to get started.

fredizzimo:
I'm closing this, since we added
* https://github.com/neovide/neovide/pull/3110

Issue #2203, opened 2023-12-25
Scroll animations being skipped when using some plugins


Mixaz017:
**Describe the bug**
Scroll animations do not play correctly with some plugins in 0.12, even though it used to work in previous versions. When this happens, scrolling is instant, instead of the viewport scrolling smoothly.

**To Reproduce**
1. Install [nvim-scrollview](https://github.com/dstein64/nvim-scrollview) (some other plugins are affected as well, I'm guessing it has to do with a floating buffer overlapping with the scrolling buffer)
2. Open a file
3. Scroll around the buffer

**Expected behavior**
Scroll animations play like it used to in previous versions

**Screenshots**
Scrolling line by line with nvim-scrollview:

https://github.com/neovide/neovide/assets/31244102/a0f97c79-f6a7-44cb-9c79-1241f4595b08

Even if I disable nvim-scrollview, scroll animation is skipped when [nvim-treesitter-context](https://github.com/nvim-treesitter/nvim-treesitter-context) floating buffer disappears (when pressing C-u in this case):

https://github.com/neovide/neovide/assets/31244102/58cba0e7-702d-4945-a392-05daae0a9534

Doing the same in 0.11.1:

https://github.com/neovide/neovide/assets/31244102/1e165070-dce8-47b2-97b8-7058af06cd92

https://github.com/neovide/neovide/assets/31244102/9beede16-7673-4d6c-b13b-e34ace41ae6c

**Desktop (please complete the following information):**
 - OS: Arch Linux with KDE Plasma 5.27.10 (X11)
 - Neovide Version: 0.12.0
 - Neovim Version: 0.9.4

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/13764093/neovide_rCURRENT.log)

fredizzimo:
I have debugged the `nvim_scrollview` a bit. And this is what happens
```
TRACE [neovide::channel_utils] neovim_handler GridLine { grid: 5, row: 0, column_start: 167, cells: [GridLineCell { text: " ", highlight_id: Some(130), repeat: Some(2) }, GridLineCell { text: " ", highlight_id: Some(0), repeat: Some(0) }] }
[...]

PR #2208, opened 2023-12-26
fix(scrolling): don't reset the scrolling when the window position changes

fredizzimo:
The scroll animation was previously always reset when the window position changed, because the same command was used for positioning and resizing of the windows. This is now changed so that the scroll animation is kept for position or width changes, and only reset when the height changes.

This also add a tracy profiler graph of the scroll position.

* Fixes https://github.com/neovide/neovide/issues/2203

- No


MultisampledNight:
Thank you! Chooooooooooooooo chooooooooooooooo :steam_locomotive:

Issue #2201, opened 2023-12-24
Windows: Neovide 0.12 regressions


VladimirMarkelov:
**Describe the bug**

Neovim `0.9.4` (Neovide 0.12 requires 0.9.2 or newer, so it must be fine).
Neovim is in the directory `d:\Utils\neovim\bin\`, I put the new Neovide version **into the same** directory.

Comparing to the previous version of Neovide, the new version has a few regressions:

- If I am not mistaken, there was such an issue with older versions and it was fixed: Neovide 0.12 parses the command line options and unescapes them. Example: since Neovide 0.11.2 I used a command `"d:\Utils\neovim\bin\neovide.exe" --neovim-bin="d:\utils\neovim\bin\nvim.exe" --geometry=140x45` - it worked. First, I had to change, as Neovide suggested, the command line to `"d:\Utils\neovim\bin\neovide.exe" --neovim-bin="d:\utils\neovim\bin\nvim.exe" -- --geometry=140x45` (added extra ` -- ` before `--geometry`). At the same time, Neovide 0.12 started requiring different kind of slashes: `d:\Utils\neovim\bin\neovide.exe" --neovim-bin="d:/utils/neovim/bin/nvim.exe" -- --geometry=140x45`. Only after all those changes I managed to start Neovide 0.12.
- But it was not the only issue. Now I have no clue how to pass `geometry`. If I do not add an extra `--`, Neovide complains that the extra options must be passed after `--`. When I add ` -- ` before `--geometry`, Neovim starts with a default tiny window. How to set window size now?
- The last issue: since 0.12 Neovide started complaining about unknown options in `ginit.nvim`. It contains commands for `neovim-qt`, like `call GuiWindowMaximized(1)` and `GuiFont!`. Neovide 0.12 complains at both command, so I had to comment them out, because the way I use for `init.vim` does not work here. I tried to add this:

```
if exists("g:nvy")
else
	call GuiWindowMaximized(1)
	GuiFont! JuliaMono:h11
endif
```

But Neovide keeps complaining about non-existing functions. How to mark commands in `ginit.vim` so Neovide skips them?

**To Reproduce**
1. `cd d:\utils\neovim\bin`
2. `neovide` - works fine, but the windows is tiny 
3. `neovide --neovim-bin="d:\utils\neovim\bin\nvim.exe"` - crashes with:

![image](https://github.com/neovide/neovide/assets/1577186/972ef505-1066-4dc0-8479-6fbc70022f67)

[...]

PR #2210, opened 2023-12-26
fix: Don't split the neovim-bin with shlex on non-wsl windows

fredizzimo:
The shlex crate does not handle windows path separator correctly, so don't try to use it on Windows, except when targeting WSL, and Windows paths are involved.

* Fixes https://github.com/neovide/neovide/issues/2201

- No


MultisampledNight:
Can't test on Windows, but code looks good to me. Thank you!

MultisampledNight:
Chooooooooooooooooooo choooooooooooooo :steam_locomotive:

PR #2213, opened 2023-12-28
feat(config): watch config and font features

TwIStOy:
After some discussion in #1920 

This pr includes

- Watch `config.toml`. Only `font` can be hot-reloaded now.
- Font settings now looks like [`alacritty`](https://alacritty.org/config-alacritty.html).
- Support `font.features` settings

- Feature

- No

### DIFF
--- a/src/settings/config.rs
+++ b/src/settings/config.rs
@@ -67,17 +67,7 @@ impl Config {
     }
 
     pub fn watch_config_file(init_config: Config, event_loop_proxy: EventLoopProxy<UserEvent>) {
-        let (tx, rx) = mpsc::channel();
-        let mut watcher = notify::RecommendedWatcher::new(
-            tx,
-            notify::Config::default().with_compare_contents(true),
-        )
-        .unwrap();
-        watcher
-            .watch(&config_path(), notify::RecursiveMode::NonRecursive)
-            .unwrap();
-        std::mem::forget(watcher);
-        std::thread::spawn(move || watcher_thread(init_config, rx, event_loop_proxy));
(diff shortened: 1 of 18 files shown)


fredizzimo:
Thank you very much for taking time to implement this!

I don’t have acess to a computer to run Neovide at the moment, so I can’t really test it and give feedback at the moment.

But I think someone else will take a look shortly. Maybe some documentation would be good to add, so we can better see if we want to make some small tweaks to the schema.

src/settings/config.rs
```diff
@@ -42,16 +51,69 @@ pub struct Config {
...
+        let mut watcher = notify::RecommendedWatcher::new(
+            tx,
+            notify::Config::default().with_compare_contents(true),
+        )
+        .unwrap();
+        watcher
+            .watch(&config_path(), notify::RecursiveMode::NonRecursive)
+            .unwrap();
+        std::mem::forget(watcher);
+        std::thread::spawn(move || {
```
MultisampledNight, line 85: Might fit better as a standalone function in order to avoid too much nesting.

src/settings/config.rs
```diff
@@ -42,16 +51,69 @@ pub struct Config {
...
+            let mut previous_config = init_config;
+            loop {
+                match rx.recv() {
+                    Ok(_) => {
```
MultisampledNight, line 89: Could make use of early return.

MultisampledNight (commented):
Thank you _a lot_, for the past and this PR! That's a lot of work, whew. I'll take care of what I annotated and then merge this!

MultisampledNight:
Chooooooooooo chooooooooooo! Thank you! :steam_locomotive:

PR #2215, opened 2023-12-28
Direct3D Rendering on Windows

fredizzimo:
This switches to a Direct3D based renderer and finally makes the Windows render loop smooth with almost no frame drops, at least on my machine. 

Using Direct3D will also improve the compatibility with older hardware.

NOTE:
* You can still use the old opengl based rendering by passing the  `--opengl` option. 
* The `--no-vsync` option does not work properly at the moment
* There might be bugs when the device is lost/computer sleeps/monitor turned off
* The device selection code is somewhat naive
* The code probably needs some cleanup, and that's the biggest reason for the draft status. I will probably split this PR into two, one pure refactoring, and another one that adds the functionality.
* This also seem to reduce the stuttering in some Wayland compositors like Mutter/Gnome at the cost of a very, very slight possible additional latency after pressing a key, less than a frame
* `Asus GPU Tweak III` causes Neovide to stutter when it’s running. Not because it messes with the GPU, but because it seem to do something to the Windows messaging system, all calls, even `PeekMesage`, with should be almost instant,  take a long time to complete. In the case of `PeekMessage` sometimes a couple of ms, and several of those in a row will cause dropped frames. So, I suggest closing the program or find an alternative. Neovide is probably not the only application affected.
* The gpu profiling is almost unusably slow, and I think we would need access to the skia internal command queues to make it faster. But maybe there’s another way?

- No, it should not

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -195,7 +195,7 @@ checksum = "a507401cad91ec6a857ed5513a2073c82a9b9048762b885bb98655b306964681"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.55",
+ "syn 2.0.57",
 ]
 
 [[package]]
@@ -206,9 +206,9 @@ checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
 name = "autocfg"
-version = "1.1.0"
+version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
+checksum = "f1fdabc7756949593fe60f30ec81974b613357de856987752631dea1e3394c80"
 
 [[package]]
 name = "backtrace"
@@ -250,7 +250,7 @@ dependencies = [
  "regex",
  "rustc-hash",
  "shlex",
- "syn 2.0.55",
+ "syn 2.0.57",
  "which 4.4.2",
 ]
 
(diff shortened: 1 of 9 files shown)


A-Lamia:
had really bad lag with OpenGL and AMD on windows, now using this branch 98% of the lag gone a small hick up every once in a while but other then that ill be using this branch from now on amazing work! :)

sid-6581:
Hmm.. On my laptop, this is very laggy and considerably less smooth. I haven't done any more investigation to find the cause yet, but I figured I should mention it.

fredizzimo:
> Hmm.. On my laptop, this is very laggy and considerably less smooth. I haven't done any more investigation to find the cause yet, but I figured I should mention it.

My initial guess is that it selects the wrong graphics adapter, as mentioned that code is pretty naive currently

A-Lamia:
can this be rebased on to master or will it explode ?

fredizzimo:
@A-Lamia, I just pushed the re-based version. I had already done it locally, so it was easy.

Kethku:
I've been using this for a week and as far as I can tell its just a straight improvement. Excited to get this merged

fredizzimo:
I will create a new PR for the refactoring part first to make things easier to review.

Then there might be some small cleanup needed. And I think we should also solve the adapter selection problem, I think that's what @sid-6581 are seeing, before merging.

fredizzimo:
The code could also be a lot cleaner if we switch to `windows-rs` first? Should we do that?

Kethku:
@sid-6581 I merged this as its a very large improvement on my machines which span a variety of specs. Can you try again with current main maybe with profiling enabled? I'd like to dig into what is going wrong

sid-6581:
> @sid-6581 I merged this as its a very large improvement on my machines which span a variety of specs. Can you try again with current main maybe with profiling enabled? I'd like to dig into what is going wrong

So far so good after an initial test on my main machine with a 4090. I'll keep using it and trying it on my laptops as well to make sure everything looks good there as well.

sid-6581:
@Kethku I've done some more testing, and here are my findings:

* On my main machine with a 4090, speed is fine, but rendering speed was also never a problem previously, so I don't really see much of an improvement. The big problem I have with this PR on my main machine is that Alt+Enter is hijacked to toggle fullscreen, which is a problem because I have that mapped in neovim.
* On my laptop with hybrid Intel/NVidia 3080 Ti graphics, it still runs noticeably slower than the previous renderer, and latency while typing is slow enough to be frustrating to use. If I run neovide with `--opengl` it runs fine again. Curiously, Alt+Enter does not toggle fullscreen on this machine, even though I compiled exactly the same commit.

Unfortunately I had to revert back to the commit before this PR to make neovide usable on both machines again.

Issue #2214, opened 2023-12-28
Regression on font sizing on neovide 0.12.0


borisfaure:
**Describe the bug**
Font size in font settings `set guifont=PragmataPro\ Mono\ Liga:h10` not taken into account with neovide-0.12.0

**To Reproduce**
1. start neovide

**Expected behavior**
The font appears large and any change to the guifont parameter is not taken into account. This worked fine on 0.11.2

**Screenshots**
.

**Desktop (please complete the following information):**
 - OS: Gentoo Linux
 - Neovide Version 0.12.0 (or 5743356 that introduced this regression)
 - Neovim Version 0.9.4

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
The file is too large, but here is `grep -i font neovide_rCURRENT.log` 
```
54:TRACE [neovide::channel_utils] neovide::editor::EditorCommand NeovimRedrawEvent(OptionSet { gui_option: GuiFont("") })
55:TRACE [neovide::channel_utils] neovide::editor::EditorCommand NeovimRedrawEvent(OptionSet { gui_option: GuiFontWide("") })
67:TRACE [neovide::editor] Option set GuiFont("")
69:TRACE [neovide::editor] Option set GuiFontWide("")
90:TRACE [neovide::channel_utils] alloc::vec::Vec<neovide::renderer::DrawCommand> [FontChanged(""), LineSpaceChanged(0), DefaultStyleChanged(Style { colors: Colors { foreground: Some(Color4f { r: 1.0, g: 1.0, b: 1.0, a: 1.0 }), background: Some(Color4f { r: 0.0, g: 0.0, b: 0.0, a: 1.0 }), special: Some(Color4f { r: 1.0, g: 0.0, b: 0.0, a: 1.0 }) }, reverse: false, italic: false, bold: false, strikethrough: false, blend: 0, underline: None })]
296:TRACE [neovide::channel_utils] neovide::editor::EditorCommand NeovimRedrawEvent(OptionSet { gui_option: GuiFont("PragmataPro Mono Liga:h10") })
323:TRACE [neovide::editor] Option set GuiFont("PragmataPro Mono Liga:h10")
[...]

PR #2218, opened 2023-12-28
fonts/caching_shaper: use scale_factor when computing fudge_factor

borisfaure:
Closes: #2214

- Fix

- No

### DIFF
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -36,6 +36,7 @@ pub struct CachingShaper {
     blob_cache: LruCache<ShapeKey, Vec<TextBlob>>,
     shape_context: ShapeContext,
     scale_factor: f32,
+    /// factor get pixel-perfect width
     fudge_factor: f32,
     linespace: i64,
     font_info: Option<(Metrics, f32)>,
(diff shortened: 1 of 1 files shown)


src/renderer/fonts/caching_shaper.rs
```diff
@@ -164,12 +165,13 @@ impl CachingShaper {
                 "Font width: {:.2}px (avg: {:.2}px)",
                 font_width, metrics.average_width
             );
-            self.fudge_factor = font_width.round() / font_width;
+            self.fudge_factor = font_width.round() / font_width / self.scale_factor;
```
fredizzimo, line 168: I’m not sure that using the scale factor here is correct, simce the width should already be scaled with that, and so should the pixels.

But I don’t have a development environment to test it with at the moment, so I might be wrong.

fredizzimo: After looking at it again, I still think I'm correct. Dividing by the scale factor here will just make the fudge factor wrong, unless the scale is close to 1.

What do you think @borisfaure? Why was it added?

fredizzimo: Setting `self.info` to none should be enough. And the move of the code ensures that the debug output is wrong.

Setting it to none is needed, since otherwise cached information will be used.

borisfaure: I agree it looks weird. I'll do more tests on it with multiple screens and which should bring me different scale values

borisfaure:
Closing this PR for the moment since it needs more work

PR #2238, opened 2024-01-05
fix: don't hide random text

MultisampledNight:
@falcucci noticed that https://github.com/neovide/neovide/pull/2208 for him re-introduced the problem with text being hidden randomly. I couldn't reproduce this myself, but apparently this patch works for him.

- Fix
  - Regression as introduced by https://github.com/neovide/neovide/pull/2208

- No


MultisampledNight:
(merging this myself since I deem it quite important for 0.12.1)

falcucci:
thank you for this 💯

PR #2239, opened 2024-01-05
Draft: Add native border on floating window 

TwIStOy:
- Feature 

- No

Draw a rectangle rounding floating windows. Also change color when entering that window.

### DIFF
--- a/src/renderer/cursor_renderer/mod.rs
+++ b/src/renderer/cursor_renderer/mod.rs
@@ -477,4 +477,8 @@ impl CursorRenderer {
     pub fn get_current_position(&self) -> Point {
         self.destination
     }
+
+    pub fn get_cursor_window(&self) -> u64 {
+        self.cursor.parent_window_id
+    }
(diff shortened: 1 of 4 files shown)


fredizzimo:
I have not reviewed the code yet, but I think this is a good feature.

I'm just not sure about the configuration. It might be good enough for initial support, but I think we eventually should be able to configure this, and the shadows, which are similar, in the same way as native floating window borders. So, that it's window specific and can be applied to just the floating windows that need it.

For that we need to discuss with the Neovim team on what's the best approach. Perhaps, a new setting guiborder, with some defaults. It would also be nice to render the current native borders, using our own renderer.  The border propery as it is, is not enough, since it does not allow combining shadows with regular border styles.

TwIStOy:
Yeh, we need some more information about these floating windows from Neovim. It's not easy to determine if the native border should be rendered or not now. This PR now adds a lot of borders on these floating windows.

PR #2241, opened 2024-01-05
fix: #2233 by replacing echo command with api call

### LINKED ISSUE #2233: Error in init.lua freezes Neovide
**Describe the bug**
Any lua error in init.lua freezes Neovide window, that is the Neovide window becomes unresponsive for keyboard and mouse inputs and has to be terminated by Windows task manager.

**To Reproduce**
1. insert `asset(false)` in top of init.lua
2. run Neovide
3. See error

**Expected behavior**
Neovide prints error messages and then continue without freezing

**Screenshots**
화면 캡처 2024-01-03 061027

 - OS: Windows 11 (10.0.22631)
 - Neovide Version 0.12.0
 - Neovim Version 0.9.5

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/13813959/neovide_rCURRENT.log)

**Additional context**
Initially I noticed this when changing settings for some plugins, and found out freezing can be reproduced consistently with minimal setup `assert(false)`
Running nvim in Windows terminal with identical setup doesn't crash or freeze Neovim

genzj:
- Fix

- No

### DIFF
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -46,6 +46,32 @@ pub enum SerialCommand {
     },
 }
 
+fn has_nvim_version(metadata: &Value, major: i64, minor: i64) -> bool {
+    if let Some(map) = metadata.as_map() {
+        if let Some(Some(y)) = map
+            .iter()
+            .find(|&(k, _)| k.as_str() == Some("version"))
+            .map(|(_, value)| value.as_map())
+        {
+            let mut actual_major: i64 = -1;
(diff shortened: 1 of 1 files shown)


fredizzimo:
This is good. I think alternatively we could perhaps test the api_level, but I'm not exactly sure how it works, so this is good at least for now. For new functions we could also check the existense of the function, or the parameters, but neither is enough in this case.

hutterm:
@fredizzimo what's the plan when this fix will be released?

Issue #2240, opened 2024-01-05
config.toml setting for specifying fallback fonts


comiluv:
**Is your feature request related to a problem? Please describe.**
Neovide previously enabled users to set up fallback fonts for CJK characters, emojis, etc. within the `init.lua` file under `vim.o.guifont`. This was achieved by specifying fallback fonts after the default fonts. For instance, an example setup would be `vim.o.guifont = "consolas, Noto_Color_Emoji:h11"`, indicating a fallback to the Noto Color Emoji font for character sets not available in Consolas.
However, with the release of Neovide 0.12.1, users now have the ability to configure fonts and font features in the `config.toml` file. While this enhancement is a great addition to Neovide, it currently lacks the capability for users to establish fallback fonts to accommodate diverse requirements.

**Describe the solution you'd like**
Add an option to specify a fallback font in config.toml file

**Describe alternatives you've considered**
I think support for fallback fonts in config.toml would be best for symmetry

TwIStOy:
You can also specify fallback fonts in `config.toml`. Like,

```toml
[font]
normal = [ "Symbols Nerd Font Mono", "FiraCode" ]
```

Sorry for the document of this feature is not clear.

TwIStOy:
Ah, It's an issue about parsing these configurations. Please try #2244. It seems work now!

BatmanAoD:
It seems that even in `config.toml`, fallback is "noisy" as described in this issue: https://github.com/neovide/neovide/issues/2533

Is that intentional? I use the same vim/neovide config across all my machines, and don't necessarily have my preferred fonts set up on all of them, and my preferred pre-installed font depends on the OS. So the "silent fallback" behavior was very convenient for me.

PR #2244, opened 2024-01-07
fix: #2240 by swap item order in `FontDescriptionSettings`

TwIStOy:
- Fix

- No

### DIFF
--- a/src/settings/font.rs
+++ b/src/settings/font.rs
@@ -16,8 +16,8 @@ pub enum SimpleFontDescription {
 #[derive(Debug, Clone, Deserialize, PartialEq)]
 #[serde(untagged)]
 pub enum FontDescriptionSettings {
-    Single(SimpleFontDescription),
     Vec(Vec<SimpleFontDescription>),
+    Single(SimpleFontDescription),
 }
 
 #[derive(Debug, Clone, Deserialize, PartialEq)]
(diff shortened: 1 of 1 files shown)


src/settings/font.rs
```diff
@@ -132,3 +132,28 @@ impl From<FontSettings> for FontOptions {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_font_settings() {
```
fredizzimo, line 141: Can you also add a test for specifing a single font, to make sure that case didn't break instead?

src/settings/font.rs
```diff
@@ -132,3 +132,28 @@ impl From<FontSettings> for FontOptions {
...
+            "size": 20
+        }
+        "#;
+
+        let settings: FontSettings = serde_json::from_str(settings).unwrap();
+        match settings.normal {
+            FontDescriptionSettings::Vec(fonts) => {
+                assert_eq!(fonts.len(), 2);
+                // assert_eq!(fonts[0].family, "Consolas");
+                // assert_eq!(fonts[1].family, "Noto Emoji");
```
fredizzimo, line 154: Is there a reason why these are not tested, and commented out instead?

fredizzimo (changes requested):

fredizzimo (approved):

Kethku:
Thanks for your fix!

PR #2246, opened 2024-01-08
feat: support all font weights and slants

TwIStOy:
- Feature

- No


fredizzimo:
Thank You!

Do you think it would be possible to also support numerical values, perhaps `Weight400`, for the same as normal for example? At least weight supports it, and perhaps slant as well.

This was already attempted here, but stalled due to not having a good way of configuring it
* https://github.com/neovide/neovide/pull/2121

TwIStOy:
Make variable font weight can be set in font's style now!

fredizzimo:
Great! I think this is good once some documentation is added with the supported options.

fredizzimo (approved):
Thank you, this looks good to me. But I'm leaving it open for a while before merging, to allow other people to review.

falcucci (approved):

Kethku:
Thanks for the change! Works great

Issue #2234, opened 2024-01-03
Freeze when opening neo-tree (?)


Jomik:
**Describe the bug**
When I open my Go project, dotfiles or pretty much any project. I can open individual files, but as soon as I open `neo-tree` it freezes. I cannot resize the window or anything.
I initially believed it to be any project, because it freezes as soon as I open a directory.
But I do believe it is something to do with the rendering that happens when neo-tree is opened.
I do not have the error in Neovide 0.11.2

**To Reproduce**
1. Open a project folder with `neovide ~/projects/foo`, open neo-tree
2. Observe frozen Neovide screen.

**Expected behavior**
Neovide to show neovim as normal :)

**Screenshots**
![image](https://github.com/neovide/neovide/assets/699655/7bc008fb-e118-4256-b91d-f4ca04b3a96b)

**Desktop (please complete the following information):**
 - OS: MacOS 14.2.1
 - Neovide Version: 0.12.0
 - Neovim Version: NVIM v0.10.0-dev-2021+g[ee2127363](https://github.com/neovim/neovim/commit/ee2127363)-Homebrew

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
They are quite long, made a gist : https://gist.github.com/Jomik/e90f9b0460caa76e7f2ef3fccc9af5ee

**Additional context**

Mentalist:
Experiencing (almost) the same issue (mine instantly crashes), it happens when launching an instance of neovide with a directory path as argument or when triggering Neo-tree.nvim manually. This behavior is not reproducible in the kitty and alacritty terminal emulators nor Neovim-Qt.

- OS: Arch Linux (Kernel 6.6.10-arch-1-1)
- Neovide Version: 0.12.1
- Neovim Version: v0.10.0-dev-2048+g367e52cc79

Commands:

```sh
$ RUST_BACKTRACE=1 cargo run -- --log ./ # instant crash
$ RUST_BACKTRACE=1 cargo run -- --log # crash on toggle Neo-tree.nvim
```

<details>
  <summary>neovide_backtrace.log</summary>
  
  ```2024-01-07 21:11:45 - Neovide panicked with the message 'attempt to calculate the remainder with a divisor of zero'. (File: /rustc/e51e98dde6a60637b6a71b8105245b629ac3fe77/library/core/src/num/mod.rs; Line: 455, Column: 5)
   0: neovide::main::{{closure}}
             at src/main.rs:78:25
[...]

PR #2247, opened 2024-01-08
fix: #2234 by adding !self.scrollback_lines.is_empty() check

Mentalist:
Fixes https://github.com/neovide/neovide/issues/2234 as per suggestion by [fredizzimo](https://github.com/fredizzimo).

- Fix

- No


fredizzimo:
Thank you!

PR #2256, opened 2024-01-09
Convert Windows paths to WSL when launching in WSL mode

### LINKED ISSUE #2194: Neovide does not gracefully handle windows paths in wsl mode.
**Describe the bug**
Neovide does not gracefully handle windows paths in wsl mode. 

**To Reproduce**
On WSL1 with [automount enabled](https://learn.microsoft.com/en-us/windows/wsl/wsl-config#wslconf):
1. Set `wsl = true` in config
2. Double click in a file associated with neovide on the windows side.

**Expected behavior**
Either by using [`/bin/wslpath -a`](https://stackoverflow.com/questions/53474575/how-to-translate-the-wslpath-home-user-to-windows-path#53548414), or by [working across filesystems](https://learn.microsoft.com/en-us/windows/wsl/filesystems), or by falling back to using nvim from the windows side.

 - OS: Windows 11
 - neovide 0.11.2
 - NVIM v0.9.4; Build type: MinSizeRel; LuaJIT 2.1.0-beta3

TRACE [neovide] Neovide version: 0.11.2
DEBUG [neovide::bridge::command] Running which command: "wsl" "$SHELL" "-lc" "which nvim"
DEBUG [neovide::bridge::command] Starting neovim with: Command { std: "wsl" "$SHELL" "-lc" "/usr/bin/nvim --embed -p", kill_on_drop: false }

Holzhaus:
This uses the ~~`wslpath2`~~ (EDIT: `wslpath-rs`) crate to convert Windows paths to WSL paths if WSL mode is active. If conversion of a path fails, it will be retained as-is and passed to neovim unchanged (as before).

Fixes #2194.

- Fix

- No, I don't think so.


Holzhaus:
**Disclaimer:** I didn't perform a manual test on an actual windows machine yet (I don't use Windows at home). However, I implemented a test and it works on CI. Maybe some can volunteer (@pedrohgmacedo maybe?). Otherwise I can check if this works next time I have access to a windows machine.

pedrohgmacedo:
I used the build artifacts to test it on my machine, and it works, but it also has the side effect of creating a Terminal window when launching, which is a regression from #2125.

Holzhaus:
> I used the build artifacts to test it on my machine, and it works, but it also has the side effect of creating a Terminal window when launching, which is a regression from #2125.

Yeah, that is likely a bug in the wslpath2 package: https://github.com/michidk/wslpath2/blob/develop/src/lib.rs#L74-L77

I'll file a PR, and once they released a fix I can update the dependency here.

Holzhaus:
https://github.com/michidk/wslpath2/pull/1

pedrohgmacedo:
Can you create a build with the patched wslpath2? Then I can test it. Their last commit was 9 months ago, waiting for them might take some time.

Holzhaus:
> Can you create a build with the patched wslpath2? Then I can test it. Their last commit was 9 months ago, waiting for them might take some time. 

Done. If you could test if the issue is fixed, I'd appreciate that.

pedrohgmacedo:
I've been using the patched version and it no longer has the regression. I could not find any other problem yet, though I will keep using it and if I find anything I'll comment here.

fredizzimo:
I'm not sure about using wslpath2 for this. The implementation is trivial https://github.com/michidk/wslpath2/blob/develop/src/lib.rs, and even more when we can hardcode the arguments. Furthermore, it does not look like anyone will look at the PR for removing the console. So, I think it's better to get rid of the implementation and call the `wslpath` executable directly.

But even then, I'm not sure it's a good idea. On my system calling `wsl wslpath 'c:\\users'` takes 3 to 4 seconds and we don't want to make the slow startup of Neovide with WSL even slower. So, we have two options:
1. Instead of doing multiple WSL calls, we could use a shell script or similar, which finds and launches Neovim. I think this is something we want to do anyway in order to speed up the launch.  But it's out of scope for this PR, and I think quite hard to take everything we support into account and work exactly like before.
2. The path could be converted by code, especially the Windows to Linux direction, should be quite straightforward, just turn the dashes around, change the drive to `/mnt`, and maybe some escaping.

pedrohgmacedo:
> I'm not sure about using wslpath2 for this. The implementation is trivial https://github.com/michidk/wslpath2/blob/develop/src/lib.rs, and even more when we can hardcode the arguments. Furthermore, it does not look like anyone will look at the PR for removing the console. So, I think it's better to get rid of the implementation and call the `wslpath` executable directly.
> 
> But even then, I'm not sure it's a good idea. On my system calling `wsl wslpath 'c:\\users'` takes 3 to 4 seconds and we don't want to make the slow startup of Neovide with WSL even slower. So, we have two options:
> 
>     1. Instead of doing multiple WSL calls, we could use a shell script or similar, which finds and launches Neovim. I think this is something we want to do anyway in order to speed up the launch.  But it's out of scope for this PR, and I think quite hard to take everything we support into account and work exactly like before.
> 
>     2. The path could be converted by code, especially the Windows to Linux direction, should be quite straightforward, just turn the dashes around, change the drive to `/mnt`, and maybe some escaping.


The /mnt/c mount point can be customized in wsl. I don't know if people actually do that, but it is possible. 
You can do it through [wsl.config](https://learn.microsoft.com/en-us/windows/wsl/wsl-config#:~:text=%23%20Sets%20the%20directory%20where%20fixed%20drives%20will%20be%20automatically%20mounted.%20This%20example%20changes%20the%20mount%20location%2C%20so%20your%20C%2Ddrive%20would%20be%20/c%2C%20rather%20than%20the%20default%20/mnt/c.%20%0Aroot%20%3D%20/), and even through fstab.

I wrote a lua script to convert the path at #2194, but the problem is that the filename is passed unescaped to the neovide process, so all the slashes are gone. 
But at the neovide level, it should be a simpler problem.

fredizzimo:
> The problem is not really finding neovim, that already works, the problem to me was associating a file with neovide, and then starting it from wsl, the file cannot be opened.

My comment was referring to doing everything we do in https://github.com/neovide/neovide/blob/main/src/bridge/command.rs, in a single shell script instead, at least for WSL. Then we only need to launch WSL once, which saves a lot of times. And for the WSL path conversion, we could escape the windows path on the Neovide side, send that as an argument to the shell script, which would then be passed on to wslpath before given as an argument to Neovim.

But, since that solution is quite complex, I think we could perhaps assume the `/mnt` mount point for now.

Holzhaus:
Alright, looking into a pure-rust implementation of wslpath so that we don't need to call executables.

Holzhaus:
I wrote a little crate named [`wslpath-rs`](https://crates.io/crates/wslpath-rs) because I figured that kind of functionality might be desired by other projects as well.

I updated and rebased this PR accordingly. Please have a look.

pedrohgmacedo:
> I wrote a little crate named [`wslpath-rs`](https://crates.io/crates/wslpath-rs) because I figured that kind of functionality might be desired by other projects as well.
> 
> I updated and rebased this PR accordingly. Please have a look.

I have been using it, and found no problems so far. Haven't noticed any particular improvements on startup time. 

> > The problem is not really finding neovim, that already works, the problem to me was associating a file with neovide, and then starting it from wsl, the file cannot be opened.
> 
> My comment was referring to doing everything we do in https://github.com/neovide/neovide/blob/main/src/bridge/command.rs, in a single shell script instead, at least for WSL. Then we only need to launch WSL once, which saves a lot of times. And for the WSL path conversion, we could escape the windows path on the Neovide side, send that as an argument to the shell script, which would then be passed on to wslpath before given as an argument to Neovim.
> 
> But, since that solution is quite complex, I think we could perhaps assume the `/mnt` mount point for now.

I wonder if startup time could be improved by using a non login shell on wsl? Or maybe even allowing the customization of that, say, by reusing a the sshd daemon, and so on...

fredizzimo:
@Holzhaus, I will check this and hopefully merge during the weekend.

@pedrohgmacedo
> I wonder if startup time could be improved by using a non login shell on wsl? Or maybe even allowing the customization of that, say, by reusing a the sshd daemon (it starts up faster than the raw wsl.exe), and so on...

I think we need a login shell, since the shell will be the entrypoint to WSL. But we can make sure that we only start it once, by moving all the logic of the `bridge/command.rs` to the shell script. That is still quite slow, especially the first time, if systemd is enabled. But that can be improved by making it start with Windows.

And to further improve it we could use ssh as you said, or perhaps even a custom daemon that starts with systemd.

pedrohgmacedo:
> @Holzhaus, I will check this and hopefully merge during the weekend.
> 
> @pedrohgmacedo
> 
> > I wonder if startup time could be improved by using a non login shell on wsl? Or maybe even allowing the customization of that, say, by reusing a the sshd daemon (it starts up faster than the raw wsl.exe), and so on...
> 
> I think we need a login shell, since the shell will be the entrypoint to WSL. But we can make sure that we only start it once, by moving all the logic of the `bridge/command.rs` to the shell script. That is still quite slow, especially the first time, if systemd is enabled. But that can be improved by making it start with Windows.
> 
> And to further improve it we could use ssh as you said, or perhaps even a custom daemon that starts with systemd.

On my machine, I'm using wsl1, which does not even have systemd. It is still noticeable.

Holzhaus:
> > I wrote a little crate named [`wslpath-rs`](https://crates.io/crates/wslpath-rs) because I figured that kind of functionality might be desired by other projects as well.
> > 
> > I updated and rebased this PR accordingly. Please have a look.
> 
> I have been using it, and found no problems so far. Haven't noticed any particular improvements on startup time. 

I found a problem but I don't think it's caused by this PR: when opening a file with spaces in the path, multiple partial paths are passed to neovim instead. Might be an issue with how windows handles command line arguments or something. I associated some file types with neovide and then just double-clicked a file in Windows Explorer.

pedrohgmacedo:
> I found a problem but I don't think it's caused by this PR: when opening a file with spaces in the path, multiple partial paths are passed to neovim instead. Might be an issue with how windows handles command line arguments or something. I associated some file types with neovide and then just double-clicked a file in Windows Explorer. I think this shouldn't block this PR though.

You need to always add \" \" to the file path you are passing to neovim in wsl.

fredizzimo:
@Holzhaus

I totally forgot about this. But looking at it again, I think everything is good for merging, but I will do some more checks after the conflicts are resolved. Would you mind doing that?

Holzhaus:
> @Holzhaus
> 
> I totally forgot about this. But looking at it again, I think everything is good for merging, but I will do some more checks after the conflicts are resolved. Would you mind doing that?

Sure, I'll have a look soon.

As explained above, even with this PR merged it will still not work to double click a file in Windows Explorer if the file path contains spaces due to quoting issues (neovide will open multiple tabs split by space instead). 

I think this issue is not caused by my code, as the tests work fine: https://github.com/Holzhaus/wslpath-rs/blob/efcd6e37a2873da36d40a06915ca653c147115ea/src/lib.rs#L66 and can be fixed separately.

Holzhaus:
@fredizzimo Done. Please have a look.

Issue #2175, opened 2023-12-12
If dragging, there should be a window details recorded


9mm:
**Describe the bug**

I'm not quite sure what the bug is, as I got this crash report while opening a large file (may have hit another key on accident). I have no idea what I did to cause it.

Another comment had some ideas on how to reproduce but I don't know any windows that involve mouse interactions, or even "hidden windows", but I'm open to try and reproduce with some actual commands, if someone wants to put exact steps

https://github.com/neovide/neovide/pull/2167#issuecomment-1851578744

**Desktop (please complete the following information):**
 - OS: 14.1.2 (23B92)
 - Neovide Version NVIM v0.9.4
 - Neovim Version Main
 - 
[neovide_backtraces.log](https://github.com/neovide/neovide/files/13648723/neovide_backtraces.log)

AThePeanut4:
I've just run into the same crash, and I've (eventually) managed to figure out consistent reproduction steps (on Linux). This might not be exactly be what happened to cause your crash, but it's giving me the same "If dragging, there should be a window details recorded" panic message.

1. Most importantly, do not move the mouse _at all_ during the entire procedure. It's probably easiest to open Neovide using a keyboard shortcut directly or through something like `drun`, or from a terminal. The cursor does need to be positioned where the Neovide window will actually appear though.
2. Open Neovide.
3. Click and hold the right mouse button, and then move the mouse.
4. Neovide should then crash immediately, and a "If dragging, there should be a window details recorded" panic message should be added to `neovide_backtraces.log`.

I should also note that after this crash, the `neovide --no-fork` process and its child `nvim` process remain running, even though the window disappears. If I run it from a terminal with `neovide --no-fork` and cause the panic, I get the panic message printed, but the process doesn't actually exit.

I've also come across a different, but likely related, crash. This one doesn't add a backtrace at all, but if you run it through a terminal with `--no-fork` then it prints a stack overflow error message before `SIGABRT`ing.
1. Run Neovide from a terminal using `neovide --no-fork`.
2. Move the mouse (otherwise the next step won't show a context menu at all, presumably related to the above issue).
3. Right click somewhere (not hold) to show the Paste/Select All/etc context menu.
4. Move the mouse to inside the context menu, then click and hold the right mouse button.
6. At this point, the context menu should "teleport" to the top left corner of the screen.
7. While still holding the right mouse button, move the mouse.
[...]

PR #2257, opened 2024-01-10
Fix issue with recursive floating window anchors

Kethku:
Neovim sometimes requests that a grid be anchored to itself when using the context menu. This PR addresses this by catching the recursive case in the editor and logging a warning.
Fixes https://github.com/neovide/neovide/issues/2175

- Fix

- No


Kethku:
Ah makes sense. I will update this PR to fix that root issue

fredizzimo:
I'm going to merge this fix. And I'm preparing another PR for the other issue in #2175.

PR #2259, opened 2024-01-10
support some font weights by ':set guifont=...'

Passw:
https://rust-skia.github.io/doc/skia_safe/type.FontMgr.html#method.family_names

FontMgr::family_names() get font names between Windows and Ubuntu differently

For Cascadia Code fonts,
on Windows will get 'Cascadia Code PL' ..., which does not contain font weight 
on Ubuntu   will get 'Cascadia Code PL Light' ..., which contains font weight 'Light'

This commit is a workaround to use `:set guifont=Cascadia_Code_PL_Light:h10`

- Fix

- No

### DIFF
--- a/src/renderer/fonts/font_options.rs
+++ b/src/renderer/fonts/font_options.rs
@@ -187,6 +187,53 @@ impl FontOptions {
         for font in font_options.normal.iter_mut() {
             font.style = style.clone();
         }
+        for font in font_options.normal.iter_mut() {
+            match font.family.split_whitespace().last() {
+                Some("Thin") => {
+                    font.family = font.family[..font.family.len() - 5].to_string();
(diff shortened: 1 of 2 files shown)


fredizzimo:
Thanks, but I don't think we want to merge this. The reason we moved the extended font configuration to the config file was to be able to support more options than what's posible with guifont. We only keep guifont for backward compatibility.

So
```toml
[font]
normal = [{family ="Cascadia Code PL", style="Light"]
```

Should work on both Windows and Linux, and load `Cascadia Code PL Light` if it exists, otherwise the closest version. Additionally if it's a variable weight font it will use the correct configuration. This is consistent with how you select a font in a word processor for example, you don't select the light version manually there either.

Issue #2003, opened 2023-08-30
Neovide doesn't close until after clipboard has been changed externally


isovector:
**Describe the bug**

After yanking to the unnamedplus clipboard, neovide hangs on exit until some other program changes the clipboard.

**To Reproduce**

Given a minimal config, `~/.config/nvim/init.vim`:

```viml
autocmd!
set clipboard+=unnamedplus
```

then start neovide. Type:

```
yy:q!<CR>
```

Neovide will now hang, refusing to exit. It will immediately quit after copying to the clipboard from some other application.

**Expected behavior**

Neovide should not hang until the clipboard changes.

**Desktop (please complete the following information):**
 - OS: Linux 6.1.38-1-MANJARO #1 SMP PREEMPT_DYNAMIC Wed Jul  5 23:49:30 UTC 2023 x86_64 GNU/Linux
 - Neovide Version 0.11.1
 - Neovim Version v0.9.1

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/12477151/neovide_rCURRENT.log)

**Additional context**

Both lines of the config are necessary.

fredizzimo:
Can you test this with terminal mode Neovim? Neovide itself should use the default Neovim clipboard integration unless you use wsl or a tcp connection to Neovim.

More specifically it should use `wl-paste` on Wayland, which can hang in certain situations, https://github.com/neovim/neovim/issues/24470.

fredizzimo:
You could try to disable Neovide`s clipboard, although it should already be disabled.

vim: `let g:neovide_no_custom_clipboard = v:true`
lua: `vim.g.neovide_no_custom_clipboard = true`

Are you on Wayland or X11, and which desktop environment / window manager? We need to know which clipboard provider is used https://github.com/neovim/neovim/blob/b2c3e9c72ef12a5bc276dec78cdc8dfbd2f5a24b/runtime/autoload/provider/clipboard.vim#L69

And if that is running, blocking the exit of the process. Can you check that with `ps all`?

isovector:
Interestingly, `xclip` isn't running on a fresh open of neovide, but does start immediately when I yank. Furthermore, it's in that same `Ss` status as soon as I yank anything (before I go to quit neovide.)

Running `nvim` reports the same clipboard, and also starts the same `xclip` process, but doesn't wait for `xclip` to close before exiting.

isovector:
[...]

PR #2265, opened 2024-01-12
fix: fix hang on neovim exit

sid-6581:
- Fix

- No

This is basically #2130 again. I wanted a cleaner fix, but a lot of the async and threading stuff will need to be reworked. In the mean time there have been enough reports of hangs to justify a quick fix so it's usable.

I tried a couple of things that worked for me before, but an added complication now is the task spawning in `start_ui_command_handler`. If those are stuck in a `recv()`, they just won't exit.

I'm thinking instead of `RUNNING_TRACKER` we need to use cancellation tokens or something tokio can select on so we can break out of futures that will never return. Checking `RUNNING_TRACKER.is_running()` does nothing when stuck in an await, and we shouldn't rely on channels dropping at the end of a scope to break out of these loops, because that drop can possibly not happen if stuck in another await. That is exactly what's happening here.

After working on this for a bit I also realized that no timeout is really needed, because tokio doesn't kill the child process when the runtime shuts down. Not unless we call `kill_on_drop()`, which we don't. This change just lets neovide exit instead of waiting forever in stuck awaits.

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -12,10 +12,7 @@ use log::{error, info};
 use nvim_rs::{error::CallError, Neovim, UiAttachOptions, Value};
 use rmpv::Utf8String;
 use std::{io::Error, ops::Add, sync::Arc};
-use tokio::{
-    runtime::{Builder, Runtime},
-    task::JoinHandle,
-};
+use tokio::runtime::{Builder, Runtime};
 use winit::event_loop::EventLoopProxy;
 
 use crate::{
@@ -34,14 +31,8 @@ pub use ui_commands::{send_ui, start_ui_command_handler, ParallelCommand, Serial
 const INTRO_MESSAGE_LUA: &str = include_str!("../../lua/intro.lua");
 const NEOVIM_REQUIRED_VERSION: &str = "0.9.2";
 
-enum RuntimeState {
-    Idle,
-    Attached(JoinHandle<()>),
-}
-
 pub struct NeovimRuntime {
-    runtime: Runtime,
-    state: RuntimeState,
+    runtime: Option<Runtime>,
 }
 
 fn neovim_instance() -> Result<NeovimInstance> {
@@ -158,8 +149,7 @@ impl NeovimRuntime {
         let runtime = Builder::new_multi_thread().enable_all().build()?;
 
         Ok(Self {
-            runtime,
-            state: RuntimeState::Idle,
+            runtime: Some(runtime),
         })
     }
 
@@ -168,20 +158,16 @@ impl NeovimRuntime {
         event_loop_proxy: EventLoopProxy<UserEvent>,
         grid_size: Option<Dimensions>,
     ) -> Result<()> {
-        assert!(matches!(self.state, RuntimeState::Idle));
         let handler = start_editor(event_loop_proxy);
-        let session = self.runtime.block_on(launch(handler, grid_size))?;
-        self.state = RuntimeState::Attached(self.runtime.spawn(run(session)));
+        let runtime = self.runtime.as_ref().unwrap();
+        let session = runtime.block_on(launch(handler, grid_size))?;
+        runtime.spawn(run(session));
         Ok(())
     }
 }
 
 impl Drop for NeovimRuntime {
     fn drop(&mut self) {
-        if let RuntimeState::Attached(join_handle) =
-            std::mem::replace(&mut self.state, RuntimeState::Idle)
-        {
-            let _ = self.runtime.block_on(join_handle);
-        }
+        self.runtime.take().unwrap().shutdown_background();
     }
 }


fredizzimo:
I think, we can merge this if someone does a release before I'm back to my computer in one week.

But I think we need another solution, so that we can react to neovim crashing unexpectedly, which I think could cause similar hangs. This probably needs a patch to nvim-rs, to detect that the process exits and close the pipes, which will cause the reads to abort with zero bytes read, and exit.

I'm not sure I will be able to repeat it, but it seems like hangs might happen when there are partial messages in flight when neovim exits.

fredizzimo (approved):

sid-6581:
> I think, we can merge this if someone does a release before I'm back to my computer in one week.
> 
> But I think we need another solution, so that we can react to neovim crashing unexpectedly, which I think could cause similar hangs. This probably needs a patch to nvim-rs, to detect that the process exits and close the pipes, which will cause the reads to abort with zero bytes read, and exit.
> 
> I'm not sure I will be able to repeat it, but it seems like hangs might happen when there are partial messages in flight when neovim exits.

Totally agreed, I'm working on a proper fix right now. I've reworked the `RUNNING_TRACKER` to use cancellation tokens, and I'm instrumenting `nvim-rs` to see if anything can be done inside it or if it's a tokio issue. I'm still not sure where the issue ultimately lies.

fredizzimo:
If this function is hit, the. we should at least be able to detect that the process quits and close the pipes inside nvim-rs

https://github.com/tokio-rs/tokio/blob/84c5674c601dfc36ab417ff0ec01763c2dd30a5c/tokio/src/process/windows.rs#L166-L169

fredizzimo:
This is probably very much related https://stackoverflow.com/questions/10841840/win32-readfile-from-pipe-block-even-after-child-terminated

fredizzimo:
I was in a bit rush, and just found that link, so I only had time to post it. But it's a documented fact that read will never return zero, if there's a duplicated handle somewhere

https://learn.microsoft.com/en-gb/windows/win32/ipc/pipe-handle-inheritance?redirectedfrom=MSDN

So, we should not need to detect that the process exits and close, just ensure that we, or nvim-rs never duplicates the handles, without closing.

I don't think any cancellation token system is needed

sid-6581:
From what I've been able to tell from tracing through code, I don't see any handles being duplicated. Also, this is more of a race condition that happens occasionally, I would imagine if it was the above issue it would consistently deadlock. I'll check again, though.

I also believe that I did check the Child `JoinHandle` for process exit the last time I was working on this, and I don't believe that future was completed even when the process was gone, but I may be misremembering that and will need to double check.

As for the cancellation token, it (or a broadcast channel) is very much needed to shut things down cleanly. Like I mentioned earlier, checking a bool in an async loop only works when conditions are perfect. When conditions aren't perfect, you get deadlocks. To cleanly cancel a task you need a proper async primitive so the async runtime can break out of an await. Otherwise your code checking the bool will never even get run, because the task is just parked waiting to get woken up.

I've implemented it and cleaned things up a bit, and it works great. It now shuts down cleanly and doesn't hang, and the runtime can exit normally because the tasks are completed. It's also a lot more explicit and easier to understand. I'll make a PR for that later, but it will be a bigger one. There's still more cleanup and testing. In the current release I've also experienced a lot of hangs while using neovide, not just when shutting down, so I'm trying to debug those as well while I'm at it.

Here are Tokio's recommendations for a graceful shutdown, which is what I'm doing now:

https://tokio.rs/tokio/topics/shutdown

fredizzimo:
I don't see what imperfect conditions there could be without bugs, we either receive a message or the channel is closed.

The same for the internal read and writes of the pipes, they either succeed or return 0 and fail. There are no windows bugs that explains the hangs, otherwise basically everything will break, applications have been written with blocking io for a very long time.

I also don't like workarounds for something which can't be explained, the only reason I accepted this current workaround is that the bug seems serious enough for a quick fix. We will have to prove that there is an unfixable bug in either tokio, nvim-rs or the operation system first.

fredizzimo:
Note that I can accept that we need cancellation tokens to exit with imperfect network conditions, where the connection remains open, but has a very high latency or is very slow. Or when neovim gets stuck in an infinite loop or something and does not respond.

But neither of those are the case here.

fredizzimo:
I have quite hard time debugging this on my ipad, and without ever reproducing the issue itself. But by reading the code, it appears like the handle duplication happens here 
https://github.com/tokio-rs/tokio/blob/84c5674c601dfc36ab417ff0ec01763c2dd30a5c/tokio/src/process/windows.rs#L253

sid-6581:
I really don't like quick workarounds either, which is why I'm working on a proper solution. I just think the proper solution is to anticipate the unanticipated and code defensively. The way I see it, there are multiple reports of hangs in various circumstances, issues that I personally run into many, many times every day as I try to get work done. Issues I don't have in any other neovim UI. So yes, applications have been written with blocking IO for a long time, but these things are hard to get right, and certain patterns are considered best practices for a reason.

In the case of cancellation tokens, that's a very idiomatic way to ensure a clean shutdown in multiple languages. It also doesn't cost anything, performance wise, and it makes the code more clear, not less. So it's really an overall win. Is it needed under perfect conditions? Not really, but the world is isn't perfect, as evidenced by the bug reports. My whole career has been spent writing very heavily multithreaded/async servers that need to run 24/7 for months or years. I've learned these lessons the hard way, and I'm spending my free time trying to fix what are real issues reported by multiple people and experienced by myself as well. I think neovide is great, and I'm just trying to help improve it and make it the best it can be. 

Your work made me come back to neovide after not being able to use it for a long time, and I really appreciate all your efforts. I think neovide has more potential than any other neovim UI, which is why I want to contribute. (OK, also because I love rust.)

falcucci (approved):

falcucci (approved):

fredizzimo:
I think I will merge this later today, since we don't have any other workaround.

But it worries me that this leaks Neovim processes, which could cause issues with Neovim swap files for example. See https://github.com/neovide/neovide/issues/2003#issuecomment-1897932362. But I was not able to reproduce that, for me it only leaks xclip or xsel, which is not quite as bad.

sid-6581:
> I think I will merge this later today, since we don't have any other workaround.
> 
> But it worries me that this leaks Neovim processes, which could cause issues with Neovim swap files for example. See [#2003 (comment)](https://github.com/neovide/neovide/issues/2003#issuecomment-1897932362). But I was not able to reproduce that, for me it only leaks xclip or xsel, which is not quite as bad.

In reality, the only time the runtime will be dropped is when `RUNNING_TRACKER` is already signaling a clean shutdown or when the pipe is broken. At the point where the runtime shutdown happens, there's no more communicating with neovim in any way. Also, dropping the runtime doesn't forcibly kill the neovim process, so it's not like we're interrupting any neovim shutdown processes in any way.

As for leaking neovim processes, if neovim doesn't shut down cleanly at the point we're dropping the runtime, I'd consider that a bug in neovim and not something we can do anything about. We could forcibly kill the process, but that's not something we want to do.

fredizzimo:
Ok, merging this, but we do have to find the root cause. 

Well, I think I have found it, but because I can't reproduce I have no way to verify. The rust standard library uses WaitForSingleObject wrong, it's not valid to wait for file handles like they do.

See my investigation here
* https://github.com/neovide/neovide/issues/2299#issuecomment-1907411309

PR #2272, opened 2024-01-14
fix: shadow artifacts on macOS

falcucci:
- Feature

- No

The shadow artifacts are being visible when we unfocus/focus the app on transparent mode.

macOS only.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -45,7 +45,6 @@ use std::time::SystemTime;
 use time::macros::format_description;
 use time::OffsetDateTime;
 use winit::event_loop::EventLoopProxy;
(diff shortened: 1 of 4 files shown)


fredizzimo (approved):
This looks good if it works and it's enough to only invalidate the shadows when the window is created or the shadows are changed. But I thought we had to do it every time we render something.

falcucci:
in the future we can find a workaround to have the shadow still in transparent mode, but for now it seems to be a winit bugfix or even the AppKit itself.

Fortunately we don't need to invalidate the shadows every rendered time.

Kethku:
tyvm

PR #2280, opened 2024-01-17
fix: font debug is incorrect because self.info() is not refreshed

borisfaure:
- Fix

- No

Very small fix while I investigate a bit more #2214

### DIFF
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -171,6 +171,7 @@ impl CachingShaper {
             self.fudge_factor = font_width.round() / font_width;
             debug!("Fudge factor: {:.2}", self.fudge_factor);
             font_size = self.current_size();
+            self.font_info = None;
             debug!("Fudged font size: {:.2}px", font_size);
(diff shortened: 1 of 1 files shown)


src/renderer/fonts/caching_shaper.rs
```diff
@@ -171,6 +171,7 @@ impl CachingShaper {
             self.fudge_factor = font_width.round() / font_width;
             debug!("Fudge factor: {:.2}", self.fudge_factor);
             font_size = self.current_size();
+            self.font_info = None;
```
fredizzimo, line 174: I think this is still buggy, since `info()` is called below below in the debug statement, before the new font loader is created, `self.info` will therefore be cached with the old size again.

If that causes other problems though, then it's possible that the problem is something else than I inititially thought in https://github.com/neovide/neovide/issues/2214

fredizzimo (changes requested):
I think we should close this PR, and properly fix everything in the original one https://github.com/neovide/neovide/pull/2218. It's not worth merging something until it works completely.

PR #2298, opened 2024-01-22
fix: When counting leading spaces for space skip optimization count codepoints

ZyX-II:
- Fix

- No (I assume change in 5be2acec746309c0b6be4cd61312b9503c39baa5 does not count as breaking)

Result from running `cargo +stable-x86_64-pc-windows-msvc run -- --grid '1x25' --log -- -u NONE -i NONE -N --cmd 'set fillchars=eob:\  cmdheight=0 nolist noshowcmd laststatus=0' --cmd "call setline(1, map(range(0x09, 0x0D) + [0x20, 0x85, 0xA0, 0x1680] + range(0x2000, 0x200A) + [0x2028, 0x2029, 0x202F, 0x205F, 0x3000], 'printf(''%s%04x'', nr2char(v:val), v:val)'))" --cmd 'autocmd FocusGained * redraw! | sleep 1'` with this branch:

image

Result from running the same command, but with `cargo +stable-x86_64-pc-windows-msvc run --` replaced with `neovide` to get neovide 0.12.1 from scoop:

image

### DIFF
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -386,7 +386,7 @@ impl CachingShaper {
 
         let mut resulting_blobs = Vec::new();
 
-        trace!("Shaping text: {}", text);
+        trace!("Shaping text: {:?}", text);
 
         for (cluster_group, font_pair) in self.build_clusters(&text, style) {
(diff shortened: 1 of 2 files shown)


fredizzimo:
Thank you!

PR #2311, opened 2024-01-26
fix: Determine the window under the mouse on the fly

fredizzimo:
Previously it was only checked when the mouse moved, but that does not work when new windows are created without moving the mouse.

So, this is fixed by re-checking he window when clicking or scrolling.

* fix https://github.com/neovide/neovide/issues/2175

- No

### DIFF
--- a/src/window/mouse_manager.rs
+++ b/src/window/mouse_manager.rs
@@ -43,6 +43,11 @@ fn mouse_button_to_button_text(mouse_button: &MouseButton) -> Option<String> {
     }
 }
 
+struct DragDetails {
+    draw_details: WindowDrawDetails,
+    button: String,
+}
+
 #[derive(Debug)]
 struct TouchTrace {
     start_time: Instant,
@@ -52,19 +57,16 @@ struct TouchTrace {
 }
 
 pub struct MouseManager {
-    dragging: Option<String>,
+    drag_details: Option<DragDetails>,
     drag_position: GridPos<i32>,
 
     has_moved: bool,
-    position: GridPos<i32>,
-    relative_position: GridPos<i32>,
+    position: PixelPos<f32>,
 
     scroll_position: GridPos<f32>,
 
     // the tuple allows to keep track of different fingers per device
     touch_position: HashMap<(DeviceId, u64), TouchTrace>,
-
-    window_details_under_mouse: Option<WindowDrawDetails>,
 
     mouse_hidden: bool,
     pub enabled: bool,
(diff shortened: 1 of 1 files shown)


src/window/mouse_manager.rs
```diff
@@ -60,6 +60,11 @@ fn mouse_button_to_button_text(mouse_button: &MouseButton) -> Option<String> {
     }
 }
 
+struct DragDetails {
+    draw_details: WindowDrawDetails,
+    button: String,
```
Kethku, line 65: Given that this can only be a static set of options, maybe this could be an enum?

src/window/mouse_manager.rs
```diff
@@ -69,40 +74,76 @@ struct TouchTrace {
 }
 
 pub struct MouseManager {
-    dragging: Option<String>,
+    drag_details: Option<DragDetails>,
```
Kethku, line 77: Renaming to drag_details obscures the difference between this field and the drag_position field. So coming to this code as somebody new, its not as immediately obvious why drag_details might have a value or not.

Maybe we could call it current_drag_details or active_drag_details?

src/window/mouse_manager.rs
```diff
@@ -69,40 +74,76 @@ struct TouchTrace {
...
+            .filter(|details| {
+                position.x >= details.region.left
+                    && position.x < details.region.right
+                    && position.y >= details.region.top
+                    && position.y < details.region.bottom
+            })
+            .last()
+    }
+
+    fn get_relative_position(
```
Kethku, line 126: Minor thing, but would it make sense to put this as a function on the renderer which takes a position and returns a relative position?

Take it or leave it

src/window/mouse_manager.rs
```diff
@@ -232,26 +234,40 @@ impl MouseManager {
...
+        x: f32,
+        y: f32,
+        keyboard_manager: &KeyboardManager,
+        renderer: &Renderer,
+    ) {
         if !self.enabled {
             return;
         }
 
+        let draw_details = self.get_window_details_under_mouse(renderer);
```
Kethku, line 268: draw details seems like it could have a more descriptive name. Maybe relevant_window_details? Or just window_details?

Kethku (approved):
Mostly minor suggestions, so feel free to ignore. LGTM

falcucci (approved):

src/window/mouse_manager.rs
fredizzimo: I think it's slightly too specialized, since it takes a global position in pixels, then clamps the coordinates to an active Neovim window, and returns the relative grid coordinates. So, I left it as it is.

src/window/mouse_manager.rs
fredizzimo: I changed it to `window_details`

src/window/mouse_manager.rs
fredizzimo: I changed it to the winit `MouseButton` enum, at the cost of an additional string conversion each time the drag message is sent.

src/window/mouse_manager.rs
fredizzimo: I kept `drag_details` , but renamed the other variables to `grid_position` and `window_position`, which better reflects what they mean.

Issue #2300, opened 2024-01-23
Neovide cannot work well with hover.nvim


nullptr-yxw:
[hover.nvim](https://github.com/lewis6991/hover.nvim) is a plugin to display lsp and other sources, just like vim.lsp.buf.hover. But it can support mouse hover as well. However, the mouse hover cannot work well with neovide.

Behaviour of it in windows terminal (can work as except):

https://github.com/neovide/neovide/assets/96286337/be9c5715-0c7f-4572-9c32-2b90796ab557

Behaviour of it in neovide (have no response to the mouse):

https://github.com/neovide/neovide/assets/96286337/e2df7a01-82db-449e-8f2b-e9a9f925384f

However, if I use the plugin via keyboard keymap, it can work well, so the plugin has been loaded and setup successfully:

https://github.com/neovide/neovide/assets/96286337/b4276584-dbcd-403a-ac99-3d0afbc0e001

I have a look at the source code of `hover.nvim`, the code of keymap setting is [Setup and Installation](https://github.com/lewis6991/hover.nvim/blob/a7b7aba3ebf92a85e4cb6ebf1c4ee67862c5980a/README.md#setup-and-installation), and the source code of mouse hover is [require('hover').hover_mouse](https://github.com/lewis6991/hover.nvim/blob/a7b7aba3ebf92a85e4cb6ebf1c4ee67862c5980a/lua/hover/actions.lua#L311-L331)
```lua
-- copied from lewis6991/hover.nvim
local timer --- @type uv.uv_timer_t
function M.hover_mouse()
  timer = timer or assert(vim.uv.new_timer())

  local buf = api.nvim_get_current_buf()

  M.close(buf)

  local config = get_config()

  timer:start(config.mouse_delay, 0, vim.schedule_wrap(function()
    local pos = vim.fn.getmousepos()
    M.hover {
      providers = config.mouse_providers,
      relative = 'mouse',
      pos = { pos.line, pos.column }
    }
  end))
end
```
I have taken a test, and find that both `<MouseMove>`, `vim.fn.getmousepos` and `vim.uv.timer` can work well with neovide, so I have no idea about why the bug appears.

**Desktop (please complete the following information):**
 - OS: win11
 - Neovide Version: `0.12.1`(the latest release)
 - Neovim Version: `NVIM v0.10.0-dev-2164+gdc466f9a6`

fredizzimo:
We have a couple of other issues that are almost certainly caused by the same underlying bug.
[...]

Issue #2222, opened 2023-12-29
Mouse position functions don't work in latest release


PyDev19:
**Describe the bug**
I recently downloaded Neovide 0.12.0 on Windows 11 through the MSI installer. I am using AstroNvim, and they have a clickable fold column. The fold column utilizes the `vim.fn.screenstring function`, which retrieves the string under the mouse. I tested this function while moving the mouse around the screen, and it consistently returned blank strings instead of the character under the cursor. This issue does not exist in Neovide 0.11.2 on Windows 11. I don't believe it's a problem with Neovim, as I tested it in the terminal and it worked correctly. Additionally, I tested this on Windows 10 to confirm that it's not an issue exclusive to Windows 11.

**To Reproduce**
1. Download Neovide 0.12.0 on Windows 11 or 10
2. Use `vim.fn.screenstring(row, col)` with the mouse position as argument
3. Check to see if the output in blank or if it returns the character under the mouse

**Expected behavior**
It should return the character under the mouse when using the `screenstring` function

**Desktop (please complete the following information):**
 - OS: Windows 11 & 10
 - Neovide Version 0.12.0
 - Neovim Version 0.6.4

Here is the neovide log while I was clicking the foldcolumns:
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/13795608/neovide_rCURRENT.log)

**Additional context**
I didn't test any other mouse position functions but I do think that some updates changed the way mouse position functions are handled which causes some errors and returns nothing when called on.

fredizzimo:
I’m not sure why it works in he terminal. We disabled sending of the mouse movement events when it’s disabled (https://neovim.io/doc/user/options.html#'mousemoveevent' ) here https://github.com/neovide/neovide/pull/2111

So enabling that should fix the issue. But I’m not sure what the correct behaviour should be, so if someoene can find a definite specification, it would be good. Otherwise, we need to clarify it with the neovim team. Note, we do send the cursor position with clicks.

fredizzimo:
Note that if we always send the mouse position, we will cause a regression for this bug https://github.com/neovide/neovide/issues/1838. So perhaps, additionally a Neovim fix is needed to get the same behaviour as the terminal mode.

tan-wei:
I can confirm that only with neovide (0.12.1) I will meet this issue.

fredizzimo:
This is very likely the same issue as
* #2300 

[...]

PR #2315, opened 2024-01-27
Send options on startup as well as on changed except for columns and rows

Kethku:
Updates option reporting to send the initial values of options as well as when they are changed.

This manually excludes lines and columns as they are set dynamically on startup and our logic requires that we set them ourselves before receiving any changes.

fixes https://github.com/neovide/neovide/issues/2222
fixes https://github.com/neovide/neovide/issues/2300

- Fix

- No

### DIFF
--- a/lua/init.lua
+++ b/lua/init.lua
@@ -78,8 +78,19 @@ for _,global_variable_setting in ipairs(args.global_variable_settings) do
 end
 
 for _,option_setting in ipairs(args.option_settings) do
+    -- Ignore initial values of lines and columns because they are set by neovim directly.
+    -- See https://github.com/neovide/neovide/issues/2300
+    if option_setting ~= "lines" and option_setting ~= "columns" then
+        vim.api.nvim_create_autocmd({ "VimEnter" }, {
+            pattern = option_setting,
+            once = true,
+            nested = true,
+            callback = function()
+                rpcnotify("option_changed", option_setting, vim.o[option_setting])
+            end
+        })
+    end
     vim.api.nvim_create_autocmd({ "OptionSet" }, {
-        pattern = option_setting,
         once = false,
         nested = true,
         callback = function()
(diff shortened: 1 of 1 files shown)


lua/init.lua
```diff
@@ -78,8 +78,19 @@ for _,global_variable_setting in ipairs(args.global_variable_settings) do
...
+            pattern = option_setting,
+            once = true,
+            nested = true,
+            callback = function()
+                rpcnotify("option_changed", option_setting, vim.o[option_setting])
+            end
+        })
+    end
     vim.api.nvim_create_autocmd({ "OptionSet" }, {
-        pattern = option_setting,
```
fredizzimo, line 82: Doesn't the removal of this cause it to be run for all option changes now.? And since it's done in a loop many times per option.

lua/init.lua
```diff
@@ -78,8 +78,19 @@ for _,global_variable_setting in ipairs(args.global_variable_settings) do
 end
 
 for _,option_setting in ipairs(args.option_settings) do
+    -- Ignore initial values of lines and columns because they are set by neovim directly.
+    -- See https://github.com/neovide/neovide/issues/2300
+    if option_setting ~= "lines" and option_setting ~= "columns" then
+        vim.api.nvim_create_autocmd({ "VimEnter" }, {
```
fredizzimo, line 84: Perhaps it would make sense to move `VimEnter` outside the loop, And have an internal loop for notifying all settings instead?

lua/init.lua
Kethku: Mmmm yes. Obvious good point.

lua/init.lua
Kethku: Yup will do that

fredizzimo (approved):

fredizzimo:
Thank you! I tested this and it seems to work fine.

PR #2316, opened 2024-01-27
feat: Add colored output to `--help`

last-partizan:
- Feature

- No

----------

Today i was building neovide, and instead of `cargo r -- --help`, i typed `cargo r --help`. And cargo help is colored!

I immediately wanted to have the same colored help in neovide. So, i added it!

They're using `cyan` for command names, and i've used defaults from clap docs, because in our case cyan looks not so fitting.

Take a look!

image

-------

One thing to be aware of - i expected `Cargo.lock` to be updated, after changing `Cargo.toml`, but it was not updated. Is there something else i need to do to update it?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ tracy-callstack-inlines = ["tracy-client-sys?/callstack-inlines"]
 anyhow = { version = "1.0.75", features = ["backtrace"] }
 async-trait = "0.1.53"
 backtrace = "0.3.67"
-clap = { version = "4.3.19", features = ["cargo", "derive", "env"] }
+clap = { version = "4.4.18", features = ["cargo", "derive", "env", "color"] }
 copypasta = "0.8.1"
 csscolorparser = "0.6.2"
 derive-new = "0.5.9"
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -3,15 +3,26 @@ use std::{iter, mem};
 use crate::{dimensions::Dimensions, frame::Frame, settings::*};
 
 use anyhow::Result;
-use clap::{builder::FalseyValueParser, ArgAction, Parser};
+use clap::{
+    builder::{styling, FalseyValueParser, Styles},
+    ArgAction, Parser,
+};
 
 #[cfg(target_os = "windows")]
 pub const SRGB_DEFAULT: &str = "1";
 #[cfg(not(target_os = "windows"))]
 pub const SRGB_DEFAULT: &str = "0";
 
+fn get_styles() -> Styles {
+    styling::Styles::styled()
+        .header(styling::AnsiColor::Green.on_default() | styling::Effects::BOLD)
+        .usage(styling::AnsiColor::Green.on_default() | styling::Effects::BOLD)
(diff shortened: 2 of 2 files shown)


fredizzimo:
Can you test that it also works with our gui version? Using `neovide --no-fork --help`. I don't think you need to fix it here, but at least disable it, there if it doesn't work.

> One thing to be aware of - i expected Cargo.lock to be updated, after changing Cargo.toml, but it was not updated. Is there something else i need to do to update it?

Clap 4.4.18, is already in Cargo.lock, and the rest of the dependencies were updated yesterday, so I think it's normal.

last-partizan:
`neovide --no-fork --help` works fine.

I think there's no difference between `cargo r -- --help` and `neovide --help`, first is just shortcut for build and run, with args after `--` going to the binary.

MultisampledNight (approved):

MultisampledNight:
> One thing to be aware of - i expected Cargo.lock to be updated, after changing Cargo.toml, but it was not updated. Is there something else i need to do to update it?

Nope, all ok. `Cargo.lock` already contains `4.4.18`, even before this PR.

Background: When writing `depname = "X.Y.Z"` in `Cargo.toml`, that's actually an alias for `depname = "^X.Y.Z"`, which basically means "feel free to upgrade `Y` (minor) and `Z` (patch) however you like but just make sure it's _at least_ what I specified". See

- https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio
- https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements

fredizzimo:
> `neovide --no-fork --help` works fine.
> 
> I think there's no difference between `cargo r -- --help` and `neovide --help`, first is just shortcut for build and run, with args after `--` going to the binary.

Sorry, I meant `--fork`, but now I realize that won't work either. You need to launch it detached from the terminal to show the Window. I don't remember how to properly do that, but at least editing the desktop file should work. 

Alternatively, you can temporarily comment out the isatty check

last-partizan:
@fredizzimo i do not understand what do you want to check?

`--fork` and `--no-fork` has nothing to do with `--help`, and i just added colors for `--help`.

When using `--no-fork` to launch GUI without detaching from the terminal, it works and logs as usual. When using `--help` it ignores other options and shows help in the terminal.

```
./target/release/neovide --no-fork README.md
[flexi_logger][ERRCODE::Time] flexi_logger has to work with UTC rather than with local time, caused by IndeterminateOffset
    See https://docs.rs/flexi_logger/latest/flexi_logger/error_info/index.html#time
```

fredizzimo:
When there's no tty attached we take this codepath

https://github.com/neovide/neovide/blob/879172ec071a9d02279a50aaeb38072fe258efe7/src/error_handling.rs#L107

And I'm not sure our custom error drawing supports colors. And yes, the help can be displayed there, if the user edits the desktop file for example, and provides the wrong parameters.

last-partizan:
@fredizzimo oh, sure! i edited desktop file, and it works fine without additional modifications.

image

image

fredizzimo:
Thank you!

PR #2319, opened 2024-01-27
feat: implement dynamic window title visibility option

falcucci:
PR needed to complement #2168 

- Feature

- No

### DIFF
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -61,6 +61,10 @@ pub struct CmdLineSettings {
     #[arg(long = "no-multigrid", env = "NEOVIDE_NO_MULTIGRID", value_parser = FalseyValueParser::new())]
     pub no_multi_grid: bool,
 
+    /// Sets title hint for the window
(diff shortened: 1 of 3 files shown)


fredizzimo:
I think, I would rather see that we fix https://github.com/neovide/neovide/pull/2126, to allow an empty titestring if set explicitly. `Neovide` would still remain the default if unset 

This way it works on all platforms, and we don't need any custom options

falcucci:
@fredizzimo got it working.

notice that:

- IF `title` is enabled and `titlestring` is empty, neovim sends `[No Name] - NVIM` as value;
- IF `title` is 0 then it is really empty as expeted;

fredizzimo:
Hm.. Yes, it does not seem to work the way I imagined. 

I guess we need an option to hide the title then, I don't think we should modify the default one. The option could just force it to empty, since that should be crossplatform

falcucci:
@fredizzimo done, work just fine. If you want to release you can use this to complement #2168 

I'll be idle for now, come back tomorrow

fredizzimo:
I rebased and fixed a warning from https://github.com/neovide/neovide/pull/2168 on non-macos platforms

Issue #1279, opened 2022-04-20
neovide_confirm_quit option not working when pressing command+q on macos


dineshKumar777:
**Describe the bug**
https://github.com/neovide/neovide/pull/1247
Option does not seem to work on MacOS 12.3, neither via command-q or via clicking the menu option.

**To Reproduce**
1. enable option let g:neovide_confirm_quit=1 and g:neovide_confirm_quit=v:true
2. make changes in buffer
3. click command-q or via clicking the menu option
4. neovide closes without save prompt

**Expected behavior**
show save prompt when quit
Note: In windows, alt+f4 shows save prompt.
More info on https://github.com/rust-windowing/winit/issues/41

**Desktop (please complete the following information):**
 - OS: [e.g. macOS 12.3]
 - Neovide Version [e.g. 0.8.0]
 - Neovim Version [e.g. 0.7.0]

MultisampledNight:
Do we even want to explicitly handle that or just let the user configure it?

dineshKumar777:
@MultisampledNight nvim-qt prompts on cmd-q, so I thought this is the default behavior. since this mac specific I am not sure about this one. As per winit#41 discussion, application authors has to handle this behavior.

[Alacritty implementation](https://github.com/alacritty/alacritty/blob/9bbb296d1dc4f72e845695f22e903bcc1d66d5a3/alacritty/src/config/bindings.rs#L720)

Liverm0r:
I remapped cmd+q (and lots of other cmd hotkeys, like cmd+v/c/w/s for paste/copy/close/save respectively) for neovide to be alt+q, and mapped alt+q to exit in vim.

[karabiner settings](https://github.com/Liverm0r/dotfiles/blob/master/.config/karabiner/assets/complex_modifications/1234504.json)

I gues you know how to remap vim keys yourself :)

9mm:
I still have this issue, was there something that i need to do to get this to work? id love to adjust both the window position as well as the side to always be on right half of screen

9mm:
@D00mch did you find a way to remap command+Q/W/T in neovide? i was trying to add tabs back and windows like macvim

fredizzimo:
I think this might be fixed in Neovide main. We recently updated to latest winit there.

And https://github.com/rust-windowing/winit/issues/41 was marked as fixed by https://github.com/rust-windowing/winit/pull/1583. I don't know if we need to do anything special to enable the menu though.

9mm:
[...]

PR #2323, opened 2024-01-29
feat: add window menu on mac

polachok:
- Feature

- No

This PR adds Window menu so the app feels more native and adds some potentially useful items, like tiling.
Also make cmd+m shortcut work.

Screenshot 2024-01-29 at 12 03 54

I'm not very sure about the event loop part, open to suggestions.

### DIFF
--- a/src/window/macos.rs
+++ b/src/window/macos.rs
@@ -3,6 +3,7 @@ use icrate::{
         NSApplication, NSColor, NSEvent, NSEventModifierFlagCommand, NSEventModifierFlagControl,
         NSEventModifierFlagOption, NSMenu, NSMenuItem, NSView, NSViewMinYMargin,
         NSViewWidthSizable, NSWindow, NSWindowStyleMaskFullScreen, NSWindowStyleMaskTitled,
+        NSWindowTabbingModeDisallowed,
     },
     Foundation::{MainThreadMarker, NSObject, NSPoint, NSProcessInfo, NSRect, NSSize, NSString},
 };
@@ -68,6 +69,10 @@ impl MacosWindowFeature {
             },
             _ => panic!("Not an appkit window."),
         };
+        // Disallow tabbing mode to prevent the window from being tabbed.
+        unsafe {
+            ns_window.setTabbingMode(NSWindowTabbingModeDisallowed);
+        }
 
         if let Ok(color) = &SETTINGS
             .get::<WindowSettings>()
--- a/src/window/macos.rs
+++ b/src/window/macos.rs
@@ -1,9 +1,10 @@
 use icrate::{
     AppKit::{
-        NSApplication, NSColor, NSEvent, NSMenu, NSMenuItem, NSView, NSViewMinYMargin,
-        NSViewWidthSizable, NSWindow, NSWindowStyleMaskFullScreen, NSWindowStyleMaskTitled,
+        NSApplication, NSColor, NSEvent, NSEventModifierFlagCommand, NSEventModifierFlagOption,
+        NSMenu, NSMenuItem, NSView, NSViewMinYMargin, NSViewWidthSizable, NSWindow,
+        NSWindowStyleMaskFullScreen, NSWindowStyleMaskTitled,
     },
-    Foundation::{MainThreadMarker, NSPoint, NSRect, NSSize, NSString},
+    Foundation::{MainThreadMarker, NSObject, NSPoint, NSProcessInfo, NSRect, NSSize, NSString},
 };
 use objc2::{declare_class, msg_send_id, mutability::InteriorMutable, rc::Id, sel, ClassType};
 
@@ -12,6 +13,7 @@ use raw_window_handle::{HasRawWindowHandle, RawWindowHandle};
 use winit::event::{Event, WindowEvent};
 use winit::window::Window;
 
+use crate::bridge::{send_ui, ParallelCommand};
 use crate::{
     cmd_line::CmdLineSettings, error_msg, frame::Frame, renderer::WindowedContext,
     settings::SETTINGS, window::UserEvent,
@@ -177,17 +179,41 @@ impl MacosWindowFeature {
     }
 }
 
-pub struct Menu {
-    menu_added: bool,
-}
+declare_class!(
+    struct QuitHandler;
 
-impl Default for Menu {
-    fn default() -> Self {
-        Menu { menu_added: false }
+    unsafe impl ClassType for QuitHandler {
+        type Super = NSObject;
+        type Mutability = InteriorMutable;
+        const NAME: &'static str = "QuitHandler";
+    }
+
+    unsafe impl QuitHandler {
+        #[method(quit:)]
+        unsafe fn quit(&self, _event: &NSEvent) {
+            send_ui(ParallelCommand::Quit);
+        }
+    }
+);
+
+impl QuitHandler {
+    pub fn new(_mtm: MainThreadMarker) -> Id<QuitHandler> {
+        unsafe { msg_send_id![Self::alloc(), init] }
     }
 }
 
+pub struct Menu {
+    menu_added: bool,
+    quit_handler: Id<QuitHandler>,
+}
+
 impl Menu {
+    pub fn new(mtm: MainThreadMarker) -> Self {
+        Menu {
+            menu_added: false,
+            quit_handler: QuitHandler::new(mtm),
+        }
+    }
     pub fn ensure_menu_added(&mut self, ev: &Event<UserEvent>) {
         if let Event::WindowEvent {
             event: WindowEvent::Focused(_),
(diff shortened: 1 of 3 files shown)


src/window/mod.rs
```diff
@@ -343,7 +371,21 @@ pub fn main_loop(
...
+            Event::WindowEvent {
+                event: winit::event::WindowEvent::Focused(_),
+                ..
+            },
+            false,
+        ) = (&e, menu_added)
+        {
+            add_window_menu();
+            menu_added = true;
+        }
```
falcucci, line 388: the main loop function is getting long, I would like to see the same approach you did at `add_window_menu` making macOS specific changes external.

here for example, if we intend to add those kind of features on menu, as the project evolves would be hard to track and maintain.

I would do something like this to manage the menu state:

```rust

/// this should probably be moved to `/window/macos.rs` crate
#[cfg(target_os = "macos")]
struct MenuState {
    menu_added: bool,
}

impl MenuState {
    fn new() -> Self {
        Self { menu_added: false }
    }

    fn ensure_menu_added(&mut self, e: &Event<UserEvent>) {
        if let Event::WindowEvent {
            event: winit::event::WindowEvent::Focused(_),
            ..
        } = e
        {
            if !self.menu_added {
                add_window_menu();
                self.menu_added = true;
            }
        }
    }
}
```

and then use it in a simple way like:

```rust
#[cfg(target_os = "macos")]
let mut menu_state = MenuState::new();

event_loop.run(move |e, window_target| {
    #[cfg(target_os = "macos")]
    menu_state.ensure_menu_added(&e);

    if e == Event::LoopExiting {
        return;
    }

    if !RUNNING_TRACKER.is_running() {
        save_window_size(&window_wrapper);
        window_target.exit();
    } else {
        window_target.set_control_flow(update_loop.step(&mut window_wrapper, Ok(e)).unwrap());
    }
})
```

falcucci (changes requested):

src/window/mod.rs
polachok: Ok, did that

falcucci (approved):

fredizzimo:
I will review this later, but just for awareness, there was an earlier attempt of this by @9mm, trying a couple of different approaches, but I don't know how they compare, or if we should combine some of the things tried.
* https://github.com/neovide/neovide/pull/2055

Kethku:
It also seems like whatever implementation should at least reference https://neovim.io/doc/user/gui.html to see if theres some way we can pull from the neovim native menu setup or maybe emulate their approach to adding things dynamically

falcucci:
@Kethku that looks interesting. As it says, neovim doesn't create them by default. I think we could go against that and add or simplest default menu items (as it is e.g in this PR) and in case the user uses their own :emenus we try to sync them adding it dynamically in a separate tab?

I am not a menu user tho but some information there can help, maybe?

@fredizzimo hmm i don't see such a difference in comparison to this PR you've mentioned, regarding to the simplest way that this one is implemented without any external lib for example. I appreciate @9mm PR tho, his opinion here would be welcome.

src/window/macos.rs
```diff
@@ -175,3 +176,52 @@ impl MacosWindowFeature {
...
+            if !self.menu_added {
+                self.add_window_menu();
+                self.menu_added = true;
+            }
+        }
+    }
+
+    #[cfg(target_os = "macos")]
+    fn add_window_menu(&self) {
+        use cocoa::appkit::{NSApplication, NSMenu, NSMenuItem};
```
fredizzimo, line 202: This should probably use icrate instead of cocoa, like 
* https://github.com/neovide/neovide/pull/2351

fredizzimo:
I think this should include at least `⌘-Q`, which was one of the main reasons why @9mm attempted:
* https://github.com/neovide/neovide/pull/2055

polachok:
Okay, I rewrote it with icrate and `⌘-Q` handling

9mm:
Quite cool, great work! @polachok not sure if you know how to do this (as I'm not familiar with OSX dev .. or rust), which is why I stopped, but in this draft PR i tried to add dragging and dropping files into the dock icon, which uses `openFile:`. I stopped though, as I didn't find a good way to do this.

In the PR though, I did manage to switch to a more comprehensive build setup that allows 2 new things: 1) completely customizing info.plist which I believe is a precursor to dragging files. 2) dmg packager which is a nice to have

Do you happen to have a good way to know how to do this? I feel like this is a hugely needed feature and there are a number of issues related to it, I dunno if you have any insight into how this might be possible (no pressure though if you are busy/not interested, etc). I noticed a few other issues in winit mentioning things like "fruitbasket" but when I looked at those, it seems like managing 2 different event loops is required, which is where I stopped as it wasnt a simple thing.

https://github.com/neovide/neovide/pull/2191

fredizzimo:
@9mm, could you take a look at this? 

It should now support `⌘-Q`,  which hopefull finllay fixes this:
*  https://github.com/neovide/neovide/issues/1279

So, it would be great to get some confirmation that this indeed fixes that.

9mm:
Definitely!

9mm:
So far things I notice:

* in full screen the red/yellow/green window buttons are hidden. This is zero issue for me, and probably because I'm using "NEOVIDE_FRAME=buttonless", but I figure it's worth pointing out anyway. You can still exit by going to window > exit full screen
* Command+Q works great! including the clean shutdown:
* The application name should be captialized to `Neovide` though
* About neovide should have something good in it as well, i put macvims for comparison

image


image

image

image

9mm:
It does work in default `NEOVIDE_FRAME=full` however, so thats great:

image

fredizzimo:
I forgot to ask, are you also able to close the window without a frame, so we can close this one as well:
* https://github.com/neovide/neovide/issues/2087

9mm:
On this branch, when running:

    NEOVIDE_FRAME=none cargo run --release

And typing `:q` it still crashes

fredizzimo:
Ah, yes we most likely need to move the window creation inside the event loop. If that does not work, I think we need to report the bug to winit.

9mm:
Ahh i do see one problem... the app title at the top is definitely a show stopper, I just noticed my statusline is much bigger. This occurs in buttonless which is what i use:

```
NEOVIDE_FRAME=buttonless cargo run --release
```

Title at top:

screenshot-20240218-hbgvzWBc

Much bigger status line which is definitely not good:

image

polachok:
>About neovide should have something good in it as well, i put macvims for comparison

I think you're running it with `cargo run`, if you use `cargo bundle` and open the bundle instead, it looks better.

9mm:
Awesome:

image

9mm:
^ That also fixes the capitalized name

polachok:
>Ahh i do see one problem... the app title at the top is definitely a show stopper

Have you tried the main branch? I don't think this PR affects the title in any way

9mm:
@polachok yes you are right, this is on main now :/ Im not sure when this was introduced, i have been using a built version from main thats maybe 2 months old

9mm:
https://github.com/neovide/neovide/issues/2319

this maybe, let me have a look

9mm:
Thats it, if I pass `NEOVIDE_TITLE_HIDDEN=true` the title hides but unfortunately the command line has a lot more wasted space. @fredizzimo is this something obvious or easy I can change that im missing?

fredizzimo:
It's more likely this:
* https://github.com/neovide/neovide/pull/2168

I'm not sure what the normal titlebar height is, but it does look a bit thick in the screenshots there. 


For the commandline size, maybe the padding is added both to the top and bottom in that PR?

fredizzimo:
But if it's indeed caused by that, then we should create a new issue. Since it does not affect if this can be merged or not.

9mm:
Awesome, I made an issue. the only remaining thing here i see is the `buttonless` frame in full size, but i dont know if that's even fixable as im assuming buttonless removes all those buttons.

One other thing I noticed, to activate full screen the hotkey is currently FN+F. it might be better to do the typical hotkey of  `ctrl+command+f`

image

polachok:
Okay, I added ctrl+command+f binding, menu still shows 🌐 F though 🤷🏼

polachok:
Anything else I need to do to get this merged?

fredizzimo:
I thought you wanted to fix the wrong shortcut display, so I haven't looked at it further.


We also need some macOS user to check the final solution.

falcucci:
@polachok could you disable tabbing items from the Window menu?

you can do that changing the tab mode completely using

```rust
// Disallow tabbing mode to prevent the window from being tabbed.
unsafe {
    ns_window.setTabbingMode(NSWindowTabbingModeDisallowed);
}
```

This is important to avoid issues related to tabs right now. People can get confuse when they see those options and have a misunderstanding about any tabs support.

polachok:
Done

falcucci:
LGTM

falcucci:
@polachok thank you!

PR #2334, opened 2024-02-02
fix: Handle some more numpad keys

hhirtz:
- Fix. Some numpad keys are not mapped to the correct neovim key. With this patch, neovide honors numlock for [`NumpadDecimal`], and sets the correct neovim keycode for [`NumpadComma`] and [`NumpadEqual`].

- No. 

[`NumpadComma`]: https://docs.rs/winit/0.29.10/winit/keyboard/enum.KeyCode.html#variant.NumpadComma
[`NumpadDecimal`]: https://docs.rs/winit/0.29.10/winit/keyboard/enum.KeyCode.html#variant.NumpadDecimal
[`NumpadEqual`]: https://docs.rs/winit/0.29.10/winit/keyboard/enum.KeyCode.html#variant.NumpadEqual

This is a follow up of #2032

See also [neovim docs][0].

Please note that I don't have a numpad with `<kEqual>` and `<kComma>` keys, I could only test kPoint (which is currently incorrectly handled as `<kDel>` even with numlock on).

[0]: https://github.com/neovim/neovim/blob/1bf645918e94e7e8f770592484164f1ee303f24e/runtime/doc/intro.txt#L377-L379

### DIFF
--- a/src/window/keyboard_manager.rs
+++ b/src/window/keyboard_manager.rs
@@ -95,7 +95,23 @@ impl KeyboardManager {
             KeyCode::NumpadSubtract => Some("kMinus"),
             KeyCode::NumpadAdd => Some("kPlus"),
             KeyCode::NumpadEnter => Some("kEnter"),
-            KeyCode::NumpadDecimal => Some("kDel"),
+            KeyCode::NumpadEqual => Some("kEqual"),
+            KeyCode::NumpadComma => match key_event.logical_key.as_ref() {
+                Key::Character(",") => Some("kComma"),
+                Key::Character(".") => Some("kPoint"),
+                _ => None,
+            },
+            KeyCode::NumpadDecimal => {
+                if is_numlock_key {
+                    match key_event.logical_key.as_ref() {
+                        Key::Character(",") => Some("kComma"),
+                        Key::Character(".") => Some("kPoint"),
+                        _ => None,
+                    }
+                } else {
+                    Some("kDel")
+                }
+            }
             KeyCode::Numpad9 => {
                 KeyboardManager::handle_numpad_numkey(is_numlock_key, "k9", "kPageUp")
             }


fredizzimo:
Thank you, but I think this is still wrong, since both `NumpadDecimal` and `NumpadComma` are locale dependent. I wonder if `KeyEvent.text` has the correct translated character to display https://docs.rs/winit/0.29.10/winit/event/struct.KeyEvent.html#structfield.text?

hhirtz:
As far as I can tell winit's `KeyCode::NumpadDecimal` is equivalent to neovim's `kPoint`.

winit's `KeyCode` denotes physical key on a keyboard, not text. What neovim might print when fed `kPoint` might be locale-dependant though.

It [seems] winit have taken the naming from [a w3.org spec][w3], which has this great illustration:

illustration of 3 different numpad layouts

You can see `NumpadDecimal` (bottom-right on all layouts) and `NumpadComma` (brazilian layout only) are two different keys. And from neovim's source code (more specifically its implementation of kitty keyboard protocol), it seems that `kPoint` really is `NumpadDecimal`.

When neovim is in a terminal, it uses [the kitty keyboard protocol][kbp] to get the physical key when possible, and in [neovim' source][nvsrc] `kPoint` points to `KP_DECIMAL`  from the kitty keyboard protocol (`KP_DECIMAL` is 57409, [table]):

https://github.com/neovim/neovim/blob/be1d09c4272212ea9b354c900603568d238b4ab3/src/nvim/tui/input.c#L101-L107

Now kitty's protocol does not go in as much details as the w3 spec (no illustrations and misses the comma key), but terminals that implement it (including winit-based terminals wezterm, alacritty and rio) do translate `KP_DECIMAL` either to winit's `NumpadDecimal`, or to "." (not ","):

- wezterm uses NumpadDecimal: https://github.com/wez/wezterm/blob/5c172e4ed2508873817876a718be3920c83fdf9d/termwiz/src/input.rs#L284
- alacritty checks for numpad and uses ".": https://github.com/alacritty/alacritty/blob/2d2b894c3b869fadc78fce9d72cb5c8d2b764cac/alacritty/src/input/keyboard.rs#L408
- rio follows the same approach as alacritty: https://github.com/raphamorim/rio/blob/f953f5ce012d70e3818334b9e54ba1794adbafe0/frontends/rioterm/src/bindings/bindings_winit/kitty_keyboard_protocol.rs#L57
- foot maps KP_DECIMAL to `XKB_KEY_KP_Decimal`, an equivalent of `NumpadDecimal`: https://codeberg.org/dnkl/foot/src/commit/44c0cf594b4674e92b9dd1646d7a2d43c401e5d8/kitty-keymap.h#L57

so i think it's pretty safe to say that `KeyCode::NumpadDecimal` is equivalent to `KP_DECIMAL` and thus to `kPoint`.

[kbp]: https://sw.kovidgoyal.net/kitty/keyboard-protocol
[table]: https://sw.kovidgoyal.net/kitty/keyboard-protocol/#functional-key-definitions
[nvsrc]: https://github.com/neovim/neovim/blob/be1d09c4272212ea9b354c900603568d238b4ab3/src/nvim/tui/input.c#L101
[seems]: https://github.com/rust-windowing/winit/blob/4d4d6e5052877d9578bac73f52eaf323e9089998/src/platform_impl/windows/keyboard.rs#L1127-L1130
[w3]: https://www.w3.org/TR/uievents-code/#key-numpad-section

fredizzimo:
I think you are wrong.

In terminal nvim I get a comma or period depending on the keyboard layout. I tested both alacritty and neovim-qt and both report `<kPoint>` on an English layout, but with a Swedish layout they return `,`. I believe both are buggy though, and they should return `<kComma>`, on the Swedish layout.

But with your branch I always get period, well actually `<kPoint>`. It's also wrong during regular typing, no matter which keyboard layout is used, it always types periods.

That's not surprising, since there's no translation https://github.com/neovim/neovim/blob/be1d09c4272212ea9b354c900603568d238b4ab3/src/nvim/getchar.c#L1505-L1508, and it cannot be done, since Neovim could be run on a remote computer with no idea of what keyboard layout is active.

So, we will have to do it on the Neovide side. Alacritty deals with it by reading the text of the `logical_key` as seen in the code you linked.

fredizzimo:
Note, I was only testing `NumpadDecimal`, since I don't have a `NumpadComma` key. But `NumpadpadComma` is also either a period or comma, depending on which thousand separator is used in the language, and is called `VK_SEPARATOR` on windows. It might even map to space on some layouts, but I'm not sure.

src/window/keyboard_manager.rs
```diff
@@ -95,7 +95,24 @@ impl KeyboardManager {
             KeyCode::NumpadSubtract => Some("kMinus"),
             KeyCode::NumpadAdd => Some("kPlus"),
             KeyCode::NumpadEnter => Some("kEnter"),
-            KeyCode::NumpadDecimal => Some("kDel"),
+            KeyCode::NumpadEqual => Some("kEqual"),
+            KeyCode::NumpadComma => match key_event.logical_key.as_ref() {
+                Key::Character(",") => Some("kComma"),
+                Key::Character(".") => Some("kPoint"),
+                _ => key_event.logical_key.to_text(),
```
fredizzimo, line 102: I think it's better to return None here, that way `format_normal_key` will be called and used, which probably works better.

And the same thing for `NumpadDecimal`

hhirtz: Here you go

fredizzimo:
Thank you!

Issue #2299, opened 2024-01-23
Crashes on Windows 10


Zhniing:
**Describe the bug**
Neovide crashes after opening an empty buffer and typing something.

Also try to open a file (about 200 lines), and scroll the buffer via the mouse wheel. Crashes as well.

**To Reproduce**
1. Run `neovide --log -- --clean` in PowerShell
2. Type something
3. Neovide freezes 1 or 2 seconds
4. Neovide crashes (window closed)

**Expected behavior**
Don't crash.

**Screenshots**
I captured a video (1.84MB in `.mp4`) of the whole process but, seems unable to upload.

**Desktop:**
 - OS: Windows 10 22H2
 - Neovide Version: build from [3e575fcd](https://github.com/neovide/neovide/commit/3e575fcd21863fdaa1fd14d23db649738694af97) (via `cargo install --path .`)
 - Neovim Version: 0.9.5 (via `scoop install neovim`)

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/14022054/neovide_rCURRENT.log)

**Additional context**
This may be a stupid issue. I tried various versions of both Neovide and Neovim, and always crashed. I just have no idea what happened. Could you help me out?

Thank you for making this impressed project!

fredizzimo:
Do you have a neovide_backtraces.log file, in the working directory where you started neovide from?

Zhniing:
> Do you have a neovide_backtraces.log file, in the working directory where you started neovide from?

No. I start Neovide in an empty directory. There is only one file (i.e. neovide_rCURRENT.log) generated after Neovide crashed.

fredizzimo:
Ok, can you try first setting the environment variable `RUST_BACKTRACE=1` and then run `cargo run -- --clean --log`, that should print the backtrace to the terminal, and run a debug build.

Zhniing:
Tried as the following process:

1. Run `$env:RUST_BACKTRACE=1` in PowerShell to set the environment variable
2. Run `cargo run -- --clean --log` and get this error:

```
error: unexpected argument '--clean' found

  tip: to pass '--clean' as a value, use '-- --clean'

Usage: neovide.exe [OPTIONS] [FILES_TO_OPEN]... [-- [NEOVIM_ARGS]...]

For more information, try '--help'.
error: process didn't exit successfully: `target\debug\neovide.exe --clean --log` (exit code: 2)
```

[...]

Issue #2182, opened 2023-12-15
Neovide Windows freezes when quitting


Parsifa1:
## **Describe the bug**

Neovide Windows 11 WSL2 not responding when quit

## **To Reproduce**

1. Open WSL terminal
1. Run ` neovide.exe --wsl `
1. Quit neovide either using ` :q ` or clicking close button
1. Neovide becomes freezed

## **Expected behavior**

exit neovim and neovide as expected

## **Screenshots**

![image](https://github.com/neovide/neovide/assets/31800073/95afc09d-c22b-4d2d-ba75-d0e3491e904e)

## **Desktop (please complete the following information):**

 - OS: Windows11 23H2
 - Neovide Version : 0.11.2
 - Neovim Version: 
 ` nightly `
 ```
 NVIM v0.10.0-dev-1846+gf31f260f0c6
Build type: RelWithDebInfo
LuaJIT 2.1.1702233742
Run "nvim -V1 -v" for more info
```

 **Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/13687097/neovide_rCURRENT.log)

fredizzimo:
I'm using Neovide  in Windows and WSL all the time, and don't see the problem, so it's hard to tell what's wrong.

One possibility is the clipboard hang, described here https://github.com/neovide/neovide/issues/2003. On WSL you only need Win32yank.exe, so if you have xlclip or similar installed, perhaps you could try to uninstall that. It also causing problems with the copy and paste functionality, since it has a higher priority than Win32Yank, or at least it used to have when I last checked.

Parsifa1:
In fact, I only used win32yank, and if I didn't use win32yank, ^M would appear on each line due to the difference between Windows and Linux.

![image](https://github.com/neovide/neovide/assets/31800073/51d0aa39-0f7a-4628-a95d-900cc04dab48)

Parsifa1:
I also tried close ` vim.opt.clipboard = "unnamedplus" `, and it comes again

![image](https://github.com/neovide/neovide/assets/31800073/cbc23ecc-069e-4455-9fae-18a9bd3ff608)

Parsifa1:
[...]

PR #2336, opened 2024-02-03
Update nvim-rs, remove superflous arc-wrapping

KillTheMule:
- Codestyle
- Other

- No

Updated `nvim-rs` to the latest release. I noticed you're wrapping the `Neovim` instance in `Arc`. That's not really necessary, since `nvim-rs` does `Arc`-wrapping for all components anyways.

There might be performance implications - this removes an indirection, but cloning the `Neovim` instance now needs to bump 3 counters instead of one. My assumption would be that this is a win, but there aren't any benchmarks in place to decide this.Your choice :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1758,9 +1758,9 @@ dependencies = [
 
 [[package]]
 name = "nvim-rs"
-version = "0.6.0"
+version = "0.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9156b04d79079937abae1eb912669c48c360fb5d86138052a9bbae9c5220a71e"
+checksum = "45fcd98ddbb0c866684dacec0d99a7c623c1784b0facfd44cfee17cc79aa81b2"
 dependencies = [
  "async-trait",
  "futures 0.3.30",
@@ -1770,6 +1770,7 @@ dependencies = [
  "rmpv",
  "tokio",
  "tokio-util",
+ "winapi",
 ]
 
 [[package]]
@@ -2775,9 +2776,9 @@ checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
 
 [[package]]
 name = "tokio"
-version = "1.35.1"
+version = "1.36.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c89b4efa943be685f629b149f53829423f8f5531ea21249408e8e2f8671ec104"
+checksum = "61285f6515fa018fb2d1e46eb21223fff441ee8db5d0f1435e8ab4f5cdb80931"
 dependencies = [
  "backtrace",
  "bytes 1.5.0",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ log = "0.4.16"
 lru = "0.7.5"
 neovide-derive = { path = "neovide-derive", version = "0.1.0" }
 num = "0.4.1"
-nvim-rs = { version = "0.6.0", features = ["use_tokio"] }
+nvim-rs = { version = "0.7.0", features = ["use_tokio"] }
 parking_lot = "0.12.0"
 rand = "0.8.5"
 raw-window-handle = "0.5.0"
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -11,7 +11,7 @@ use itertools::Itertools;
 use log::{error, info};
 use nvim_rs::{error::CallError, Neovim, UiAttachOptions, Value};
 use rmpv::Utf8String;
-use std::{io::Error, ops::Add, sync::Arc};
+use std::{io::Error, ops::Add};
 use tokio::runtime::{Builder, Runtime};
 use winit::event_loop::EventLoopProxy;
 
@@ -109,7 +109,7 @@ async fn launch(handler: NeovimHandler, grid_size: Option<Dimensions>) -> Result
     let should_handle_clipboard = settings.wsl || settings.server.is_some();
     setup_neovide_specific_state(&session.neovim, should_handle_clipboard).await?;
 
-    start_ui_command_handler(Arc::clone(&session.neovim));
+    start_ui_command_handler(session.neovim.clone());
     SETTINGS.read_initial_values(&session.neovim).await?;
 
     let mut options = UiAttachOptions::new();
--- a/src/bridge/session.rs
+++ b/src/bridge/session.rs
@@ -37,7 +37,7 @@ impl NeovimSession {
         let io_handle = spawn(io);
 
         Ok(Self {
-            neovim: neovim,
+            neovim,
             io_handle,
         })
     }
--- a/src/bridge/session.rs
+++ b/src/bridge/session.rs
@@ -4,7 +4,6 @@
 use std::{
     io::{Error, ErrorKind, Result},
     process::Stdio,
-    sync::Arc,
 };
 
 use nvim_rs::{error::LoopError, neovim::Neovim, Handler};
@@ -23,7 +22,7 @@ type BoxedReader = Box<dyn AsyncRead + Send + Unpin + 'static>;
 type BoxedWriter = Box<dyn AsyncWrite + Send + Unpin + 'static>;
 
 pub struct NeovimSession {
-    pub neovim: Arc<Neovim<NeovimWriter>>,
+    pub neovim: Neovim<NeovimWriter>,
     pub io_handle: JoinHandle<std::result::Result<(), Box<LoopError>>>,
 }
 
@@ -38,7 +37,7 @@ impl NeovimSession {
         let io_handle = spawn(io);
 
         Ok(Self {
-            neovim: Arc::new(neovim),
+            neovim: neovim,
             io_handle,
         })
     }
--- a/src/bridge/session.rs
+++ b/src/bridge/session.rs
@@ -36,10 +36,7 @@ impl NeovimSession {
             Neovim::<NeovimWriter>::new(reader.compat(), Box::new(writer.compat_write()), handler);
         let io_handle = spawn(io);
 
-        Ok(Self {
-            neovim,
-            io_handle,
-        })
+        Ok(Self { neovim, io_handle })
     }
 }
 
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -1,4 +1,4 @@
-use std::sync::{Mutex};
+use std::sync::Mutex;
 
 use log::trace;
 
--- a/src/bridge/ui_commands.rs
+++ b/src/bridge/ui_commands.rs
@@ -1,4 +1,4 @@
-use std::sync::{Arc, Mutex};
+use std::sync::{Mutex};
 
 use log::trace;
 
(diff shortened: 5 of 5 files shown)


falcucci:
@KillTheMule thank you! i'll just make sure to use it for a bit to certify everything works flawlessly before approving this.

KillTheMule:
Done.

fredizzimo:
@Zhniing
If I'm not mistaken (I haven't looked at the changes in detail) the file handles should no longer be asynchrounous, and run into the potential rust standard library bug I found here https://github.com/neovide/neovide/issues/2299#issuecomment-1907411309. My analysis there is slightly wrong, though`NtReadFile` allows `WaitForSingleObject` in certain cases, which I think should have been fulfilled, but I see no other possible reason for your crash,

@sid-6581, can you revert your exit hang fix and test with this? I strongly believe that one was also caused by the same bug. And if not, so much has changed in `nvim-rs` that something else might have fixed it.

sid-6581:
> @Zhniing, could you test this? If I'm not mistaken (I haven't looked at the changes in detail) the file handles should no longer be asynchrounous, and run into the potential rust standard library bug I found here [#2299 (comment)](https://github.com/neovide/neovide/issues/2299#issuecomment-1907411309). My analysis there is slightly wrong, though`NtReadFile` allows `WaitForSingleObject` in certain cases, which I think should have been fulfilled, but I see no other possible reason for your crash,
> 
> @sid-6581, can you revert your exit hang fix and test with this? I strongly believe that one was also caused by the same bug. And if not, so much has changed in `nvim-rs` that something else might have fixed it.

I'll check! I'm also experiencing a lot of hangs now while I'm in the middle of using neovide (not when closing), so I'll be running in the debugger while working and checking on both issues.

fredizzimo:
@KillTheMule, thanks. I guess it didn't occur to us that it was cloneable.

fredizzimo:
@sid-6581, I think there's a temporary freeze caused by https://github.com/neovide/neovide/pull/2188, it might not skip ahead when the computer has been sleeping for example. But that one recovers after a while.

sid-6581:
@fredizzimo, the freezes I've observed happen a lot, and I haven't seen it recover. From my recollection it also happens without locking the computer or anything like that. If I'm not careful to kill the whole process tree, I get a lot of orphan nvim processes left behind.

I just hope it's easily reproducible in a debug build, and not just a race condition that mostly happens in the production build.

fredizzimo:
I wish I could help you with that, but all three of my Windows (11) machines are completely stable. One of them with Neovide uptimes of weeks, it's probably one week even now, but that's on the D3D branch.

sid-6581:
OK, I tried reverting my fix on this PR and got neovide to hang on the first attempt, just like before. The nvim process is gone, and neovide still doesn't detect it as gone.

@KillTheMule have you ever seen this problem?

(Also, totally unrelated observation, but neovide no longer compiles on the latest nightly. Something has a dependency on ahash 0.7.7, which is where the problem is.)

KillTheMule:
> @KillTheMule have you ever seen this problem?

I couldn't really follow along, discussions are quite long... can you spell that out for me? Or link to a specific comment that tells me what you're talking about?

sid-6581:
> > @KillTheMule have you ever seen this problem?
> 
> I couldn't really follow along, discussions are quite long... can you spell that out for me? Or link to a specific comment that tells me what you're talking about?

Apologies, I wasn't sure if you had followed the previous discussions. Here is some context:

https://github.com/neovide/neovide/pull/2130

TLDR; sometimes when the nvim process exits (cleanly, by all appearances), the stdin/stdout pipes aren't broken, and `nvim-rs` doesn't break out of its I/O loops. This causes the runtime to never shut down, because it's waiting for the `nvim-rs` task.

sid-6581:
> @sid-6581, here's a fix for the nightly toolchain:
> 
> * [chore: update lru crate to fix building with the nightly toolchain #2346](https://github.com/neovide/neovide/pull/2346)

Great, that was quick! Thanks!

KillTheMule:
> TLDR; sometimes when the nvim process exits (cleanly, by all appearances), the stdin/stdout pipes aren't broken, and `nvim-rs` doesn't break out of its I/O loops. This causes the runtime to never shut down, because it's waiting for the `nvim-rs` task.

Ok, I have not seen this before. It's windows only, and you're the only one to reproduce it? I'll try to see if I can reproduce it... and then I need a windows dev environment... might take some time, but I'll look.

I do agree though that the loop @fredizzimo points to looks hellishly suspicious. As far as I can tell though that amounts to busy waiting, i.e. it should put some load on a core. Can you confirm this? Should also fill up the log with debug messages...

fredizzimo:
@KillTheMule, 
This is what I'm most suspicious of currently. I don't think the fix here is correct:
* https://github.com/rust-lang/rust/issues/98947

Since @Zhniing encountered the exact same issue here https://github.com/neovide/neovide/issues/2299, with the hashes matching the 1.75.0 toolchain version.

IMO they should use an event object there. 

Although the hang is different from the crash, it could be explained by the same thing, but in this case `WaitForSingleObject` never returning. 

The last line in this image posted by @sid-6581, also seem to indicate that it's blocking exactly there
image

sid-6581:
> > TLDR; sometimes when the nvim process exits (cleanly, by all appearances), the stdin/stdout pipes aren't broken, and `nvim-rs` doesn't break out of its I/O loops. This causes the runtime to never shut down, because it's waiting for the `nvim-rs` task.
> 
> Ok, I have not seen this before. It's windows only, and you're the only one to reproduce it? I'll try to see if I can reproduce it... and then I need a windows dev environment... might take some time, but I'll look.
> 
> I do agree though that the loop @fredizzimo points to looks hellishly suspicious. As far as I can tell though that amounts to busy waiting, i.e. it should put some load on a core. Can you confirm this? Should also fill up the log with debug messages...
> 
> (e) I guess I can't even read my own loop, for now I think it's fine. Gonna look into the hang.

There have been a lot of reports of hangs on exit, so unfortunately, it's not just me. I can just assume other's hangs are caused by the same issue, but I can't say for sure without anyone firing up a debugger and checking. When it hangs there's no busy waiting, it's stuck in a Win32 wait function. No debug messages in the log, which makes sense, because it's just stuck in OS land.

fredizzimo:
@sid-6581 


Is it stuck in this function if you are on nightly? 
https://github.com/rust-lang/rust/blob/8ace7ea1f7cbba7b4f031e66c54ca237a0d65de6/library/std/src/sys/pal/windows/handle.rs#L250

Or here in 1.75
https://github.com/rust-lang/rust/blob/82e1608dfa6e0b5569232559e3d385fea5a93112/library/std/src/sys/windows/handle.rs#L250

That code should probably create an event pass that to NtReadFile, and wait for the event, not wait for the file handle directly.

There are some special situations where waiting for the handle is allowed, but I'm not sure they hold here, especially since we are dealing with pipes and not filea https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-ntreadfile

> If the caller opened the file with the SYNCHRONIZE flag set in DesiredAccess, the calling thread can synchronize to the completion of the read operation by waiting on the file handle, FileHandle. The handle is signaled each time that an I/O operation that was issued on the handle completes. However, the caller must not wait on a handle that was opened for synchronous file access (FILE_SYNCHRONOUS_IO_NONALERT or FILE_SYNCHRONOUS_IO_ALERT). In this case, NtReadFile waits on behalf of the caller and does not return until the read operation is complete. The caller can safely wait on the file handle only if all three of the following conditions are met:
The handle was opened for asynchronous access (that is, neither FILE_SYNCHRONOUS_IO_XXX flag was specified).
The handle is being used for only one I/O operation at a time.
NtReadFile returned STATUS_PENDING.

KillTheMule:
> > > @KillTheMule have you ever seen this problem?
> > 
> > 
> > I couldn't really follow along, discussions are quite long... can you spell that out for me? Or link to a specific comment that tells me what you're talking about?
> 
> Apologies, I wasn't sure if you had followed the previous discussions. Here is some context:
> 
> #2130
> 
> TLDR; sometimes when the nvim process exits (cleanly, by all appearances), the stdin/stdout pipes aren't broken, and `nvim-rs` doesn't break out of its I/O loops. This causes the runtime to never shut down, because it's waiting for the `nvim-rs` task.

I could not reproduce that. I'm starting neovide from powershell, when the window opens I hit 'a' (having `nnoremap a :qa<Enter>` in $MYVIMRC), and everything seems fine. Is that how I need to do it? How many tries should this roughly need to reproduce?

sid-6581:
> > > > @KillTheMule have you ever seen this problem?
> > > 
> > > 
> > > I couldn't really follow along, discussions are quite long... can you spell that out for me? Or link to a specific comment that tells me what you're talking about?
> > 
> > 
> > Apologies, I wasn't sure if you had followed the previous discussions. Here is some context:
> > #2130
> > TLDR; sometimes when the nvim process exits (cleanly, by all appearances), the stdin/stdout pipes aren't broken, and `nvim-rs` doesn't break out of its I/O loops. This causes the runtime to never shut down, because it's waiting for the `nvim-rs` task.
> 
> I could not reproduce that. I'm starting neovide from powershell, when the window opens I hit 'a' (having `nnoremap a :qa<Enter>` in $MYVIMRC), and everything seems fine. Is that how I need to do it? How many tries should this roughly need to reproduce?

That's how I get it to hang easily as well. I start it up and immediately hit my shortcut that does `:qa`.

Are you running the latest commit, which includes my workaround for the hang? If so, try changing line 171 in `src/bridge/mod.rs` from `self.runtime.take().unwrap().shutdown_background()` to `drop(self.runtime.take().unwrap())`. I get it to hang repeatedly by quickly triggering a `:qa` after startup.

KillTheMule:
No I'm running 0.12.2 from the releases page. I also tried via `neovide.exe -- -c ":qa"`, but no hang.

sid-6581:
> No I'm running 0.12.2 from the releases page. I also tried via `neovide.exe -- -c ":qa"`, but no hang.

0.12.2 has my workaround for the hang, so you won't be able to reproduce it with that version.

KillTheMule:
Oh, ok... so I'm gonna setup a rust dev environment on windows.

sid-6581:
> Oh, ok... so I'm gonna setup a rust dev environment on windows.

You can also try 0.12.1, which was before the workaround.

KillTheMule:
Thanks for that hint, something on MS's end seems to be wrong, so I can't install right now...

That being said, I also could not reproduce with 0.12.1. Darn...

Zhniing:
> @Zhniing, could you test this?
> If I'm not mistaken (I haven't looked at the changes in detail) the file handles should no longer be asynchrounous, and run into the potential rust standard library bug I found here [#2299 (comment)](https://github.com/neovide/neovide/issues/2299#issuecomment-1907411309). My analysis there is slightly wrong, though`NtReadFile` allows `WaitForSingleObject` in certain cases, which I think should have been fulfilled, but I see no other possible reason for your crash,

@fredizzimo, I built from [this commit](https://github.com/KillTheMule/neovide/commit/ecf05df9fe27ae7d291df584ff78b2277713dec3) with Rust 1.76.0, and the Neovide crashed as before unfortunately. The call stack is the same as [#2299 (comment)](https://github.com/neovide/neovide/issues/2299#issuecomment-1907186068)

fredizzimo:
Thank you, I will merge this now to get more testing before we make a new release. 

It seems like it did not fix any of the hangs and crashes that we had, so I was a bit optimistic, but nevertheless removing the linebuffering seems like a good change. 

Let's continue discussing the hang issues in:
* https://github.com/neovide/neovide/issues/2182

And the Windows 10 crash in
* https://github.com/neovide/neovide/issues/2299

I still suspect they are the same issue, but let's keep them separate until proven otherwise.

Issue #1891, opened 2023-06-07
Make all GUI-related options configurable through a local config file


Hubro:
**Is your feature request related to a problem? Please describe.**

When connecting to remote Neovim instances using Neovide, it becomes an issue that Neovide can only be configured through Neovim itself. This is because when you have a remote Neovim instance running, it's *probably* because you have multiple computers you want to connect with, and those computers likely have different screen sizes, resolutions and refresh rates.

**Describe the solution you'd like**

There is an open PR (hopefully soon merged) for adding a config file specifically for Neovide, so settings can be set directly rather than through Neovim:

https://github.com/neovide/neovide/pull/1119

It would be great if all settings that directly affect Neovide was configurable using this config file.

Primarily I'm thinking of:

- `guifont`
- `neovide_refresh_rate`
- `neovide_refresh_rate_idle`
- `neovide_transparency`

**Describe alternatives you've considered**

I've considered writing a complex Neovim config that applies different Neovide settings based on the connected client, but that would require being able to identify which client is connected, which is currently not possible.

**Additional context**

\-

fredizzimo:
Sorry, I did not think properly when I marked it as a "good first issue".  Configuring those through the config file is not straightforward. All of those can be modified at runtime, so they are better handled through `init.vim` / `init.lua`.  None of those are a command line parameter either.

It might be possible to refactor them as command line / config file options, but then we get all kinds of problems of which settings have the priority and so on.

escwxyz:
> Sorry, I did not think properly when I marked it as a "good first issue". Configuring those through the config file is not straightforward. All of those can be modified at runtime, so they are better handled through `init.vim` / `init.lua`. None of those are a command line parameter either.
> 
> It might be possible to refactor them as command line / config file options, but then we get all kinds of problems of which settings have the priority and so on.

Regarding https://github.com/neovide/neovide/pull/2168, so how can I set `frame` to `transparent` without using the CLI? It seems that the settings inside `config.toml` got overriden.

PR #2343, opened 2024-02-06
feat: add option to disable window transparency

emonadeo:
Single line change that sets the window transparency based on the user configured transparency, instead of always setting it to true.

(Also removes a redundant `with_transparent` call)

On macOS this slightly changes the window's appearance for opaque backgrounds.

Comparison between Neovide, Kitty and this Pull Request on macOS

(The lack of the borders around the window is something that always has bothered me when using Neovide, because almost all macOS windows have it)

### What kind of change does this PR introduce?
- Feature

### Did this PR introduce a breaking change? 
- No

### DIFF
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -78,6 +78,10 @@ pub struct CmdLineSettings {
     #[arg(long = "no-idle", env = "NEOVIDE_IDLE", action = ArgAction::SetFalse, value_parser = FalseyValueParser::new())]
     pub idle: bool,
 
+    /// Disable transparent backgrounds. If enabled, ignores the `neovide_transparency` vim global
+    #[arg(long = "no-transparency", env = "NEOVIDE_NO_TRANSPARENCY", action = ArgAction::SetTrue, value_parser = FalseyValueParser::new())]
+    pub no_transparency: bool,
+
     /// Disable opening multiple files supplied in tabs (they're still buffers)
     #[arg(long = "no-tabs")]
     pub no_tabs: bool,
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -19,6 +19,7 @@ use winit::event::Event;
 
 use crate::{
     bridge::EditorMode,
+    cmd_line::CmdLineSettings,
     dimensions::Dimensions,
     editor::{Cursor, Style},
     profiling::{tracy_named_frame, tracy_zone},
@@ -100,6 +101,8 @@ pub struct Renderer {
     profiler: profiler::Profiler,
     os_scale_factor: f64,
     user_scale_factor: f64,
+
+    no_transparency: bool,
 }
 
 /// Results of processing the draw commands from the command channel.
@@ -111,6 +114,7 @@ pub struct DrawCommandResult {
 impl Renderer {
     pub fn new(os_scale_factor: f64, init_font_settings: Option<FontSettings>) -> Self {
         let window_settings = SETTINGS.get::<WindowSettings>();
+        let cmd_line_settings = SETTINGS.get::<CmdLineSettings>();
 
         let user_scale_factor = window_settings.scale_factor.into();
         let scale_factor = user_scale_factor * os_scale_factor;
@@ -124,6 +128,8 @@ impl Renderer {
 
         let profiler = profiler::Profiler::new(12.0);
 
+        let no_transparency = cmd_line_settings.no_transparency;
+
         Renderer {
             rendered_windows,
             cursor_renderer,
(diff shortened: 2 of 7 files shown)


fredizzimo:
Thanks,

I think this is a good idea, but I'm not sure that this actually works properly as it's implemented, since `g:neovide_transparency` can be set after the window is created. 

We also need to check how this works on the other platforms.

emonadeo:
Unfortunately setting the transparency after constructing the window [does not work on X11](https://docs.rs/winit/0.29.10/winit/window/struct.Window.html#method.set_transparent) and after testing it myself also doesn't work on macOS. Haven't tested on Windows or Wayland.

I see these options to solve this:

1. Reconstruct the window on every change.
2. Move neovide_transparency to a command line flag. (Breaking change)
   Related: #1891
   (I would even make an argument that all neovide configuration should happen outside of neovim, since having multiple places is confusing imo, but that's a different discussion)
3. Drop this PR (since it probably doesn't matter to most users anyways)

falcucci:
hm I'm not sure if I like it.

I mean, this change just means you want to toggle the transparency.

I would prefer a cmd arg such as `--enable-transparency` and avoid to mix what each setting values aims to be. 

`neovide_transparency` measure the opacity.

window builder settings shouldn't have their values mixed with runtime window settings.

emonadeo:
> hm I'm not sure if I like it.
> 
> 
> 
> I mean, this change just means you want to toggle the transparency.
> 
> 
> 
> I would prefer a cmd arg such as `--enable-transparency` and avoid to mix what each setting values aims to be.
> 
> 
> 
> `g:neovide_transparency` measure the opacity.
> 
> 
> 
> window builder settings shouldn't have their values mixed with runtime window settings. (with no good reason)

An additional cmd arg is a bad idea imho because it adds unnecessary complexity and allows for invalid states (e.g. `--transparent false` and `g:neovide_transparency < 1.0`), when realistically the boolean transparent can be derived from the opacity.

Also there is no clear line between window builder settings and runtime settings because some window settings (like transparent, haven't checked if there are others) are both depending on the OS.

falcucci:
I don't see any complexity, again you just want to toggle transparency and this is what it should do.

Actually the cmd arg should be `--disable-transparency` once all users have it enabled by default. If used, we should just do what it means and that's it, there is no conflict or invalid states.

The only trade-off to decide should be related to `--frame transparent` which must or not be applied to. 

Also, many users update `g:neovide_transparency` during runtime accordingly with back dropped window and necessities.

Rebuilding the window every time it changes isn`t a good option either.

emonadeo:
idk, I am always to add new settings to apps, but I really like the idea of having an inverse option. It's explicit and opt-in so it doesn't affect current users.

Looking at the existing command line options I named it `no-transparency` instead of `disable-transparency`.

It overrides (not overwrites) the `neovide_transparency` vim global and shows a warning whenever the user tries to set `neovide_transparency` to <1.0 while `no-transparency` is set.

fredizzimo:
Sorry for being late about this, but what about calling the setting `--border`? That would be a bit more descriptive and explain what it actually does. 

With the documentation explaining that transparency is not supported. We also need to determine which `--frame` options can be combined with this.

I would like to keep the possibility of supporting transparency with this option by perhaps drawing the border ourselves, or something else, and also support any possible unsupported `--frame` options in combination with this. Maybe at some point make it default and then have a corresponding `--no-border` setting to disable it. Perhaps other platforms could also be supported later.

emonadeo:
I looked into the source code of winit a little bit and it turns out that macOS _does_ support changing transparency after constructing the window, but it looks like this is a bug in winit.

A brief comparison:
[winit/platform_impl/macos/window.rs#494](https://docs.rs/winit/0.29.10/x86_64-apple-darwin/src/winit/platform_impl/macos/window.rs.html#494)
```rust
// set via `with_transparent` when initially constructing the window builder
if attrs.transparent {
    this.setOpaque(false);
    this.setBackgroundColor(&NSColor::clear());
}
```
Compared to when you call `set_transparent` after window construction:
[winit/platform_impl/macos/window.rs#587](https://docs.rs/winit/0.29.10/x86_64-apple-darwin/src/winit/platform_impl/macos/window.rs.html#587)
```rust
pub fn set_transparent(&self, transparent: bool) {
    self.setOpaque(!transparent)
}
```
It looks like it never changes the background color after window construction, so if the window is not initially transparent it defaults to a grey background color (`NSColor::windowBackgroundColor()`), which for some reason is also what causes the border.

Dropping this PR in favor of **TBD**, which achieves the border at runtime and without any command line options.

PR #2344, opened 2024-02-07
feat: add support for hiding window title from config file

### LINKED ISSUE #2342: `title-hidden = true` in config file has no effect
Setting `title-hidden = true` in the config file does not hide the title.
It works however when running neovide with `--title-hidden` flag.

### Config
`~/.config/neovide/config.toml`
```toml
frame = "buttonless"
title-hidden = true
```
### Command
```console
neovide
```
### Result
image

### Config
`~/.config/neovide/config.toml`
```toml
frame = "buttonless"
```
### Command
```console
neovide --title-hidden
```
### Result
image

## Expected behavior
The result of Exhibit A to match the result of Exhibit B

## Technical Details
 - OS: macOS 14.2.1
 - Neovide 0.12.2
 - NVIM 0.9.5

#### `~/.config/neovide/config.toml`
```toml
frame = "buttonless"
title-hidden = true

[font]
normal = "GeistMono Nerd Font Mono"
size = 16

[font.features]
"GeistMono Nerd Font Mono" = ["-liga"]
```

#### Neovide specific vim options
```
vim.g.neovide_cursor_vfx_mode = "railgun"
vim.g.neovide_cursor_vfx_particle_density = 14.0

vim.g.neovide_padding_top = 48
vim.g.neovide_padding_bottom = 48
vim.g.neovide_padding_right = 48
vim.g.neovide_padding_left = 48

vim.g.neovide_floating_blur_amount_x = 8.0
vim.g.neovide_floating_blur_amount_y = 8.0
vim.g.neovide_floating_shadow = false
```

### Log

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/14176398/neovide_rCURRENT.log)

falcucci:
- add a new field `title_hidden` to the `config` struct
- set the `neovide_title_hidden` environment variable based on the value of `title_hidden`

- Fix

- No

fixes #2342

### DIFF
--- a/src/settings/config.rs
+++ b/src/settings/config.rs
@@ -47,6 +47,7 @@ pub struct Config {
     pub frame: Option<Frame>,
     pub theme: Option<String>,
     pub font: Option<FontSettings>,
+    pub title_hidden: Option<bool>,
 }
 
 #[derive(Debug, Clone, PartialEq)]
@@ -100,6 +101,9 @@ impl Config {
         }
         if let Some(theme) = &self.theme {
             env::set_var("NEOVIDE_THEME", theme);
+        }
+        if let Some(title_hidden) = &self.title_hidden {
+            env::set_var("NEOVIDE_TITLE_HIDDEN", title_hidden.to_string());
         }
     }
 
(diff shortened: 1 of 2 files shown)


fredizzimo:
I think we should create some macro, so that we can just annotate the settings we want to be exposed. But that's something for another PR.

I'm also not sure about the priority here, to me it feels like environment variables should have a higher priority, so that you can have temporary overrides for a shell session. So that's also something to think about.

But this is good for now, merging, thank you!

PR #2345, opened 2024-02-07
fix: bug of restoring window position on macos.

### LINKED ISSUE #2303: On MacOS, Neovide window moves up after each relaunch
**Describe the bug**
On MacOS, the Neovide window opens at a slightly higher position after each relaunch.

**To Reproduce**
1. Launch Neovide, quit, relaunch
2. After each relaunch, the Neovide window is slightly above its previous position.

**Expected behavior**
The Neovide window should be at the same position after relaunch.

 - macOS Sonoma 14.3
 - Neovide Version 0.12.1
 - Neovim Version 0.9.5

crupest:
Fix #2303 

According to winit docs at https://docs.rs/winit/latest/winit/window/struct.WindowBuilder.html#method.with_position, `with_position` sets the position of top left corner of **content** rect instead of **window** rect on macOS, so a `set_outer_position` call is needed.

### DIFF
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -245,6 +245,11 @@ pub fn create_window(
     let gl_window = build_window(winit_window_builder, event_loop);
     let window = &gl_window.window;
 
+    #[cfg(target_os = "macos")]
+    if let Some(previous_position) = previous_position {
+        window.set_outer_position(previous_position);
+    }
+
     // Check that window is visible in some monitor, and reposition it if not.
     window.current_monitor().and_then(|current_monitor| {
         let monitor_position = current_monitor.position();


falcucci:
@crupest thank you for troubleshooting! but could you share some video/gif?

I am not able to reproduce.

fredizzimo:
It's strange that the function works differently on macOS, and then they have another function that makes it work the same...

I guess Winit did not want to break backwards compatibility or something.

falcucci (approved):

falcucci:
@fredizzimo Approved. that's what I have thought as well.

fredizzimo:
Thank you!

PR #2346, opened 2024-02-07
chore: update lru crate to fix building with the nightly toolchain

fredizzimo:
Updates the `lru` package from 0.7.5 to 0.12.2. 

The old package caused an old `hasbrown` version to be used, which in turn caused an old version of `ahash` to be used, which does not build on the current nightly version of the rust toolchain.

- No

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -35,17 +35,6 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "ahash"
-version = "0.7.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


fredizzimo:
We should find some way to keep our dependencies up to date, and at least try to prevent having multiple versions of the same dependencies in our lockfile. In this case we had duplicates of both `hashbrown` and `ahash`

fredizzimo:
Winit also pulls in a lot of unneeded stuff, wasm, some web stuff and and `android-activity` for example.

fredizzimo:
I will try to clean up some other duplicates as well in this PR, so putting it into draft for now

fredizzimo:
Let's ignore that, in many cases it does not seem to be possible, and I think we need some tooling for this rather than doing it by hand.

Issue #2122, opened 2023-11-10
Ghost text on macOS


Gutawer:
**Describe the bug**
The text that used to be in the window seems to get stuck in the background after changing the buffer in any way (i.e. scrolling) with transparency on:
<img width="1034" alt="image" src="https://github.com/neovide/neovide/assets/9727338/cf116e0c-be26-4bf3-afbe-607521713868">
The effect can be "reset" by going into the macOS desktop switcher, but this shows a different but similar-vibe visual bug:
<img width="1034" alt="image" src="https://github.com/neovide/neovide/assets/9727338/c38c63ee-4a01-46bc-9b2c-72847ad48c91">

**To Reproduce**
I used a very stripped down (one-line) Neovim config to make sure this wasn't more complicated:

```lua
vim.g.neovide_transparency = 0.5
```

Then I created a buffer with random text and scrolled it. I then went into the macOS desktop switcher (gesture is three fingers upwards on the mouse pad) while focused on the Neovide window and it shows a new stuck background.

**Expected behavior**
No stuck text.

**Screenshots**
Screenshots in description.

**Desktop (please complete the following information):**

- OS: macOS 12.7.1
- Neovide Version: latest git commit (installed via `cargo install --git`)
- Neovim Version: 0.9.4

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/13318120/neovide_rCURRENT.log)

**Additional context**
My macOS version is a bit old, could just be an incompatibility? If it doesn't happen on modern macOS then feel free to close and I'll work out upgrading.

fredizzimo:
See this issue https://github.com/neovide/neovide/issues/2094

PR #2351, opened 2024-02-09
feat(macos): display border for opaque backgrounds

emonadeo:
Reworked solution initially done in #2343. (See details there)

* moves the macOS specific shadow invalidation to the `macos` module
* additionally set background color and opaqueness of the window
* renames it to `update_transparency` to better reflect what it does
* use `icrate` instead of `cocoa` for AppKit bindings in line with winit (which uses icrate internally) and #2168

- Feature
- Refactor

- No

### DIFF
--- a/src/window/macos.rs
+++ b/src/window/macos.rs
@@ -66,11 +66,9 @@ impl MacosWindowFeature {
             _ => panic!("Not an appkit window."),
         };
 
-        if let Ok(color) = &SETTINGS
-            .get::<WindowSettings>()
-            .background_color
-            .parse::<Color>()
-        {
+        let window_settings = SETTINGS.get::<WindowSettings>();
+
+        if let Ok(color) = &window_settings.background_color.parse::<Color>() {
             error_msg!(concat!(
                 "neovide_background_color has now been deprecated. ",
                 "Use neovide_transparency instead if you want to get a transparent window titlebar. ",
@@ -117,12 +115,16 @@ impl MacosWindowFeature {
 
         let is_fullscreen = unsafe { ns_window.styleMask() } & NSWindowStyleMaskFullScreen != 0;
 
-        MacosWindowFeature {
+        let macos_window_feature = MacosWindowFeature {
             ns_window,
             titlebar_click_handler,
             extra_titlebar_height_in_pixel,
             is_fullscreen,
-        }
+        };
+
+        macos_window_feature.update_transparency(window_settings.transparency);
+
+        macos_window_feature
     }
 
     // Used to calculate the value of TITLEBAR_HEIGHT, aka, titlebar height in dpi-independent length.
@@ -172,6 +174,22 @@ impl MacosWindowFeature {
             0
         } else {
             self.extra_titlebar_height_in_pixel
+        }
+    }
+
+    /// Updates window properties affected by transparency
+    /// (shadow, background color and opaqueness)
+    pub fn update_transparency(&self, transparency: f32) {
+        let opaque = transparency >= 1.0;
+
+        unsafe {
+            let background_color = match opaque {
+                true => NSColor::windowBackgroundColor(),
+                false => NSColor::clearColor(),
+            };
+            self.ns_window.setBackgroundColor(Some(&background_color));
+            self.ns_window.setHasShadow(opaque);
+            self.ns_window.setOpaque(opaque);
         }
     }
 }
--- a/src/window/window_wrapper.rs
+++ b/src/window/window_wrapper.rs
@@ -19,7 +19,7 @@ use crate::{
 };
 
 #[cfg(target_os = "macos")]
-use crate::window::{invalidate_shadow, set_window_blurred};
+use crate::window::set_window_blurred;
 
 #[cfg(target_os = "macos")]
 use super::macos::MacosWindowFeature;
@@ -236,10 +236,11 @@ impl WinitWindowWrapper {
                 }
             }
             #[cfg(target_os = "macos")]
-            WindowSettingsChanged::Transparency(opacity) => {
-                log::info!("transparency changed to {}", opacity);
-                invalidate_shadow(self.windowed_context.window());
-                set_window_blurred(self.windowed_context.window(), opacity);
+            WindowSettingsChanged::Transparency(transparency) => {
+                log::info!("transparency changed to {}", transparency);
+                let window = self.windowed_context.window();
+                self.macos_feature.update_transparency(transparency);
+                set_window_blurred(window, transparency);
             }
             #[cfg(target_os = "macos")]
             WindowSettingsChanged::WindowBlurred(window_blurred) => {
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -9,9 +9,6 @@ mod window_wrapper;
 #[cfg(target_os = "macos")]
 mod macos;
 
-#[cfg(target_os = "macos")]
-use cocoa::base::id;
-
 #[cfg(target_os = "linux")]
 use std::env;
 
@@ -28,12 +25,6 @@ use winit::window::Window;
 
 #[cfg(target_os = "macos")]
 use winit::platform::macos::WindowBuilderExtMacOS;
-
-#[cfg(target_os = "macos")]
-use objc::{msg_send, sel, sel_impl};
-
-#[cfg(target_os = "macos")]
-use raw_window_handle::{HasRawWindowHandle, RawWindowHandle};
 
 #[cfg(target_os = "linux")]
 use winit::platform::wayland::WindowBuilderExtWayland;
@@ -140,27 +131,6 @@ pub fn set_window_blurred(window: &Window, opacity: f32) {
     let opaque = opacity >= 1.0;
 
     window.set_blur(window_blurred && !opaque);
-}
-
-/// Force macOS to clear shadow of transparent windows.
-#[cfg(target_os = "macos")]
-pub fn invalidate_shadow(window: &Window) {
-    use cocoa::base::NO;
-    use cocoa::base::YES;
-
-    let window_transparency = &SETTINGS.get::<WindowSettings>().transparency;
-    let opaque = *window_transparency >= 1.0;
-
-    let raw_window = match window.raw_window_handle() {
-        #[cfg(target_os = "macos")]
-        RawWindowHandle::AppKit(handle) => handle.ns_window as id,
-        _ => return,
-    };
-
-    let value = if opaque { YES } else { NO };
-    unsafe {
-        let _: id = msg_send![raw_window, setHasShadow: value];
-    }
 }
 
 pub fn create_window(
@@ -276,9 +246,6 @@ pub fn create_window(
 
     #[cfg(target_os = "macos")]
     set_window_blurred(window, SETTINGS.get::<WindowSettings>().transparency);
-
-    #[cfg(target_os = "macos")]
-    invalidate_shadow(window);
 
     gl_window
 }


fredizzimo:
@crupest, since you wrote some of the code that was modified, could you take a look at this? 

From my point of view this looks good.

crupest:
A great work! Looks good to me!

crupest (approved):

falcucci:
cool, gonna review tonight or tomorrow morning when I am back home

falcucci:
I have just reproduced the changes, but the border are still based on `g:neovide_transparency` values.

The border should be optional [here](https://github.com/neovide/neovide/pull/2351/files#diff-502f1d5dda478a0ff74b01587399520cb0db87937b69f5af1363c0731e95812fR186-R190) and disabled by default taking the `NSColor::windowBackgroundColor()` value.

As we have discussed at #2343 and taking part of @fredizzimo suggestion about changing it to a `--border` setting indeed sounds the perfect fit.

falcucci (changes requested):
I have just reproduced the changes, but the border are still based on `g:neovide_transparency` values only.

The border should be optional [here](https://github.com/neovide/neovide/pull/2351/files#diff-502f1d5dda478a0ff74b01587399520cb0db87937b69f5af1363c0731e95812fR186-R190) and disabled by default taking the `NSColor::windowBackgroundColor()` value.

As we have discussed at #2343 and taking part of @fredizzimo suggestion about changing it to a `--border` setting indeed sounds the perfect fit.

emonadeo:
While the border is the main motivation, it is just a side effect of setting the background color to the macos default. It's not guaranteed to stay this way with new macos releases.

```
if (--border is set)
    setBackgroundColor(windowBackgroundColor)
else
    setBackgroundColor(clearColor)
```
just feels wrong

What about introducing `--no-background`? Imo the default should be as close to native macos as possible, while giving the user options to turn it off. Might as well add a `--no-shadow` flag.

Alternatively, since this works at runtime this could also be vim globals instead.

With this in mind do you want me to stick to `--border` (or `--no-border)?

falcucci:
> Might as well add a --no-shadow flag.

Due to issues like https://github.com/neovide/neovide/issues/2122 we are forcing our users to disable the shadow o transparent mode only. Adding an option for that is asking for having open issues again, let's keep it like that since the majority users has no complains at all. (but this can be addressed to a proper moment)

About the border: you can continue using `g:neovide_transparency >= 1.0` AND the toggle option to enable it.

I don't see why it feels wrong, I don't see a reason of forcing white borders to users once we never had it before - we can take me as an example. I have a whole black window theme and a white border wouldn't fit my layout. This is the default for us since always.

Adding border option should be incremental to the end-user. if the user likes the border, we bring an option to add it.

emonadeo:
I just noticed that this overrides the deprecated `g:neovide_background_color` property, which also sets the background color when the window is created. Also, despite it being a vim global, it does not support runtime changes.

I now added the missing support for `g:neovide_background_color` runtime changes and also logic so it takes priority over `g:neovide_show_border`.

We could also remove `g:neovide_background_color` altogether, but then this would be a breaking change.

> Due to issues like https://github.com/neovide/neovide/issues/2094 https://github.com/neovide/neovide/issues/2122 we are forcing our users to disable the shadow on transparent mode only. Adding an option for that is asking for having open issues again, let's keep it like that since the majority users has no complains at all. (but this can be addressed to a proper moment)

Oh I am aware of this, I never intended `--no-shadow`/`g:neovide_no_shadow` to work for `g:neovide_transparency < 1.0`. Same with `--border`/`g:neovide_show_border`.

emonadeo:
So there is another quirk with the borders.

```rust
setBackgroundColor(something_non_transparent);
setHasShadow(true);
```
will always draw a border.

This luckily only affects the deprecated `g:neovide_background_color`.
I "fixed" this by disabling shadows, as this how it behaves in the current version of Neovide.

Apple is slowly making me lose my mind.

src/window/macos.rs
```diff
@@ -174,4 +165,53 @@ impl MacosWindowFeature {
...
+                true => NSColor::windowBackgroundColor(),
+                false => NSColor::clearColor(),
+            };
+            self.ns_window.setBackgroundColor(Some(&ns_background));
+            self.ns_window.setHasShadow(opaque);
+            // Setting the window to opaque upon creation shows a permanent subtle grey border on the top edge of the window
+            self.ns_window.setOpaque(opaque && show_border);
+            self.ns_window.invalidateShadow();
+        }
+    }
```
falcucci, line 216: honestly, this function can be improved. It's getting too long and repetitive due to the deprecated background behavior. we should probably delete the deprecated code asap. not for now.

I think we can improve the legibility creating self descriptive code.

suggestion:


```rust
/// Print a deprecation warning for `neovide_background_color`
pub fn display_deprecation_warning(&self) {
    error_msg!(concat!(
        "neovide_background_color has now been deprecated. ",
        "Use neovide_transparency instead if you want to get a transparent window titlebar. ",
        "Please check https://neovide.dev/configuration.html#background-color-deprecated-currently-macos-only for more information.",
    ));
}

#[deprecated(
    since = "0.12.2",
    note = "This function will be removed in the future."
)]
pub fn handle_legacy_background(
    &self,
    color: Color,
    show_border: bool,
    ignore_deprecation_warning: bool,
) {
    if !ignore_deprecation_warning {
        self.display_deprecation_warning();
    }
    let [red, green, blue, alpha] = color.to_array();
    unsafe {
        let opaque = alpha >= 1.0;
        let ns_background = NSColor::colorWithSRGBRed_green_blue_alpha(red, green, blue, alpha);
        self.ns_window.setBackgroundColor(Some(&ns_background));
        // If the shadow is enabled and the background color is not transparent, the window will have a grey border
        // Workaround: Disable shadow when the background color is not transparent
        self.set_window_properties(opaque, show_border);
    }
}

/// Set background color, shadow and transparency/opaqueness of the window
pub fn handle_background(&self, transparency: f32, show_border: bool) {
    unsafe {
        let opaque = transparency >= 1.0;
        // Setting the background color to `NSColor::windowBackgroundColor()`
        // makes the background opaque and draws a grey border around the window
        let ns_background = match opaque && show_border {
            true => NSColor::windowBackgroundColor(),
            false => NSColor::clearColor(),
        };
        self.ns_window.setBackgroundColor(Some(&ns_background));
        self.set_window_properties(opaque, show_border);
    }
}

/// Sets opacity and shadow properties of a window.
///
/// If `opaque` is true, the window will have a shadow and may show a grey top border.
///
/// Please note, creating the window with `opaque` will result in a subtle grey border at the top.
///
/// # Arguments
///
/// * `window` - A mutable reference to the window object
/// * `opaque` - Flag to decide if window should have shadow and potential border
/// * `show_border` - Only if `opaque == true`, this decides if a subtle grey border should be shown
pub fn set_window_properties(&self, opaque: bool, show_border: bool) {
    unsafe {
        self.ns_window.setHasShadow(opaque);
        self.ns_window.setOpaque(opaque && show_border);
        self.ns_window.invalidateShadow();
    }
}

/// Set background color, shadow and transparency/opaqueness of the window
pub fn set_background(
    &self,
    transparency: f32,
    show_border: bool,
    background_color: String,
    ignore_deprecation_warning: bool,
) {
    match background_color.parse::<Color>() {
        Ok(color) => {
            self.handle_legacy_background(color, show_border, ignore_deprecation_warning)
        }
        _ => self.handle_background(transparency, show_border),
    }
}
```

website/docs/configuration.md
```diff
@@ -269,6 +269,24 @@ Lua:
...
+let g:neovide_show_border = v:true
+```
+
+Lua:
+
+```lua
+vim.g.neovide_show_border = true
+```
+
+Draw a grey border around the window, as seen on most native macOS apps.
```
falcucci, line 286: just describe what it does and the conditions it should apply.

for example the opacity rule.

something like:

```suggestion
Draw a grey border around opaque windows only.
```

falcucci (changes requested):

src/window/macos.rs
emonadeo: I think it makes it harder to read because now there are three layers of function calls to set the background.
Splitting responsibility into `handle_background` and `handle_legacy_background` is a good idea but `set_window_properties` would only work for `setOpaque` and `invalidateShadow`.
The `setHasShadow` call is different for legacy and non-legacy:
```rust
// Legacy:
// If the shadow is enabled and the background color is not transparent, the window will have a grey border
// Workaround: Disable shadow when the background color is not transparent
self.ns_window.setHasShadow(opaque && show_border);
```
```rust
// Non-Legacy:
self.ns_window.setHasShadow(opaque);
```
At this point I don't think it's worth it to outsource it into a new function.
(See my new commit for the full example and let me know if you think that's good)

Also two things I don't think I quite understand:

>```rust
>/// If `opaque` is true, the window will have a shadow and may show a grey top border.
>///
>/// Please note, creating the window with `opaque` will result in a subtle grey border at the top.
>```
Aren't these two the same thing except for „creating the window with“?

>```rust
>/// # Arguments
>///
>/// * `window` - A mutable reference to the window object
>```
I don't think there is a mutable window reference argument.

falcucci: with your newest commits it looks already better.

I wasn't precise about the suggestions once it was just an idea to give you a direction.

It makes sense to refactor better later to remove th deprecated code.

it was a take it or leave it.

thank you!

emonadeo: ah okay, thanks a lot for the terrific help!

falcucci (approved):
@emonadeo great work! approved, thank you

src/window/window_wrapper.rs
```diff
@@ -236,10 +236,44 @@ impl WinitWindowWrapper {
...
+                let show_border = window_settings.show_border;
+                let background_color = window_settings.background_color;
+                self.macos_feature.set_background(
+                    transparency,
+                    show_border,
+                    background_color,
+                    true,
+                );
+                set_window_blurred(self.windowed_context.window(), transparency);
             }
```
fredizzimo, line 277: Maybe we should add `macos_feature.handle_windows_settings` and move all these there (and also the WindowBlurred function?

I think that would be cleaner than adding `#[cfg(target_os = "macos")]` to each match.

And perhaps also read the window settings inside the functions themselves, the match is easier to read if it's kept short.

falcucci: I particularly finds the older version cleaner to read, even if it has little more lines.

Isolated behavior for each setting, easier to debug, `log.info` has been removed... or maybe I am wrong.

It feels like overengineering our code trying to be as more generic as possible where we don't need to? or is just not the right moment?

A mix of feelings.

emonadeo: Yeah, matching `changed_setting` twice (once for all platforms and once for macOS) isn't super pretty. But it has two advantages:
1. macOS specific config updates are now all handled in `macos.rs`
2. shared logic for all platforms while still allowing additional platform specific updates/overrides

About 1.: I initially loved the idea of having all macOS specific window settings inside `macos.rs` but thinking about it again there is something nice about not having to jump between files and functions to read the code for what is arguably the same concern (i.e. observing window setting changes). It probably isn't more idiomatic than just doing `#[cfg(target_os = "macos")]`. Reverted this in 22b88b3.

And yes, I did remove `log.info` (oops) as it all went through the same branch now:
```rust
match changed_setting {
    WindowSettingsChanged::BackgroundColor(_)
    | WindowSettingsChanged::ShowBorder(_)
    | WindowSettingsChanged::Transparency(_)
    | WindowSettingsChanged::WindowBlurred(_) => {
        self.update_background(window, ignore_deprecation_warning);
    }
    _ => {}
}
```

falcucci: usually I use this kind of multiple match pattern for testing purposes or decoding many similar range of types that has similar ends. 

for our use case I much prefer the reverted code you had just done.

LGTM as it is now.

fredizzimo: The issue with mixing code from different platforms is more a maintainability issue, than a readability one here.

Most people don't have the ability to compile and test on macOS, so it's easy to refactor some code and forget to change the macOS parts. That has happened both to me and others before, so it's a real problem. Now, isolating the platform parts into their own files don't make us immune to the problem, but the breakages get more limited, and apply to only the API that the platform specific code use. But you are more free to refactor the rest of the code, split functions, rename local variables and so on.

It also causes problems with warnings for unused variables, unused use declarations and so on. 

That said, this latest iteration is far better than the original code, mostly due to moving the blur code away to where it belongs.  But also, by removing the noise from the match statements itself. So, I can accept https://github.com/neovide/neovide/pull/2351/commits/22b88b3e7a5c2027d715d849d38ac19e413033c0, but I personally prefer the version before that, also for the reason that the Window wrapper itself have no reason to deal with those events at all, they are completely unrelated to the functionality of it.

But it can become a readability issue as well, here it only works because, we are only special casing macOS here, if more platforms need special cases, especially for the same settings, it will quickly become a mess. Look at https://github.com/neovide/neovide/blob/main/src/renderer/vsync/mod.rs, I have written that myself, but it's a mess. This one from https://github.com/neovide/neovide/pull/2359, is very slightly better, but only works because the code is really simple https://github.com/neovide/neovide/blob/8c2e80c9ff653bc56224a29d14bfc6a854ef7afe/src/renderer/vsync/mod.rs

falcucci: why we just don't avoid the multiple match pattern and just move it as it is to `macOS.rs` then?

I agree about the mess that it can become using only `#[cfg]`, definitely.

maybe this would be better for everyone.

and jesus @fredizzimo that makes my eyes hurt. I guess your color scheme should handle it as ghost texted.

CleanShot 2024-02-23 at 01 01 24@2x

emonadeo: To put my two cents in, I think we should start a separate issue on this.

I briefly looked into it but this looks to be a large scale refactor.
For macos alone migrating to a separate module is big, due to the hierarchy between window creation, the window wrapper, the main loop, the macos window wrapper and how all of them currently use `#[cfg]`.

It might be worth to look at how [winit](https://github.com/rust-windowing/winit/blob/master/src/platform_impl/mod.rs) and [gpui](https://github.com/zed-industries/zed/tree/main/crates/gpui) do this.

For now, what is the consensus here? Move back to `macos.rs` or leave as-is?

falcucci: in case of moving to macos.rs I would avoid the multiple match pattern to keep the log.info for each setting changes for example and to split responsibilities of each event once this is interactive.

besides that, I have nothing else to add/share.

fredizzimo: A separate call is fine
`macos_window_feature.handle_settings_changed(changed_setting)`

That way the macos code csn independently decide which settings to handle.

fredizzimo: And if you rename it to something more generic, we could use the same pattern as winit, and it would become for example `platform_window.handle_settings_changed`

fredizzimo:
Thank you!

PR #2356, opened 2024-02-10
doc: instructions of profiling with tracy.

crupest:
Add documentation of instructions of profiling neovide with tracy.

This is the first time I write docs for public. Please point out any spelling or grammar mistakes or inappropriate voice, etc.

### DIFF
--- a/website/docs/troubleshooting.md
+++ b/website/docs/troubleshooting.md
@@ -36,3 +36,63 @@
     mixed refresh rate out of the box), if so,that's because X11 does not support mixed refresh
     rate well. You may be able to fix this through your compositor or by switching to wayland.
     As a temporary work around, you may set `g:neovide_refresh_rate` to the lower value.
+
+## Performance Profiling
+
+If you encounter a performance problem like frame rate stuttering, besides attaching a log file
+when reporting bugs, [tracy](https://github.com/wolfpld/tracy) profiling data will also be very
+useful and can usually help developers to troubleshoot the bug much faster. Here is how you can
+collect tracy data.
+
+1. *Install tracy.* Windows users can download it at
+[its GitHub release page](https://github.com/wolfpld/tracy/releases). Linux and macOS users can
+install it with package manager. Otherwise, you may have to build it yourself following tracy
+docs.
+
+2. *Build a profiling version of Neovide.* Follow
+[the installation page](https://neovide.dev/installation.html) to install all required
+dependencies and Rust SDK. Download or clone
+[source code of Neovide](https://github.com/neovide/neovide). Build it with following commands.
+Note that you need to specify **both** `--profile profiling` and `--features profiling`, so that
+Neovide is built for a profiling version. Or, you can skip these commands, and let `cargo run`
+in step 5 build it automatically before running.
+
+    ```sh
+    cd [neovide-source-dir]
+    cargo build --profile profiling --features profiling
+    ```
+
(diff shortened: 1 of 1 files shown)


website/docs/troubleshooting.md
```diff
@@ -36,3 +36,59 @@
...
+reporting bugs, [tracy](https://github.com/wolfpld/tracy) profiling data will also be very useful and
+can usually help developers to troubleshoot the bug much faster. Here is how you can collect tracy data.
+
+1. Install tracy. Windows users can download it at
+[its GitHub release page](https://github.com/wolfpld/tracy/releases). Linux and macOS users can install
+it with package manager. Otherwise, you may have to build it yourself following tracy docs.
+
+2. Install [Rust SDK](https://www.rust-lang.org).
+
+3. Download or clone [source code of Neovide](https://github.com/neovide/neovide). Build it with
```
fredizzimo, line 52: I think we can remove the build instructions here. Since `cargo run` will automatically build, and it simplifies the steps needed.

website/docs/troubleshooting.md
```diff
@@ -36,3 +36,59 @@
...
+
+If you encounter a performance problem like frame rate stuttering, besides attaching a log file when
+reporting bugs, [tracy](https://github.com/wolfpld/tracy) profiling data will also be very useful and
+can usually help developers to troubleshoot the bug much faster. Here is how you can collect tracy data.
+
+1. Install tracy. Windows users can download it at
+[its GitHub release page](https://github.com/wolfpld/tracy/releases). Linux and macOS users can install
+it with package manager. Otherwise, you may have to build it yourself following tracy docs.
+
+2. Install [Rust SDK](https://www.rust-lang.org).
```
fredizzimo, line 50: I think we should also refer to this page https://neovide.dev/installation.html#installation, for more prerequisites for building from source

fredizzimo (commented):
Thank you, this is great, I have been thinking about writing this for a long time, so it's easier to ask for more information about performance releated problems.

website/docs/troubleshooting.md
crupest: I did consider about this. But I worry that people wait for a long time of building after starting tracy and before starting neovide, which might be some kind wired (I think). So I instruct people to build Neovide here, so that they can first ensure Neovide is built successfully, and then just run a good Neovide after starting tracy.

I think I can add a note about this. And I don't think an extra step can bother too much.

website/docs/troubleshooting.md
crupest: Done!

crupest:
> Thank you, this is great, I have been thinking about writing this for a long time, so it's easier to ask for more information about performance releated problems.

That's why I write this. I have seen so many people report performance issue like FPS stuttering. But without some trace log, we can only do very limit things. So this should be a good start!

website/docs/troubleshooting.md
crupest: Emmm, now I think there are a lot of steps. It might stop people from doing this. I think I can squash some steps to make it neater.

crupest: I have merged some steps and now it should look neater and more structural.

fredizzimo:
Thank you!. This looks good, we can improve later, if it looks like people have problems following the instructions.

falcucci:
this is great indeed!

Issue #661, opened 2021-05-16
[Investigation] Loading eglGetCurrentDisplay function pointer causes segfault


j4qfrost:
While debugging the opengl branch, I discovered that it was possible to simply bypass the eglGetCurrentDisplay function to get the branch to run. While this is currently a workable solution, we may run into issues down the road.

From further investigation, I have discovered that glutin loads a glx context by default as opposed to an egl context when using X11. https://github.com/rust-windowing/glutin/blob/master/glutin/src/platform_impl/unix/x11.rs

In order to validate that the context loaded is the main cause we would need to test the function loading on a machine that uses Wayland, since glutin's implementation for Wayland defaults to egl. https://github.com/rust-windowing/glutin/blob/master/glutin/src/platform_impl/unix/wayland.rs

After the issue is confirmed, another issue should be raised with glutin to discuss possible implementation changes, or whether was should simply continue on ignoring eglGetCurrentDisplay for x11 systems. An issue should also be raised with https://github.com/rust-skia/rust-skia, since there is no test or example that catches this issue.

One curious problem that may point away from my theory that this is caused by the loading of the wrong context is that 
eglQueryString still loads fine, so the correct context may be loaded, but the issue just lies somewhere else.

j4qfrost:
TL;DR help wanted from Wayland users. Please test.

p00f:
What exactly should I do to test?

j4qfrost:
@p00f you can either use an older version of the opengl branch from before the egl patch was added or comment out the if statement for eglGetCurrentDisplay

p00f:
I set a breakpoint here:

https://github.com/Kethku/neovide/blob/3ba98d05d113c81412be7a6f52bfed5616c21075/src/window/window_wrapper/renderer.rs#L54

On stepping over, it jumped directly to:

https://github.com/Kethku/neovide/blob/3ba98d05d113c81412be7a6f52bfed5616c21075/src/window/window_wrapper/renderer.rs#L57

So I think it isn't loaded

p00f:
uh huh this happens even when i run it under xwayland

p00f:
I added debug statements
diff --git a/src/window/window_wrapper/renderer.rs b/src/window/window_wrapper/renderer.rs
index 9d2a8aa..78f4c84 100644
--- a/src/window/window_wrapper/renderer.rs
+++ b/src/window/window_wrapper/renderer.rs
@@ -1,3 +1,4 @@
+use log::debug;
[...]

Issue #491, opened 2021-03-07
OpenGL Renderer


Kethku:
I recently swapped to an OpenGL backend away from the vulkan one with the hope that it would increase linux compatibility. However it seems that some folks are having issues.

I don't have the tools or know-how required to fix these problems on my own.

The opengl binding is handled via https://github.com/rust-windowing/glutin and is done here: https://github.com/Kethku/neovide/blob/main/src/window/window_wrapper/renderer.rs which was based in part on alacritty's implementation https://github.com/alacritty/alacritty/tree/master/alacritty/src/display and the sample in skia-safe https://github.com/rust-skia/rust-skia/blob/master/skia-safe/examples/gl-window/main.rs.

If anybody is able to synthesize something that works more consistently that would be amazing.

NOTE: I reverted the main change as it wasn't ready for prime time like I thought. The code is currently held in the OpenGL branch

kiuKisas:
```
❯ ./neovide
Ignored client type property: "methods"
Ignored client type property: "attributes"
fish: “./neovide” terminated by signal SIGSEGV (Address boundary error
```

calvinkosmatka:
Building from 4aec0d5

I was able to at least get a window open on my old X201T, Ubuntu 20.04, sway wm, wayland-only with a little massaging.

1. A couple dependency issues - `typeface-open-stream` was merged upstream, so I had to change the `skia-safe` url to `git = "https://github.com/rust-skia/rust-skia"`. `flexi_logger` was causing a version conflix for `regex` so I switched it to `flexi_logger = { version = "0.15", default-features = false }`.
2. I didn't have the consolas font, so had to switch the fallback to a font I had installed (not really an issue)
3. Had to comment out `.with_double_buffer(Some(true))` at `src/window/window_wrapper/mod.rs` L425 to get a window to open, otherise it would panic with the message `thread 'main' panicked at 'called 'Result::unwrap()' on an 'Err' value: NoAvailablePixelFormat', src/window/window_wrapper/mod.rs:430:10`. This might just be a quirk of my hardware, but others might hit this as well. Also tried running with `LIBGL_ALWAYS_SOFTWARE=1` (which I need to make Alacritty run on this machine), made no difference.
4. I can't type most special characters, only `@` and `:`. Looks like the issue is that glutin WindowEvent::KeyboardInput events already take the modifier state into account and they don't have virtual_keycodes for most special characters.

Kethku:
[...]

PR #2359, opened 2024-02-14
chore: Support multiple render backends

fredizzimo:
This is mostly a refactoring in order to support multiple render backends, like D3D, and makes the actual PR that implements the D3D backend much smaller.

* WindowContext has been renamed to `SkiaRenderer`, which can have different implementations depending on the platform
* The Gpu profiling system now supports multiple backends, and code has been moved to a dedicated `opengl` implementation
* `VSyncWin` has been renamed to `VSyncWinDwm` to allow a swapchain implementation for D3D.
* The latency reduction optimization has been removed, as it is incompatible with some render backends.

- No


src/renderer/opengl.rs
```diff
@@ -16,63 +17,202 @@ use glutin::{
...
 
-pub struct Context {
-    surface: Surface<WindowSurface>,
+use crate::{profiling::tracy_gpu_zone, window::UserEvent};
+
+#[cfg(feature = "gpu_profiling")]
+use crate::profiling::{opengl::create_opengl_gpu_context, GpuCtx};
+
+pub struct OpenGLSkiaRenderer {
+    // NOTE: The destruction order is important, so don't re-arrange
```
MultisampledNight, line 48: Maybe a `SAFETY` could fit better in order to convey that a critical invariant is upheld here.

src/renderer/opengl.rs
```diff
@@ -16,63 +17,202 @@ use glutin::{
...
+        let raw_window_handle = window.raw_window_handle();
+
+        let size = clamp_render_buffer_size(&window.inner_size());
+
+        let surface_attributes = SurfaceAttributesBuilder::<WindowSurface>::new()
+            .with_srgb(Some(srgb))
+            .build(
+                raw_window_handle,
+                NonZeroU32::new(size.width).unwrap(),
+                NonZeroU32::new(size.height).unwrap(),
```
MultisampledNight, line 89: Those unwraps might very well blow up if starting in e.g. an initially minimized state, need to try that out.

src/renderer/opengl.rs
```diff
@@ -16,63 +17,202 @@ use glutin::{
...
+            window_surface.set_swap_interval(&context, SwapInterval::DontWait)
+        };
 
-    pub fn resize(&self, width: NonZeroU32, height: NonZeroU32) {
-        GlSurface::resize(&self.surface, &self.context, width, height)
+        gl::load_with(|s| get_proc_address(&window_surface, CString::new(s).unwrap().as_c_str()));
+
+        let interface = skia_safe::gpu::gl::Interface::new_load_with(|name| {
+            if name == "eglGetCurrentDisplay" {
+                return std::ptr::null();
```
MultisampledNight, line 118: Why is `eglGetCurrentDisplay` filtered out? Can't we let `get_proc_address` also find `eglGetCurrentDisplay`? (possibly this is a peculiarity of EGL, I've never used EGL)

MultisampledNight (commented):
Nothing too interesting code-wise, typical noisy GL setup. Haven't tested yet.

src/renderer/vsync/mod.rs
```diff
@@ -4,59 +4,43 @@ mod macos_display_link;
...
 
 use crate::{
-    renderer::WindowedContext, settings::SETTINGS, window::UserEvent, window::WindowSettings,
+    renderer::SkiaRenderer, settings::SETTINGS, window::UserEvent, window::WindowSettings,
 };
-use winit::event_loop::EventLoopProxy;
+use winit::{event_loop::EventLoopProxy, window::Window};
 
 #[cfg(target_os = "linux")]
 use std::env;
```
MultisampledNight, line 17: Fwiw due to the moved detection logic, this import is actually unused and yields a warning

```rust
warning: unused import: `std::env`
  --> src/renderer/vsync/mod.rs:17:5
   |
17 | use std::env;
   |     ^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
```

MultisampledNight:
Test results:

- On native wayland, everything's fine and smooth as ever.
- On Xwayland, there's a few freezes the first few seconds after starting up, where interestingly the view keeps updating (the profiler keeps moving and displaying), but the entire text contents are not responding to any keypresses for a few seconds. Maybe this doesn't come from this PR though, I haven't tested main in a while.

MultisampledNight:
After retrying a few times between main and this branch on Xwayland, I can't seem to reproduce what I found initially, and found no meaningful differences. Ignore that then.

src/renderer/opengl.rs
fredizzimo: I have no idea, it was introduced here 
* https://github.com/neovide/neovide/pull/631

After some discussion here
* https://github.com/neovide/neovide/issues/491#issuecomment-830762584
* https://github.com/neovide/neovide/issues/661

I found no issues in the glutin repository.

src/renderer/opengl.rs
fredizzimo: The code is just moved, so it has been the same before. `clamp_render_buffer_size` should ensure that it's not zero.

fredizzimo:
I'm not sure about the XWayland freeze, I can't think of any changes that could cause it, except this one https://github.com/neovide/neovide/pull/2359/commits/110ea9957a003f1fe85eec53542be0e19474da97, but that should not affect x11 or xwayland at all. 

The rest of the code should have stayed functionally the same.

src/renderer/opengl.rs
fredizzimo: I'll keep the `NOTE:` the syntax highlighter highligts it, but not `SAFETY:`

src/renderer/vsync/mod.rs
fredizzimo: That is now fixed. I was doing most of my testing on Windows, and that was part of one late change that never got properly tested on Linux.

PR #2379, opened 2024-02-23
fix: Only blur bottom window

TwIStOy:
Certain components of `nui.nvim` are displayed through an overlay of multiple floating windows. It suffices to blur only the bottommost window.

- Fix

- No


falcucci:
I am not sure if I can help, if there is a way to simulate this without using `nui.nvim` I would be on it.

fredizzimo:
I think this is a bit too risky to include as it is, since you might have overlapping floating windows on top of each other, where all of the covered windows should blur IMO. 

Do you have an example screenshot where it looks wrong without this patch?

We should probably request a new window setting from Neovim, so that plugin authors can set the blur to false when needed.

TwIStOy:
The `popup` component of `nui.nvim` supports adding padding both vertically and horizontally. This functionality is achieved by layering two floating windows together. When displaying hover documentation, `noice.nvim` utilizes the popup component from `nui.nvim`, resulting in a similar effect.

CleanShot 2024-03-02 at 22 05 39

MultisampledNight (approved):

Kethku:
I could go either way on this. I think the best case solution for this feature would be to add a way to track the blur window by window (tracking window options is a feature that we don't have right now but would unblock a large number of desirable features).

That said a large number of extensions use this multiple window trick for window borders. I personally run into the double blur when using floaterm for example.

I think merging this for now seems fine, but given that @fredizzimo pushed back, I'm going to wait for his ok

Issue #1783, opened 2023-03-01
Anyway to avoid the startup zoom out animation?


AllenDang:
**Is your feature request related to a problem? Please describe.**
In macOS, each time when I launch neovide, the window will appear with a small size and zoom out to fullscreen, I did add the --maximized, still can see the zoom out process.

**Describe the solution you'd like**
Launch neovide directly with fullscreen without a explicit zoom out.

**Additional context**
I'm using newest master branch.

yavorski:
This is happening on `windows-11` too. 

Also it is causing issues with `lualine tabline` for reason, because in the moment of starting `vim.o.columns` is equal to  `100` and `lualine tabline` is using this for its configuration, which leads to tabline not being expanded to full width.

Workaround is to start neovide with explicit size flag e.g. `neovide --size 1900x968` for example, this is working fine and then to maximize - which is a bit tedious of course.

[Config `max_length` for tabline width doc](https://github.com/nvim-lualine/lualine.nvim/blob/7d131a8d3ba5016229e8a1d08bf8782acea98852/doc/lualine.txt#L532)

In my config i just set it to `max_length = vim.o.colums` to take the full width, but it is not working in `neovide` when maximized and started with `maximized` flag.

falcucci:
@AllenDang @yavorski unable to reproduce it, could any of you send a video/gif for better comprehension?

yavorski:
@falcucci 

https://github.com/neovide/neovide/assets/3436517/981159fa-5a9f-4b10-ac28-3e4f74d39ccf

yavorski:
That is indeed strange, yes it does not play in Firefox for some reason, don't know why. 
Maybe try different browser, if that is the case.

falcucci:
Ive handle to see it using chrome tho, thank you!

I just can't reproduce it, or it opens so fast that maybe isn't even eye perceptive here.

but I guess it's taking the default inner size position grid and then maximizing it. For some it perceptible for others not that much.

falcucci:
could some of you test this candidate fix? https://github.com/neovide/neovide/pull/2389

yavorski:
@falcucci hi, I have built it with `cargo build --release` unfortunately both issues still persists, there is no difference. at least on `Windows` os.

falcucci:
did you try using --maximized?

this change only affects if you use it for now.

yavorski:
[...]

PR #2389, opened 2024-03-01
fix: winit window builder inner size conflict request

falcucci:
- Fix

- No


falcucci:
@fredizzimo take a look at this please.

I think that we are conflicting some window size requests in the window builder.

It probably doesn't bother the majority of users because it should mount the window super fast, but some of them seems to be perceptive at https://github.com/neovide/neovide/issues/1783 causing a "delay" feeling to maximize the window in case of the window was not fullscreen persisted.

PR #2397, opened 2024-03-04
Add ability to use flatpak-spawn to call nvim when running from a flatpak container

sqwxl:
Add ability to use flatpak-spawn to call nvim when running from a flatpak container.

I've also opened a pr for flathub that points to my own build for now: https://github.com/flathub/flathub/pull/5017

- Feature

- No

### DIFF
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -40,7 +40,13 @@ fn build_nvim_cmd() -> Result<TokioCommand> {
         if neovim_ok(&path, &[])? {
             return Ok(build_nvim_cmd_with_args(path, vec![]));
         }
+    } else if should_use_flatpak_spawn() {
+        return Ok(build_nvim_cmd_with_args(
+            "flatpak-spawn".to_string(),
+            vec!["--host".to_string(), "nvim".to_string()],
+        ));
     }
+
     bail!("ERROR: nvim not found!")
 }
 
@@ -203,4 +209,30 @@ fn build_nvim_cmd_with_args(bin: String, mut args: Vec<String>) -> TokioCommand 
     args.push("--embed".to_string());
     args.extend(SETTINGS.get::<CmdLineSettings>().neovim_args);
     nvim_cmd_impl(bin, args)
+}
+
+// Adapted from https://github.com/bugadani/lapce/blob/0c0a37ab1ceda829ba9691149038a1e7f282cb6d/lapce-proxy/src/terminal.rs#L378
+#[cfg(target_os = "linux")]
+fn should_use_flatpak_spawn() -> bool {
+    use std::path::Path;
+
+    const FLATPAK_INFO_PATH: &str = "/.flatpak-info";
+
+    // The de-facto way of checking whether one is inside of a Flatpak container is by checking for
+    // the presence of /.flatpak-info in the filesystem
+    if !Path::new(FLATPAK_INFO_PATH).exists() {
+        return false;
+    }
+
+    // Ensure flatpak-spawn --host can execute a basic command
+    StdCommand::new("flatpak-spawn")
+        .arg("--host")
+        .arg("true")
+        .status()
+        .map(|status| status.success())
+        .unwrap_or(false)
+}
+#[cfg(not(target_os = "linux"))]
+fn should_use_flatpak_spawn() -> bool {
+    return false;
 }


sqwxl:
Hi @fredizzimo, thanks for your reply.

I'd love to contribute to making flatpak-ed neovide a reality. As you can see from the referenced Flathub PR, they are asking for lot's of changes. Most I can do myself, others will need Neovide maintainers. If you can help with the application to get approval for launching binaries on the host, that would be great.

I've been playing around with my own build for the past couple days and it's been pretty nice. Here are some of my observations:

1. Neovide runs really well from within the flatpak container. It did lock up on me 2 or 3 times, but I haven't investigated why that happened and I don't know if it's a issue with "native" neovide as well.
1. Allowing neovide to use `flatpak-spawn` to run the host's `nvim` binary is key. The alternative -- i.e. bundling `nvim` within the container -- is severely crippling and barely usable. In this scenario, not only is the sand-boxed `nvim` not configured like the host's, it also doesn't have access to any other binaries a user would expect (git, node, python, rg, etc).
2. I couldn't copy and paste to and from the system clipboard. I'm sure there's a way to unlock this, probably at the application level, like with `flatpak-spawn`, but I haven't looked into it yet.

Going forward, perhaps it would make sense for me to open a parent issue for tracking the tasks involved in getting neovide published to Flathub. This issue would become one of its subtasks. What do you think? I'd be happy to get the ball rolling.

sqwxl:
**Update**: Following some [feedback](https://github.com/flathub/flathub/pull/5017#issuecomment-1989781021) from a Flathub maintainer, it was suggested that the flatpak Neovide should still work when the host doesn't have neovim installed (or it just isn't accessible from the container for some reason).
The expected behavior, imho, should be to fallback to a bundled version of neovim when the host isn't available. I'll implement the changes in this branch.

fredizzimo:
@sqwxl, 

I will look at the other issues a bit later, but just quick comment for now

Notice that you will need some other dependencies as well, at least some of the provider ones:
https://neovim.io/doc/user/provider.html

Without at least the clipboard integration, copy and paste from the system clipboard won't work. I think that should be the bare minimum. 

I think we also need to update the documentation, with info that running shell commands, the built-in terminal and LSP integration won't work without an externally installed nvim, since the bundled nvim won't know to use flatpak-spawn for those.

sqwxl:
@fredizzimo I already added a clipboard provider in the latest manifest which you can try out by following by installing the test build: `flatpak install --user https://dl.flathub.org/build-repo/89216/dev.neovide.Neovide.flatpakref`

As for the other provider they seem to work out of the box (although that obviously won't be the case with the sandboxed neovim)
Screenshot from 2024-03-12 14-05-15

MultisampledNight (approved):

Kethku:
> **Update**: Following some [feedback](https://github.com/flathub/flathub/pull/5017#issuecomment-1989781021) from a Flathub maintainer, it was suggested that the flatpak Neovide should still work when the host doesn't have neovim installed (or it just isn't accessible from the container for some reason). The expected behavior, imho, should be to fallback to a bundled version of neovim when the host isn't available. I'll implement the changes in this branch.

Is a bundled copy of neovim correct here? That would be 120mb of data that for most uses doesn't actually get run and also includes many small files that can slow down the unpacking/unzipping steps that I presume flatpak uses. I don't personally use flatpak/flathub so I don't have a good sense for what the expected behavior would be. Just figured I'd ask

fredizzimo:
> > **Update**: Following some [feedback](https://github.com/flathub/flathub/pull/5017#issuecomment-1989781021) from a Flathub maintainer, it was suggested that the flatpak Neovide should still work when the host doesn't have neovim installed (or it just isn't accessible from the container for some reason). The expected behavior, imho, should be to fallback to a bundled version of neovim when the host isn't available. I'll implement the changes in this branch.
> 
> Is a bundled copy of neovim correct here? That would be 120mb of data that for most uses doesn't actually get run and also includes many small files that can slow down the unpacking/unzipping steps that I presume flatpak uses. I don't personally use flatpak/flathub so I don't have a good sense for what the expected behavior would be. Just figured I'd ask

I agree, it does not look like they requested to embed Neovim, just to make it more clear in the description that it's required. Perhaps we could even improve our error message when running from Flatpaks? 

I think we also need to be able to launch the flatpak version of Neovim if the terminal one is not available. And that should be preferred even if we embed it. Is that possible?

Issue #2112, opened 2023-11-08
Can not start by spotlight or double click App icon , but can launch from cli: /Applications/neovide.app/Contents/MacOS/neovide


finviman:
**Describe the bug**
can not start by alfred or click App icon , nothing happened. But can start from cli: ./Applications/neovide.app/Contents/MacOS/neovide

**To Reproduce**
In MacOS, Finder, go to Application, double click Neovide, nothing happened

**Expected behavior**
Can be launched by alfred or spotlight.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOS 13.5.2 
 - Neovide Version 0.11.2
 - Neovim Version 0.9.4

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
there are 2 log files, both of them are too large. only TRACE log, no ERROR log.
**Additional context**

vasac:
I had the same problem. Culprit was my `.bash_profile` or `.bashrc` config which printed something to `stderr` which then causes `neovide` to abort startup at it expects empty `stderr`.

falcucci:
> I had the same problem. Culprit was my `.bash_profile` or `.bashrc` config which printed something to `stderr` which then causes `neovide` to abort startup at it expects empty `stderr`.

This is a reasonable answer. I had it in the past as-well, same issue.

Issue #1679, opened 2022-12-16
Neovide.app launch on MacOS (Apple Silicon) via Spotlight/Alfred is broken


owittek:
neovide 0.10.3
NVIM v0.8.1

I've already tried adding the NVIM_BIN env to .zprofile (as an alternative to https://github.com/neovide/neovide/issues/1373#issuecomment-1179268069 ) and tried the fix described in Troubleshooting. Symlinking is a workaround but it breaks telescope and might affect other plugins.

____________________________________

> I found a section in [Homebrew documentation](https://docs.brew.sh/FAQ#my-mac-apps-dont-find-homebrew-utilities) talking about this issue. Hope it helps.

This seems to work for opening the GUI in finder but it is still not possible to open it from spotlight/alfred. Symlinking nvim to /usr/local/bin ( ln -s /opt/homebrew/bin/nvim /usr/local/bin/nvim ) fixes that but neovide will always launch in the /usr/local/bin folder which seems to break the telescope plugin.

_Originally posted by @owittek in https://github.com/neovide/neovide/issues/1242#issuecomment-1355747526_

aromeronavia:
Same here

chrisgrieser:
As far as I can tell, opening via Alfred works now. (neovide 0.13)

ttollet:
A temporary workaround:

1. Open Automator
2. Click "New Document"
3. Select "Application" then click "Choose"
4. Search and add "Run Shell Script"
5. Paste the following: `open -a /opt/homebrew/bin/neovide --args $1`
   - Assumes neovide is installed via homebrew
   - Opens neovide as an application
   - Accepts other files as arguments to ensure "Open With" will work
7. In the top right of the input box, set _Pass Input_ to "as arguments"
8. Click _File > Save_

Now:
- Neovide can be opened from Spotlight search.
- Neovide supports the "Open With" menu option.

Only caveat is this approach shows Neovide without the nice Neovide icon until opened.

falcucci:
At least to me, no problem at all.

https://github.com/neovide/neovide/assets/33763843/a4d74c47-0a15-4227-a1bc-e076d5c0125b

Issue #1604, opened 2022-10-18
CLI does not start a proper instance of Neovide when called from Alfred


chrisgrieser:
**Describe the bug**
When using `neovide` from the Terminal, it properly starts up the app, but when using it from inside a third party app like [Alfred](https://www.alfredapp.com/)  (Mac), the Neovide app starts but the window black and unresponsive.

This is not a path issue, since I added the path of the neovide CLI. Also, this does not seem to be an Alfred issue, since calling CLIs from a script triggered by Alfred works for any other CLI I have ever used.

**To Reproduce**
Minimal Alfred Run-Script-Object: 
<img width="799" alt="Pasted image 2022-10-18 12 26 38" src="https://user-images.githubusercontent.com/73286100/196405599-02c17010-5aa9-40bc-861e-48b2ec7bf80d.png">

**Expected behavior**
Neovide should open normally.

**Desktop (please complete the following information):**
```
macOS 12.5.1 (M1)
neovim 0.8.0 (homebrew)
Neovide 0.10.1 (homebrew)
Alfred 5.0.3
```

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/9809420/neovide_rCURRENT.log)

chrisgrieser:
issue seems to have been fixed with 0.13. Thanks!

PR #2405, opened 2024-03-07
fix: execute login before starting login shell

falcucci:
This must solve most problems when neovide interact as a NON tty from the macOS application perspective persisting user sessions.

The following changes has been made:

- execute a non interactive login shell properly;
	
  After some research, using `/usr/bin/login` is the proper way for tty creation (hello alacrity, kitty), this is also important for script creation which depends from different conditions to be applied in case of tty command execution. I could avoid for example some mistakes at my scripts after identify this improvement.
  
  This also fixes https://github.com/neovide/neovide/issues/2407 which is a desirable behavior for macOS users to apply when they launch Neovide from the Finder or the Dock without any neovim setup changes to work out of the box.
  
  It forks a login shell with the user environment correctly such as its PATH values and uses the default macOS shell `/bin/zsh` in case we miss the shell common identification instead of `/bin/sh` to be spawned.

- update neovide installation instructions for macOS users;

  mainly homebrew users are getting confused about it afterwards, regarding to their shell setup, this instruct them better and avoid possible open issues like https://github.com/neovide/neovide/issues/1373;
  
- update `neovim_ok` function to handle errors properly;
  
  in the previous implementation the function wasn't working as expect after some debugging scenarios, such as returning OK when stderr is empty and stdout is succeed but also empty.
  
  The dynamic of the function was also simplified for a better maintainability.

- Fix

- No

fixes https://github.com/neovide/neovide/issues/2407 https://github.com/neovide/neovide/issues/1373 https://github.com/neovide/neovide/issues/1604 https://github.com/neovide/neovide/issues/2112 https://github.com/neovide/neovide/issues/1859 https://github.com/neovide/neovide/issues/1679


website/docs/installation.md
```diff
@@ -62,6 +62,17 @@ line:
 brew install --cask neovide
 ```
 
+Neovide register launched shells taking the users preferred shell into account.
```
Kethku, line 65: I think this should read "Neovide registers launched shells taking the user's preferred shell into account."

But that said, I'm not sure I understand what this is trying to tell the end user.

falcucci: thank you!

src/bridge/command.rs
```diff
@@ -57,14 +60,27 @@ fn create_platform_shell_command(command: &str, args: &[&str]) -> StdCommand {
...
+        let mut result = StdCommand::new("/usr/bin/login");
+
+        // We use a special flag to tell login not to prompt us for a password, because we're
+        // going to spawn it as the current user anyway. The addition of "p",
+        // preserves the environment.
+        // -f: Bypasses authentication for the already-logged-in user.
+        // -l: Skips changing directory to $HOME and prepending '-' to argv[0].
+        // -p: Preserves the environment.
+        // -q: Forces quiet logins, as if a .hushlogin is present.
+        result.args(["-flpq", &user, &shell, "-lc", &exec]);
```
fredizzimo, line 86: I think we should use `/bin/zsh` here like in Alacritty. The user shell does not necessarily have the exec command, so it would fail for those shells. There should also be no need to run that as a login shell (we don't need to do it twice), so `-c` is enough.

src/bridge/command.rs
```diff
@@ -79,53 +95,68 @@ fn create_platform_shell_command(command: &str, args: &[&str]) -> StdCommand {
...
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    let stderr = String::from_utf8_lossy(&output.stderr);
+
+    let error_regex = Regex::new(r"your \d+x\d+ screen size is bogus. expect trouble").unwrap();
+    let (_, non_matching_stderr_lines): (Vec<_>, Vec<_>) =
+        stderr.lines().partition(|line| error_regex.is_match(line));
+
+    let unexpeted_output = !output.status.success()
+        || stdout.is_empty()
+        || !(stdout.starts_with("NVIM v") && stderr.is_empty())
```
fredizzimo, line 148: I think the `stderr.is_empty()` condition should be removed here, we want it to error out in all cases when stdout is wrong.

The `stdout.is_empty()` condition is also redundant

src/bridge/command.rs
```diff
@@ -180,20 +214,40 @@ fn platform_which(bin: &str) -> Option<String> {
     None
 }
 
+#[cfg(target_os = "macos")]
+pub fn is_tty() -> bool {
+    !std::io::stdout().is_terminal()
+}
+
 #[cfg(target_os = "macos")]
 fn nvim_cmd_impl(bin: String, mut args: Vec<String>) -> TokioCommand {
```
fredizzimo, line 223: It seems like we should be able to share the implementation between this and `create_platform_shell_command`

The only difference seems to be that one of the functions constructs a `TokioCommand` and the other one a `std::Command`. So perhaps we should create a function that builds just the arguments?

src/bridge/setup.rs
```diff
@@ -100,6 +108,11 @@ pub async fn setup_neovide_specific_state(
         .await
         .context("Error when running Neovide init.lua")?;
 
+    #[cfg(target_os = "macos")]
+    setup_startup_directory(nvim)
```
fredizzimo, line 112: Could we move this to the command implementation instead? Something like `cd $HOME;nvim` but a little bit more advanced to deal with all the arguments.

This way any arguments that changes the startup directory for `nvim` works correctly, for example `neovide -- -c "cd proj"`

fredizzimo (changes requested):

fredizzimo:
The title of this PR confuses me, there's nothing in this that executes a TTY or even simulates a TTY. Perhaps "On macOS, execute login before starting the login shell"

src/bridge/command.rs
falcucci: Sure, I am actually against about forcing our users into a specific shell. It means they will be obliged to maintain the .zshenv which is something we shouldn't.

I'll remove the `exec` and launch as their own shell.

Also exec is not really needed here to make it work.

falcucci: > There should also be no need to run that as a login shell (we don't need to do it twice), so -c is enough.

thank you for remembering this but unfortunately it didn't load the correct PATH for both shells I've tested using only `-c`, it is basically not loading their shell configuration accordingly with `-l`

I also agree it's redundant but seems like alacrity is doing the same to make it work.

src/bridge/command.rs
```diff
@@ -37,11 +37,14 @@ fn build_nvim_cmd() -> Result<TokioCommand> {
         }
 
         bail!("ERROR: NEOVIM_BIN='{}' was not found.", cmdline);
-    } else if let Some(path) = platform_which("nvim") {
+    }
+
+    if let Some(path) = platform_which("nvim") {
```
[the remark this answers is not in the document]
falcucci: done

src/bridge/command.rs
falcucci: done

src/bridge/command.rs
falcucci: done

src/bridge/setup.rs
falcucci: that was a good idea! done.

src/bridge/command.rs
falcucci: done

src/bridge/setup.rs
falcucci: @fredizzimo the idea is good but actually that implementation was wrong.

any neovim command passing through neovide should be executed already when we launch the `--embed` instance so 1. it would be duplicated and 2. it should be made by the command line parser as it is right now.

I also forgot that some TTY could pass nvim commands such as the startup directory (hello Neohub), so as it was implemented, those TTY would break.

I think it's better to ignore *any* neovim command and just handle the startup path received in case there is one. If not we understand the the dafault is $HOME.

falcucci: @fredizzimo done some refactor to make it work respecting the `--cmd` or `-c` nvim commands.

Executing it in lua was the only way to make it respect the startup directories and the priority of commands tho.

I've tested using Finder, Dock and Neohub. (the terminal sessions shouldn't be affected)

src/bridge/command.rs
```diff
@@ -42,90 +42,148 @@ fn build_nvim_cmd() -> Result<TokioCommand> {
...
-
-                if is_wsl {
-                    bail!("{error_message_prefix}wsl '$SHELL' -lc '{bin} -v'");
-                } else {
-                    bail!("{error_message_prefix}$SHELL -lc '{bin} -v'");
-                }
-            }
-            return Ok(true);
-        }
+    let output = cmd.output();
```
Kethku, line 155: I think this could be written as
```rust
let Ok(output) = cmd.output()
else {
    return Ok(false);
};
```

Or something like that. I'm confident there is a way to write this that doesn't split the error check and then use the ? operator in two steps.

src/bridge/command.rs
```diff
@@ -42,90 +42,148 @@ fn build_nvim_cmd() -> Result<TokioCommand> {
...
+    // But a lossy conversion is OK for our purposes
+    let output = output?;
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    let stderr = String::from_utf8_lossy(&output.stderr);
+
+    let error_regex = Regex::new(r"your \d+x\d+ screen size is bogus. expect trouble").unwrap();
+    let (_, non_matching_stderr_lines): (Vec<_>, Vec<_>) =
+        stderr.lines().partition(|line| error_regex.is_match(line));
+
+    let unexpeted_output = !output.status.success()
```
Kethku, line 171: spelling: unexpected

src/bridge/command.rs
```diff
@@ -181,19 +239,20 @@ fn platform_which(bin: &str) -> Option<String> {
...
-        Ok(args) => args,
-        Err(_) => panic!("Failed to join arguments"),
-    };
-    let mut cmd = TokioCommand::new(shell);
-    if env::var_os("TERM").is_none() {
-        cmd.arg("-l");
-    }
-    cmd.arg("-c");
-    cmd.arg(&args);
+pub fn is_tty() -> bool {
```
Kethku, line 242: Maybe I'm missing something dumb, but shouldn't this return true if stdout is a terminal? I think I'm missing something. Is it possible we could rename this function to something that makes that confusion less likely?

Kethku:
The last thing that stands out to me is a bit of a nitpick, but theres enough macos specific code here that I wonder if those functions should get pulled out into their own module that is conditionally compiled as a group. Up to you

src/bridge/command.rs
falcucci: Actually you are right. In cases like Finder/Dock/Neohub it should be considered as non tty, my confusion. So we should basically update the code to

```rust
#[cfg(target_os = "macos")]
pub fn is_tty() -> bool {
    std::io::stdout().is_terminal()
}
```

which teturns `true` if the descriptor/handle refers to a terminal/tty.

Regarding the function name, keeping it like it is, has the right term now.

falcucci:
@Kethku regarding to the macOS specific code I strongly agree. I suggest we create a specific technical issue for that so we can split it up all together in another PR so we can review that separately.

I like the idea of platform tree structure just like winit did [here](https://github.com/rust-windowing/winit/blob/master/src/platform/macos.rs)

src/bridge/mod.rs
```diff
@@ -31,6 +31,9 @@ pub use ui_commands::{send_ui, start_ui_command_handler, ParallelCommand, Serial
 const INTRO_MESSAGE_LUA: &str = include_str!("../../lua/intro.lua");
 const NEOVIM_REQUIRED_VERSION: &str = "0.9.2";
 
+#[cfg(target_os = "macos")]
+const TTY_STARTUP_DIRECTORY: &str = "$HOME";
```
fredizzimo, line 35: This should perhaps be named as `HOME_DIRECTORY`, the naming is confusing right now, since it's used when you are running from the GUI, not the TTY.

Furthermore, we already use the `dirs` crate, so this could be
https://docs.rs/dirs/latest/dirs/fn.home_dir.html

src/bridge/mod.rs
```diff
@@ -44,6 +47,84 @@ fn neovim_instance() -> Result<NeovimInstance> {
...
+    match path {
+        arg if PathBuf::from(&arg).is_dir() => arg.to_string(),
+        arg if PathBuf::from(&arg).is_file() => Path::new(&arg)
+            .parent()
+            .unwrap_or_else(|| Path::new(TTY_STARTUP_DIRECTORY))
+            .to_string_lossy()
+            .to_string(),
+        _ => TTY_STARTUP_DIRECTORY.to_string(),
+    }
+}
```
fredizzimo, line 119: I think the home dir should be the default, unless a file or directory is opened, on all platforms, not just macOS,  when running from the GUI.

But that looks like it's already the case by default, is it not on macOS? What is it if you don't change it here?

src/bridge/command.rs
falcucci: nice catch! done

src/bridge/mod.rs
falcucci: good catch, I'll replace it to use the `dirs` crate instead

src/bridge/mod.rs
falcucci: on macOS if you launch from Finder or Dock the default is the root path `/`

falcucci: that's why I made it specifically to macOS

src/bridge/mod.rs
```diff
@@ -44,6 +47,84 @@ fn neovim_instance() -> Result<NeovimInstance> {
...
+///
+/// Any nvim --cmd or -c argument is handled as a command to be executed.
+///
+/// Conditions:
+///
+/// - Is not a TTY session.
+/// - Argument is a directory, it becomes the startup directory.
+/// - Argument is a file, its parent directory becomes the startup directory.
+/// - Neither directory nor file, $HOME is used.
+#[cfg(target_os = "macos")]
```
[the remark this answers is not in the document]
falcucci: just confirm to me that the default startup path in windows or linux (if you have it) aren't $HOME launching from a non tty, then I make it platform agnostic.

Kethku:
> @Kethku regarding to the macOS specific code I strongly agree. I suggest we create a specific technical issue for that so we can split it up all together in another PR so we can review that separately.
> 
> I like the idea of platform tree structure just like winit did [here](https://github.com/rust-windowing/winit/blob/master/src/platform/macos.rs)

https://github.com/neovide/neovide/issues/2439

src/bridge/mod.rs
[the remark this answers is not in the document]
fredizzimo: It looks like my comment has jumped around, it was meant for the `handle_command_arg_as_path_or_default`, which should be possible to support on all platforms even if the default startup directory isn't changed otherwise.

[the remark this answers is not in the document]
falcucci: that was it, makes sense @fredizzimo 

done!

fredizzimo:
Thanks you!

falcucci:
NOTE: the custom startup dir feature has been reverted due to possible macOS incompatibilities and bugs to determine the non-tty instance of itself. This feature is not also desirable as neovide should always be launched and [recognized](https://github.com/neovide/neovide/blob/fcf5e87bf8a75727ee75da961ce98b2f51fd0cb2/src/main.rs#L239-L248) as non-tty as default.

Some specific nvim commands regressions were also found during the usability.

Issue #2375, opened 2024-02-22
Differences between Kitty and Neovide


ddevaal:
I want Neovide to look the same as vim running under Kitty. But I have no idea how to fix the issues below:

On the left you see Kitty with Lazyvim (nevermind the blur). On the right Neovide. Somehow there are horizontal lines in the LazyVim logo?

Also the 'lightning' icon seems to have a different color but this might be a LazyVim issue.

![LAZYVIM kopie](https://github.com/neovide/neovide/assets/74067645/a6f4ea30-c25a-4432-92d2-fba09b6b8a0d)

I'm using the JetBrainsMono Nerd Font. I have configured this font in `~/.config/neovide/config.toml`:

```
frame = "transparent"

[font]
normal = ["JetBrainsMono Nerd Font"]
size = 13
```

Popup windows are not fully transparent. How can I remove the black background? Kitty/Vim popup windows are fully transparent so I'm not sure if this is a Telescope problem. Also what is going on with the statusbar? 
 
![Neovide](https://github.com/neovide/neovide/assets/74067645/77c59e8b-9fa4-4115-b62c-958714e7365c)

These are the only settings I have configured in LazyVim:

```
vim.g.guifont = "JetBrainsMono Nerd Font:h13"
vim.g.neovide_transparency = 0.5
return {
  "tokyonight.nvim",
  opts = {
    transparent = true,
    styles = {
      sidebars = "transparent",
      floats = "transparent",
    },
  },
}
```

I'm running Neovide 0.12.2 (macOS 14.3.1).

Thanks for any support. Neovide is amazing! ❤️

emonadeo:
The popups not being fully transparent is because of how neovide renders [multigrid](https://neovide.dev/command-line-reference.html?highlight=multigrid#multigrid) I think.
Run Neovide with `--no-multigrid`, or put `no-multigrid = true` into your ~/.config/neovide/config.toml to turn it off.
Afaik the floating window transparency is tied to your `neovide_transparency`, so I don't think you can have it look like kitty without disabling multigrid. (Correct me if I am wrong)

xzbdmw:
If I set telescope winblend to 100, the font below floating window will not shadow, which is totally invisible if mutigrid is disabled, it is the biggest difference of neovidetransparency and winblend.

ddevaal:
> The popups not being fully transparent is because of how neovide renders [multigrid](https://neovide.dev/command-line-reference.html?highlight=multigrid#multigrid).
> 
[...]

Issue #1676, opened 2022-12-13
Multigrid: z-index is wrong for Telescope modal


ndan:
With multigrid enabled. When I open a `Telescope file_browser` then press `<esc>?` to open help. The help window opens behind the Telescope. 

* I can reproduce it on my custom config and lunarvim.
* It doesn't matter which Telescope command you use, file_browser or any other command
* No bug in a terminal

Multigrid disabled
<img width="1792" alt="Screenshot 2022-12-13 at 11 38 49" src="https://user-images.githubusercontent.com/760394/207222071-0c9109c9-823d-4536-ae7e-b69febc2a529.png">

Multigrid enabled
<img width="1792" alt="Screenshot 2022-12-13 at 11 51 52" src="https://user-images.githubusercontent.com/760394/207222205-f24602e6-303c-4179-88c2-a116e4ac9bd0.png">

**Desktop:**
 - OS: macOS 13.0.1
 - Neovide Version: 0.10.3
 - Neovim Version: 0.8.1

Kethku:
I believe this is fixed. Please reopen if problems persist

Kethku:
Can you send a link to your config or a minimal repro?

sid-6581:
> Can you send a link to your config or a minimal repro?

Here is my config: https://github.com/sid-6581/NeovimConfig

FloatStream:
I also have this problem, and I tried a minimal repro:
```lua
{
    'nvim-telescope/telescope.nvim',
    branch = '0.1.x',
    dependencies = { 'nvim-lua/plenary.nvim' },
    opts = {}
}
```
I use Lazy.nvim to manage my plugins. The above only calls `require(telescope).setup()` to initialize it. The test results are as follows:
Multigrid ON
![multigrid](https://github.com/neovide/neovide/assets/69284622/b36c1148-44c2-4b62-ad90-c8b7b9511505)
Multigrid OFF
![no-multigrid](https://github.com/neovide/neovide/assets/69284622/6dc89ed3-5e7f-464b-bd25-68cd79babeb4)

FloatStream:
> To be honest, I'm not sure if this is a Neovide bug, both windows specify a zindex of 50, and the neovim documentation does not say anything about the ordering of windows with the same index.
> 
> ~So, I think a bug report should be opened in the LazyVim repository.~ The bug is actually in Telescope

Ok, I'll try to find the answer on Telescope's project or open an issue. Thank you

sid-6581:
[...]

PR #2408, opened 2024-03-07
Allow user to specify zindex so that some floating window is treated as normal

xzbdmw:
- Feature
  Some plugins use floating windows to simulate a normal window, and they may not have anticipated that neovide will render additional shadows, resulting in strange visual effects
such as 
https://github.com/nvim-treesitter/nvim-treesitter-context/issues/415#issuecomment-1983852128
https://github.com/neovide/neovide/issues/2375#issuecomment-1969194912
, so I make a small change to let mutigrid ignore specific z-index windows, the result is like this:
截屏2024-03-08 06 03 15
On the left is the normal multi grid, and on the right is the `vim.g.neovide_no_multigrid_zindex=20` that I set to make it completely transparent(no shadow no winblend no background color) and no longer appear like a glass

I don't know if this feature is good or bad, It is my first real code pr in github :)

- No


src/renderer/mod.rs
```diff
@@ -47,6 +47,7 @@ pub struct RendererSettings {
     floating_blur_amount_y: f32,
     floating_shadow: bool,
     floating_z_height: f32,
+    no_multigrid_zindex: u64,
```
MultisampledNight, line 50: Since a user might want to have this apply to multiple zindices, how about making this a list of indices instead?

src/renderer/rendered_window.rs
```diff
@@ -367,6 +367,7 @@ impl RenderedWindow {
         let transparent_floating = self.anchor_info.is_some() && has_transparency;
 
         if self.anchor_info.is_some()
+            && self.anchor_info.as_ref().unwrap().sort_order != settings.no_multigrid_zindex
```
MultisampledNight, line 370: Might want to combine these two? As in e.g.:

```suggestion
        if self.anchor_info
            .as_ref()
            .map_or(false, |info| info.sort_order != settings.no_multigrid_zindex)
```

MultisampledNight (commented):
Sounds very useful to me, thank you!!

I'm not aware of any API in Neovim available to control this, and there are uses indeed where having a plugin's floating window be treated as being part of the main grid would be desirable.

MultisampledNight:
(fwiw you can just run `cargo fmt` and commit to make the rustfmt lint stop complaining)

MultisampledNight:
Sidenote for future PRs (no biggie here, really just for the future): It is generally preferred to make a new branch on your fork for each new PR. This allows both

1. parallel work on your fork, if you'd happen to have any other interesting things to tackle
2. easier testing locally, since now the tester can just do `git switch your-branch` instead of doing `git checkout -b xzbdmw-main xzbdmw/main` (thus having to watch out where to push on changes)

Optionally see https://github.com/neovide/neovide/pull/1468#issuecomment-1215667467 for further details.

src/renderer/mod.rs
Kethku: Maybe it would make sense to say that all zindex values below the configured one will be treated as flat?

Tbh I kinda think a name like `flatten_floating_windows_below_zindex` or something more concise but in that direction might help communicate what the setting actually does. Multigrid is a feature for neovim guis that might not make sense to end users.

xzbdmw: > Maybe it would make sense to say that all zindex values below the configured one will be treated as flat?
> 
> Tbh I kinda think a name like `flatten_floating_windows_below_zindex` or something more concise but in that direction might help communicate what the setting actually does. Multigrid is a feature for neovim guis that might not make sense to end users.

There is a usage scenario where a floating window can be used as a preview and can do some editing. Generally, the zindex of this type of window is set very low.
For example, I set glance.nvim's zindex to 10 and treesitter-context's zindex to 20, and keep glance's preview's shadow like this:
image
That being said, a specified zindex would allow this kind of customization.

xzbdmw:
> Sidenote for future PRs (no biggie here, really just for the future): It is generally preferred to make a new branch on your fork for each new PR. This allows both
> 
> 1. parallel work on your fork, if you'd happen to have any other interesting things to tackle
> 2. easier testing locally, since now the tester can just do `git switch your-branch` instead of doing `git checkout -b xzbdmw-main xzbdmw/main` (thus having to watch out where to push on changes)
> 
> Optionally see [#1468 (comment)](https://github.com/neovide/neovide/pull/1468#issuecomment-1215667467) for further details.
Very detailed explaination! I changed the name to flatten_floating_zindex as @Kethku suggested, and make it a comma separated list such as `vim.g.neovide_flatten_floating_zindex = "20,30,40" which is the same way font is configured.

fredizzimo:
Before merging this, I would like to see another approach being tried, which I think fixes this issue and also these
* https://github.com/neovide/neovide/pull/2379 @TwIStOy
* https://github.com/nvim-telescope/telescope.nvim/issues/2603 

Currently we draw each floating window separately, sorting by z-index, then by x position, and finally by y position. With each window blending together with everything being drawn so far. Especially the sort by x and y position is quite random and arbitrary, not much better than completely random.

We should instead change the drawing to be done by z-index layers, to draw the complete full layer before blending with the lower ones and applying shadows to the combined outline. That way as long as the windows have the same z-index they will be treated as one. Some plugins might have to be changed to support that, but it should always be possible to do so, without breaking the regular terminal rendering.

There are two special cases that needs to be handled, since the windows with the same z-index can have conflicting contents.
1. The windows draw something with the foreground color to the same cell. This is really a bug in either the user configuration or the plugin. But perhaps the most logical thing for Neovide to do in this situation to draw the text on top of each other (in random order). In practice this should not be a problem, since properly made plugins will only have text in one of the overlapping windows.
2. The windows draws with different colors or blend factors. To support the [telescope case](https://github.com/nvim-telescope/telescope.nvim/issues/2603 ), render the background with the lowest blend factor (the least transparent) one. If there are still conflicts, randomly by deterministically, maybe by the x/y sort, select one of them.

NOTE: That this only solves a subset of what this PR attempts to do. But I'm not sure if configuring a z-index, is the best way to disable the shadows from a subset of the windows. I think we need something more configurable that the theme and plugins can control. Maybe some additional windows settings in Neovim itself or something else.

It also does not solve this issue:
https://github.com/neovide/neovide/issues/1676, which I think only telescope can fix.

Would either you @xzbdmw, or @TwIStOy, be able to work on that?

TwIStOy:
That's a great idea! I will take a look and try to implement it on Monday.

xzbdmw:
If I understand correctly, this approach mainly solves the problem of multiple overlapping(or adjacent) floating windows, and there will still be a combined shadow (if they have the same zindex), I wander these two features can coexisting together(better rendering and don't render at all).

TwIStOy:
IMO, disables the shadow and blur of a specific z-index window seems somewhat unusual. I notice that all floating window related ui-events contains a win field. Therefore could we use the win field to decide whether to apply shadow and blur to that floating window? This also solves an issue with another plugin [colorful-winsep](https://github.com/nvim-zh/colorful-winsep.nvim), which uses floating windows to create these colorful separators.

xzbdmw:
Hmm, https://github.com/nvim-zh/colorful-winsep.nvim can already be disabled by setting zindex,
before :
image
after:
截屏2024-03-24 16 47 29

> disables the shadow and blur of a specific z-index window seems somewhat unusual

because it is the only "unique" property of a floating window

> Therefore could we use the win field to decide whether to apply shadow and blur to that floating window?

Hmm how could a user configure that win field to say they don't want shadow from the specified plugin

TwIStOy:
> because it is the only "unique" property of a floating window

However, z-index is not really the unique property. Currently, there are a lot of floating windows using the same z-index. Thus, I question whether using the z-index might inadvertently disable other windows that should have shadow and blur.

> Hmm how could a user configure that win field to say they don't want shadow from the specified plugin?

Obviously, the buffers in those special windows corresponded to specific filetypes. So, I think a simple `autocmd` is enough.

xzbdmw:
Seems like this suggestion https://github.com/neovide/neovide/issues/2375#issuecomment-1969222013
but it is not possible to disable shadow for specified window

> This also solves an issue with another plugin [colorful-winsep](https://github.com/nvim-zh/colorful-winsep.nvim)

Could you show a working examples using autocmd?

fredizzimo:
We definitely want to allow disabling the shadows from specific windows, as I already mentioned in my previous message.

But I don't think that using the `z-index` is the best way, unless we can create a global standard that is followed across the Neovim community. But even then, it has limitations, like you might want to disable the shadows for some window outside that range. 

I can think of two solutions, which are basically the same, but specified in different places. That would be to add a `shadow_type` configuration option to either the highlight group or the floating window configuration. A similar option could be added for the background blur. This also allows us to implement different types of shadows and blurs in the future and have everything absolutely configurable. Of course, it also requires some kind of standardization, but that could happen more incrementally, and mostly implemented in the color schemes, where this kind of stuff belongs IMO.

This needs support on the Neovim side though, but I'm planning to add a PR that adds support for arbitrary arguments to the highlight groups, which are passed to GUI implementations. This allows us to experiment without updating Neovim, and when things stabilize, standardize those arguments, and make them official. This would also allow us to fix the transparency issue we have, since we need more expressiveness than the default highlight groups can provide.

TwIStOy:
> Could you show a working examples using autocmd? Because when a WinEnter or something happends, if you disable these shadows, the entire window will have no shadow, say, cmp's completion menue shadow.

Ohhh, It's my mistake. Upon reviewing the content of ui-events, I found that `win` in ui-events differs from `win` retrieved from `vim.api`. Thus, implement it using `autocmd` might not be feasible.

Kethku:
Im sorta inclined to close this. The "render same z-index together" PR solves a part of this problem by rendering sequential zindex floating windows together.

I fully agree that we need a way to configure how windows are rendered on a window by window basis but this PR doesn't really solve this in my mind in a way thats approachable to end users. My impression is that the Z-Index is not a thing that most users think about.

I'm going to close for now, but feel free to reopen if more discussion is needed

Issue #342, opened 2020-08-28
Add support for background images


oxysoft:
I think given that Neovide supports crazy vfx and cursor animations, it wouldn't be too out of place to support background images as well. A good use case is to set a faint and soft tiling pattern at low opacity which can be very sexy! For example a pattern like this:

![image](https://user-images.githubusercontent.com/3945277/91506278-bd348000-e89f-11ea-9673-e119402627b5.png)

Bonus point if there's support for a multiply operation instead of transparency, which lets black & white patterns play well with any color scheme.

j4qfrost:
Hmm, wondering if we should add a "sexy" tag.

danielnehrig:
or just add opacity kinda like terminals it do like kitty or alacritty

xbladesub:
Only two reasons why do I still use iTerm:
1) Custom line height support 
2) Background images support

norflin321:
looking forward to this feature 😄

agoodfellow123:
Why stop with images? I usually use neovide with transparency + my animated background, since I'm on linux I can't blur the neovide transparent background but I solved that with blurring the wallpaper itself instead:

[Screencast_20240405_223440.webm](https://github.com/neovide/neovide/assets/36235265/d882bc76-9a0e-43fc-a13e-2548187678b0)

PR #2419, opened 2024-03-14
Add background image

mysty0:
This relates to #342 
- Add background image that can be specified in config with `vim.g.neovide_background_image = "E:\\downloads\\102075572_p0.jpg"`
- Slightly bump skia version to add support for new image loading API

It will have the same blur and opacity settings as single color background and look like this:
image

- Feature

- No


src/window/settings.rs
```diff
@@ -16,6 +17,7 @@ pub struct WindowSettings {
     pub touch_deadzone: f32,
     pub touch_drag_timeout: f32,
     pub background_color: String,
+    pub background_image: String,
```
MultisampledNight, line 20: Should this be in `src/renderer/mod.rs` in `RendererSettings` instead? (alongside with `background_color`)

src/renderer/mod.rs
```diff
@@ -148,12 +157,14 @@ impl Renderer {
         self.cursor_renderer.prepare_frame()
     }
 
-    pub fn draw_frame(&mut self, root_canvas: &Canvas, dt: f32) {
+    pub fn draw_frame(&mut self, root_canvas: &Canvas, dt: f32, width: f32, height: f32) {
```
MultisampledNight, line 160: Might be able to use `Canvas::base_layer_size` instead of passing `width` and `height` by parameter?

(fwiw you can look up the things available in `skia_safe` on skia.org, e.g. [`SkCanvas::getBaseLayerSize`] corresponds to `Canvas::base_layer_size`. The docs for `skia_safe` are available directly by running `cargo doc --open` locally)

src/renderer/rendered_window.rs
```diff
@@ -353,11 +354,44 @@ impl RenderedWindow {
...
+        let scale_x = image_width / screen_rect.width();
+        let scale_y = image_height / screen_rect.height();
+
+        // Calculate the portion of the image that corresponds to the window's position
+        let src_x = scale_x * (window_rect.left() - screen_rect.left());
+        let src_y = scale_y * (window_rect.top() - screen_rect.top());
+        let src_width = scale_x * window_rect.width();
+        let src_height = scale_y * window_rect.height();
+
+        let src_rect = Rect::from_xywh(src_x, src_y, src_width, src_height);
```
MultisampledNight, line 377: Since this implies unequal scaling on both axes, could this be made a cropping operation, such that both axes are always scaled equally and the longest side gets pixels cut off?

MultisampledNight (commented):
Thank you!

pidgeon777:
Very cool, I was thinking of the same functionality just yesterday.

I've a question though, what will happen if Neovide and background images sizes are different? For example:

- Background image will be stretched
- Background image will be padded

etc.

src/window/settings.rs
mysty0: I think it makes more sesnse to keep it in window settings as it modifies the window property rather than how the window is being rendered

src/renderer/mod.rs
mysty0: With the latest renderer changes using `base_layer_size` makes things much more cleaner however it doesn't cover full screen. There is some space left in the bottom that isn't covered by the image. 
image
I don't think it's very critical but it would be nice if image could cover the whole screen. Do you know how this can be achived?

mysty0:
> Thank you!

Thank you for the review @MultisampledNight! Sorry for the wait I just noticed that there is some activity on this pr. I merged latest changes and implimented your suggestions.

mysty0:
> Very cool, I was thinking of the same functionality just yesterday.
> 
> I've a question though, what will happen if Neovide and background images sizes are different? For example:
> 
> * Background image will be stretched
> * Background image will be padded
> 
> etc.

Before it was stretching however in latest commit I made it to keep aspect ratio, fill the screen and center the image and crop the longest side in case if it overflows as suggested [here](https://github.com/neovide/neovide/pull/2419#discussion_r1536172165)

> I wonder that If I make Neovide fullscreen and pass a background image sharing the same screen resolution, no image stretches / paddings will occur? That would be the "ideal" case.

Yes it would fill the screen without any stretching and padding

MultisampledNight:
(fwiw I am not familiar enough with the neovide codebase anymore to qualitatively review this PR anymore, sorry)

src/window/renderer.rs
Kethku, line 1: this change seems unintentional

fredizzimo:
@pidgeon777, I started the CI run, the artifacts should be there once it finishes

pidgeon777:
I just did a quick test with these settings:

```lua
  vim.g.neovide_transparency = 0.8
  -- vim.g.neovide_transparency = 1.0
  vim.g.neovide_background_image = "C:\\Users\\<NAME>\\Downloads\\pexels-photo-1483054.jpeg"
```

Background image is now shown but it makes Neovide extremely laggy, even with simple cursor movements in normal mode.

pidgeon777:
When utilizing an image with a resolution of 1920x1080 as a background, the performance and smoothness of Neovide appear to remain unaffected (but more tests should be done).

I would like to inquire about the configuration settings of Neovide, specifically pertaining to the transparency of the background image. Could you provide guidance on how to adjust these settings so that:

- When the transparency is set to 0, the background image is entirely obscured, and only the Neovim background color is visible.

Example:

image

- Conversely, when the transparency is set to 1, the background image is fully displayed without any overlay of the Neovim background color:

image

To achieve the above image, I simply moved the Neovide window on the top of a Chrome background image.

But it would be great to obtain the second setup by setting the path to the background image in Neovide, and toggle between the amount of image and the neovide background, with the transparency option.

pidgeon777:
The aforementioned test was conducted using the artifact linked below. This version of the software did not yet include the background image feature. However, it did allow for the adjustment of the Neovide background transparency:

[Artifact Link](https://github.com/neovide/neovide/actions/runs/8991931441#artifacts)

mysty0:
> But it would be great to obtain the second setup by setting the path to the background image in Neovide, and toggle between the amount of image and the neovide background, with the transparency option.

You can do it with `neovide_background_transparency` option, it controls how transparent is build-in neovide background color. `neovide_transparency` controls how transparent is the whole window. 

This is also probably why you can't see the background image. Try setting `neovide_background_transparency` to lower values like 0.5 to have a mix between a neovide's background color and image.


>Background image is now shown but it makes Neovide extremely laggy, even with simple cursor movements in normal mode.

I tried with 8k image but unfortunatly I wasn't able to reproduce it. Could you send your image that you are testing?

pidgeon777:
You're right, I didn't know about this new `neovide_background_transparency` option.

Below the background image causing the lag:

pexels-photo-1483054

pidgeon777:
I have conducted some tests using the artifacts available at [Neovide Actions Run](https://github.com/neovide/neovide/actions/runs/9061048547#artifacts). I collected some observations regarding it:

1. **Background Image Stretching**: The background image appears to stretch when the Neovide window is resized.

2. **Background Image Replication**: The background image is replicated for each window (including Trouble, window split, etc.). This could lead to a cluttered appearance, particularly when multiple windows are open.

Given these observations, I have a couple of suggestions for potential improvements:

**For Issue 1 (Background Image Stretching):**

- Investigate the possibility of maintaining the aspect ratio of the background image during window resizing. This could prevent distortion and ensure the image remains visually appealing regardless of window size. For example, by adding some padding. It would be nice to leave this strech feature enabled by specifying a configuration option.

**For Issue 2 (Background Image Replication):**

- Consider allowing users to assign a unique background image to each window. This could enhance customization and allow users to differentiate between windows more easily. It seems to be something technically doable, given that the background image is currently replicated for each window.

- Alternatively, consider assigning a single background image to the Neovide background only, regardless of the number of open splits. This would prevent the replication of the background image for each window split, which currently occurs.

fredizzimo:
> * Consider allowing users to assign a unique background image to each window. This could enhance customization and allow users to differentiate between windows more easily. It seems to be something technically doable, given that the background image is currently replicated for each window.

I think this is quite hard currently. We don't have any support for window specific options in Neovide. And the Neovim gui protocol does not really support it easily either, which means that if we use the Neovim api to do it, things could get out of sync.

It's something that we want to add support for in the future though, for example, we should be able to control the shadows, and blur effects per window.

pidgeon777:
Understand, so for now a simple "hack" to achieve the same effect could be just set `vim.g.neovide_background_transparency` to a value and put a static background under the `Neovide` window

fredizzimo:
@mysty0, are you able to test if a smaller image is fast, I mean the cost should be way less than a ms? 

In that case, the solution is probably to either scale the image first when loading, or draw in sections loading different parts of the image.

mysty0:
> @mysty0, are you able to test if a smaller image is fast, I mean the cost should be way less than a ms?
> 
> In that case, the solution is probably to either scale the image first when loading, or draw in sections loading different parts of the image.

Yep, for 1080p image I get around 1.5us for background rendering

pidgeon777:
Does it also support .webp images?

mysty0:
> Does it also support .webp images?

No, I don't think it supports them

mysty0:
I was able to optimize it to render for less than 1ms for 10k image while text editing. However it still can go up to 15ms when resizing windows but it doesn't really feel that laggy. To further optimize I can probably prescale image to maximum monitor size but it would involve a lot of platform specific code and I don't think it really worth it.

pidgeon777:
I have a question regarding the changes made in the previous commit. Previously, we had the capability to load high-resolution images, but this process was time-consuming and resulted in significant lag in Neovide.

With the new implementation, my understanding is that the image loading time has been reduced. However, I'm curious about the impact on Neovide's performance. Will it still be influenced by the size of the image, albeit to a lesser extent than before?

To put it more succinctly, if we load a high-resolution image as a background and do not resize the Neovide window, will the overall performance, such as cursor movement and scrolling, remain the same as if no background image was loaded at all? Or rather, would loading a 10K image as a background result in the same performance as loading a 2K image?

src/window/window_wrapper.rs
```diff
@@ -172,6 +176,35 @@ impl WinitWindowWrapper {
         wrapper
     }
 
+    fn load_image(image: &str) -> Option<Image> {
+        if image.is_empty() {
+            None
+        } else {
+            let skia_data = Data::from_filename(std::path::Path::new(image))?;
```
fredizzimo, line 183: I think we might want to use https://docs.rs/shellexpand/latest/shellexpand/fn.tilde_with_context.html here so that you can easily refer to the home directory

src/renderer/rendered_window.rs
```diff
@@ -373,16 +385,31 @@ impl RenderedWindow {
...
+                &pixel_region,
+                screen_rect,
+            );
+            // Self::draw_window_background_image(
+            //     &paint,
+            //     root_canvas,
+            //     background,
+            //     &pixel_region,
+            //     screen_rect,
+            // );
```
fredizzimo, line 402: There's an extra comment here.

fredizzimo (changes requested):
This still needs documentation, including instructions that tells that the image is only shown when `g_neovide_transparency` is less than 1.0

A-Lamia:
is there a way for this to tile lets say i have a small pattern image 250x250 pixels or will all images always be centered with no repeat ?

fredizzimo:
There's one more thing that I noticed.

There's some strange interaction between `neovide_transparency` and `neovide_background_transparency`. Even without a background image the `neovide_background_transparency` affects how transparent the window is. 

I suggest that we use one of these three alternatives, I'm not really sure which one is the best
1. Make a breaking change so that `neovide_transparency` affects everything else, but not the final window transparency. And the `neovide_background_transparency` controls that. With both set to opaque, then the background image replaces the Neovide default background on the root window, but fully opaque.
2. Change `neovide_background_transparency` to `neovide_background_image_transparency`, and let it control only the background image.  Otherwise, it works like 1.
3. Let `neovide_transparency` control all the transparency. I think even in this case I would prefer that the background image is always visible when set and replaces the default background on root windows.

mysty0:
> is there a way for this to tile lets say i have a small pattern image 250x250 pixels or will all images always be centered with no repeat ?

No, unfortunately its not possible at the moment

mysty0:
> Make a breaking change so that neovide_transparency affects everything else, but not the final window transparency. And the neovide_background_transparency controls that. With both set to opaque, then the background image replaces the Neovide default background on the root window, but fully opaque.

Initially I wanted to do it like this but then decided on the current aproach to not break backwards compatibility

> Change neovide_background_transparency to neovide_background_image_transparency, and let it control only the background image. Otherwise, it works like 1.

Currently it is potentially possible to control three different transparencies: window, image and color. Right now  `neovide_transparency` changes all three of them and `neovide_background_transparency` adds more transparancy to color.  Also `neovide_background_transparency` in current implimentation is also affecting the window transparancy which I am sure how to fix.

The rendering is happening in the same order: window, image, color. So option that is changing only image transparancy wouldn't make much sense as the image would only be visible if `neovide_transparency` is less than 1 and less visible with the new setting. However I can change the render order and render image on top of color background and then it would make more sense to introduce this option.

> Let neovide_transparency control all the transparency. I think even in this case I would prefer that the background image is always visible when set and replaces the default background on root windows.

But in this case it is not possible to control the brightness of the image

---
I think the second option is the most optimal one. I will try to change render order and add setting to change image transparancy

kwikenkwak:
Any updates on this? Would love to see this feature.

pidgeon777:
> Any updates on this? Would love to see this feature.

I agree, I tested that feature some time ago and it worked really good. I hope it will be merged.

fredizzimo:
We are still waiting for this to be implemented
> I think the second option is the most optimal one. I will try to change render order and add setting to change image transparancy

It can't be merged the way it currently works:
> There's some strange interaction between `neovide_transparency` and `neovide_background_transparency`. Even without a background image the `neovide_background_transparency` affects how transparent the window is.

But actually, since we are going to do [breaking changes to the transparency settings]( https://github.com/neovide/neovide/pull/2769). I think I would actually prefer option 1. of my suggestion, as it gives more control, even without a background image. I don't think this would conflict my suggestions [here]( https://github.com/neovide/neovide/pull/2769), it just gives a quick configuration option if you just want to have a transparent background.

tskj:
I would also love this feature!

(proud daily user of neovide ❤️)

VortexViking:
It isn't working for me. Is it because I run the flatpak version?

fredizzimo:
@VortexViking, it hasn't been merged yet, because it's not ready. I don't remember exactly how, but it breaks the default transparency settings.

@mysty0, now when thinking about it, I previously suggested this
> Make a breaking change so that neovide_transparency affects everything else, but not the final window transparency. And the neovide_background_transparency controls that. With both set to opaque, then the background image replaces the Neovide default background on the root window, but fully opaque.

But now when thinking about it, it does not need to be a breaking change
1. If no background image is set, `neovide_transparency` and `neovide_normal_opacity` works like before
2. If a background image is set, `neovide_transparency` works like before, but there's an opaque background image below instead of the desktop. `neovide_background_transparency` (or rather, let's name it as `neovide_background_opacity`) controls how transparent the image itself is and how much of the desktop you see through. I suspect in most cases,users want to leave it at the default fully opaque value, to avoid seeing two images on top of each other.
3.  `neovide_normal_opacity` also works as before, so if you want to replace all the `Normal` highlights with the background image you could set that to 0.

VortexViking:
> Any updates on this, and anyway we can help, at least others because i do not know Rust (yet) sadly :o

Why not have an AI do it LOL.

epicseven-cup:
I can't tell if that's sarcasm 😔

VortexViking:
> I can't tell if that's sarcasm 😔

It is. It would not go well having chatgpt data mine our codebase.

VanshNexus0212:
when is this going to be merged

fredizzimo:
> when is this going to be merged

When it works without breaking regular transparency settings. Sadly @mysty0, seems to have stopped working on it, so someone else probably need to step up.

intermediate-developer:
> > when is this going to be merged
> 
> When it works without breaking regular transparency settings. Sadly @mysty0, seems to have stopped working on it, so someone else probably need to step up.

What about you? You seem like a good guy to continue this.

fredizzimo:
I think that the generic image support here could cover this use case as well
* https://github.com/neovide/neovide/pull/3039

So, we don't necessarily need specialized background image support, when that's ready. I think the negative z index support of the Kitty protocol could cover the use case https://sw.kovidgoyal.net/kitty/graphics-protocol/. For the official Neovim protocol, something better could be implemented that is aware the Neovim z-indices.

fredizzimo:
I'm closing this since it's no longer worked on.

Note that https://github.com/neovide/neovide/pull/3067 has also been stalled, so if someone wants to continue working on this, feel free to do so.

PR #2421, opened 2024-03-16
feat: Smooth blink animation

agraven:
Smoothly transition between the on and off state of the cursor

TODO:
- Make this optional behind a setting
- Don't require manually setting `--no-idle` to actually make the animation visible

[Screencast from 2024-03-16 16-33-33.webm](https://github.com/neovide/neovide/assets/23525639/2af1c202-d361-4040-aa52-60bc748c4da1)

- Feature

- No


src/renderer/cursor_renderer/blink.rs
```diff
@@ -82,6 +83,32 @@ impl BlinkStatus {
...
     }
 
+    /// Calculate the opacity the cursor should be drawn with when smooth cursor blink is enabled.
+    /// `0.0` is fully transparent, `1.0` is fully opaque.
+    pub fn opacity(&self) -> f32 {
+        let now = Instant::now();
+        if self.state == BlinkState::Waiting {
+            return 1.0;
+        }
+        let total = self.get_delay().as_secs_f32();
```
MultisampledNight, line 93: (might be a problem here with f32 if it's running for 2 weeks or more, the game Celeste has issues with that — I need to test around a bit locally though. ignore this for now)

website/docs/configuration.md
```diff
@@ -781,6 +781,24 @@ If the value is \<=0 then the cursor will be invisible. This setting takes effec
...
+
+```vim
+let g:neovide_cursor_smooth_blink = v:true
+```
+
+Lua:
+
+```lua
+vim.g.neovide_cursor_smooth_blink = true
+```
```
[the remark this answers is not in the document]
agraven: I hadn't noticed that convention, I think consistency is the most important thing. I'll update it.

src/renderer/cursor_renderer/blink.rs
agraven: `get_delay` will never return more than `max(blinkwait, blinkon, blinkoff)` in milliseconds, so I don't think that concern applies in this case.

MultisampledNight: Sounds solid then! Nevermind!

MultisampledNight (approved):

MultisampledNight:
When testing, I've found that the new option causes the cursor to disappear when configured to be non-blinking/static. Minimal repro is

1. Launch `neovide -- --clean`
2. `:let neovide_cursor_smooth_blink=v:true`

Causing the cursor to disappear.

This is an issue, since in some configurations, it might be desirable to have the cursor static in one mode but not in another.

`self.get_delay()` in `src/renderer/cursor_renderer/blink.rs:93` seems to return 0 in that case, causing `remaining / total` to be actually `0 / 0` ⇒ `NaN`. I'd guess it might make sense to add a guard condition for `is_static` in `fn opacity`?

agraven:
The change
```diff
@@ -63,7 +63,7 @@ impl BlinkStatus {
         let current_cursor = self.current_cursor.as_ref().unwrap();
 
         if is_static(current_cursor) {
-            self.state = BlinkState::On;
+            self.state = BlinkState::Waiting;
             ShouldRender::Wait
         } else {
             if self.transition_time <= now {
```

i.e. setting the `BlinkState` for when blinking is disabled to `BlinkState::Waiting`. This doesn't seem to cause any issue after some quick testing, and it would integrate better with the changes to the update loop, on top of feeling more semantically correct to me in general, since `BlinkState::On` implies there is an ongoing blinking process, while `BlinkState::Waiting` more accurately represents that blinking is inactive, just indefinitely instead of for `blinkwait` milliseconds

MultisampledNight:
Looks good to me, thank you again!! Merging...

hudgins:
I love this, but I've noticed it causes Neovide to report 15% CPU usage, where it is normally at about 1%.

agraven:
@hudgins That does seem like a bit much. Could you check if you get the same CPU usage when running neovide with `--no-idle` and smooth blink disabled?

PR #2438, opened 2024-03-24
fix: Use the new ext_elementtype extension to detect floating window borders and winbars

fredizzimo:
This fixes the issues with winbar and floating bar scrolling, by using the upcoming `ext_elementtype` UI extension. The corresponding PR is here:
* https://github.com/neovim/neovim/pull/28016

It will also need a new version of `nvim-rs` before it can be merged.

Currently it does not work properly with nvim 0.9.x, but it will be fixed to revert to the old behaviour when that is detected.

The API information stuff should also probably be extracted to its own separate PR.

- No


Kethku:
Closing as the upstream extension has been rejected

Issue #2433, opened 2024-03-23
Handle blur setting on wayland


agraven:
**Is your feature request related to a problem? Please describe.**
As noted in `winit`'s documentation for [`Window::set_blur`](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_blur), the blur setting works for wayland desktops with support for the `org_kde_kwin_blur_manager` protocol, i.e. at the very least KDE.

The `g:neovide_window_blurred` setting is currently only applied to the window in MacOS, but can be applied in the above cases as well. I would suggest calling `set_blur` on the winit window based on the neovide setting unconditionally, since it will at worst be a no-op.  

**Describe the solution you'd like**
Apply the `set_blur` winit setting at least on wayland, and ideally everywhere 

**Describe alternatives you've considered**
I suppose things could be kept as is

**Additional context**
Here's an example of what it looks like on KDE with neovide patched to unconditionally enable blur and with transparency set to `0.9`

![Screenshot_20240323_163547](https://github.com/neovide/neovide/assets/23525639/3ccde8a5-2d97-490f-8809-66095a3e3a5e)

fredizzimo:
We have a couple of issues open already:
* https://github.com/neovide/neovide/issues/2078
* https://github.com/neovide/neovide/issues/1936

And a start of an implementation here:
* https://github.com/neovide/neovide/pull/2104

So I think I will close this issue.

PR #2440, opened 2024-03-24
feat: Apply winit window blur setting on all platforms

agraven:
Note that the window will only actually be blurred if the underlying platform supports it

Could be considered either of these
- Fix
- Feature

- No

### DIFF
--- a/src/window/macos.rs
+++ b/src/window/macos.rs
@@ -227,7 +227,6 @@ impl MacosWindowFeature {
             background_color,
             show_border,
             transparency,
-            window_blurred,
             ..
         } = SETTINGS.get::<WindowSettings>();
         match background_color.parse::<Color>() {
@@ -236,8 +235,6 @@ impl MacosWindowFeature {
             }
             _ => self.update_ns_background(transparency, show_border),
         }
-        let opaque = transparency >= 1.0;
-        window.set_blur(window_blurred && !opaque);
     }
 
     pub fn handle_settings_changed(&self, window: &Window, changed_setting: WindowSettingsChanged) {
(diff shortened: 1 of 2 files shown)


fredizzimo:
I think this needs to be done slightly differently on Windows, but if you don't have Windows, it can be done separately
* https://github.com/rust-windowing/winit/pull/3257

agraven:
That feature doesn't appear to have been actually been released yet. But since it appears to "replace" the blur setting, maybe enabling the blur in windows as well should be reconsidered.

fredizzimo:
Ah, yes it looks like only bugfixes goes to the 0.29 branch. So, I guess the windows support has to wait until 0.30 is out. We could also use @Kethku's implementation on Windows https://github.com/neovide/neovide/pull/2104 until then.

src/window/window_wrapper.rs
```diff
@@ -137,7 +146,7 @@ impl WinitWindowWrapper {
             font_changed_last_frame: false,
             saved_inner_size,
             saved_grid_size: None,
-            ime_enabled,
+            ime_enabled: input_ime,
```
fredizzimo, line 149: This can't be moved without creating a proper new function for the window wrapper, since it does some additional initialization.

But I think that since we now have a proper settings notification system, we wouldn't need to store the `ime_enabled` state in the `WindowWinitWrapper` at all.

src/window/macos.rs
```diff
@@ -246,21 +243,14 @@ impl MacosWindowFeature {
...
-            #[cfg(target_os = "macos")]
             WindowSettingsChanged::Transparency(transparency) => {
                 log::info!("transparency changed to {}", transparency);
                 self.update_background(window, true);
             }
-            #[cfg(target_os = "macos")]
-            WindowSettingsChanged::WindowBlurred(window_blurred) => {
-                log::info!("window_blurred changed to {}", window_blurred);
-                self.update_background(window, true);
-            }
```
fredizzimo, line 263: I think we still need to keep this to call `update_background` , I seem to remember that was required from the discussions here:
* https://github.com/neovide/neovide/pull/2168

@crupest might know better

agraven: Right, update_ns_background still uses it, it's just updating the winit setting that's superfluous.

fredizzimo (approved):
I haven't tested this locally, but the code looks good now. We should also probably test that it still works on macOS before merging.

fredizzimo:
I just tested this on KDE plasma 6.0.2, and it works fine. At first, I was confused, because it only seemed to have a very slight effect of changing the background color, but it turned out that the blur strength was turned to the maximum on my system, but I don't think I have changed any defaults. Anway changing that to something less strong, works well.

agraven:
I also tested it on plasma 5, posted a screenshot in #2433

src/window/macos.rs
crupest: Actually this was introduced in #2351 by @emonadeo . I'm not really quite familiar with the codes here.

crupest:
Works good on my mac. Though I don't use blur before.

Screenshot 2024-03-26 at 22 40 14

falcucci (approved):

fredizzimo:
Thank you!

PR #2441, opened 2024-03-25
feat: tabs option as ENV variable and config file position + prefix cleanup for CLI argument

### LINKED ISSUE #2348: `no-tabs = true` in `~/.config/neovide/config.toml` does not work
**Describe the bug**

Neovide opens files in tabs. I dislike this and want to disable it permanently. Documentation suggests that I can put `no-tabs = true` in `~/.config/neovide/config.toml` and it will act as if I passed `--no-tabs` on the command line.

It does not.

**To Reproduce**

1. `neovide *.md` in a directory with multiple markdown files. Observe that the files are open in tabs.
2. Add `~/.config/neovide/config.toml` containing `no-tabs = true`.
3. `neovide *.md` again. Observe that the files still open in tabs.

**Expected behavior**

Files do not open in tabs and do not waste horizontal space on the vim "tab bar".

- OS: macOS 14.3 23D56 arm64
- Neovide 0.12.2
- nvim 0.9.5

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/14205105/neovide_rCURRENT.log)

**Additional context**

Trying to use `~/Library/Application Support/neovide/config.toml` does not work, either.

nazriel:
- Feature
- Documentation

- No

Fixes #2348

### DIFF
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -79,7 +79,7 @@ pub struct CmdLineSettings {
     pub idle: bool,
 
     /// Disable opening multiple files supplied in tabs (they're still buffers)
-    #[arg(long = "no-tabs")]
+    #[arg(long = "no-tabs", env = "NEOVIDE_NO_TABS", value_parser = FalseyValueParser::new())]
     pub no_tabs: bool,
 
     /// Request sRGB when initializing the window, may help with GPUs with weird pixel
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -78,9 +78,13 @@ pub struct CmdLineSettings {
     #[arg(long = "no-idle", env = "NEOVIDE_IDLE", action = ArgAction::SetFalse, value_parser = FalseyValueParser::new())]
     pub idle: bool,
 
+    /// Enable opening multiple files supplied in tabs [DEFAULT]
+    #[arg(long = "tabs", env = "NEOVIDE_TABS", action = ArgAction::SetTrue, default_value = "1", value_parser = FalseyValueParser::new())]
+    pub tabs: bool,
+
     /// Disable opening multiple files supplied in tabs (they're still buffers)
-    #[arg(long = "no-tabs", env = "NEOVIDE_NO_TABS", value_parser = FalseyValueParser::new())]
-    pub no_tabs: bool,
+    #[arg(long = "no-tabs", action = ArgAction::SetTrue, value_parser = FalseyValueParser::new())]
+    _no_tabs: bool,
 
     /// Request sRGB when initializing the window, may help with GPUs with weird pixel
     /// formats. Default on Windows.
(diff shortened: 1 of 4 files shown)


src/settings/config.rs
```diff
@@ -105,6 +106,9 @@ impl Config {
         if let Some(title_hidden) = &self.title_hidden {
             env::set_var("NEOVIDE_TITLE_HIDDEN", title_hidden.to_string());
         }
+        if let Some(no_tabs) = &self.no_tabs {
+            env::set_var("NEOVIDE_NO_TABS", no_tabs.to_string());
```
fredizzimo, line 110: It's more logical if the environment variable is `NEOVIDE_TABS`, and you can set `0` and `1`. We do the same for other similar variables like `NEOVIDE_VSYNC=0/1`.

We also try to make the actual argument accept both forms `--tabs` and `--no-tabs` for example. That way you can always temporarily overwrite the environment variable or configuration file setting from the command line.

nazriel: There seems to be a little bit of inconsistency.
Some parameters have form of title-bar=on/off, another no-fork=on/off.

I feel same way about `NEOVIDE_TABS` but my idea was not to change convention already in place and just propagate option.

So, is fork=on/off, idle=on/off, tabs=on/off the way moving forward?

nazriel: Yes but I think docs still do not reflect most of it, for example:
https://github.com/neovide/neovide/blob/main/website/docs/command-line-reference.md?plain=1#L115 - no information about --no-fork vs --fork

Anyways, I changed it like you suggested. 
Let me know what you think.

BTW, I don't like the current precedence (CLI arg > config_file > ENV) where config_file has precedence over env variable.

fredizzimo: I agree with both of your points, the documentation could be clearer, we should fix that at some point. 

And also, that the order should be CLI > ENV > config instead, but I'm not sure how much of a breaking change that would be. I somehow missed that the order was what it is when the config file support was added.

src/cmd_line.rs
```diff
@@ -159,14 +163,9 @@ impl Default for CmdLineSettings {
...
-    let maybe_tab_flag = (!cmdline.no_tabs).then(|| "-p".to_string());
-
-    cmdline.neovim_args = maybe_tab_flag
-        .into_iter()
-        .chain(mem::take(&mut cmdline.files_to_open))
-        .chain(cmdline.neovim_args)
-        .collect();
+    if cmdline._no_tabs {
+        cmdline.tabs = false;
+    }
```
fredizzimo, line 168: I had to check if clap would support this out of the box now, but apparently, it's still unsupported.
https://github.com/clap-rs/clap/issues/815. 

It looks like we might be able to refactor our code to something like that workaround though, and also deal with the options and environment variables at the same time. Maybe even autogenerate the config file support or something.

But that's not something for this PR.

fredizzimo (approved):

falcucci (approved):
LGTM

PR #2442, opened 2024-03-25
feat: render same z-index together

TwIStOy:
A preliminary implementation was previously discussed in issue #2408. Close #2379.

- Feature

- No

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,8 @@
 .DS_Store
 website/book
 *.AppImage
+.envrc
+flake.nix
+flake.lock
+.direnv/
+.devenv/
--- a/.gitignore
+++ b/.gitignore
@@ -6,8 +6,3 @@
 .DS_Store
 website/book
 *.AppImage
-.envrc
-flake.nix
-flake.lock
-.direnv/
-.devenv/
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -4,6 +4,7 @@ pub mod fonts;
 pub mod grid_renderer;
 pub mod opengl;
 pub mod profiler;
+mod rendered_layer;
 mod rendered_window;
 mod vsync;
 
@@ -16,6 +17,7 @@ use std::{
     sync::Arc,
 };
 
+use itertools::Itertools;
 use log::error;
 use skia_safe::{Canvas, Point, Rect};
 use winit::{
@@ -29,6 +31,7 @@ use crate::{
     dimensions::Dimensions,
     editor::{Cursor, Style},
     profiling::{tracy_create_gpu_context, tracy_named_frame, tracy_zone},
+    renderer::rendered_layer::{group_windows, FloatingLayer},
     settings::*,
     window::{ShouldRender, UserEvent},
     WindowSettings,
@@ -64,6 +67,7 @@ pub struct RendererSettings {
     debug_renderer: bool,
     profiler: bool,
     underline_stroke_scale: f32,
+    group_zindex_step: u64,
 }
 
 impl Default for RendererSettings {
@@ -82,6 +86,7 @@ impl Default for RendererSettings {
             debug_renderer: false,
             profiler: false,
             underline_stroke_scale: 1.,
+            group_zindex_step: 5,
         }
     }
 }
@@ -175,7 +180,10 @@ impl Renderer {
             root_canvas.clip_rect(clip_rect, None, Some(false));
         }
 
-        let windows: Vec<&mut RenderedWindow> = {
+        let settings = SETTINGS.get::<RendererSettings>();
+        let zindex_step = settings.group_zindex_step;
+
+        let (root_windows, floating_layers) = {
             let (mut root_windows, mut floating_windows): (
                 Vec<&mut RenderedWindow>,
                 Vec<&mut RenderedWindow>,
@@ -187,28 +195,80 @@ impl Renderer {
 
             root_windows
                 .sort_by(|window_a, window_b| window_a.id.partial_cmp(&window_b.id).unwrap());
-
             floating_windows.sort_by(floating_sort);
 
-            root_windows.into_iter().chain(floating_windows).collect()
+            let mut floating_layers = vec![];
+
+            let mut base_zindex = 0;
+            let mut current_windows = vec![];
+
+            for window in floating_windows {
+                let zindex = window.anchor_info.as_ref().unwrap().sort_order;
+                log::debug!("zindex: {}, base: {}", zindex, base_zindex);
+                if zindex - base_zindex > zindex_step {
+                    if !current_windows.is_empty() {
+                        for windows in group_windows(current_windows, font_dimensions) {
+                            floating_layers.push(FloatingLayer {
+                                sort_order: base_zindex,
+                                windows,
+                            });
+                        }
+                        current_windows = vec![];
+                    }
+                    base_zindex = zindex;
+                }
+                current_windows.push(window);
+            }
+            if !current_windows.is_empty() {
+                for windows in group_windows(current_windows, font_dimensions) {
+                    floating_layers.push(FloatingLayer {
+                        sort_order: base_zindex,
+                        windows,
+                    });
+                }
+            }
+            for layer in &mut floating_layers {
+                layer.windows.sort_by(floating_sort);
+                log::debug!(
+                    "layer: {:?}",
+                    layer
+                        .windows
+                        .iter()
+                        .map(|w| (w.id, w.anchor_info.as_ref().unwrap().sort_order))
+                        .collect_vec()
+                );
+            }
+
+            (root_windows, floating_layers)
         };
 
-        let settings = SETTINGS.get::<RendererSettings>();
-        let mut floating_rects = Vec::new();
-
-        self.window_regions = windows
+        let root_window_regions = root_windows
             .into_iter()
             .map(|window| {
                 window.draw(
                     root_canvas,
+                    default_background.with_a((255.0 * transparency) as u8),
+                    font_dimensions,
+                )
+            })
+            .collect_vec();
+
+        let floating_window_regions = floating_layers
+            .into_iter()
+            .flat_map(|mut layer| {
+                layer.draw(
+                    root_canvas,
                     &settings,
                     default_background.with_a((255.0 * transparency) as u8),
                     font_dimensions,
-                    &mut floating_rects,
                 )
             })
-            .collect();
+            .collect_vec();
 
+        self.window_regions = root_window_regions
+            .into_iter()
+            .chain(floating_window_regions)
+            .collect();
         self.cursor_renderer
             .draw(&mut self.grid_renderer, root_canvas);
 
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -86,7 +86,7 @@ impl Default for RendererSettings {
             debug_renderer: false,
             profiler: false,
             underline_stroke_scale: 1.,
-            group_zindex_step: 5,
+            group_zindex_step: 3,
         }
     }
 }
@@ -205,7 +205,7 @@ impl Renderer {
             for window in floating_windows {
                 let zindex = window.anchor_info.as_ref().unwrap().sort_order;
                 log::debug!("zindex: {}, base: {}", zindex, base_zindex);
-                if zindex - base_zindex > zindex_step {
+                if zindex - base_zindex >= zindex_step {
                     if !current_windows.is_empty() {
                         for windows in group_windows(current_windows, font_dimensions) {
                             floating_layers.push(FloatingLayer {
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -86,7 +86,7 @@ impl Default for RendererSettings {
             debug_renderer: false,
             profiler: false,
             underline_stroke_scale: 1.,
-            group_zindex_step: 3,
+            group_zindex_step: 2,
         }
     }
 }
--- a/src/renderer/rendered_layer.rs
+++ b/src/renderer/rendered_layer.rs
@@ -149,6 +149,7 @@ impl<'w> FloatingLayer<'w> {
         let save_layer_rec = SaveLayerRec::default().bounds(&bound_rect).paint(&paint);
 
         root_canvas.save_layer(&save_layer_rec);
+        root_canvas.clear(default_background.with_a(255));
 
         let regions = self
             .windows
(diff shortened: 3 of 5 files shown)


TwIStOy:
@fredizzimo Could you please review it? I believe the pr's functionality has been basically implemented.

Now, `noice.nvim` hover doc and `telescope` looks like:

CleanShot 2024-03-27 at 11 19 00

CleanShot 2024-03-27 at 11 19 16

Kethku:
This is awesome. I'm excited to dig in

fredizzimo:
The linter error should be gone if you rebase to master.

falcucci:
I'll test it properly soon and do some troubleshooting but a quick trial makes neovide breaks during the launching here on macOS.

TwIStOy:
> The linter error should be gone if you rebase to master.

Rebased!

fredizzimo:
Thanks, I will review this later today or tomorrow.

src/renderer/rendered_layer.rs
```diff
@@ -0,0 +1,608 @@
...
+
+    pub fn uniform_background_blend(&self) -> u8 {
+        self.windows
+            .iter()
+            .filter_map(|window| window.get_smallest_blend_value())
+            .min()
+            .unwrap_or(0)
+    }
+
+    fn _draw_shadow(&self, root_canvas: &Canvas, path: &Path, settings: &RendererSettings) {
```
Kethku, line 195: nit: Why is an underscore prefix used here? If its to indicate that this function is internal, I think it should be removed since it doesn't match the style for the rest of the project (barring the profiling code where this pattern has snuck in)

fredizzimo: BTW, for the profiling code, there's a reason. The `_LocationData` and `_Zone` have to be public but are only supposed to be used indirectly through the macros.

src/renderer/mod.rs
```diff
@@ -52,6 +55,7 @@ pub struct RendererSettings {
     debug_renderer: bool,
     profiler: bool,
     underline_stroke_scale: f32,
+    group_zindex_step: u64,
```
Kethku, line 58: This new setting needs documentation added here: https://github.com/neovide/neovide/blob/main/website/docs/configuration.md#underline-automatic-scaling

Kethku:
Ok this PR is a big improvement, I'm excited to get it merged.

One thing that stands out though is that as it is built currently if a window doesn't overlap at a given index but is directly adjacent, they will still draw shadows that overlap leading to some somewhat odd effects. As an example, with nvim-notify:
image

Imho we should render all of the windows at a given z-index all at once with the same shadow regardless of whether they actually intersect. That way theres no overlapping shadows and the clip used to render the shadow applies to the whole layer. I think its possible this approach would simplify the code a bit as well.

src/renderer/mod.rs
```diff
@@ -70,6 +74,7 @@ impl Default for RendererSettings {
             debug_renderer: false,
             profiler: false,
             underline_stroke_scale: 1.,
+            group_zindex_step: 3,
```
fredizzimo, line 77: I there a reason why we need this, or at least why we need to have it as 3 as default. I would prefer if we could limit the effects to the same z-index. 

This might mean that some plugins have to change slightly, but if it's possible it's still beter IMO, since any hardcoded numbers have the potential to cause problems.

fredizzimo (commented):
I have not yet checked the silhouette code in detail. But the rest looks good IMO. Did you work out the algorithms from yourself, or did you borrow it from somewhere?

Kethku:
In testing this locally I encountered some weird rendering artifacts.

1. Occasionally in ways I haven't been able to reproduce yet, entire layers will flicker or stop rendering which is very jaring
2. Similarly hard to reproduce, I encountered floating windows getting clipped and was able to capture a gif of it.
a22100e2-dd13-41ea-839c-fe20e4888fa3
Suspiciously the shadow doubles up when the graphical error is happening which says to me maybe the clip path is getting computed incorrectly?

TwIStOy:
> Imho we should render all of the windows at a given z-index all at once with the same shadow regardless of whether they actually intersect. That way theres no overlapping shadows and the clip used to render the shadow applies to the whole layer. I think its possible this approach would simplify the code a bit as well.

The shadows should consistently fall on the contours of the polygons created by the overlap of multiple windows. A group of non-overlapping windows has the potential to form multiple contours, complicating the computation process.
      
>    Suspiciously the shadow doubles up when the graphical error is happening which says to me maybe the clip path is getting computed incorrectly?

I've refactored the part of the code responsible for calculating the contours. This issue should now be resolved.

src/renderer/mod.rs
TwIStOy: In fact, this value could just be set to 2, as the z-index difference between the border and content windows in nui.nvim is 1. This parameter should be customizable, considering there may be cases where users layer windows with more z-index levels, and we should support such cases.

src/renderer/mod.rs
TwIStOy: Add docs!

.gitignore
```diff
@@ -6,3 +6,8 @@
 .DS_Store
 website/book
 *.AppImage
+.envrc
+flake.nix
+flake.lock
+.direnv/
+.devenv/
```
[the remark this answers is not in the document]
TwIStOy: Remove these lines~

TwIStOy:
> I have not yet checked the silhouette code in detail. But the rest looks good IMO. Did you work out the algorithms from yourself, or did you borrow it from somewhere?

The algorithm used to determine whether two regions fall within the same group is a [disjoint-set algorithm](https://en.wikipedia.org/wiki/Disjoint-set_data_structure). Everything else was developed by me.

falcucci:
I am still enable to test on macOS. Panicking at launching still.

```
2024-03-06 22:17:13 - Neovide panicked with the message 'failed printing to stdout: Resource temporarily unavailable (os error 35)'. (File: library/std/src/io/stdio.rs; Line: 1055, Column: 9)
```

it seems to be a concurrency issue.

fredizzimo:
@falcucci, that seems a bit unrelated, are you able to determine where it comes from? Maybe one of the new `log::debug!`statements?  That might be too verbose anyway, our logging is not really configured correctly IMO, we always log with max verbosity by default.

Or perhaps just the fact that the tests are using `println`. But other than that, I don't see anything that could directly cause it here, but it might have changed some timings of course.

falcucci:
@fredizzimo when I mean test it I don't mean running the tests itself, but compiling this branch. Neovide launches and panicks.

I'll paste the full log trace after lunch time.

fredizzimo:
Yes, and I think that's not something directly related to this branch, even if it triggers it. It might just expose some old bug.

falcucci:
Okay, got it.

Anyway, it means that we need to fix this unrelated bug to make this viable at least on macOS. Otherwise I can't really replicate those improvements.

let's talk about it somewhere else

falcucci:
the log trace just to register.

[neovide_backtraces.log](https://github.com/neovide/neovide/files/14828927/neovide_backtraces.log)

fredizzimo:
@falcucci, it lokks like you have a custom `nvim-rs` `/Users/falcucci/clones/nvim-rs/src/rpc/model.rs:60:5`

That line does not try to print anything on master https://github.com/KillTheMule/nvim-rs/blob/master/src/rpc/model.rs

falcucci:
@fredizzimo sorry, out-dated log file. I've updated the comment.

src/renderer/rendered_window.rs
```diff
@@ -684,6 +670,63 @@ impl RenderedWindow {
         self.scroll_delta = 0;
     }
 
+    fn map_lines<F>(&self, mut f: F)
```
Kethku, line 673: Theres a todo here that needs replaced. But also I think this function should be rewritten and renamed to iter_lines. It should look something like this:
```rust
fn iter_lines<'a>(&'a self) -> Box<dyn Iterator<Item = &'a Rc<RefCell<Line>>> + 'a> {
    let scroll_offset_lines = self.scroll_animation.position.floor() as isize;
    let height = self.grid_size.height as isize;
    if height == 0 {
        return Box::new(std::iter::empty());
    }

    let top_border_lines = self
        .actual_lines
        .iter_range(0..self.viewport_margins.top as isize)
        .flatten();
    let actual_line_count = self.actual_lines.len() as isize;
    let bottom_border_lines = self
        .actual_lines
        .iter_range(
            actual_line_count - self.viewport_margins.bottom as isize..actual_line_count,
        )
        .flatten();

    if self.scrollback_lines.is_empty() {
        Box::new(top_border_lines.chain(bottom_border_lines))
    } else {
        let scrollable_lines = self
            .scrollback_lines
            .iter_range(scroll_offset_lines..scroll_offset_lines + height + 1)
            .flatten();

        Box::new(
            top_border_lines
                .chain(scrollable_lines)
                .chain(bottom_border_lines),
        )
    }
}
```

Bonus points if another similar function is introduced called iter_scrollback_lines which is identical but targets the scrollback_lines rather than actual lines and can be used in the prepare_lines function

Kethku:
> 2. Similarly hard to reproduce, I encountered floating windows getting clipped and was able to capture a gif of it.
>    a22100e2-dd13-41ea-839c-fe20e4888fa3
>          
>            [
>              
>            
>                a22100e2-dd13-41ea-839c-fe20e4888fa3
>              ](https://private-user-images.githubusercontent.com/3323631/318228611-99687189-fa58-4dcf-b8fd-a724b7260266.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTIwOTI1MjcsIm5iZiI6MTcxMjA5MjIyNywicGF0aCI6Ii8zMzIzNjMxLzMxODIyODYxMS05OTY4NzE4OS1mYTU4LTRkY2YtYjhmZC1hNzI0YjcyNjAyNjYuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDQwMiUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA0MDJUMjExMDI3WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9ZTNlMmMyZmZhYzAxZGY3NGRiN2NmYzRjODdiYTJjNTg5OGM3ZWU2ZGUwNjlmNWY3MjcyOWQ1MjRhOWZhYzc3MyZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.t7hD5XH3aA1DGKJKaLy6iDmlwMKDeWfv1UjrBe_icTo)
>            
>            
>              
>                
>                  
>                
>                
>                  
>                  
>                
>              
>              [
>                
>                  
>                
>              ](https://private-user-images.githubusercontent.com/3323631/318228611-99687189-fa58-4dcf-b8fd-a724b7260266.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTIwOTI1MjcsIm5iZiI6MTcxMjA5MjIyNywicGF0aCI6Ii8zMzIzNjMxLzMxODIyODYxMS05OTY4NzE4OS1mYTU4LTRkY2YtYjhmZC1hNzI0YjcyNjAyNjYuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDQwMiUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA0MDJUMjExMDI3WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9ZTNlMmMyZmZhYzAxZGY3NGRiN2NmYzRjODdiYTJjNTg5OGM3ZWU2ZGUwNjlmNWY3MjcyOWQ1MjRhOWZhYzc3MyZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.t7hD5XH3aA1DGKJKaLy6iDmlwMKDeWfv1UjrBe_icTo)
>            
>          
>    Suspiciously the shadow doubles up when the graphical error is happening which says to me maybe the clip path is getting computed incorrectly?

This issue still happens even with above changes

TwIStOy:
> This issue still happens even with above changes

The strange thing is, why is the word under the cursor also blurred? Is there a special floating window there? Could you attach the log of this part?

https://github.com/TwIStOy/neovide/blob/5659b80dbfdea43f3af2d744be60334d2e161859/src/renderer/mod.rs#L230-L240

src/renderer/rendered_window.rs
TwIStOy: Update!

Kethku:
> > This issue still happens even with above changes
> 
> The strange thing is, why is the word under the cursor also blurred? Is there a special floating window there? Could you attach the log of this part?
> 
> https://github.com/TwIStOy/neovide/blob/5659b80dbfdea43f3af2d744be60334d2e161859/src/renderer/mod.rs#L230-L240

That blur is unrelated to the issue and a part of nvim saga which I use for lsp features. The issue I'm keying on is the strange cut off floating window. I'll try to get a log of the relevant part in a bit.

Kethku:
However thats a super good point and probably the cause of the problem in the sense that there are two floating windows that are adjacent to each other. So in that sense it is related

fredizzimo:
To me it looks like the area covered is not rectangular, there's the actual doc window, and the symbol it refers to. 

But the blurring effect, or some of the other effects can only deal with rectangles rendered as two triangles. But because of this non-rectangular area, one of the triangles gets rendered on the symbol, and the other one on the documentation window.

Kethku:
I think I'll pull the branch down and write some basic tests for it. If my hunch is correct it should be pretty easy to write a failing test case which should reveal some things

Kethku:
Closing in favor of https://github.com/neovide/neovide/pull/2467. Will continue work on it there

PR #2443, opened 2024-03-25
chore: Upgrade dependencies and remove unused

agraven:
Three things in life are inevitable: death, taxes, and dependency upgrades.

All crates with breaking version bumps had all of their usages carefully referenced against relevant change logs to ensure no unexpected (documented) changes in behavior were introduced.  

Some of the removed deps are still in the crate graph, but are not direct dependencies, and the additional constraints on their versions does not appear to achieve anything useful.

I refrained from touching some crates like `skia-safe` and dependencies specific to non-linux platforms

- Other

- No (correct me if I'm wrong on this one, not sure if you consider major dependency upgrades breaking for an application)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,17 +31,17 @@ anyhow = { version = "1.0.75", features = ["backtrace"] }
 async-trait = "0.1.53"
 backtrace = "0.3.67"
 clap = { version = "4.4.18", features = ["cargo", "derive", "env", "color"] }
-copypasta = "0.8.1"
+copypasta = "0.10.1"
 csscolorparser = "0.6.2"
-derive-new = "0.5.9"
+derive-new = "0.6.0"
 dirs = "5.0.0"
-flexi_logger = { version = "0.22.3", default-features = false }
+flexi_logger = { version = "0.28.0", default-features = false }
 futures = "0.3.21"
 gl = "0.14.0"
 glutin = "0.31.1"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -55,28 +55,26 @@ rmpv = "1.0.0"
...
-winit = { version = "=0.29.10", features = ["serde"] }
+which = "6.0.1"
+winit = { version = "=0.29.15", features = ["serde"] }
 xdg = "2.4.1"
 notify = "6.1.1"
 regex = "1.10.3"
 
 [dev-dependencies]
 approx = "0.5.1"
-mockall = "0.11.0"
```
fredizzimo, line 77: Good catch, it looks like this was not even used when it was introduced here:
* https://github.com/neovide/neovide/pull/261

fredizzimo (approved):
No, this is great, we should do this more often, and maybe even start using something like [dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates). We already use it for security alerts, and this PR fixes the current alert about the `mio` package that we have there. 

The changes also look good, but I think we will leave this open for a while to allow for more testing before merging.

Cargo.toml
agraven: You could also consider running cargo-udeps in your CI, it caught some of the unused ones for me.

fredizzimo:
Thank you!

PR #2454, opened 2024-03-30
fix: Maximize window on creation on linux

agraven:
The code notes a flickering issue when creating the window in a maximized state, but I haven't been able to reproduce this flicking on any Linux environment. I tested GNOME, KDE, and sway, and none of them exhibited any flickering. I therefore changed the behavior so that the window is maximized on creation in accordance with persistent window settings or config file/command.

This also resolves the mild annoyance of the window existing in a blank, non-maximized state for a short while before initialization is finished, since [set_visible](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_visible) has no effect on Wayland

TODO: More X11 testing

- Fix

- No

### DIFF
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -142,8 +142,15 @@ pub fn create_window(
         Some(PersistentWindowSettings::Windowed { position, .. }) => Some(position),
         _ => None,
     };
+    #[cfg(target_os = "linux")]
+    let maximized = matches!(window_settings, Some(PersistentWindowSettings::Maximized))
+        || cmd_line_settings.geometry.maximized;
+    // Unfortunately we can't maximize here, because winit shows the window momentarily causing
+    // flickering
+    #[cfg(not(target_os = "linux"))]
+    let maximized = false;
 
-    log::trace!("Settings initial_window_size {:?}", initial_window_size);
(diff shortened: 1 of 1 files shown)


fredizzimo:
The reason why it's not maximized on creation, is because it creates a very noticeable white flash on Windows. There's a bug in winit that shows the window for one frame when set to hidden and maximized.

But even if we disable this on Windows, and merge your PR, it will only be a temporary fix, so I'm not sure if it's worth merging.

The next version of Winit will require us to create the window inside the event loop which also fixes some macOS bugs in the current version. So, our plan is to run the event loop and Neovim event processing without a window initially, and only create the window when it's time to show. That's the same place as our current code sets it as visible. That way we can create the window with the correct flags and attributes without any flickering.

Would you want to work on that instead?

agraven:
Sure thing, that definitely seems like a more ideal solution. I also managed to reproduce an unpleasant visual artifact when testing on plasma and gnome on X11, so for this to even make sense it would also require a runtime check that neovide is running under wayland, at which point this workaround becomes far more trouble than it's worth.

fredizzimo:
Since https://github.com/neovide/neovide/pull/2562 is now merged, I think we can close this one.

PR #2457, opened 2024-04-02
fix: set startup directory failed is not an error

TwIStOy:
Fix launching neovide with a "fake" neovim instance `neovide --neovim-bin nvim.dev.local`:

nvim.dev.local:
```bash
#!/bin/bash
ssh dev.local "bash -l -c \"nvim $@\"" 
```

- Fix

- No

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -98,7 +98,12 @@ pub async fn setup_tty_startup_directory(
         cmd = format!("{} | {}", cmd, handle_command_arg(pos, neovim_args));
     }
 
-    nvim.command(cmd.as_str()).await
+    match nvim.command(cmd.as_str()).await {
+        Ok(_) => {}
+        Err(e) => log::error!("Error setting startup directory: {}", e),
+    }
+
+    Ok(())
 }
 
 fn get_startup_directory(path: &str) -> Option<String> {


Kethku:
This seems reasonable. Though I am curious about what the usecase is for a fake neovim instance. Seems interesting

Kethku:
Seems good to me. Ty

fredizzimo:
It's a reasonable fix, but I think there's another bug hiding here. The startup directory is not supposed to be set based on the `--neovim-bin` argument, only based on actual files and directories passed as arguments. 

@falcucci

falcucci:
this is indeed a good missing error treatment.

> It's a reasonable fix, but I think there's another bug hiding here. The startup directory is not supposed to be set based on the --neovim-bin argument, only based on actual files and directories passed as arguments.

I think it has nothing with `--neovim-bin` and also the startup dir should follow for non tty only.

I am just not able to replicate it just creating that "fake" neovim instance as mentioned.

A startup path must be given somehow.

the following command wouldn't be even considered as a non tty:

```bash
neovide --neovim-bin nvim.dev.local
``` 

I have created an instance as a connected fake server and nothing wrong happened aswell,

It also depends about the platform. macOS would take $HOME in this case and execute it.

So I cannot suppose a hided bug here, we need to replicate it first.

PR #2461, opened 2024-04-03
feat: add support of Metal renderer on macOS.

crupest:
Add Metal renderer on macOS.

Currently this is a very naive implementation. I created this pr when the window at least shows things. I'm very unfamiliar with Metal development. So if you have any advice, I will extremely appreciate it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,13 +101,16 @@ windows = { version = "0.56.0", features = [
 ] }
 windows-registry = "0.2.0"
 
-[target.'cfg(not(target_os = "windows"))'.dependencies]
-skia-safe = { version = "0.75.0", features = ["gl", "textlayout"] }
-
 [target.'cfg(target_os = "macos")'.dependencies]
 objc2 = "0.5.2"
 objc2-foundation = { version = "0.2.2", features = [ "NSUserDefaults" ] }
-objc2-app-kit = { version = "0.2.2", features = [ "NSLayoutConstraint" ] }
+objc2-app-kit = { version = "0.2.2", features = [ "NSLayoutConstraint", "NSColorSpace", "objc2-quartz-core" ] }
+objc2-quartz-core = { version = "0.2.2", features = [ "objc2-metal", "CALayer", "CAMetalLayer" ] }
+objc2-metal = { version = "0.2.2", features = [ "MTLCommandQueue", "MTLCommandBuffer" ] }
+skia-safe = { version = "0.75.0", features = ["metal", "gl", "textlayout"] }
+
+[target.'cfg(not(any(target_os = "windows", target_os = "macos")))'.dependencies]
+skia-safe = { version = "0.75.0", features = ["gl", "textlayout"] }
 
 [target.'cfg(target_os = "windows")'.build-dependencies]
 winres = "0.1.12"
--- a/src/renderer/metal.rs
+++ b/src/renderer/metal.rs
@@ -1,0 +1,215 @@
+use objc2::{rc::Retained, runtime::ProtocolObject};
+use objc2_app_kit::NSColorSpace;
+use objc2_foundation::{CGFloat, CGSize};
+use objc2_metal::{
+    MTLCommandBuffer, MTLCommandQueue, MTLCreateSystemDefaultDevice, MTLDevice, MTLDrawable,
+    MTLTexture,
+};
+use objc2_quartz_core::{CAMetalDrawable, CAMetalLayer};
+use skia_safe::{
+    gpu::{
+        self,
+        mtl::{BackendContext, TextureInfo},
+        surfaces::wrap_backend_render_target,
+        DirectContext, SurfaceOrigin,
+    },
+    Canvas, ColorSpace, ColorType, Surface,
+};
+use winit::{event_loop::EventLoopProxy, window::Window};
+
+use crate::{
+    profiling::tracy_gpu_zone,
+    renderer::{SkiaRenderer, VSync},
+    window::{macos::get_ns_window, UserEvent},
+};
+
+struct MetalDrawableSurface {
+    pub _drawable: Retained<ProtocolObject<dyn CAMetalDrawable>>,
+    pub metal_drawable: Retained<ProtocolObject<dyn MTLDrawable>>,
+    pub surface: Surface,
+}
+
+impl MetalDrawableSurface {
+    fn new(
+        drawable: Retained<ProtocolObject<dyn CAMetalDrawable>>,
+        context: &mut DirectContext,
+    ) -> MetalDrawableSurface {
+        tracy_gpu_zone!("MetalDrawableSurface.new");
+
+        let texture = unsafe { drawable.texture() };
+        let texture_info = unsafe { TextureInfo::new(Retained::as_ptr(&texture).cast()) };
+        let backend_render_target = gpu::backend_render_targets::make_mtl(
+            (texture.width() as i32, texture.height() as i32),
+            &texture_info,
+        );
+
+        let metal_drawable =
+            unsafe { Retained::cast::<ProtocolObject<dyn MTLDrawable>>(drawable.clone()) };
+
+        let surface = wrap_backend_render_target(
+            context,
+            &backend_render_target,
+            SurfaceOrigin::TopLeft,
+            ColorType::BGRA8888,
+            ColorSpace::new_srgb(),
+            None,
+        )
+        .expect("Failed to create skia surface with metal drawable.");
+
+        MetalDrawableSurface {
+            _drawable: drawable,
+            metal_drawable,
+            surface,
+        }
+    }
+
+    fn metal_drawable(&self) -> &ProtocolObject<dyn MTLDrawable> {
+        &self.metal_drawable
+    }
+}
+
+pub struct MetalSkiaRenderer {
+    window: Window,
+    _device: Retained<ProtocolObject<dyn MTLDevice>>,
+    command_queue: Retained<ProtocolObject<dyn MTLCommandQueue>>,
+    metal_layer: Retained<CAMetalLayer>,
+    _backend: BackendContext,
+    context: DirectContext,
+    metal_drawable_surface: Option<MetalDrawableSurface>,
+}
+
+impl MetalSkiaRenderer {
+    pub fn new(window: Window, srgb: bool, vsync: bool) -> Self {
+        log::info!("Initialize MetalSkiaRenderer...");
+
+        let draw_size = window.inner_size();
+        let ns_window = get_ns_window(&window);
+
+        unsafe {
+            ns_window.setColorSpace(Some(
+                if srgb {
+                    NSColorSpace::sRGBColorSpace()
+                } else {
+                    NSColorSpace::deviceRGBColorSpace()
+                }
+                .as_ref(),
+            ));
+        }
+
+        let device = unsafe {
+            Retained::retain(MTLCreateSystemDefaultDevice())
+                .expect("Failed to create Metal system default device.")
+        };
+        let metal_layer = unsafe {
+            let metal_layer = CAMetalLayer::new();
+            metal_layer.setDevice(Some(&device));
+            metal_layer.setPresentsWithTransaction(false);
+            metal_layer.setFramebufferOnly(false);
+            metal_layer.setDisplaySyncEnabled(vsync);
+            metal_layer.setOpaque(false);
+
+            let ns_view = ns_window.contentView().unwrap();
+            ns_view.setWantsLayer(true);
+            ns_view.setLayer(Some(&metal_layer));
+
+            metal_layer
+                .setDrawableSize(CGSize::new(draw_size.width as f64, draw_size.height as f64));
+            metal_layer
+        };
+
+        let command_queue = device
+            .newCommandQueue()
+            .expect("Failed to create command queue.");
+
+        let backend = unsafe {
+            BackendContext::new(
+                Retained::as_ptr(&device).cast(),
+                Retained::as_ptr(&command_queue).cast(),
+            )
+        };
+
+        let context = gpu::direct_contexts::make_metal(&backend, None).unwrap();
+
+        MetalSkiaRenderer {
+            window,
+            _device: device,
+            metal_layer,
+            command_queue,
+            _backend: backend,
+            context,
+            metal_drawable_surface: None,
+        }
+    }
+
+    fn move_to_next_frame(&mut self) {
+        tracy_gpu_zone!("move_to_next_frame");
+
+        let drawable = unsafe {
+            self.metal_layer
+                .nextDrawable()
+                .expect("Failed to get next drawable of metal layer.")
+        };
+
+        self.metal_drawable_surface = Some(MetalDrawableSurface::new(drawable, &mut self.context));
+    }
+}
+
+impl SkiaRenderer for MetalSkiaRenderer {
+    fn window(&self) -> &Window {
+        &self.window
+    }
+
+    fn flush(&mut self) {
+        tracy_gpu_zone!("flush");
+
+        self.context.flush_and_submit();
+    }
+
+    fn swap_buffers(&mut self) {
+        tracy_gpu_zone!("swap buffers");
+
+        let command_buffer = self
+            .command_queue
+            .commandBuffer()
+            .expect("Failed to create command buffer.");
+        command_buffer.presentDrawable(
+            self.metal_drawable_surface
+                .as_mut()
+                .expect("No drawable surface now.")
+                .metal_drawable(),
+        );
+        command_buffer.commit();
+
+        self.metal_drawable_surface = None;
+    }
+
+    fn canvas(&mut self) -> &Canvas {
+        tracy_gpu_zone!("canvas");
+
+        self.move_to_next_frame();
+
+        self.metal_drawable_surface
+            .as_mut()
+            .expect("Not metal drawable surface now.")
+            .surface
+            .canvas()
+    }
+
+    fn resize(&mut self) {
+        tracy_gpu_zone!("resize");
+
+        let window_size = self.window.inner_size();
+        unsafe {
+            self.metal_layer.setDrawableSize(CGSize::new(
+                window_size.width as CGFloat,
+                window_size.height as CGFloat,
+            ));
+        }
+
+        self.window.request_redraw();
+    }
+
+    fn create_vsync(&self, _proxy: EventLoopProxy<UserEvent>) -> VSync {
+        VSync::MacosMetal()
+    }
+}
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -11,6 +11,9 @@ mod vsync;
 #[cfg(target_os = "windows")]
 pub mod d3d;
 
+#[cfg(target_os = "macos")]
+mod metal;
+
 use std::{
     cmp::Ordering,
     collections::{hash_map::Entry, HashMap},
@@ -498,6 +501,8 @@ pub enum WindowConfigType {
     OpenGL(glutin::config::Config),
     #[cfg(target_os = "windows")]
     Direct3D,
+    #[cfg(target_os = "macos")]
+    Metal,
 }
 
 pub struct WindowConfig {
@@ -521,7 +526,19 @@ pub fn build_window_config(
         }
     }
 
-    #[cfg(not(target_os = "windows"))]
+    #[cfg(target_os = "macos")]
+    {
+        let cmd_line_settings = SETTINGS.get::<CmdLineSettings>();
+        if cmd_line_settings.opengl {
+            opengl::build_window(window_attributes, event_loop)
+        } else {
+            let window = event_loop.create_window(window_attributes).unwrap();
+            let config = WindowConfigType::Metal;
+            WindowConfig { window, config }
+        }
+    }
+
+    #[cfg(not(any(target_os = "windows", target_os = "macos")))]
     {
         opengl::build_window(window_attributes, event_loop)
     }
@@ -549,6 +566,10 @@ pub fn create_skia_renderer(
         }
         #[cfg(target_os = "windows")]
         WindowConfigType::Direct3D => Box::new(d3d::D3DSkiaRenderer::new(window.window)),
+        #[cfg(target_os = "macos")]
+        WindowConfigType::Metal => {
+            Box::new(metal::MetalSkiaRenderer::new(window.window, srgb, vsync))
+        }
     };
     tracy_create_gpu_context("main_render_context", renderer.as_ref());
     renderer
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1478,6 +1478,8 @@ dependencies = [
  "objc2",
  "objc2-app-kit",
  "objc2-foundation",
+ "objc2-metal",
+ "objc2-quartz-core",
  "parking_lot",
  "rand 0.8.5",
  "raw-window-handle",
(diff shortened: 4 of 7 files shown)


Cargo.toml
```diff
@@ -98,12 +98,16 @@ winapi = { version = "0.3.9", features = [
...
 skia-safe = { version = "0.68.0", features = ["gl", "d3d", "textlayout"] }
 
-[target.'cfg(not(target_os = "windows"))'.dependencies]
-skia-safe = { version = "0.68.0", features = ["gl", "textlayout"] }
-
 [target.'cfg(target_os = "macos")'.dependencies]
 icrate = { version = "0.0.4", features = [ "apple", "Foundation", "Foundation_NSThread", "AppKit", "AppKit_NSColor", "AppKit_NSEvent", "AppKit_NSView", "AppKit_NSWindow", "AppKit_NSViewController", "AppKit_NSMenu", "AppKit_NSMenuItem", "AppKit_NSOpenPanel", "AppKit_NSScreen", "Foundation_NSArray" ] }
 objc2 = "0.4.1"
+metal = "0.27.0"
+skia-safe = { version = "0.68.0", features = ["metal", "gl", "textlayout"] }
```
fredizzimo, line 105: It think you should be able to use the standard `gl`, `textlayout`, feature here. We don't use anything from the `metal` and `gl` features, even if the rendering is done through that.

There's no `metal`, `gl`, `textlayout` binary combination as far as I can see https://github.com/rust-skia/skia-binaries/releases, so it would slow down the compilation.

crupest: Though there is no `metal`, `gl`, `textlayout` combination there, this is exactly what we want. Just like on Windows, there are two renderer, OpenGL and D3D. There are also two renderer, OpenGL and Metal on macOS. Without either one, it does not compile.

I don't know why there is no such combination. Maybe due to OpenGL is obsolete on macOS.

fredizzimo: I was probably wrong, and it's indeed needed. I remembered wrong how it was done on Window.

I guess we have to ask for binaries from `skia-safe` then.

falcucci:
@crupest I've done some changes to make skia-safe working at my MacBook targeting `aarch64-apple-darwin` to make it work properly as it tries to build from source.

The build from source definitely doesn't work at least for this platform, even their examples fails to build it so I have forked and it seems that *the latest skia-safe works well from **upstream** changes*.

I'll send the changes on this PR to help it make more compatible on both platforms I guess?

That's nice work and I am willing to merge til we finish our development at https://github.com/neovide/vide which will be also good to compare the performance and when we get it integrated afterwards.

In anyway I guess we will need to wait til the 1.79 skia-safe version to bump the release, for now let's use the upstream.

src/renderer/metal.rs
```diff
@@ -0,0 +1,208 @@
...
+
+        let device = unsafe {
+            Retained::retain(MTLCreateSystemDefaultDevice())
+                .expect("Failed to create Metal system default device.")
+        };
+        let metal_layer = unsafe {
+            let metal_layer = CAMetalLayer::new();
+            metal_layer.setDevice(Some(&device));
+            metal_layer.setPixelFormat(if srgb {
+                MTLPixelFormat::BGRA8Unorm_sRGB
```
falcucci, line 95: The awkward behavior I got here is that not setting it at all for sRGB solves the problem I had with non vivid colors so maybe it could be better to only disable sRGB if requested to and avoid the BGRA8Unorm_sRGB due to compilation errors.

At least on M3 silicon it doesn't identify as an available option (81)

```rust
if !srgb {
    metal_layer.setPixelFormat(MTLPixelFormat::BGRA8Unorm)
};
```

this would be enough

crupest: > At least on M3 silicon it doesn't identify as an available option (81)

I'm not sure. Based on https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf, M3 silicon should support `MTLPixelFormat::BGRA8Unorm_sRGB`.

The reason why I write this is simply that I guess it based on the **name**. Actually I have no clear idea about it. This is very naive, indeed.

I'm totally unfamiliar about sRGB things. And I need to spend some time on investigating it. But it may take some time for me.

crupest:
@falcucci, thanks for forwarding this.

> The build from source definitely doesn't work at least for this platform using 0.75.0, even their examples fails to build it so I have forked and it seems that the latest skia-safe works well from upstream changes.

The problem both I encountered and in CI is that Python recently released 3.13 version. I use homebrew and it upgrades Python to 3.13. In this version, it removes some deprecated module. However, skia of currently used version uses one of them, aka, `pipes` for building. I build it by rolling back Python version to 3.12 and it is then OK.

I'm not sure about your actual problem. But if you see some error message like `ModuleNotFoundError: No module named 'pipes'`. Then it is the same one. You can try rollback Python version (either by homebrew or a Python version manager) and build it again.

src/renderer/metal.rs
falcucci: what I am sure is that it is getting an error for `MTLPixelFormat::BGRA8Unorm_sRGB`
```bash
fatal error: "Invalid MTLPixelFormat: 81
```

seems that we should *not* change the pixel format, once we handle the sRGB on our skia [surface](https://github.com/neovide/neovide/pull/2461/files#diff-bd0d9f27119e48a1f8ca5898037d7f50ee5325f8418e884a749514b165919ce3R48-R55)

another thing is that at least for me the `BGRA8Unorm` is the default pixel format and it sounds correct, setting the pixel format again here, make it loose the skia surface we have applied before somehow.

falcucci: maybe we should even remove the `metal_layer.setPixelFormat`

falcucci:
@crupest the main problem to me was how they were handling the `ninja` dependency.

It seems they remove it from the built in injection and uses it as external now which makes it work perfectly from skia-safe master upstream.

falcucci:
@crupest seems that we don't have any breaking change on upstream changes, everything works smoothly to me, with no need for any python version management trick.

src/renderer/metal.rs
```diff
@@ -0,0 +1,208 @@
...
+            } else {
+                MTLPixelFormat::BGRA8Unorm
+            });
+            metal_layer.setPresentsWithTransaction(false);
+            metal_layer.setFramebufferOnly(false);
```
falcucci, line 100: I wonder if setting the `setFramebufferOnly(true)` would improve performance?

src/renderer/metal.rs
falcucci: Adding the sRGB support under the hood would be better as default on macOS 

```rust
unsafe {
    ns_window.setColorSpace(Some(NSColorSpace::deviceRGBColorSpace().as_ref()));
}
```

And I just tested again, and not setting the pixel format at all works perfectly fine here.

src/renderer/mod.rs
```diff
@@ -521,7 +526,19 @@ pub fn build_window_config(
         }
     }
 
-    #[cfg(not(target_os = "windows"))]
+    #[cfg(target_os = "macos")]
+    {
+        let cmd_line_settings = SETTINGS.get::<CmdLineSettings>();
+        if cmd_line_settings.opengl {
```
falcucci, line 532: we need some documentation for this

falcucci, line 532: also, we need some documentation for this

falcucci (changes requested):

9mm:
Random question: I'm not familiar with metal renderer. What kind of new cool stuff will this PR provide when running Neovide under macOS + Metal, vs what it is using now?

Cargo.toml
```diff
@@ -101,13 +101,16 @@ windows = { version = "0.56.0", features = [
...
-skia-safe = { version = "0.75.0", features = ["gl", "textlayout"] }
-
 [target.'cfg(target_os = "macos")'.dependencies]
 objc2 = "0.5.2"
 objc2-foundation = { version = "0.2.2", features = [ "NSUserDefaults" ] }
-objc2-app-kit = { version = "0.2.2", features = [ "NSLayoutConstraint" ] }
+objc2-app-kit = { version = "0.2.2", features = [ "NSLayoutConstraint", "objc2-quartz-core" ] }
+objc2-quartz-core = { version = "0.2.2", features = [ "objc2-metal", "CALayer", "CAMetalLayer" ] }
+objc2-metal = { version = "0.2.2", features = [ "MTLCommandQueue", "MTLCommandBuffer" ] }
+skia-safe = { version = "0.75.0", features = ["metal", "gl", "textlayout"] }
```
falcucci, line 110:
```suggestion
skia-safe = { git = "https://github.com/rust-skia/rust-skia.git", branch = "master", features = ["metal", "gl", "textlayout"] }
```

we can assume to target upstream and bump the `0.79.0` afterwards when available which is coming very soon, we are free on code changes from the client side which has no breaking changes and only improvements made by skia-safe team.

we can also setup our CI correctly with no need of python version regression to make it work.

falcucci (changes requested):

falcucci:
> Random question: I'm not familiar with metal renderer. What kind of new cool stuff will this PR provide when running Neovide under macOS + Metal, vs what it is using now?

@9mm it's a first class citizen choice. Apple has deprecated OpenGL as the official graphical api. Using Metal, apps on macOS ecosystem win by far when we speak about performance and smoothness due to the seamless integration on Apples specific hardware, which is the opposite of OpenGL purpose to be a generalist cross-platform.

Both are quite similar tho, but Metal brings you a more native experience due to a to low overhead and high performance under the hood.

That said, we are approaching here a more specific support to improve the macOS users experience with the option to choose OpenGL if needed, but in the future we are planing to integrate https://github.com/neovide/vide officially which abstracts most of the graphic apis available and make the user choose any of the options if *supported by its platform*. (must be discussed)

src/renderer/mod.rs
crupest: Do you mean in code or in document of neovide. I'll add the latter later.

src/renderer/metal.rs
crupest: I just simply follow `skia-safe` example at https://github.com/rust-skia/rust-skia/blob/f5d0181db8f7383539553f3daa4064c8f85cb993/skia-safe/examples/metal-window/main.rs#L158. I have no idea about this. But if you can prove this, we can change it to true definitely.

src/renderer/metal.rs
crupest: > maybe we should even remove the metal_layer.setPixelFormat

I think we still need to explicitly specify the pixel format and align it with the one of surface, which is specified at about line 52, rather than depend on the default behavior, which might be changed by Apple in the future. Though change of default is very unlike to happen, we explicitly specify it does no harm.

```
let surface = wrap_backend_render_target(
    context,
    &backend_render_target,
    SurfaceOrigin::TopLeft,
    ColorType::BGRA8888,
    ColorSpace::new_srgb(),
    None,
)
```

> Adding the sRGB support under the hood would be better as default on macOS

I think this change is good and I'll apply it later. But I still have no clear idea about this because of a lack of knowledge on Apple's API.

Cargo.toml
crupest: I'm afraid of whether it is proper to change the version to use git. This is much unstable and would lead to some building problem

crupest: And also probably a break of function suddenly without clear information on this.

crupest:
> @crupest the main problem to me was how they were handling the `ninja` dependency.
> 
> It seems they remove it from the built in injection and uses it as external now which makes it work perfectly from skia-safe master upstream.

@falcucci  OK. I always install ninja on my computer since it is a very fundamental tool of building code, so I might miss this problem.

crupest:
> Random question: I'm not familiar with metal renderer. What kind of new cool stuff will this PR provide when running Neovide under macOS + Metal, vs what it is using now?

@9mm For a machine that can run Neovide very fluently, there is no much difference. But for those with limit performance, especially laptop, this would improve a lot. The point is specifiy by @falcucci

> it's a first class citizen choice. Apple has deprecated OpenGL as the official graphical api.

src/renderer/metal.rs
crupest: And this is the comment in the example on this. I copy it here for convenience.

```rust
// Disabling this option allows Skia's Blend Mode to work.
// More about: https://developer.apple.com/documentation/quartzcore/cametallayer/1478168-framebufferonly
layer.set_framebuffer_only(false);
```

crupest: OK, I see. Set it to `false` (the default value) improves performance. The disadvantage is

> However, you may not sample, read from, or write to those textures.

So if we don't use such functions, then we can remove it and leave it `false` as default.

crupest:
@falcucci By the way, really thanks for your effort of investigation on these problems, when I have no much time to dive into them.

Cargo.toml
crupest: I mean in the future, not in current version.

src/renderer/metal.rs
crupest: > Adding the sRGB support under the hood would be better as default on macOS
> 
> ```rust
> unsafe {
>     ns_window.setColorSpace(Some(NSColorSpace::deviceRGBColorSpace().as_ref()));
> }
> ```
> 
> And I just tested again, and not setting the pixel format at all works perfectly fine here.

And according to https://developer.apple.com/documentation/quartzcore/cametallayer/1478170-colorspace, maybe it is better to set `colorspace` on `CAMetalLayer` instead of on `NSWindow`? Just my guess.

Cargo.toml
fredizzimo: Yes, we should not use the git version, only the official versions have binaries available, and that's what 99% of the people are using when building Neovide. 

However, we should remove the opengl support on macOS, or request for that combination from skia-safe, if we still want to support it. There are no binaries for that combination available here https://github.com/rust-skia/skia-binaries/releases. That will fix the build problems for @falcucci, I think, but if I remember correctly @crupest has proxy problems that prevents the usage of the pre-built binaries.

crupest: > but if I remember correctly @crupest has proxy problems that prevents the usage of the pre-built binaries.

Actually, no. The proxy problem is I failed to access Discord. And with a proper proxy to make network work, everything works as intended. Prebuilt binary for Windows will be downloaded directly without being built from scratch. The problem is still there is no OpenGL+Metal prebuilt binary.

> However, we should remove the opengl support on macOS, or request for that combination from skia-safe, if we still want to support it.

I would choose to request for this combination from skia-safe. Or at least, we can get the reason why this combination is missing if it is intentional.

falcucci:
> I always install ninja on my computer since it is a very fundamental tool of building code, so I might miss this problem

@crupest to be clear, ninja still must be installed but by yourself now, since they have removed it on upstream from their local `depot_tools` to avoid these compatibilities errors I had.

falcucci:
> By the way, really thanks for your effort of investigation on these problems, when I have no much time to dive into them.

@crupest thank you for the contribution! It is really stable for me as-well and erases many of the performance problems I had on long code sessions.

src/renderer/metal.rs
falcucci: we can leave it as false then, my question was also out of curiosity :)

src/renderer/mod.rs
falcucci: I mean in the command-line [documentation](https://neovide.dev/command-line-reference.html)

Cargo.toml
fredizzimo: One thing we could do is to make metal a feature flag instead of command line parameter, and release two versions of Neovide. That way we never need to build the unsupported configurations.

src/renderer/mod.rs
crupest: Got it. Will add it later.

Cargo.toml
falcucci: I wasn't clear with my request I guess.

*we are not merging using the git target version* 

this change is to take advantage of upstream changes and bump the release `0.79.0` when is available, so we can setup our CI correctly installing ninja with less efforts and tricks.

The change I would still request and @fredizzimo is right is about change the branch from master but add a pinned to a specific commit (the current latest).

Regarding the OpenGL deprecation on neovide, I think we should still offer both, in the end of the day the final user will install the .dmg or just the binary in their PATH and in case they need to build it from source is just a matter of install ninja and that's it.

src/renderer/metal.rs
fredizzimo: I think skia internally need to read from these when you show a semi-transparent floating window for example, so it needs to be set to false.

src/renderer/metal.rs
falcucci: As an example Alacrity is using `NSWindow` to specify it. https://github.com/alacritty/alacritty/pull/6602

falcucci: > I think we still need to explicitly specify the pixel format and align it with the one of surface, which is specified at about line 52, rather than depend on the default behavior, which might be changed by Apple in the future. Though change of default is very unlike to happen, we explicitly specify it does no harm.

specifying the `metal_layer.setPixelFormat` as `BGRA8Unorm_sRGB` get's an error as I mentioned. IMO applying `BGRA8Unorm_sRGB` get's no difference but try to fix that, otherwise just apply the `BGRA8Unorm` to disable the sRGB.

Cargo.toml
crupest: > One thing we could do is to make metal a feature flag instead of command line parameter, and release two versions of Neovide. That way we never need to build the unsupported configurations.

@fredizzimo This is good for us, developers, but causing a lot of trouble for users and delivering.

The users have to download two versions to test and then choose one. If they want to switch temporarily because of some bugs or other things, it is more inconvenient.

For delivering, like package manager homebrew, they have to deliver 2 versions with different names. It might confuse users a lot.

So I would rather to choose building skia and not use feature.

src/renderer/metal.rs
crupest: @falcucci 

> As an example Alacrity is using NSWindow to specify it. https://github.com/alacritty/alacritty/pull/6602

That's good. I'll apply it. And I will also read some codes of alacritty to make this behavior correct. It is a very nice way to handle this by following alacritty.

> specifying the metal_layer.setPixelFormat as BGRA8Unorm_sRGB get's an error as I mentioned. IMO applying BGRA8Unorm_sRGB get's no difference but try to fix that, otherwise just apply the BGRA8Unorm to disable the sRGB.

Sorry, I might mislead you. I mean set `BGRA8Unorm` always explicitly rather than rely on default behavior. Though very unlikely, the default value might change by Apple in the future. And this format must align with skia surface's one. Explicitly doing this makes it clearer.

Cargo.toml
falcucci: @crupest I agree about building skia, deliver two apps unfortunately is out of our intention so far.

I would take building skia as our prior solution but also *make a request to support both features to improve compilation time* but this is not a blocker for us.

falcucci: we can still deprecate OpenGL for macOS in another future version, and then use their binaries.

crupest: > I would take building skia as our prior solution but also make a request to support both features to improve compilation time but this is not a blocker for us.

@falcucci Agree with you.

But one thing to notice is that current CI failure is not about ninja. It is python version problem. But we can use an GitHub Action to set python version before building.

src/renderer/metal.rs
falcucci: setting `BGRA8Unorm` explicitly breaks the sRGB surface, and the colors are no vivid anymore.

fredizzimo: I'm don't know how it currently looks like, but for this PR, I think we should try to keep the colors exactly the same as before. 

We can make the `srgb` and `no-srgb` option do as expected in another PR. Probably so that `no-srgb` is the default, and emulates the old Neovide behaviour, while `srgb` has more dull colors, and matches the `iterm2` colors.

Cargo.toml
falcucci: the CI is currently failing because you are still using the `0.75.0` and not the upstream `0.79.0` and we need to install ninja anyway afterwards, there is no need to manage python version since it gets the latest.

steps to fix are probably:
- update skia to upstream using the last pinned commit;
- add ninja installation to our ci pipelines;

falcucci (changes requested):

Cargo.toml
falcucci: the CI is currently failing because you are still using the `0.75.0` and not the upstream `0.79.0` and we need to install ninja anyway afterwards, there is no need to manage python version since it gets the latest.

steps to fix are probably:
- update skia to upstream using the last pinned commit;
- add ninja installation to our ci pipelines;

src/renderer/metal.rs
falcucci: @fredizzimo Agree. that's why I am suggesting this

```rust
if !srgb {
    metal_layer.setPixelFormat(MTLPixelFormat::BGRA8Unorm)
};
```

to keep the default behavior.

we can troubleshoot anything else afterwards.

Cargo.toml
fredizzimo: I don't think we should merge anything that requires building from source, since that's a big complication for macOS contributor. We have four options.
1. Remove the opengl support
2. Keep the opengl support, but make it a feature flag, and add both variations to the CI and the release
3. Request for a new binary that includes both opengl and meta
4. Host our own skia binaries somewhere else.

src/renderer/metal.rs
falcucci: yes @fredizzimo is right, it definitely need to be false, it breaks the floating window blur. just checked myself.

Cargo.toml
falcucci: I actually have no problem to remove OpenGL support. Metal is much more stable for macOS.

falcucci: > big complication for macOS contributor.

I totally agree, that's a point I forgot. based on that if we don't have both features mashup from the skia team, I would stick to remove OpenGL support.

src/renderer/metal.rs
crupest: > I'm don't know how it currently looks like, but for this PR, I think we should try to keep the colors exactly the same as before.

@fredizzimo Agree with you, actually. This pr should never be merged until we get clear on this problem totally. At least make it work.

Though I never use color space personally, I think many people care about it very much.

9mm:
Awesome, thanks for the explanations and hard work!

src/renderer/metal.rs
crupest: Done! Please have a double check on functions both for --no-srgb and --srgb. Thanks!

crupest:
To forward this.

We should wait for an update of rust-skia version both in itself and neovide to continue this pr?

As I can see, the current latest version of rust-skia was made before I added the new combination of prebuilt binary for this pr. So,

1. rust-skia needs to make a new release for the new combination making effects.
2. neovide needs to upgrade rust-skia dependency version.
3. We rebase this. If CI passes, we merge this.

Is this correct?

@falcucci You can edit this pr directly when you find everything is ok, in case I have no response.

fredizzimo:
We probably don't need to wait for long for a new skia-safe release, this was merged 3 days ago
https://github.com/rust-skia/rust-skia/pull/1046

When that's out, I think we should make a separate PR that updates it, and then we can merge this one.

fredizzimo:
The binaries are out, https://github.com/rust-skia/skia-binaries/releases/tag/0.80.0, but the main package is not yet released.

falcucci:
@fredizzimo that's good, release is coming soon in upstream then.

falcucci:
@fredizzimo https://github.com/neovide/neovide/pull/2873

falcucci (approved):
Great contribution made by @crupest and we are finally merging.

thank you very much to collaborate and for all dedication to make neovide even better. 

kudos for you!

Issue #2387, opened 2024-02-28
Exiting neovide on windows leaves orphaned neovim processes around


kareigu:
**Describe the bug**
Closing neovide by any normal means results in 2 neovim processes being left running indefinitely.
After a few days I have had upwards of 10 that are taking up a sizeable amount of memory for nothing.

**To Reproduce**
1. Start neovide
2. Close it
3. Look at task manager, 2 neovim processes are still running

**Expected behavior**
Close down the neovim processes during teardown.

**Screenshots**
![image](https://github.com/neovide/neovide/assets/31644842/ec931277-0382-4eb4-8803-90efaafc2fa1)
![image](https://github.com/neovide/neovide/assets/31644842/315b30a2-be26-43ed-bb06-5271965e1f37)
After closing neovide with :q
![image](https://github.com/neovide/neovide/assets/31644842/37c9f8ea-98d5-4426-9ea3-6cde39aa730d)

**Desktop (please complete the following information):**
 - OS: Windows 11 22635.3140
 - Neovide Version: 0.12.2
 - Neovim Version: 0.10.0-dev-2465+ga7788c2e2 , latest nightly on scoop at this time

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
<details>
<summary>Log from receiving Q KeyEvent</summary>

```log
TRACE [neovide::window::keyboard_manager] KeyEvent {
    physical_key: Code(
        KeyQ,
    ),
    logical_key: Character(
        "q",
    ),
    text: Some(
        "q",
    ),
    location: Standard,
    state: Pressed,
    repeat: false,
    platform_specific: KeyEventExtra {
        text_with_all_modifers: Some(
            "q",
        ),
        key_without_modifiers: Character(
            "q",
        ),
    },
}
TRACE [neovide::window::keyboard_manager] Key pressed q ModifiersState(0x0)
TRACE [neovide::channel_utils] UICommand Serial(Keyboard("q"))
TRACE [neovide::bridge::ui_commands] In Serial Command
TRACE [neovide::bridge::ui_commands] Keyboard Input Sent: q
TRACE [neovide::bridge::handler] Neovim request: "neovide.quit"
INFO [neovide::running_tracker] Quit with code 0: Quit from neovim
DEBUG [neovide::settings::window_size] Saved Window Settings: {"window":{"Windowed":{"position":{"x":370,"y":201},"pixel_size":{"width":1838,"height":1025},"grid_size":{"width":183,"height":51}}}}
[...]

Issue #2223, opened 2023-12-30
Seg fault when exiting with --no-fork (not really critical)


DADA30000:
**Describe the bug**
When exiting with --no-fork, there is seg fault message

**To Reproduce**
1. Open neovide with --no-fork
2. Exit
**Expected behavior**
No seg fault message
**Screenshots**

https://github.com/neovide/neovide/assets/73874445/1bc4536c-acd7-41bb-ad4b-b9af247dfd1a

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Neovide Version: 0.12.0
 - Neovim Version: 0.9.4

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/13798057/neovide_rCURRENT.log)

**Additional context**
I think it happened after some update, because I don't remember that happening before

fredizzimo:
Yes, there`s a known exit crash when using some wlroots based compositors.

Alacritty had the same issue https://github.com/alacritty/alacritty/issues/7146

And the fix is also known, we need to do something similar to this https://github.com/alacritty/alacritty/pull/7220/files#diff-c32f38542d52895124f53e3425f4243a05391d1d2bd8640f371ac6e427be5d70R1594-R1610

Issue #2452, opened 2024-03-29
Neovide crashing while "working" Value not an Array: '{"kind": "syntax", "hi_name": "SignColumn"... }


lucax88x:
**Describe the bug**
This morning neovide just keep crashing because of treesitter highlight change (check the logs below)

**To Reproduce**
Checking logs, might related to redraw of neovide?

**Expected behavior**
To not crash

**Desktop (please complete the following information):**
 - OS: 14.4.1
 - Neovide Version 0.12.2
 - Neovim Version v0.10.0-dev-2739+ge2224a793

[crash log](https://gist.github.com/lucax88x/1d36ed14f4fa6657499ab4b75809b697)

fredizzimo:
This is either some regression in Neovim nighly or some new even we don't handle. I will take a look at it soon, but do you remember when you last updated Neovim before today?

lucax88x:
no more than 2/3 days

fredizzimo:
Unfortunately. I can't reproduce this here, it's probably caused by some plugin, can you try to run `neovide -- --clean` and see if it happens?

fredizzimo:
Most likely the `nvim-treesitter-context` plugin, I will try to install that here

CertainLach:
Also experiencing with `nvim_input_mouse` when I try to enable https://github.com/lewis6991/hover.nvim

```
Caused by:
    0: Error decoding response to request 'nvim_input_mouse'
    1: Error while decoding
    2: Value not an Array: '{"kind": "syntax", "hi_name": "GruvboxRedUnderline", "id": 167}'
```

(I understand this is not a neovide issue per se)

fredizzimo:
@CertainLach, thanks, is that plugin enough to reproduce the issue? If you follow this:
https://github.com/neovim/neovim/blob/master/contrib/minimal.lua, 

If you prefer you can alternatively create a single file `lazy.nvim` configuration that can be launched with `NVIM_APPNAME=notarray neovide`, the config then goes into `.config/notarray` instead of the normal`.config.nvim`

lucax88x:
Really hard to reproduce sadly, it happens randomly.

vargad:
I'm having the same issue with neovide 0.12.2, and the latest neovim v0.10.0-dev-2806+g9af355964-dirty. I'm not using treesitter nor hover, not sure which plugin triggers it.
```
ERROR [neovide::bridge::ui_commands] Mouse Scroll Failed

Caused by:
[...]

PR #2463, opened 2024-04-04
fix!: Perform a clean exit again

fredizzimo:
The main loop now waits for Neovim to quit before exiting. This allows error messages to be shown and dealt with, instead of leaking nvim processes, as neovide will block here https://github.com/neovim/neovim/blob/15a2dd9e963784123273ec830e82e24ecea4ad0b/src/nvim/main.c#L802-L807

It primarily waits for the stdio stream to finish, but due to bugs, like https://github.com/neovim/neovim/issues/26743, that does not always happen, so it additionally waits for the process to exit, and forces a quite if the io stream has not finished within 2 seconds after that. **If you use Neovim nightly with https://github.com/neovim/neovim/pull/28157, it should not happen**

An exit crash that happens on some Wayland systems has also been fixed, by properly shutting down the rendering system before the event loop finishes.

* Fixes https://github.com/neovide/neovide/issues/2387
* Fixes https://github.com/neovide/neovide/issues/2223

The special handling of just detaching from remote connections when closing the Window, which was added here https://github.com/neovide/neovide/pull/687 has been removed.  It would uneccessarily complicate this change, and I found that re-attaching causes a lot of bugs, like invisible text, the cursor showing the wrong position, the screen not updating and so on, so it's not very usable anyway.

NOTE: That it's still possible to detach manually by the following command: `:call chanclose(g:neovide_channel_id)`

Some hangs that @sid-6581 has reported on Windows might also start to appear again. But it will not hang compltely, just for 2 seconds. If that still happens, the actual cause should be determined and fixed. The same applies to other platforms as well of course.

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -12,8 +12,11 @@ use itertools::Itertools;
 use log::{error, info};
 use nvim_rs::{error::CallError, Neovim, UiAttachOptions, Value};
 use rmpv::Utf8String;
-use std::{io::Error, ops::Add};
-use tokio::runtime::{Builder, Runtime};
+use std::{io::Error, ops::Add, time::Instant};
+use tokio::{
+    runtime::{Builder, Runtime},
+    task::JoinSet,
+};
 use winit::event_loop::EventLoopProxy;
 
 use crate::{
@@ -28,13 +31,16 @@ pub use api_info::*;
 pub use command::create_nvim_command;
 pub use events::*;
 pub use session::NeovimWriter;
-pub use ui_commands::{send_ui, start_ui_command_handler, ParallelCommand, SerialCommand};
+pub use ui_commands::{
+    send_ui, shutdown_ui, start_ui_command_handler, ParallelCommand, SerialCommand,
+};
 
 const INTRO_MESSAGE_LUA: &str = include_str!("../../lua/intro.lua");
 const NEOVIM_REQUIRED_VERSION: &str = "0.9.2";
 
 pub struct NeovimRuntime {
-    runtime: Option<Runtime>,
+    runtime: Runtime,
+    join_set: JoinSet<()>,
 }
 
 fn neovim_instance() -> Result<NeovimInstance> {
@@ -77,7 +83,11 @@ pub async fn show_error_message(
     nvim.echo(prepared_lines, true, vec![]).await
 }
 
-async fn launch(handler: NeovimHandler, grid_size: Option<GridSize<u32>>) -> Result<NeovimSession> {
+async fn launch(
+    handler: NeovimHandler,
+    grid_size: Option<GridSize<u32>>,
+    join_set: &mut JoinSet<()>,
+) -> Result<NeovimSession> {
     let neovim_instance = neovim_instance()?;
 
     let session = NeovimSession::new(neovim_instance, handler)
@@ -110,7 +120,7 @@ async fn launch(handler: NeovimHandler, grid_size: Option<GridSize<u32>>) -> Res
     setup_neovide_specific_state(&session.neovim, should_handle_clipboard, &api_information)
         .await?;
 
-    start_ui_command_handler(session.neovim.clone(), &api_information);
+    start_ui_command_handler(session.neovim.clone(), &api_information, join_set);
     SETTINGS.read_initial_values(&session.neovim).await?;
 
     let mut options = UiAttachOptions::new();
@@ -148,12 +158,17 @@ async fn run(session: NeovimSession, proxy: EventLoopProxy<UserEvent>) {
     proxy.send_event(UserEvent::NeovimExited).ok();
 }
 
+async fn wait(join_set: &mut JoinSet<()>) {
+    while join_set.join_next().await.is_some() {}
+}
+
 impl NeovimRuntime {
     pub fn new() -> Result<Self, Error> {
         let runtime = Builder::new_multi_thread().enable_all().build()?;
 
         Ok(Self {
-            runtime: Some(runtime),
+            runtime,
+            join_set: JoinSet::new(),
         })
     }
 
@@ -163,15 +178,22 @@ impl NeovimRuntime {
         grid_size: Option<GridSize<u32>>,
     ) -> Result<()> {
         let handler = start_editor(event_loop_proxy.clone());
-        let runtime = self.runtime.as_ref().unwrap();
-        let session = runtime.block_on(launch(handler, grid_size))?;
-        runtime.spawn(run(session, event_loop_proxy));
+        let session = self
+            .runtime
+            .block_on(launch(handler, grid_size, &mut self.join_set))?;
+        self.join_set
+            .spawn_on(run(session, event_loop_proxy), self.runtime.handle());
         Ok(())
     }
 }
 
 impl Drop for NeovimRuntime {
     fn drop(&mut self) {
-        self.runtime.take().unwrap().shutdown_background();
+        log::info!("Starting neovim runtime shutdown");
+        let start = Instant::now();
+        shutdown_ui();
+        self.runtime.block_on(wait(&mut self.join_set));
+        let elapsed = start.elapsed().as_millis();
+        log::info!("Neovim runtime shutdown took {elapsed} ms");
     }
 }
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -12,15 +12,10 @@ use itertools::Itertools;
 use log::info;
 use nvim_rs::{error::CallError, Neovim, UiAttachOptions, Value};
 use rmpv::Utf8String;
-use std::{
-    io::Error,
-    ops::Add,
-    time::{Duration, Instant},
-};
+use std::{io::Error, ops::Add, time::Duration};
 use tokio::{
     runtime::{Builder, Runtime},
     select,
-    task::JoinSet,
     time::timeout,
 };
 use winit::event_loop::EventLoopProxy;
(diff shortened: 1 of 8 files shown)


src/running_tracker.rs
```diff
@@ -10,34 +10,26 @@ lazy_static! {
...
     pub fn quit_with_code(&self, code: i32, reason: &str) {
-        self.exit_code.store(code, Ordering::Relaxed);
-        self.running.store(false, Ordering::Relaxed);
+        self.exit_code.store(code, Ordering::Release);
         info!("Quit with code {}: {}", code, reason);
     }
 
-    pub fn is_running(&self) -> bool {
-        self.running.load(Ordering::Relaxed)
-    }
```
fredizzimo, line 38: The running tracker only tracks the exit code now, so maybe it could be renamed or replaced by variable. But I left that for future refactoring.

src/bridge/mod.rs
```diff
@@ -236,25 +252,52 @@ async fn launch(handler: NeovimHandler, grid_size: Option<Dimensions>) -> Result
     res.map(|()| session)
 }
 
-async fn run(session: NeovimSession) {
-    match session.io_handle.await {
-        Err(join_error) => error!("Error joining IO loop: '{}'", join_error),
-        Ok(Err(error)) => {
-            if !error.is_channel_closed() {
-                error!("Error: '{}'", error);
```
fredizzimo, line 244: I will add this message back. I thought it wasn't important, but it's actually, the message printed in this bug for example: 
* https://github.com/neovide/neovide/issues/2452

src/bridge/ui_commands.rs
```diff
@@ -276,15 +278,15 @@ impl AsRef<str> for UiCommand {
 }
 
 struct UIChannels {
-    sender: LoggingSender<UiCommand>,
+    sender: Mutex<Option<LoggingSender<UiCommand>>>,
```
Kethku, line 281: These Mutex Options over concurrent capable types kill me a little, but after thinking about it for a bit nothing cleaner came to mind so I think its reasonable to leave it.

One thing we might consider is to use a parkinglot RwLock instead since the base case is just reading and we only write when the app closes. But tbh thats a micro optimization to take it or leave it

sid-6581:
I just tried this PR, and got a hard lockup immediately. It hangs forever, not just for 2 seconds.

image

src/bridge/mod.rs
```diff
@@ -236,25 +252,52 @@ async fn launch(handler: NeovimHandler, grid_size: Option<Dimensions>) -> Result
...
+    let mut session = session;
+
+    if let Some(process) = session.neovim_process.as_mut() {
+        let neovim_exited = select! {
+            _ = &mut session.io_handle => false,
+            _ = process.wait() => {
+                log::info!("The Neovim process quit before the IO stream, waiting two seconds");
+                true
+            }
+        };
```
sid-6581, line 265: I tried this approach as one of my first attempts when trying to fix this problem, but `wait()` doesn't detect that the process is gone.

src/bridge/mod.rs
```diff
@@ -236,25 +252,52 @@ async fn launch(handler: NeovimHandler, grid_size: Option<Dimensions>) -> Result
...
+        // So wait for some additional time, both to make the bug obvious and to prevent incomplete
+        // data.
+        if neovim_exited {
+            let sleep = sleep(Duration::from_millis(2000));
+            tokio::pin!(sleep);
+            select! {
+                _ = session.io_handle => {}
+                _ = &mut sleep  => {
+                    log::info!("The IO stream was never closed, forcing Neovide to exit");
+                }
```
sid-6581, line 280: Wouldn't `tokio::time::timeout()` be simpler here?

src/bridge/mod.rs
```diff
@@ -263,16 +306,23 @@ impl NeovimRuntime {
...
         Ok(())
     }
 }
 
 impl Drop for NeovimRuntime {
     fn drop(&mut self) {
-        self.runtime.take().unwrap().shutdown_background();
+        log::info!("Starting neovim runtime shutdown");
+        let start = Instant::now();
+        shutdown_ui();
```
sid-6581, line 323: Shutting down the UI explicitly from the runtime drop doesn't feel very clean. IMHO the runtime should be concerned only about itself, and shutting down other things should be done elsewhere.

As an aside, I'm not sure why the runtime is created manually instead of just using `tokio::main`. I did some refactoring earlier and eliminated the entire `NeovimRuntime` and just used `tokio::main`, and it was a lot cleaner.

src/bridge/mod.rs
fredizzimo: On Linux this workaround is needed for Neovim 0.9.5 without my fix https://github.com/neovim/neovim/pull/28157. There it correctly detects both the stdio stream close and process exit.

src/bridge/mod.rs
fredizzimo: Maybe, but it looks like the timeout consumes the future, and I want to re-use it after that.

src/bridge/mod.rs
fredizzimo: The function is probably a bit poorly named, it shuts down the UI event handlers, not the whole UI system. And we do have to wait for them here, otherwise they start to print stuff to stderr instead of the logging system, as the rest of the runtime shuts down while they continue running.

fredizzimo:
@sid-6581, We really need to find the root cause of not being able to detect either the process exit or the closing of the io streams. That's something that I know for sure works in Windows, tons of programs are absolutely dependent on that, and I have written a few myself and not something we should work around and cause more bugs anymore.

And those two are the only two signals we can rely on, since we need to keep the Neovide UI active, in order to be able to respond to errors that can happen after `VimLeave` and waits for user input here https://github.com/neovim/neovim/blob/15a2dd9e963784123273ec830e82e24ecea4ad0b/src/nvim/main.c#L802-L807

Unfortunately, I can't help much with that, since I'm unable to repeat it on any of my systems. It does seem to affect a few others, but for most it looks like the problem was with the corrupt shada files, which is now fixed so that the error message is shown and we can send the inputs. So, I think the problem is very rare.

sid-6581:
@fredizzimo I should have some more free time now to look at this again. I'll try to debug this again, I have a couple of things I'd like to try. At least I can still easily reproduce it, which is good.

fredizzimo:
Great thanks, I would start by checking if it's stuck here:
https://github.com/rust-lang/rust/blob/a4b11c8e6057bd47f8d241baafd5ae1e813d62fd/library/std/src/sys/pal/windows/handle.rs#L250

That's where I think they might violate one or more of the conditions required https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-ntreadfile

> The caller can safely wait on the file handle only if all three of the following conditions are met:

> The handle was opened for asynchronous access (that is, neither FILE_SYNCHRONOUS_IO_XXX flag was specified).
> The handle is being used for only one I/O operation at a time.
> NtReadFile returned STATUS_PENDING.

Furthermore, this document says that asynchronous IO on pipes can only be performed using overlapped operations https://learn.microsoft.com/en-us/windows/win32/ipc/synchronous-and-overlapped-input-and-output#enabling-asynchronous-operation
`NtReadFile` supports it, but they are not using it for some reason.

sid-6581:
@fredizzimo I'm in the debugger now, and that's exactly where it's stuck. I need to dig further, but it looks like it's stuck waiting for a handle that the neovide process no longer has open. Here you see it's waiting for handle 0x384:

image

But it doesn't appear that neovide still has that handle open:

image

I can't quite track it back to anything specific to see what it's actually trying to wait on, since there are no clues in the current call stack, but I'll keep digging.

sid-6581:
I forgot to add `-a` as an argument in the screenshot above. I reproduced the hang again, and when I show all the handles it shows that it hangs on named pipe 350 here:

image

That leads me to https://github.com/rust-lang/rust/blob/385fa9d845dd326c6bbfd58c22244215e431948a/library/std/src/sys/pal/windows/pipe.rs#L83

I'll backtrack from there.

fredizzimo:
That's most likely one of the pipes created here with `piped()`
https://github.com/neovide/neovide/blob/a973abb97bc3be7af9f878bab242e7f97fe0534a/src/bridge/session.rs#L84

You could maybe also try process monitor to log all the file io calls https://learn.microsoft.com/en-us/sysinternals/downloads/procmon

sid-6581:
Yep, that's what I'll try next. It's interesting that the pipe remains open when the nvim process is gone.

src/bridge/ui_commands.rs
fredizzimo: I simplified this, using `OnceLock`

src/bridge/mod.rs
fredizzimo: I changed it to a Tokio `timeout`

fredizzimo:
I will put this as draft, so that it does not accidentally get merged until the Window issue is resolved or at least identified.

sid-6581:
I'm still digging, but from what I can tell:

Sometimes both named pipes stay open after nvim exits, sometimes stdin is closed but stdout remains open.

The named pipes are create in overlapped mode (only for neovide, not the ones that are handed to neovim), so `WaitForSingleObject` should work fine on them. However, the documentation says this:

> If this handle is closed while the wait is still pending, the function's behavior is undefined

ProcMon slowed down things enough that I couldn't recreate the hang while it was running.

I'll be traveling this weekend, but I'll get back to it when I get back.

sid-6581:
Actually, the more I think about, I don't understand why the read from the overlapped named pipe is a blocking operation using `WaitForSingleObject`.

fredizzimo:
I'm re-opening this. I still think it's better to hang at the exit on a few systems than to leak Neovim processes.

sid-6581:
Looks like I need to look into this again. neovide is hanging all the time on all my machines, and it's still orphaning nvim processes. I'm not really sure why the nvim processes are leaked.

PR #2467, opened 2024-04-05
feat: render same z-index together

Kethku:
Implementation was previously discussed in issue #2408. Close #2379.
Supersedes https://github.com/neovide/neovide/pull/2442 but uses skia functions to implement path unions and removes the zindex step setting in favor of just merging floating layers with consecutive indicesj

- Feature

- No


src/utils/test.rs
```diff
@@ -0,0 +1,218 @@
+use std::collections::{BTreeMap, HashSet};
```
Kethku, line 1: I used these helper functions to make testing the above easier. I think they have value for writing tests for other portions of our codebase though so I'm leaving them in.

fredizzimo:
@Kethku, I think something went wrong with your rebase, since it shows that all of @TwIStOy's commits are unverified. But they should at least be partially verified I think, and probably fully verified if you use merge instead of re-base, so that they don't need to be modified.

src/renderer/rendered_window.rs
```diff
@@ -244,19 +260,21 @@ impl RenderedWindow {
         animating
     }
 
-    pub fn draw_surface(
-        &mut self,
-        canvas: &Canvas,
+    #[allow(clippy::type_complexity)]
+    fn split_lines(
```
fredizzimo, line 264: Maybe it would make sense to split this function into three different ones, since all three return values can be calculated completely independently? And also make it return iterators instead of vectors?

src/renderer/rendered_window.rs
```diff
@@ -721,6 +705,61 @@ impl RenderedWindow {
         self.scroll_delta = 0;
     }
 
+    fn iter_lines<'a>(&'a self) -> Box<dyn Iterator<Item = &'a Rc<RefCell<Line>>> + 'a> {
```
fredizzimo, line 708: If this is split into three different iterators, then I think these iterators could be used as the base for, the `split_lines` (or the refactored new functions) as well.

Kethku:
> @Kethku, I think something went wrong with your rebase, since it shows that all of @TwIStOy's commits are unverified. But they should at least be partially verified I think, and probably fully verified if you use merge instead of re-base, so that they don't need to be modified.

im not sure I care enough to address this unless theres some deeper reason to worry about it that I'm forgetting.

Kethku:
@TwIStOy you ok if I dont fix the unverified commits? Fixing it feels like a headache atm

TwIStOy:
Ahh, It's ok.

9mm:
Hey so after pulling main I noticed floating windows look like this:

image

Expected (when using everything except this commit):

image

9mm:
i can help you test, although im going to the sauna right now. I'm using this: https://github.com/echasnovski/mini.nvim/blob/main/readmes/mini-files.md

fredizzimo:
@9mm, are you sure it's this PR that caused the issue, and not the following?
* #2487

griwes:
Judging by the titles of the PRs alone, it seems to me that it might be an interaction of the two, not any of them specifically? I.e. an off-by-one in the sorting one, revealed by rendering changes here?

9mm:
@fredizzimo what i did was only revert the commit of 278bf6b alone (feat: render same z-index together)

I know last time I said it was one commit I was wrong, this time I tried 2 or 3 times to be sure it was the right one.

It would be nice if someone could sanity check me lol. Someone could try installing this plugin to see easily

https://github.com/echasnovski/mini.nvim/blob/main/readmes/mini-files.md#installation

And running `:lua MiniFiles.open()`

Here's what I ran:

```
➜  neovide git:(main) ✗ cargo run # problem exists
➜  neovide git:(main) git revert 278bf6b --no-commit
➜  neovide git:(main) ✗ cargo run # problem does not exist
   Compiling neovide v0.12.2 (/Users/zesty/Code/neovide)
    Finished dev [unoptimized + debuginfo] target(s) in 2.50s
     Running `target/debug/neovide`
```

9mm:
if I do `git revert deb4dd5 --no-commit`, and resolve the `use` conflict, the problem still exists

fredizzimo:
@9mm or @griwes, could you provide a log file? 

It's a bit tricky for me to try to replicate the issue myself, but I might be able to quickly spot something in the log.

9mm:
It should be at the very bottom. I have the file explorer popup mapped to `<Leader>fe` so that's what those hotkeys are at the bottom

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/15049548/neovide_rCURRENT.log)

9mm:
Heres the log when its working:

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/15049608/neovide_rCURRENT.log)

fredizzimo:
I believe this line should be changed to:
https://github.com/neovide/neovide/blob/e3c99269ded86ffd0ca1272113b5ea3e5f83a387/src/renderer/rendered_window.rs#L645

`actual_line_count - self.viewport_margins.bottom as isize..actual_line_count + 1;`

9mm:
Hmm i changed it to that and there was no change. Even if I do + 1000 there is no change

however I did test some other stuff as a sort of test, and i see if i do this:

```
            actual_line_count - self.viewport_margins.bottom as isize + 1..actual_line_count;
```

so i know it is actually doing something

image

9mm:
I dont know if this helps give a hint, but if i hover over it, it highlights both:

image

9mm:
so perhaps the border is correct but its duplicate rendering the last item?

fredizzimo:
Maybe, it's hard to say without looking properly at the code and see what it actually does. I don't have that much time right now, but perhaps @Kethku will look at it. And I should also have more time tomorrow.

9mm:
so im in xtrme nerve pain that im waitin gto go to the dr for (lol) so i cant really focus, but here:

rendered_window.rs:673

```
        line_indices.filter_map(move |i| {
            self.scrollback_lines[scroll_offset_lines + i]
                .as_ref()
                .map(|line| (i, line))
        })
```

If i delete the + i, it renders the same item which leads to the duplicate behavior, but much more exaggerated. So maybe its an off by 1 error in that area of the code

image

9mm:
Changing this `1` here to `2` makes the problem go away, but I have absolutely no idea what else that breaks:

```rust
        let line_indices = if !self.scrollback_lines.is_empty() {
            0..self.grid_size.height as isize - 2
        } else {
            0..0
        };
```

fredizzimo:
Regarding your first experiment, that seems like it's excepted, it just duplicates the same first line.

For the second one, I think it fixes it in the wrong way, removing one of the lines, essentially shrinking the window to make the border in the right place. But I think it should go the other way, the border should go one line down.

Hope you can see the doctor soon and get some medication that works.  Don't stress too much about this, but if it helps you ignoring the pain, then you can continue trying things out and maybe find the problem.

9mm:
Ah yeah I definitely dont think thats the answer haha. I was mostly just trying to figure out at least the "right path".

ill try again later, i have this pinched nerve that makes it feel like im being electrically shocked nonstop, and i need to go to the sauna again (tmi). makes it very hard to focus!

sid-6581:
I'm also seeing various rendering issues in pretty much all floating windows with borders, sometimes only when scrolling, sometimes even before scrolling. `nvim-cmp` popups, LSP hover, Lazy popup, `mini.files`, etc. all have this issue. I'll see if I get the time to investigate a bit today.

sid-6581:
@Kethku I found the issue, the problem is that `scrollback_lines.clone_from_iter(inner_view)` (both calls) results in duplicates because if `scrollback_lines` has more lines than `inner_view`, those extra lines aren't set to `None`. I haven't had a chance to dig too deep into the ring buffer code, but that behavior seems a bit surprising to me, so maybe I'm missing something. What do you think?

fredizzimo:
@sid-6581, without looking at the code, that sounds normal.

The scrollback buffer should be at least twice the size of the inner view to support smooth scrolling of a complete screen. When there are scroll events from Neovim, the scrollback buffer is rotated so that the old lines remain, and after that the new screen lines are copied to the end with `clone_from_iter`.  At this point Neovide still renders the old lines, but during the scroll animation the scrollback buffer is rotated until the newest state is completely shown.

There should be no visible `None` elements, unless you scroll more than one screen at a time. 

Another thing to note is that the actual numbers of lines rendered from the scrollback should almost always be 1 line more than the inner view, so that a partial line can be seen during the scrolling.

sid-6581:
@fredizzimo that makes sense, but in this case the last line occurs twice in the scrollback buffer, which is why the duplicate line is rendered on the bottom border. There's also no scrolling going on in my case.

9mm:
@fredizzimo so you mentioned the border is wrong but heres a screenshot again:

unhighlighted

image

highlighted

image

It seems to my eyes the border is correct, and even the original list is correct, but the duplicate item extending beyond it... i dont know if its because of the duplicate aspect you said is correct behavior (in my "config" experiment), or because the item is literally there twice

Can you explain again what you think the solution would be? I want to make sure I understand what you're saying

sid-6581:
@9mm the item is literally there twice in the scrollback buffer, that's the issue. It has one too many items, and the extra item is the duplicate.

9mm:
@fredizzimo the tank!

fredizzimo:
And just to clarify, technically there's one line of garbage in the scrollback buffer, but to avoid special casing the case where no partial lines are visible, we always render one extra line, and let the clipping rectangle take care of removing it for both partial and full lines.

And even if that was special cased, then it would glitch in a similar way during scrolling, so the clipping rectangle fix is needed anyway.

PR #2469, opened 2024-04-05
feat: Implement bell

### LINKED ISSUE #1967: Is there any bell sound system?

I wonder if `neovide` have any sound system?

- `:!tput bel` returns `tput: No value for $TERM and no -T specified` (shell returned 2) - understandable as `neovide` is not trying to replace a terminal app
- `:!echo '\a'` does nothing
- `<Tab>` on empty file does not play any bell as it does in console `nvim`

Info: I have a clean config with `set belloff=` in `$MYVIMRC` file, and there's no visual or audio bell

An actual bell reaction (audio preferred, visual accepted)

At least, working visual bell?

**Additional context**

OS: macOS Ventura 13.5 (22G74)
NVIM: v0.9.1 Release
LuaJIT: 2.1.0-beta3
Neovide: 0.11.0 stable (brew cask)

4513ECHO:
Close #1967 (currently partial)

- Feature

- No

### TODO
- Implement bell on X11/Wayland
- Implement visual bell


fredizzimo:
Another option for the bell is to use a configurable external command like Alacritty:
* https://github.com/alacritty/alacritty/issues/1528

That's probably the only easy way on X11/Wayland. On Windows and macOS it could still default to built-ins like you do here, but allow a command for more customization.

4513ECHO:
I've added `g:neovide_bell_command` option. I want to make its type `Vec<String>`, but I can't fix typing error...

fredizzimo:
> I've added `g:neovide_bell_command` option. I want to make its type `Vec<String>`, but I can't fix typing error...

Sorry for not coming back to you earlier. Supporting vectors is indeed not supported by our settings system right now. But I think a single command is good enough. You can always call a shell script if you want a more complex command.

src/editor/mod.rs
```diff
@@ -142,6 +142,14 @@ impl Editor {
                 trace!("Cursor on");
                 self.cursor.enabled = true;
             }
+            RedrawEvent::Bell => {
+                tracy_zone!("EditorBell");
+                let _ = self.event_loop_proxy.send_event(WindowCommand::Bell.into());
+            }
+            RedrawEvent::VisualBell => {
+                tracy_zone!("EditorVisualBell");
+                // TODO: Implement visual bell
```
fredizzimo, line 151: Do you need help implementing this?

src/window/window_wrapper.rs
```diff
@@ -195,6 +199,24 @@ impl WinitWindowWrapper {
                 self.mouse_manager.enabled = mouse_enabled
             }
             WindowCommand::ListAvailableFonts => self.send_font_names(),
+            WindowCommand::Bell => {
+                let bell_command = SETTINGS.get::<WindowSettings>().bell_command;
+                let bell_command = bell_command.split(" ").collect::<Vec<_>>();
+                if bell_command.is_empty() {
+                    unsafe {
+                        // TODO: How to bell on X11/Wayland?
```
fredizzimo, line 207: You can change this command to "Only custom commands are supported on X11 and Wayland"

fredizzimo (commented):
This needs rebasing on the latest master

src/editor/mod.rs
4513ECHO: Yes, I am not familiar with skia...

src/window/window_wrapper.rs
4513ECHO: I've changed.

src/window/window_wrapper.rs
```diff
@@ -175,6 +178,25 @@ impl WinitWindowWrapper {
...
             }
             WindowCommand::ListAvailableFonts => self.send_font_names(),
+            WindowCommand::Bell => {
+                unsafe {
+                    #[cfg(target_os = "macos")]
+                    NSBeep();
+                    #[cfg(target_os = "windows")]
+                    MessageBeep(MB_OK);
+                }
+                #[cfg(not(any(target_os = "macos", target_os = "windows")))]
```
fredizzimo, line 188: I think this can be enabled on all platforms. And used instead of the platform bells if defined.

src/editor/mod.rs
fredizzimo: I think you can add `draw_rect`  (https://rust-skia.github.io/doc/skia_safe/canvas/struct.Canvas.html#method.draw_rect) to `renderer::draw_frame` after everything else is rendered. Either black or white, with some alpha that animates in and out. Compare against the default background color, and select using this condition 
https://github.com/neovide/neovide/blob/4096e4c3465495ed7bd656edf8fd6db28df9c386/src/editor/mod.rs#L595-L597

The updating of the alpha value could be done in `renderer::animate_frame`

PR #2470, opened 2024-04-05
chore: Refactor the update loop

fredizzimo:
Quite heavy refactoring of the update loop, hopefully it's much easier to read and understand now. I'm planning to make a fix for a temporary freeze after the computer has been sleeping, but I wanted to get this refactored first for easier bisecting in case something breaks.

- No


Kethku:
This is a great improvement. Much more readable with names that make much more sense to me

PR #2472, opened 2024-04-06
fix: Temporary hang after sleep

fredizzimo:
This fixes a bug where Neovide stops responding for some time after the computer has gone to sleep or disconnected from a remote desktop.

There are actually two different fixes:
1. Properly prepare the frame before respecting OS or Winit initiated rendering requests. This fixes the main bug at least when disconnecting from the remote desktop on Windows. 
2. Don't simulate for more than 1 second. I don't know how to trigger this in practice, but I suspect it can happen if the computer goes to sleep while Neovide is animating. I tested this by disabling fix 1.

NOTE: Marked as draft since it depends on and should be reviewed and merged after that
* https://github.com/neovide/neovide/pull/2470

- No

### DIFF
--- a/src/window/update_loop.rs
+++ b/src/window/update_loop.rs
@@ -168,6 +168,7 @@ impl UpdateLoop {
 
     fn render(&mut self, window_wrapper: &mut WinitWindowWrapper) {
         self.pending_render = false;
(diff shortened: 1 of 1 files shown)


Kethku:
This needs rebased on main before I can easily review it as changes are mixed between the update loop changes and the actual fixes in this PR

PR #2476, opened 2024-04-08
fix!: Use points instead of pixels for the config file

fredizzimo:
* The `width` and `size` arguments in the font config are now read as points, like `guifont` and almost any other software.

Yes, users that have adjusted their size in pixels will now see too big fonts.


fredizzimo:
Hm.. It looks like there was some wrongly resolved merge conflicts or something, because the font config file is still in pixels rather than points.

PR #2484, opened 2024-04-14
chore: Use typesafe pixel and grid units

fredizzimo:
This is a big refactor that adds typesafe units almost everywhere, using the [euclid](https://docs.rs/euclid/latest/euclid/) library. The code has also been simplified in a lot of places by using primitives provided by euclid instead of individual points.

Furthermore, many of the type casts have been replaced by safer alternatives, like like `try_into().unwrap()`, crashing instead of doing something completely unexpected. In some places where invalid values are more expected, clamping is used instead.

Most calculations are also done using either floating point or signed numbers, which reduces the need for type casts, so a lot of those could be deleted.

This is a lot of changes, but the code is much easier to read and understand IMO now. It's also safer, since you usually get a compilation error instead of runtime problem if you for example think that a value is using grid coordinates, while it's in fact is using pixel coordinates. It will also help with the support for fractional grid sizes, and arbitrary font sizes. In fact, most of the coordinates are already using fractional pixels, so the change will not be that big.

This is still draft, since it includes:
* https://github.com/neovide/neovide/pull/2483

- No

### DIFF
--- a/src/renderer/cursor_renderer/mod.rs
+++ b/src/renderer/cursor_renderer/mod.rs
@@ -3,16 +3,16 @@ mod cursor_vfx;
 
 use std::collections::HashMap;
 
-use skia_safe::{op, Canvas, Paint, Path, Point};
+use skia_safe::{op, Canvas, Paint, Path};
 use winit::event::{Event, WindowEvent};
 
 use crate::{
     bridge::EditorMode,
     editor::{Cursor, CursorShape},
     profiling::{tracy_plot, tracy_zone},
-    renderer::animation_utils::*,
-    renderer::{GridRenderer, RenderedWindow},
+    renderer::{animation_utils::*, GridRenderer, RenderedWindow},
     settings::{ParseFromValue, SETTINGS},
+    units::{to_skia_point, GridPos, GridScale, PixelPos, PixelSize, PixelVec},
     window::{ShouldRender, UserEvent},
 };
 
@@ -68,10 +68,10 @@ impl Default for CursorSettings {
 
 #[derive(Debug, Clone)]
 pub struct Corner {
-    start_position: Point,
-    current_position: Point,
-    relative_position: Point,
-    previous_destination: Point,
+    start_position: PixelPos<f32>,
+    current_position: PixelPos<f32>,
+    relative_position: GridPos<f32>,
+    previous_destination: PixelPos<f32>,
     length_multiplier: f32,
     t: f32,
 }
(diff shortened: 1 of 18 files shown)


src/renderer/rendered_window.rs
```diff
@@ -152,69 +146,51 @@ impl RenderedWindow {
...
+            .min(grid_rect.max.x - grid_size.width)
+            .max(grid_rect.min.x);
 
         // For messages the last line is most important, (it shows press enter), so let the position go negative
         // Otherwise ensure that the window start row is within the screen
-        let mut y = destination.y.min(valid_rect.bottom - grid_size.y);
-        if matches!(self.window_type, WindowType::Message { .. }) {
-            y = y.max(valid_rect.top)
+        let mut y = destination.y.min(grid_rect.max.y - grid_size.height);
+        if !matches!(self.window_type, WindowType::Message { .. }) {
```
fredizzimo, line 178: This was actually a bug, the condition was the wrong way. So the top of floating windows got pushed out. 

But there is still some other bug with this floating window repositioning code. The floating border gets clipped when resizing the window. This is not a regression here though, and I don't know the cause, so it will be fixed separately.

src/renderer/cursor_renderer/mod.rs
```diff
@@ -445,10 +428,10 @@ impl CursorRenderer {
         // corners.
         let mut path = Path::new();
 
-        path.move_to(self.corners[0].current_position);
-        path.line_to(self.corners[1].current_position);
-        path.line_to(self.corners[2].current_position);
-        path.line_to(self.corners[3].current_position);
+        path.move_to(to_skia_point(self.corners[0].current_position));
```
Kethku, line 431: why isn't to_skia_point just a function on PixelPos?

src/renderer/grid_renderer.rs
```diff
@@ -162,42 +148,34 @@ impl GridRenderer {
...
-            self.draw_underline(canvas, style, underline_style, p1.into(), p2.into());
+            self.draw_underline(canvas, style, underline_style, p1, p2);
             drawn = true;
         }
 
         canvas.save();
-        canvas.clip_rect(region, None, Some(false));
-
-        let y_adjustment = self.shaper.y_adjustment();
+        canvas.clip_rect(to_skia_rect(&region), None, Some(false));
```
Kethku, line 178: same here. Imho we should have to_skia be a method here rather than a free function

src/units.rs
```diff
@@ -0,0 +1,109 @@
...
+pub type GridSize<T> = Size2D<T, Grid>;
+pub type PixelSize<T> = Size2D<T, Pixel>;
+
+pub type GridPos<T> = Point2D<T, Grid>;
+pub type PixelPos<T> = Point2D<T, Pixel>;
+
+pub type GridRect<T> = Box2D<T, Grid>;
+pub type PixelRect<T> = Box2D<T, Pixel>;
+
+pub fn to_skia_point(pos: PixelPos<f32>) -> skia_safe::Point {
```
Kethku, line 22: Ah I see. These are type definitions. In that case maybe an extension trait? Importing free functions that do this feels wrong

Kethku:
This is great. Barring a suggestion to add an extension trait for skia type conversions, this looks good to me. Once it is rebased I vote this should get merged asap

PR #2485, opened 2024-04-14
feat!: Support fractional grid and font sizes

fredizzimo:
This properly renders arbitrary font sizes, so the fudge factor is now removed. The hack for floating point font sizes is also removed, since it always caused the wrong spacing.

NOTE: This currently puts much more pressure on the Skia font cache, which can cause glitches when scrolling, so some optimization might be needed.

NOTE: Depends on a lot of other PRs, so it remains in draft until those are merged.

- Yes, the font size will be different, since no fudge factor is applied. So you might have to adjust your configuration accordingly.


fredizzimo:
I improved the skia font caching, which solves most of the performance problems, at least it's very usable on my system now. But there's still some unexplainable stuttering. I don't see any explanation in Tracy, and the GPU is also idling according to the tools that I have in hand right now.

I will try to find a way to do better GPU profiling on Linux.

Kethku:
Waiting on a rebase to review this one

src/renderer/fonts/caching_shaper.rs
```diff
@@ -373,16 +351,10 @@ impl CachingShaper {
...
-        let current_font_cache_size = font_cache_limit() as f32;
-        let percent_font_cache_used = font_cache_used() as f32 / current_font_cache_size;
-        if percent_font_cache_used > 0.9 {
-            warn!(
-                "Font cache is {}% full, increasing cache size",
-                percent_font_cache_used * 100.0
-            );
-            set_font_cache_limit((percent_font_cache_used * 1.5) as usize);
-        }
+    pub fn cleanup_font_cache(&self) {
```
Kethku, line 354: This is the only change that gives me pause. How can we be sure that the cache size constant is big enough?

Kethku:
lgtm baring a quick question about cache sizes

src/renderer/fonts/caching_shaper.rs
fredizzimo: There are a few reasons:

1. Even if the cache is too small it will still work, it just becomes a lot slower. 
2. The default cache size was 2 MB before, and we never hit that. We would have known, since there was a bug in the growing code which sets it to 1 or 2 bytes, it multiplied by percent not bytes.
3. The temporal cache during the scrolling is much bigger than the statically needed cache for just displaying the text, since it needs to store all subpixel positions encountered during the scrolling. With the tracy logger the total cache size never got bigger than around 2MB when scrolling completely through a few big buffers, so with the current code. The cleanup condition was not even met when storing all that in the cache. It would perhaps be worth doing the same with some Asian languages and a big font, and a completely full 4k+ screen, but I'm pretty sure that it's still way more than enough.
4. Even if we clean up all the cache needed for scrolling it would just be as slow as when you first start scrolling, which still has a pretty acceptable performance IMO.

fredizzimo:
Ok, it's rebased now

9mm:
So this one looks like it wrecks the menu again, but this time worse as now the shadow also appears to be off, in addition to the border:

image

fredizzimo:
Ah, sorry, I did not test it properly after resolving the conflicts. I will look at it now.

9mm:
All good 🙏🏼 Eternally grateful

fredizzimo:
https://github.com/neovide/neovide/pull/2496 should fix it.

fredizzimo:
But upon closer look, I think we might need to revert this. I think the font render quality is worse, and the text looks too thick and otherwise off.

PR #2486, opened 2024-04-14
feat!: Option key as meta configuration

9mm:
updating https://github.com/neovide/neovide/pull/2046


9mm:
i think i nuked his stuff, it was just too hard to resolve that without just starting over ... if i get this working i will try to attempt again because then i will know what to change, and ill keep his stuff

fredizzimo:
Yes, I think you should be able to keep most of the stuff in the Keyboard manager, but the rest probably needs to be rewritten anyway

9mm:
@fredizzimo so we are killing the entire old setting right? it seems like a lot of this window_wrapper code is trying to map old settings etc, but if we just make it a breaking change i assume all of this is a lot simpler

9mm:
so.... that is my first attempt. i really have no idea what a lot of this code was but i basically took a bunch of critical thinking guesses...

my test case i wrote in the original issue, i notice the Option (right) + c produces `ç` case works actually, and if i disable the setting then it produces ç for both option keys, which makes me think......... the most rudimentary raw level it is working.

Now i have to go test the different edge cases, test all settings values, etc.

Can you see if I made any blindingly stupid mistakes? I left a lot of code out that seemed to not be needed anymore.

Btw the test here did not work however... 

> Meta (left) + Command + y produces `<M-D-y>` instead of `<M-S-D-Y>`

So maybe this is a different problem unrelated to the issue, as i can definitely tell the meta key left/right IS working

===================

I can confirm that with the following setting:

```
```

Option (right) + c produces `ç`

Meta (left) + Command + y produces `<M-D-y>` instead of `<M-S-D-Y>`

I tested `OnlyLeft, OnlyRight, Both, None` and all work as expected

I dont use `M` in hotkeys much so if there is more specific test of what to look for please let me know. As of now I only tested by going:

Command + Meta (left option) + x = `<M-D-x>`

or

Command + Meta (left option) + Shift + x = `<M-D-x>` = `<M-D-X>`

fredizzimo:
@9mm, the general approach looks fine.

> @fredizzimo so we are killing the entire old setting right? it seems like a lot of this window_wrapper code is trying to map old settings etc, but if we just make it a breaking change i assume all of this is a lot simpler

Ideally, we would want that the old setting continues to work or do nothing and give a warning about deprecation. But if it's too much work, we can add it as a breaking change in the release notes, and maybe add a sticky issue.

> Btw the test here did not work however...
> Meta (left) + Command + y produces <M-D-y> instead of <M-S-D-Y>

Do you mean that pressing `Meta (left) + Command  + Shift + y` (note shift is included) still produces `<M-S-D-Y>` instead of `<M-D-y>`, I'm pretty sure that was part of the original fix. And should be fixed in your version as well.

9mm:
I take that back, it actually did work. i think i had the config wrong

also, i dont think its necessarily hard to keep the old option, but how do we manage that in the settings? i suppose it seems weird to keep old settings floating around as opposed to just putting breaking change info on the release notes as it seems like it's not a "hard" breaking change to fix, but this is my uneducated 2 cents

9mm:
specifically im referring to the key name in the struct. like we would have 2 keys for 2 different option settings

src/window/keyboard_manager.rs
```diff
@@ -68,6 +73,26 @@ impl KeyboardManager {
...
+                        }
+                        OptionAsAlt::OnlyRight => {
+                            self.modifiers.ralt_state() == ModifiersKeyState::Pressed
+                        }
+                        OptionAsAlt::None => false,
+                    };
+
+                    #[cfg(not(target_os = "macos"))]
+                    {
+                        self.meta_is_pressed = self.modifiers.state().alt_key();
```
fredizzimo, line 93: This confuses me a bit, since no one calls the alt key meta anymore. More typically the windows key is called meta. 

But I guess in this context it's ok, since Vim(Neovim) also refers to it as meta.

fredizzimo:
For the old value, in the code we should only ever use the new values, but at the parsing stage, we can support the old setting, and convert it to `both` automatically.

We can then show a warning using this system:
https://github.com/neovide/neovide/blob/fcf5e87bf8a75727ee75da961ce98b2f51fd0cb2/src/error_handling.rs#L18-L37

But support for warnings need to be added, instead of just errors.

9mm:
ooooh great idea. ok I can do that

9mm:
Note to self of things to fix:
  * parse old setting
  * `None` doesnt appear to be parsing properly as a default:

```
ERROR [neovide::window::settings] Setting neovide_input_macos_option_key_is_meta expected one of OnlyLeft, OnlyRight, Both, or None, but received Error("unknown variant `\"None\"`, expected one of `OnlyLeft`, `OnlyRight`, `Both`, `None`"): String(Utf8String { s: Ok("\"None\"") })
```

9mm:
So one area I havent really learned about yet in rust is serde/serialization/deserialization. how would you handle this part?

> ERROR [neovide::window::settings] Setting neovide_input_macos_option_key_is_meta expected one of OnlyLeft, OnlyRight, Both, or None, but received Error("unknown variant \"None\", expected one of OnlyLeft, OnlyRight, Both, None"): String(Utf8String { s: Ok(""None"") })

It looks like its returning `"None"` as a string

```rust
#[cfg(target_os = "macos")]
impl ParseFromValue for OptionAsMeta {
    fn parse_from_value(&mut self, value: Value) {
        let s = value.as_str().unwrap();
        match OptionAsAlt::deserialize(s.into_deserializer()) as Result<OptionAsAlt, value::Error> {
            Ok(oa) => *self = OptionAsMeta(oa),
            Err(e) => error!("Setting neovide_input_macos_option_key_is_meta expected one of OnlyLeft, OnlyRight, Both, or None, but received {:?}: {:?}", e, value),
        };
    }
}
```

I notice even if something is specified in the config, it comes back `Some("None")` which I think is correct, however, it also makes me think there should be no error there then, otherwise it prints the error before immediately getting the settings changed event.

9mm:
Ah, it seems the first time it does include quotes

```
[src/window/settings.rs:92:9] s = "\"None\""
ERROR [neovide::window::settings] Setting neovide_input_macos_ ......
[src/window/settings.rs:92:9] s = "OnlyLeft"
```

fredizzimo:
That seems like something that would have to be fixed in Winit, their code looks like this
https://rust-windowing.github.io/winit/src/winit/platform/macos.rs.html#429, it probably needs a serde rename attribute https://serde.rs/field-attrs.html#field-attributes

You can try by defining your own enum type, and add the correct attributes, that also allows you to rename all values to `kebab-case`, like we do here https://github.com/neovide/neovide/blob/fcf5e87bf8a75727ee75da961ce98b2f51fd0cb2/src/settings/config.rs#L37

You then need to convert that enum to the winit one, but that should be easy with a simple match statement.

9mm:
rust is intense lol

I got a lot of what you just said done, however... I suppose the same problem exists. I dont fully understand why initially it passes a string with extra quotes, and then it passes a standard string.

I do see that its properly doing kebab case... sort of.......

```
[src/window/settings.rs:104:17] value = "\"none\""
[src/window/settings.rs:104:17] value = "OnlyLeft"
```

https://github.com/neovide/neovide/pull/2486/commits/2f59222b806cee09ec63ea860761e15fe9e76c22

This is what the output is for my `dbg!` print here

9mm:
Oh nevermind...... i see, the expected values they actually set in the config are now properly kebab case as well. duh

9mm:
For the record I know I can use `(_) => ` however then I can't differentiate between an empty value, and an invalid value, because both would match that.

Is this expected behavior that `to_str` returns extra quotes if its a default?

fredizzimo:
It should not, but did you try to rename the None option yourself, above it
`#[serde(rename = "none")]

9mm:
> It should not, but did you try to rename the None option yourself, above it `#[serde(rename = "none")]`

I implemented it myself like you suggested. I think this is doing the same thing right?

```rust
#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
#[serde(rename_all = "kebab-case")]
#[cfg(target_os = "macos")]
pub enum OptionAsAlt {
    OnlyLeft,
    OnlyRight,
    Both,
    #[default]
    None,
}
```

fredizzimo:
I was thinking about this
```rust
#[serde(rename_all = "kebab-case")]
#[cfg(target_os = "macos")]
pub enum OptionAsAlt {
    OnlyLeft,
    OnlyRight,
    Both,
    #[serde(rename = "none")]
    #[default]
    None,
}
```

But I think I see the problem now, I will get back to you soon.

9mm:
Ok... no rush. thank you!

At least the part im stuck on is pretty much all the stuff I havent read about yet 💀

I see serde is quite complex and I just really hadnt researched all that yet.

fredizzimo:
No, it's not what I thought, your code looks correct. And I think other than allowing kebab case, our custom enum doesn't have that much use, and the problem is in `value.as_str()`, so nothing to do with Serde yet.

How do you set the value on the Neovim side, maybe there's something strange going on there?

9mm:
> No, it's not what I thought, your code looks correct. And I think other than allowing kebab case, our custom enum doesn't have that much use, and the problem is in `value.as_str()`, so nothing to do with Serde yet.
> 
> How do you set the value on the Neovim side, maybe there's something strange going on there?

Well, its actually *correct* when it finally loads the setting.

The problem is, _before_ the setting gets loaded, the match gets called with the extra quoted `"none"``.

So, if I have no configuration items at all in my config, I get this in my match call:

```
[src/window/settings.rs:104:17] value = "\"none\""
```

If I put this in my config:

```
  vim.g.neovide_input_macos_option_key_is_meta = "none"
```

then my match statement gets called twice with:

```
[src/window/settings.rs:104:17] value = "\"none\""
[src/window/settings.rs:104:17] value = "none"
```

9mm:
Also when i say match statement, its the one in settings.rs to parse the value. 

```rust
#[cfg(target_os = "macos")]
impl ParseFromValue for OptionAsMeta {
    fn parse_from_value(&mut self, value: Value) {
        match value.as_str() {
            Some(value) => {
                dbg!(&value);
            }
            _ => error!(
                "Setting neovide_input_macos_option_key_is_meta expected one of OnlyLeft, OnlyRight, Both, or None, but received {:?}",
                value
            ),
        }
        // match OptionAsAlt::deserialize(s.into_deserializer()) as Result<OptionAsAlt, value::Error> {
        //     Ok(oa) => *self = OptionAsMeta(oa),
        //     Err(e) => error!("Setting neovide_input_macos_option_key_is_meta expected one of OnlyLeft, OnlyRight, Both, or None, but received {:?}: {:?}", e, value),
        // };
    }
}
```

fredizzimo:
Hm. it should only be called once, what happens if you don't call it at all?

And maybe it's actually, a real None the first. time, can you check if `value.is_str()`?, Maybe it is `value.is_nil()`?

9mm:
I'm not quite sure what you mean "dont call it at all"

If I just put this:

```rust
#[cfg(target_os = "macos")]
impl ParseFromValue for OptionAsMeta {
    fn parse_from_value(&mut self, value: Value) {
        dbg!(&value);
    }
}
```

I get this:

```
[src/window/settings.rs:101:9] &value = String(
    Utf8String {
        s: Ok(
            "\"none\"",
        ),
    },
)
[src/window/settings.rs:101:9] &value = String(
    Utf8String {
        s: Ok(
            "none",
        ),
    },
)
```

If I dont put any `impl ParseFromValue` at all, then I get a warning that `parse_from_value not found for this struct`. This was code the other guy wrote in here so I assumed he was implementing required behavior to deserialize from this "messagepack" stuff whatever that is, i assume some kind of serialization format for the config or something

fredizzimo:
Sorry, I meant not setting it at all on the Neovim side.

9mm:
Ah, if I put this in my config:

```lua
  -- other
  -- vim.g.neovide_input_macos_alt_is_meta = true
  -- vim.g.neovide_input_macos_option_key_is_meta = "none"
  -- vim.g.neovide_input_macos_option_key_is_meta = "only_left"
```

then I get this:

```
[src/window/settings.rs:101:9] &value = String(
    Utf8String {
        s: Ok(
            "\"none\"",
        ),
    },
)
```

9mm:
I also just checked to make sure its not in my config file twice or anything, it doesnt appear to be the case.

fredizzimo:
Does it say `Initial value load failed` in the log file? And is there any other related log entries?

9mm:
Mmm good question. Yes I see a whole bunch of those, including the one for the option in question (the one which I commented out in the neovim config)

```
TRACE [neovide::settings] Initial value load failed for padding_bottom: Error processing request: 1 - 'Key not found: neovide_padding_bottom')
TRACE [neovide::bridge::handler] Neovim notification: "setting_changed"
TRACE [neovide::settings] Initial value load failed for input_macos_option_key_is_meta: Error processing request: 1 - 'Key not found: neovide_input_macos_option_key_is_meta')
TRACE [neovide::bridge::handler] Neovim notification: "setting_changed"
TRACE [neovide::settings] Initial value load failed for cursor_animate_command_line: Error processing request: 1 - 'Key not found: neovide_cursor_animate_command_line')
TRACE [neovide::bridge::handler] Neovim notification: "setting_changed"
TRACE [neovide::settings] Initial value load failed for refresh_rate: Error processing request: 1 - 'Key not found: neovide_refresh_rate')
TRACE [neovide::bridge::handler] Neovim notification: "setting_changed"
```

Other relevant lines:

```
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: "67:27"
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: "67"
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: "lue load failed for input_macos_option_key_is_meta: Error processing request: 1 - 'Key not found: neovide_input_macos_option_key_"
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: "lue load failed for cursor_animate_command_line: Error processing request: 1 - 'Key not found: neovide_cursor_animate_command_lin"
```

Here's some more from the `CURRENT` log file.

```
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: "TRACE [neovide::settings] Initial value load failed for floating_z_height: Error processing request: 1 - 'Key not found: neovide_floating_z_height')"
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: "TRACE [neovide::settings] Initial value load failed for cursor_antialiasing: Error processing request: 1 - 'Key not found: neovide_cursor_antialiasing')"
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: "TRACE [neovide::settings] Initial value load failed for padding_bottom: Error processing request: 1 - 'Key not found: neovide_padding_bottom')"
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: "107"
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: "TRACE [neovide::settings] Initial value load failed for input_macos_"
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: "opt"
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: "ion_key_is_meta: Error processing request: 1 - 'Key not found: neovide_input_macos_"
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: "ion_key_"
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: "TRACE [neovide::settings] Initial value load failed for cursor_animate_command_line: Error processing request: 1 - 'Key not found: neovide_cursor_animate_command_lin"
TRACE [neovide::renderer::fonts::caching_shaper] Shaping text: "TRACE [neovide::settings] Initial value load failed for refresh_rate: Error processing request: 1 - 'Key not found: neovide_refresh_rate')"
```

9mm:
I think i found where its coming from. its in his other code

```
#[cfg(target_os = "macos")]
impl From<OptionAsMeta> for Value {
    fn from(oam: OptionAsMeta) -> Self {
        let s = serde_json::to_string(dbg!(&oam.0)).unwrap();
        Value::String(Utf8String::from(s))
    }
}
```

this outputs

```
[src/window/settings.rs:106:39] &oam.0 = None
```

Then below its convering it to this:

```
    Utf8String {
        s: Ok(
            "\"none\"",
        ),
    },
)
```

TBH do i even need this OptionAsMeta struct? What if i just removed that and just used the regular enum that its wrapping? I guess I dont know the point of why he put this struct

```
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
#[cfg(target_os = "macos")]
pub struct OptionAsMeta(pub OptionAsAlt);
```

9mm:
Yeah...... i think this is the right track. I have no idea what that crazy stuff in his code was doing. I am just going to model `VfxMode`

9mm:
Ok it works! https://github.com/neovide/neovide/pull/2486/commits/475a28fd6b656fad40384557600882d2ba39ca43

I modelled CursorVfx

9mm:
> For the old value, in the code we should only ever use the new values, but at the parsing stage, we can support the old setting, and convert it to `both` automatically.
> 
> We can then show a warning using this system:
> 
> https://github.com/neovide/neovide/blob/fcf5e87bf8a75727ee75da961ce98b2f51fd0cb2/src/error_handling.rs#L18-L37
> 
> But support for warnings need to be added, instead of just errors.

Ok, regarding this above...

If I want to allow a deprecated setting:

```
  vim.g.neovide_input_macos_alt_is_meta = true
```

Is the only way to do this to basically leave `input_macos_alt_is_meta` in the `WindowSettings` struct and then wait for the settings change events?

Or is there a way even more lower level than that so I dont need to put deprecated properties into the WindowSetting struct?

9mm:
Hmm ok. how do I deal with race conditions? The problem seems to be that these settings events get triggered even if there is no config option set... therefore even if the user only has the old setting, and not the new setting, I still get 2 events which contradict:

```
[src/window/window_wrapper.rs:268:20] enabled = false
[src/window/window_wrapper.rs:262:49] option = None
```

So when I respond to `enabled`, it just gets overwritten by the default value of of the proper option.

```rust
            #[cfg(target_os = "macos")]
            WindowSettingsChanged::InputMacosOptionKeyIsMeta(option) => {
                if self.macos_option_is_meta != dbg!(option) {
                    self.set_macos_option_is_meta(option);
                }
            }
            #[cfg(target_os = "macos")]
            WindowSettingsChanged::InputMacosAltIsMeta(enabled) => {
                if dbg!(enabled) {
                    error_msg!(concat!(
                        "neovide_input_macos_alt_is_meta has now been deprecated. ",
                        "Use neovide_input_macos_option_key_is_meta instead. ",
                        "Please check https://neovide.dev/configuration.html#macos-option-key-is-meta for more information.",
                    ));
                    self.set_macos_option_is_meta(settings::OptionAsMeta::Both);
                }
            }
```

fredizzimo:
I think that you can detect it by either setting a default value, or setting a dummy value as default, which you can ignore, since no user should set it to dummy.

9mm:
ok thanks! i thought about that but wasnt sure if there was a better way w/o cluttering the interface with dummy things

9mm:
There is some exceptionally confusing stuff going on....... part of the problem is I used a lot of the previous commit without actually looking at the purpose.

For instance, i was pulling my hair out why the config file was not lining up with behavior... it seems I dont even need to call 

```
            self.skia_renderer.window().set_option_as_alt(winit_option);
```

At all.... i need to research this function because the other guys implementation here in KeyboardManager is already seemingly doing this behavior.... so basically I need to figure out why this was even here if it already works perfect with this code below

```
                    let ws = SETTINGS.get::<WindowSettings>();
                    self.meta_is_pressed = match ws.input_macos_option_key_is_meta {
                        OptionAsMeta::Both => self.modifiers.state().alt_key(),
                        OptionAsMeta::OnlyLeft => {
                            self.modifiers.lalt_state() == ModifiersKeyState::Pressed
                        }
                        OptionAsMeta::OnlyRight => {
                            self.modifiers.ralt_state() == ModifiersKeyState::Pressed
                        }
                        OptionAsMeta::None => false,
                    };
```

9mm:
Mmm interesting, so it almost seems like there are "3" states, not 2 like I was under the impression of.

This code in keyboard manager handles the `<M-D-x>` while the other call to `window()` function is what allows you to do `ç`

9mm:
So, in a way theres sort of 2 layers to this?

1. The `WindowSettings` in `settings.rs` which handles default values / set values. This affects the `KeyboardManager` knowing which key to use
2. The `WinitWindowWrapper` object and handling WindowSettings events. 

This is pretty complicated with this old setting. i think im going to nuke everything and just get it working again, period. and then try to approach it again. Ive spent like 4 hours staring at simple booleans wondering WTF is going on, its so convoluted and hard to follow that Im just going to mentally reset, and just look at it with fresh eyes once its in a working state again (without supporting the old value)

9mm:
@fredizzimo is it correct behavior for KeyboardManager to access settings directly? Or should it be accessing `WinitWindowWrapper`?

fredizzimo:
Yes, I don't see any reason for it to go through the window wrapper.

9mm:
Got it.

I'm still using `WindowWrapper` flags to know whether or not to call the winit function after I receive a WindowSetting event. I just followed the guidance of all the other WindowWrapper events. Is this still advisable? Or should I somehow just be interesting with Settings directly, and exclusively.

```rust
#[cfg(target_os = "macos")]
WindowSettingsChanged::InputMacosOptionKeyIsMeta(option) => {
    if self.macos_option_as_meta != option {
        self.set_macos_option_as_meta(option);
    }
}
#[cfg(target_os = "macos")]
WindowSettingsChanged::InputMacosAltIsMeta(enabled) => {
    if enabled {
        error_msg!(concat!(
            "neovide_input_macos_alt_is_meta has now been deprecated. ",
            "Use neovide_input_macos_option_key_is_meta instead. ",
            "Please check https://neovide.dev/configuration.html#macos-option-key-is-meta for more information.",
        ));
    }
}
```

fredizzimo:
If you can do it without flags it's better, there's less things that can go wrong that way.And if you need to query the state you could just ask Winit for it

9mm:
Ok, all flags are removed.

Everything works great... There's only one final issue...

Right now I have a simple deprecation warning for the old setting via a `WindowSettingsChanged::InputMacosAltIsMeta` listener (which also requires `input_macos_alt_is_meta` old setting on `WindowSettings`.

This is only a warning though, it doesn't attempt to set the `Both` behavior if that legacy setting is set to `true`.

```rust
#[cfg(target_os = "macos")]
WindowSettingsChanged::InputMacosOptionKeyIsMeta(option) => {
    self.set_macos_option_as_meta(option);
}
#[cfg(target_os = "macos")]
WindowSettingsChanged::InputMacosAltIsMeta(enabled) => {
    if enabled {
        error_msg!(concat!(
            "neovide_input_macos_alt_is_meta has now been deprecated. ",
            "Use neovide_input_macos_option_key_is_meta instead. ",
            "Please check https://neovide.dev/configuration.html#macos-option-key-is-meta for more information.",
        ));
    }
}
```

IF I were to attempt to set `Both` if I detect the legacy setting to `true`, that's where it starts quite hairy because...

I've actually re-typed this about 50 times and there's really no simple way to explain it... maybe it would be better to simply just ask, how would you do it?

I need to basically update 2 things....

1. I have to call `self.skia_renderer.window().set_option_as_alt()` which happens as a result of `WindowSettings` events
2. I have to set the state of `self.meta_is_pressed` with custom logic which directly looks at `SETTINGS.get::<WindowSettings>()` to detect the value

And both of these needs to stay in sync, if i were to do this legacy setting.

The tricky bit is that 1) can be set for default events even if that setting isnt even in their config, so it creates race conditions, and even if i do use a flag, that flag must be kept in sync with 2) which doesnt look at flags at all, it just looks at the WindowSettings value

fredizzimo:
The message is an error, not a warning, right? In that case I think it's fine to not do anything, it clearly tells what to do and if things don't work the users can blame themselves for not reading the message. So, it should be fine as it is.

fredizzimo:
We just need to change `deprecated` to `removed`.

9mm:
Ok cool. it is an `error_msg!`. I will change 'deprecated' to 'no longer supported' so its not implying it still works

sorry for all the tons of noise on this... I just dont want to add junk to this amazing repo, and this tbh is my first time writing actual rust 💀

9mm:
Ah beat me to it

fredizzimo:
I did not fix the macOS Clippy warnings in my PR, so do you think you can do it here? 

I think the `window` parameter can be removed completely and I don't think we need to deprecate the function, just because the feature is deprecated for the users.

9mm:
definitely, I was actually going to fix that stuff but stopped in case they were there for a reason haha.

fredizzimo:
I actually don't see any code that would fix this bug:
> Meta (left) + Command + y produces <M-D-y> instead of <M-S-D-Y>. 

As far as I can see the condition for `include_shift` has not changed. But I don't see why it would have been broken before either. Maybe it was fixed elsewhere even without these changes, otherwise I'm missing something subtle. 

Can you test that without these changes?

9mm:
Heres the most recent result of testing this PR. I copied the test cases I did above but just re-ran through them all again. LMK if this works. I am not sure the issue with shift, but heres the results:


```
vim.g.neovide_input_macos_option_key_is_meta = 'only_left'
````

Option (right) + c produces `ç`

Meta (left) + Command + y produces `<M-D-y>` instead of `<M-S-D-Y>`

Command + Meta (left option) + x = `<M-D-x>`

Command + Meta (left option) + Shift + x = `<M-D-X>`

**Comment by fredizzimo** [comment]
Ah it's probably this change:
```rust
       #[cfg(target_os = "macos")]
        if self.meta_is_pressed {
            return key_event
                .key_without_modifiers()
                .to_text()
                .map(|text| self.format_key_text(text, false));
        }
```

And yes, those results look correct. If you just can verify that `Meta (left) + Command + y` produces `<M-S-D-Y>` on main and is indeed fixed here.

fredizzimo:
Don't worry about those other clippy warnings, we will make sure to merge this first.
https://github.com/neovide/neovide/pull/2483

9mm:
Ah ok, I will revert those last 2

and i will test main quick

9mm:
I tested main. Assuming what you wrote is a typo and you meant the following, then this is what occurs on main:

Meta (left) + Command + Shift + y = `<M-S-D-Y>`

fredizzimo (approved):
Great! Thank you, everything looks good now.

9mm:
awesome!! Thank you for all your help

Issue #2464, opened 2024-04-04
#2455 regresses scrolling on certain floating windows with borders


griwes:
**Describe the bug**
Before #2455 (and after the previous heuristics-based change), floating windows like those created with nvim-cmp smooth scrolled just their contents, not their borders. Now this behavior has regressed, and neovide attempts to scroll the contents together with the floating window.

So far I have noticed this with:
* nvim-cmp;
* dropbar's navigation dropdowns;
* neozoom's zoomed in windows.

Noice's windows scroll fine, but that's probably because of the nui hack to emulate borders without actually using floating window borders (which is a whole separate discussion, but causes me grief when trying to do consistent winblend with neovide :upside_down_face:).

**To Reproduce**
1. Install nvim-cmp.
2. Configure its windows to have a border (rounded in my case, but presumably repro's with the other options too).
3. Trigger the completion window and scroll within it.

**Expected behavior**
The borders of those floating windows should not be scrolling. This is the behavior prior to #2455.

**Desktop (please complete the following information):**
 - OS: Debian sid
 - Neovide Version 0.12.2 (git head as of typing this)
 - Neovim Version v0.10.0-dev-2791+g99b3a068d

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
https://gist.github.com/griwes/3e542ad7dd304a9ef2de8647abdcf379

griwes:
@Kethku [here](https://www.dropbox.com/scl/fi/2xgk57e9pxon4j4bu6pz5/2024-04-04-03-14-14.mkv?rlkey=cdpfijp7bkzgj68vuzfdor6bz&dl=0), hopefully that's legible enough to see what's going on? Watch what happens with the top and bottom borders when I scroll.

fredizzimo:
I can repeat this here, simply by running
`:lua vim.api.nvim_open_win(0, true, {relative="win", row=10, col=10, width=100, height=40, border="single"})`

If you add a winbar to the same window with `setlocal winbar=test` it starts scrolling correctly again. And so does a floating window without borders and winbar.

Just a bottom or a top border is also buggy
`lua vim.api.nvim_open_win(0, true, {relative="win", row=10, col=10, width=100, height=40, border="=", "=", "=", "", "", "", "", ""}})` or
`lua vim.api.nvim_open_win(0, true, {relative="win", row=10, col=10, width=100, height=40, border={"", "", "", "", "=", "=", "=", ""}})`

[...]

PR #2487, opened 2024-04-15
Address viewport margin regression by sorting DrawCommand batches

Kethku:
According to neovim folks, they can and do send redraw events in any order so long as its in the same batch. This PR addresses our first issue to come of that by manually sorting the batches so that our position WindowDrawCommands always happen before all others.

Fixes https://github.com/neovide/neovide/issues/2464

- Fix

- No

### DIFF
--- a/src/editor/draw_command_batcher.rs
+++ b/src/editor/draw_command_batcher.rs
@@ -20,7 +20,7 @@ impl DrawCommandBatcher {
     }
 
     pub fn send_batch(&self, proxy: &EventLoopProxy<UserEvent>) {
-        let mut batch: Vec<DrawCommand> = self.batch.borrow_mut().split_off(0).into();
+        let mut batch: Vec<DrawCommand> = self.batch.borrow_mut().split_off(0);
         // Order the draw command batches such that window draw commands are handled first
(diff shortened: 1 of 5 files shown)


fredizzimo (commented):
There are a couple of clippy warnings that should be fixed, but other than that it looks good.

I'm slightly worried that we might process things in different order on the editor and UI side though. But as long as the end result is the same and we are actually allowed to sort it should be fine.

griwes:
@Kethku with these changes, I now see... this - looks like the last line of the floating window is duplicated over the bottom border? It looks like it duplicates when the contents of the window are updated, I think, because for dropbar, it only happens once I actually start selecting an item in the menu. But otherwise I can see this happening consistently in floating windows.
Screenshot from 2024-04-20 11-36-30

Kethku:
ah ok ill look into it. should just be an off by one error

griwes:
Having played a bit with cmp, it looks like it actually renders the last line from the _previous_ version of the contents. But yeah, definitely looks like an off by one somewhere.

fredizzimo:
That was also reported here:
* https://github.com/neovide/neovide/pull/2467#issuecomment-2067342150

But I guess this PR is the more likely one to have caused the issue.

Issue #2488, opened 2024-04-17
Migrate Windows-API Relevant Dependency from winapi to windows-rs


InfyniteHeap:
**Is your feature request related to a problem? Please describe.**
The exist [`winapi`](https://github.com/retep998/winapi-rs) crate has stopped updating for 3 years and most of functions might seems been obsolete. These functions also might have some bugs and cannot fully take advantage of Windows OS. Plus now Microsoft had published their official [`windows-rs`](https://github.com/microsoft/windows-rs) crate, so I want to migrate the dependency from `winapi` to `windows-rs`.

**Describe the solution you'd like**
Migrate the dependency from `winapi` to `windows-rs`.

**Describe alternatives you've considered**
It seems nothing better than Microsoft official's `windows-rs`. XD

**Additional context**
The reasons why I propose this suggestion:
- It can make the code easier to be read and to be maintained.
- It can make the code safer.
  - Though using `windows-rs` cannot avoid using `unsafe`, the code should still be safer because of the first reason. That is, decreasing the complexity of the `unsafe` code can decrease the appearing frequency of potential bugs that may lead to memory safety issues.

PR #2490, opened 2024-04-19
Refactor the registry-relevant functions.

InfyniteHeap:
The right click menu register functions are refactored with `windows` and `windows-registry` crates. The refactored code seems simpler and easier to be read and maintained.

- Refactor

- No

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -111,6 +111,7 @@ windows = { version = "0.56.0", features = [
     "Win32_System_Registry",
     "Win32_UI_HiDpi",
 ] }
+windows-registry = "0.1.1"
 
 [target.'cfg(not(target_os = "windows"))'.dependencies]
 skia-safe = { version = "0.68.0", features = ["gl", "textlayout"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,13 @@ time = { version = "0.3.34", features = ["macros", "formatting"] }
 tokio = { version = "1.25.0", features = ["full"] }
 tokio-util = { version = "0.7.4", features = ["compat"] }
 toml = "0.8.12"
-tracy-client-sys = { version = "0.22.0", optional = true, default-features = false, features = ["broadcast", "delayed-init", "enable", "manual-lifetime", "fibers"] }
+tracy-client-sys = { version = "0.22.0", optional = true, default-features = false, features = [
+    "broadcast",
+    "delayed-init",
+    "enable",
+    "manual-lifetime",
+    "fibers",
+] }
 unicode-segmentation = "1.9.0"
 which = "6.0.1"
 winit = { version = "=0.29.15", features = ["serde"] }
@@ -94,16 +100,37 @@ winapi = { version = "0.3.9", features = [
     "wincon",
     "winerror",
     "winuser",
-    ]}
+] }
 # for ComPtr
 wio = { version = "0.2.2" }
 skia-safe = { version = "0.68.0", features = ["gl", "d3d", "textlayout"] }
+windows = { version = "0.56.0", features = [
+    "Win32_Security",
+    "Win32_System_LibraryLoader",
+    "Win32_System_Registry",
+    "Win32_UI_HiDpi",
+] }
 
 [target.'cfg(not(target_os = "windows"))'.dependencies]
 skia-safe = { version = "0.68.0", features = ["gl", "textlayout"] }
 
 [target.'cfg(target_os = "macos")'.dependencies]
-icrate = { version = "0.0.4", features = [ "apple", "Foundation", "Foundation_NSThread", "AppKit", "AppKit_NSColor", "AppKit_NSEvent", "AppKit_NSView", "AppKit_NSWindow", "AppKit_NSViewController", "AppKit_NSMenu", "AppKit_NSMenuItem", "AppKit_NSOpenPanel", "AppKit_NSScreen", "Foundation_NSArray" ] }
+icrate = { version = "0.0.4", features = [
+    "apple",
+    "Foundation",
+    "Foundation_NSThread",
+    "AppKit",
+    "AppKit_NSColor",
+    "AppKit_NSEvent",
+    "AppKit_NSView",
+    "AppKit_NSWindow",
+    "AppKit_NSViewController",
+    "AppKit_NSMenu",
+    "AppKit_NSMenuItem",
+    "AppKit_NSOpenPanel",
+    "AppKit_NSScreen",
+    "Foundation_NSArray",
+] }
 objc2 = "0.4.1"
 
 [target.'cfg(target_os = "windows")'.build-dependencies]
--- a/src/windows_utils.rs
+++ b/src/windows_utils.rs
@@ -1,222 +1,177 @@
-use std::{
-    ffi::CString,
-    ptr::{null, null_mut},
+use windows::core::*;
+use windows::Win32::Foundation::*;
+use windows::Win32::System::LibraryLoader::GetModuleFileNameA;
+use windows::Win32::System::Registry::{
+    RegCloseKey, RegCreateKeyExA, RegDeleteTreeA, RegSetValueExA, HKEY, HKEY_CURRENT_USER,
+    KEY_READ, KEY_WRITE, REG_OPTION_NON_VOLATILE, REG_SZ,
 };
-
-use winapi::{
-    shared::{
-        minwindef::{DWORD, HKEY, MAX_PATH},
-        windef::DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2,
-    },
-    um::{
-        libloaderapi::GetModuleFileNameA,
-        winnt::{KEY_WRITE, REG_OPTION_NON_VOLATILE, REG_SZ},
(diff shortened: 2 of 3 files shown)


InfyniteHeap:
I think this commit could be a good start to migrate from `winapi` to `windows`, as what I mentioned in issue #2488.

src/windows_utils.rs
```diff
@@ -1,222 +1,177 @@
...
-            null_mut(),
-            buffer.as_mut_ptr() as *mut i8,
-            MAX_PATH as DWORD,
-        );
-        CString::from_vec_unchecked(buffer)
-            .into_string()
-            .unwrap_or_else(|_| "".to_string())
-            .trim_end_matches(char::from(0))
-            .to_string()
+        len = GetModuleFileNameA(HMODULE::default(), &mut buffer);
```
fredizzimo, line 25: I know that we used the `A` variants before, but I would prefer if we used the `W` variants, like `GetModuleFileNameW`, so that we properly handle special characters in the file path and registry keys.

src/windows_utils.rs
```diff
@@ -1,222 +1,177 @@
...
     }
+    buffer.truncate(len as usize);
+    String::from_utf8(buffer).unwrap()
```
fredizzimo, line 28: Note, when you switch to the wide char variants of the windows function, the strings are actually utf16 encoded.

src/windows_utils.rs
```diff
@@ -1,222 +1,177 @@
...
 
 fn unregister_rightclick() -> bool {
-    let str_registry_path_1 =
-        CString::new("Software\\Classes\\Directory\\Background\\shell\\Neovide").unwrap();
-    let str_registry_path_2 = CString::new("Software\\Classes\\*\\shell\\Neovide").unwrap();
     unsafe {
-        let s1 = RegDeleteTreeA(HKEY_CURRENT_USER, str_registry_path_1.as_ptr());
-        let s2 = RegDeleteTreeA(HKEY_CURRENT_USER, str_registry_path_2.as_ptr());
-        s1 == 0 && s2 == 0
+        RegDeleteTreeA(HKEY_CURRENT_USER, REGISTRY_PATH_DIRECTORY).0 == 0
```
fredizzimo, line 33: It's more correct to check for `ERROR_SUCCESS` here instead of 0.

And the same thing applies for other similar checks.

src/windows_utils.rs
```diff
@@ -1,222 +1,177 @@
...
+            None,
             &mut registry_key,
-            null_mut(),
-        ) != 0
+            None,
+        )
+        .0 != 0
         {
-            RegCloseKey(registry_key);
+            let _ = RegCloseKey(registry_key);
```
fredizzimo, line 59: The rust analyzer, recommends this pattern of `let _ = ` to ignore the result, but I think in our code base it's more common to use:
`RegCloseKey(registry_key).ok();`

src/windows_utils.rs
```diff
@@ -1,222 +1,177 @@
...
-        }
-        RegCloseKey(registry_key);
+
+        let _ = RegSetValueExA(
+            registry_key,
+            PCSTR::null(),
+            0,
+            REG_SZ,
+            Some(neovide_description.as_bytes()),
+        );
```
fredizzimo, line 69: I think we should return a failure from this function instead of ignoring the error if the setting the value fails. NOTE: The key still needs to be closed in that case.

src/windows_utils.rs
```diff
@@ -1,222 +1,177 @@
...
     unsafe {
-        let s1 = RegDeleteTreeA(HKEY_CURRENT_USER, str_registry_path_1.as_ptr());
-        let s2 = RegDeleteTreeA(HKEY_CURRENT_USER, str_registry_path_2.as_ptr());
-        s1 == 0 && s2 == 0
+        RegDeleteTreeA(HKEY_CURRENT_USER, REGISTRY_PATH_DIRECTORY).0 == 0
+            && RegDeleteTreeA(HKEY_CURRENT_USER, REGISTRY_PATH_FOLDER).0 == 0
     }
 }
 
 fn register_rightclick_directory() -> bool {
```
fredizzimo, line 38: It's more idiomatic to return a `Ressult<(), ()> here instead of a boolean indicating failure.

src/windows_utils.rs
```diff
@@ -1,222 +1,177 @@
...
-        for &(key, keytype, value_ptr, size_in_bytes) in &registry_values {
-            RegSetValueExA(
-                registry_key,
-                key,
-                0,
-                keytype,
-                value_ptr,
-                size_in_bytes as u32,
-            );
-        }
```
fredizzimo, line 98: I like the removal of the loop, which does not seem to be necessary at all here.

src/windows_utils.rs
```diff
@@ -1,222 +1,177 @@
...
-        RegCloseKey(registry_key);
+
+        let _ = RegSetValueExA(
+            registry_key,
+            PCSTR::null(),
+            0,
+            REG_SZ,
+            Some(neovide_description.as_bytes()),
```
fredizzimo, line 68: The rust strings are not null terminated, and this function requires null terminated strings, so you still need the `CString`, that the old code was using.

src/windows_utils.rs
```diff
@@ -240,6 +195,6 @@ pub fn unregister_right_click() {
 
 pub fn windows_fix_dpi() {
     unsafe {
-        SetProcessDpiAwarenessContext(DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2);
+        SetProcessDpiAwarenessContext(DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2).unwrap();
```
fredizzimo, line 198: I think we could use `expect()` instead of unwrap here, with a string describing the error. It's easier for us to know what went wrong that way, since we don't always have the callstack available when users report errors to us.

fredizzimo (changes requested):

fredizzimo:
I think it would be worth having another go trying the `windows-registry`, since it deals with the null terminating and wide characters properly. 

It's new, so there's a small possibility that it is broken, but I don't see any bug reports about it, so I think that chance is very low. It's not very widely used, but still there's 2000 downloads the past 90 days, so I would think that someone else has encountered issues.

InfyniteHeap:
Hi @fredizzimo , I've finished changing the logics.

Cargo.toml
```diff
@@ -95,16 +101,38 @@ winapi = { version = "0.3.9", features = [
...
     "winuser",
-    ]}
+] }
 # for ComPtr
 wio = { version = "0.2.2" }
 skia-safe = { version = "0.68.0", features = ["gl", "d3d", "textlayout"] }
+windows = { version = "0.56.0", features = [
+    "Win32_Security",
+    "Win32_System_LibraryLoader",
+    "Win32_System_Registry",
```
InfyniteHeap, line 111: These two redundant features (`LibraryLoader` and `Registry`) will be removed from my next PR.

src/windows_utils.rs
```diff
@@ -1,245 +1,83 @@
...
+
+    let key = CURRENT_USER.create(REGISTRY_PATH_FOLDER_COMMAND)?;
+    key.set_string("", &neovide_command)?;
+
+    Ok(())
 }
 
 pub fn register_right_click() {
-    if unregister_rightclick() {
+    if unregister_rightclick().is_ok() {
```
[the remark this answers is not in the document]
InfyniteHeap: I think this check can be removed because this check seems redundant. That is, registering keys don't need to call `unregister_rightclick()`. It can be called when removing keys from registry.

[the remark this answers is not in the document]
InfyniteHeap: > This looks a bit strange. I suggest that you still check for `is_err` here but add a check in `unregister_rightclick` so that it returns ok when there's no previous entry.

I think this check can be removed because this check seems redundant. That is, registering keys don't need to call `unregister_rightclick()`. It can be called when removing keys from registry.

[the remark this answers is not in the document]
InfyniteHeap: > This looks a bit strange. I suggest that you still check for `is_err` here but add a check in `unregister_rightclick` so that it returns ok when there's no previous entry.

I think this check can be removed because this check seems redundant. That is, registering keys don't need to call `unregister_rightclick()`. It can only be called when removing keys from registry.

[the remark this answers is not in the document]
InfyniteHeap: > This looks a bit strange. I suggest that you still check for `is_err` here but add a check in `unregister_rightclick` so that it returns ok when there's no previous entry.

I think maybe this check can be removed because it seems doesn't matter that users repeatedly register keys. In fact, in the original code, keys will always be written in registry again whether the keys are registered before, because the way that check whether the keys are registered **ONLY** tells user that some keys are registered before but doesn't avoid calling `register_rightclick_directory()` and `register_rightclick_file()`.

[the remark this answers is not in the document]
fredizzimo: Yes, that seems like a good way to do it.

fredizzimo (approved):

fredizzimo:
Thank you, this did indeed simplify the code a lot!

PR #2492, opened 2024-04-21
Use pre-commit for better linting

Holzhaus:
This PR introduces a configuration for the [pre-commit framework](https://pre-commit.com/) to establish a linter setup that is used both on CI and on the dev's machine when comitting changes. This also introduces some checks such as clippy, codespell and gitleaks.

This also cleans up the code so that it matches the linter configuration to prevent the situation that contributors have to fix unrelated code to makes the PR checks pass.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -8,28 +8,36 @@ assignees: ''
 
 <!--- NOTE: PLEASE FILL OUT TEMPLATE RATHER THAN DELETING --->
 
-**Describe the bug**
+## Describe the bug
```
Kethku, line 11: I'm not sure about this change. It converts the text to a header instead of bold which imho is better. Its a nit but the change seems unrelated unless im missing something.

I guess this was a lint? Maybe we can disable it?

Kethku:
I like most of these changes and agree that ensuring all of the lints are run is a good goal. However I'm nervous about the precommit hook. In the past I've worked on projects with them that can be more annoying than helpful especially when there are other places that can do the same check. Personally, I dislike changes that get in the way of people making contributions and playing around with the codebase. As is the CI can always run the checks and we can catch the issues in review.

But thats just my opinion. Would be good to hear from others like @fredizzimo

.github/ISSUE_TEMPLATE/bug_report.md
Holzhaus: Yes it was a lint: https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md036---emphasis-used-instead-of-a-header

Why do think bold is better? Do you prefer how it's displayed? Semantically I think the lint is right here. We could use a lower heading level though, so that it looks similar to bold text.

Holzhaus:
> However I'm nervous about the precommit hook. In the past I've worked on projects with them that can be more annoying than helpful especially when there are other places that can do the same check. Personally, I dislike changes that get in the way of people making contributions and playing around with the codebase. As is the CI can always run the checks and we can catch the issues in review.

Well, the pre-commit hook is opt-in (by running `pre-commit install` from the git repo root). A contributor doesn't *have* to use it, it's just a convenient method to ensure that all commits are clean before pushing to github and opening a PR. It's still entirely possible to just do all changes without it, open a PR and then do a clean-up commit when CI starts to complain ;-)

Just to avoid misunderstandings, pre-commit on CI works just like a regular linter. It will not require the contributors to retroactively fix old commits in their PR or something, it just checks that the latest PR state is clean.

Personally, I don't want to think about code formatting, style, etc.  This is why I find the pre-commit hooks super convenient, because they fix most of the stuff automatically for me, and warn me if manual intenvention is needed. Nothing is worse than submitting a large PR and then have to fix a bunch of code style nits by hand because CI complains :)
Since the pre-commit framework takes care of installing the project specific linters in the correct versions, there should be no difference between check results on CI and locally.

Holzhaus:
By the way, you might want to consider to use https://pre-commit.ci/ instead of the GitHub Actions workflow, then CI can commit code style fixes automatically so that users that don't use pre-commit locally don't have to clean up code smell themselves.

.github/ISSUE_TEMPLATE/bug_report.md
Holzhaus: Changed it to h4 heading. Please let me know what you think.

Holzhaus:
So you want me to rebase, or is this undesired? Then I'd close this.

falcucci:
@Holzhaus thank you, I understand the goal here but I agree with @Kethku's point of view.

I use [jj fix](https://docs.jj-vcs.dev/prerelease/config/#code-formatting-and-other-file-content-transformations) for example to handle all of lints and fmts locally for any project I am involved, in case the user forgets it the runners are going to point that out anyway, but the contributor doesn't need to worry about that barrier until they think the changes are ready to go, then jj fix it (or any git equivalent).

PR #2494, opened 2024-04-23
fix: clipping rect regression

fredizzimo:
This fixes the cacluation of the inner region clipping rectangle, which caused this regression.
 * https://github.com/neovide/neovide/pull/2467#issuecomment-2067342150

It also reduces the range of the scrollable lines iterator, which previously did not take border into account. That's not really a problem in practice, since we always need to iterate one more line than the actual grid size anyway, and the clipping rect will take care of clipping it correctly. But it could enable transparency modes when no transparent lines are visible for example.

The `floating_order` field was removed again, since it's unused and causing clippy warnings.
The `sort_order` field was also unused and removed

- No


9mm:
Amazing, this works for me !

PR #2496, opened 2024-04-23
fix: fractional border positions

fredizzimo:
https://github.com/neovide/neovide/pull/2485 did not calculate the position of the borders properly, which led to issues like this https://github.com/neovide/neovide/pull/2485#issuecomment-2073429344. This fixes that.

- No

### DIFF
--- a/src/renderer/rendered_window.rs
+++ b/src/renderer/rendered_window.rs
@@ -712,7 +712,7 @@ impl RenderedWindow {
             let mut matrix = Matrix::new_identity();
             matrix.set_translate((
                 pixel_region.min.x,
-                pixel_region.min.y + (i * grid_scale.height() as isize) as f32,
+                pixel_region.min.y + (i as f32 * grid_scale.height()),
             ));
             (matrix, line)
         })


9mm:
sweet that fixed I think

image

fredizzimo:
We reverted #2485 for now due to rendering issues. Therefore this PR is no longer needed, and this fix will be included in when those issues are fixed.

9mm:
i didnt look at what commit this originated from but i wanted to put a note as more of a reminder (right in middle of something atm)

I notice some recent commit made it where when you click inside neovim, the cursor moves to about 2 lines below where you click.

9mm:
and the same with dragging as well, as you drag the cursor drags 2 lines under the mouse

fredizzimo:
That 's probably the fractional font size one, which we now reverted.

I did test that the mouse worked properly on my system, but there are probably some special cases.

fredizzimo:
I can confirm that it's the padding that's causing the issue. The `pixel_region` function does not take that into account. If you have time, you can create a quick issue so that we don't forget it. I will fix it, maybe already tomorrow, but I think I will go to bed now.

PR #2497, opened 2024-04-23
Revert "feat!: Support fractional grid and font sizes"

Kethku:
Reverts neovide/neovide#2485 due to rendering issues


fredizzimo:
#2485, caused some regression in the text rendering quality like this:
image


So, applying this revert for now, until it's investigated further and checked if there's something we can do about it.

PR #2500, opened 2024-04-24
feat!: Support fractional fonts and grid sizes (try 2)

fredizzimo:
This is https://github.com/neovide/neovide/pull/2485 again, with some changes to improve the text clarity and performance. This bug was also fixed: https://github.com/neovide/neovide/pull/2485#issuecomment-2073429344 and the documentation updated.

This properly renders arbitrary font sizes, so the fudge factor is now removed. The hack for floating point font sizes is also 
removed, since it always caused the wrong spacing.

NOTE: This needs careful testing to make sure that we don't cause any regression in font quality. The optimal hint settings also need to be determined on different platforms, resolutions and screens. Ideally that would be read from font-config on Linux, but that's another feature support. In any case, the changed default to slight seems to match with the default of font-config, at least on my system.

- Yes, the font size will be different, since no fudge factor is applied. So, you might have to adjust your configuration accordingly.

## Notes
* This is best tested together with https://github.com/neovide/neovide/pull/2510, you can merge yourself or use my branch here https://github.com/fredizzimo/neovide/tree/fsundvik/fractional-fonts2-gamma
* The font sizes using the configuration file is wrong (they are in pixels instead of points there), see https://github.com/neovide/neovide/pull/2476


fredizzimo:
The underline thickness is inconsistent, so that's probably something that needs fixing.

fredizzimo:
On Windows, slight hinting is probably too much, it distorts the text a bit. None is better. 

And to restore the full quality, I think we need to implement gamma correct rendering. I have already been experimenting with that, and it seems to greatly increase the antialiasing quality.

fredizzimo:
This also seem to be causing bad performance regressions on Windows, probably related to the caching behaviour.

fredizzimo:
The performance and quality issues have been fixed by ensuring that the vertical grid size is always an integer size. If not, the grid size is rounded up, the way it works in other terminal emulators. It could also round to the nearest, but then it works differently than those.

PR #2505, opened 2024-04-25
Migrate dependency from `winapi` to `windows`.

InfyniteHeap:
Briefly complete migrating from `winapi` to `windows`. Known issue: Program will not show a window and closed immediately when building with `profiling` and `gpu_profiling` features enabled. On the other hand, you can still use the program when building without any features enabled. This issue will be fixed in the next commit(s).

- Refactor

- No

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ rmpv = "1.0.0"
 serde = { version = "1.0.136", features = ["derive"] }
 serde_json = "1.0.79"
 shlex = "1.1.0"
-simple_moving_average = "0.1.2"
+simple_moving_average = "1.0.2"
 spin_sleep = "1.1.1"
 strum = { version = "0.26.2", features = ["derive"] }
 swash = { version = "0.1.8", default-features = false }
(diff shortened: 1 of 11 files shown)


InfyniteHeap:
When this work completed indeed, the issue #2488 can be closed with "completed".

Cargo.toml
```diff
@@ -57,7 +57,7 @@ rmpv = "1.0.0"
 serde = { version = "1.0.136", features = ["derive"] }
 serde_json = "1.0.79"
 shlex = "1.1.0"
-simple_moving_average = "0.1.2"
+simple_moving_average = "1.0.2"
```
fredizzimo, line 60: I don't think this is in use, and can be removed

src/profiling/d3d.rs
```diff
@@ -362,10 +261,10 @@ pub fn create_d3d_gpu_context(name: &str, renderer: &D3DSkiaRenderer) -> Box<dyn
         flags: GpuContextFlags::GpuContextCalibration as u8,
         type_: GpuContextType::Direct3D12 as u8,
     };
-    let namestring = CString::new(name).unwrap();
+    // let namestring = CString::new(name).unwrap();
     let name_data = ___tracy_gpu_context_name_data {
         context: ctx_id,
-        name: namestring.as_ptr(),
+        name: name.as_ptr() as *const i8,
```
fredizzimo, line 267: This is probably your crash with the profiler, the name needs to be null terminated, that's why a CString was used.

src/renderer/d3d.rs
```diff
@@ -10,186 +7,122 @@ use skia_safe::{
...
-    };
-
-    if adapter.is_err() {
-        find_first_suitable(&|index: u32| -> Result<ComPtr<IDXGIAdapter1>, ()> {
-            call_com_fn(|adapter, _| unsafe { factory.EnumAdapters(index, adapter) })
-        })
-    } else {
-        adapter
-    }
+    unreachable!()
```
fredizzimo, line 67: I think you can put a comment here that the function returns when `EnumAdapters` has enumerated all the adapters and return a failure, so this code is never reached.

src/renderer/d3d.rs
```diff
@@ -10,186 +7,122 @@ use skia_safe::{
...
-            // actual device yet.
-            unsafe {
-                SUCCEEDED(D3D12CreateDevice(
-                    adapter as *mut IUnknown,
-                    D3D_FEATUREL_LEVEL,
-                    &ID3D12Device::uuidof(),
-                    null_mut(),
-                ))
-            }
+        if (DXGI_ADAPTER_FLAG(desc.Flags as i32) & DXGI_ADAPTER_FLAG_SOFTWARE)
```
fredizzimo, line 48: I think this can be written as `DXGI_ADAPTER_FLAG(desc.Flags as i32).contains(DXGI_ADAPTER_FLAG_SOFTWARE)`. And there are probably other similar usages that can be changed.

src/renderer/vsync/vsync_win_dwm.rs
```diff
@@ -94,7 +90,7 @@ impl VSyncWinDwm {
                     tracy_plot!("sleep_time", _sleep_time);
 
                     if redraw_requested.swap(false, Ordering::Relaxed) {
-                        let _ = proxy.send_event(UserEvent::RedrawRequested);
+                        proxy.send_event(UserEvent::RedrawRequested).unwrap();
```
fredizzimo, line 93: We don't want to crash here, if the proxy is gone, but it can be changed to 
`proxy.send_event(UserEvent::RedrawRequested).ok()`, That format seem to be more common in our code than `let _ =`

fredizzimo (changes requested):

src/profiling/d3d.rs
InfyniteHeap: Unfortunately, it still didn't work when I change back to `CString`. Even though, I'll reserve the back-change.

fredizzimo: Ok, let's deal with that later.

fredizzimo (approved):

fredizzimo:
Thank you very much for this!

Issue #2398, opened 2024-03-04
Font too thin in light background compared to chrome


xzbdmw:
**To Reproduce**
1. open a html file in chrome
[a.html.zip](https://github.com/neovide/neovide/files/14481618/a.html.zip)

2. set the same (jetbrains mono, font size 30) in neovide, and open the html file in neovide
3. see the font bold difference

**Expected behavior**
they should be the same

**Screenshots**
above is neovide, below is chrome, you can see neovide is thiner
<img width="765" alt="截屏2024-03-04 20 47 04" src="https://github.com/neovide/neovide/assets/97848247/971e4142-d955-476e-8741-93f3eff83df6">

I tested in intellij idea and vscode and found intellij render font silimlar to neovide(thiner) and vscode is similar to chrome(bolder), maybe not an issue just different rendering method
**Desktop (please complete the following information):**
 - OS: macos 14.2
 - Neovide Version 0.12.2
 - Neovim Version 0.10

xzbdmw:
I want to know is it possible make font looks like chrome:)

fredizzimo:
We usually get reports that the fonts are too bold, so in that sense this bug report is strange.

But I can repeat this on Linux as well, and the different alias options does not fix it. It's a bit puzzling, since we use Skia for font rendering, the same library as Chrome uses, so it should look the same. 

So, it's most likely a gamma issue, but I'm not sure what we need to do to fix, maybe switch to srgb rendering. This is most likely related:
https://skia.org/docs/dev/design/raster_tragedy/, and the document linked from it https://blog.johnnovak.net/2016/09/21/what-every-coder-should-know-about-gamma/

xzbdmw:
I also found neovide font weight 500(jb mono) is almost identical to kitty font weight 400 in light background, they mention gamma in text-composition document too. And kitty can specify a number to adjust that gamma.

xzbdmw:
What's the roles of skia_safe::Font and swash used as paird? Who decide the final looks?

xzbdmw:
some information https://github.com/kovidgoyal/kitty/pull/5423#issuecomment-1322240023

fredizzimo:
This should allow it to be configured the same:
* #2510

fredizzimo:
The gamma and contrast options in Neovide 0.13.0 allows for all kinds of adjustments. https://neovide.dev/configuration.html#text-gamma-and-contrast

Issue #1231, opened 2022-02-26
Font renders brighter and slightly thicker


gmr458:
**Describe the bug**
Neovide makes fonts look bold when they are actually regular or light.

**To Reproduce**
Use a regular or light font in a terminal emulator and Neovide and see the difference.

**Expected behavior**
That the font I use looks the same weight as it does in the terminal emulator.

**Screenshots**
Terminal emulator:
![alacritty](https://user-images.githubusercontent.com/53343401/155860117-306b57e7-ec28-4944-8bc4-2728b12c5b23.png)

Neovide:
![neovide](https://user-images.githubusercontent.com/53343401/155860125-b1ec65de-3960-4595-98ee-2f7a4969b600.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04.4 LTS
 - Neovide Version 0.8.0
 - Neovim Version 0.6.1

**Please run `neovide --log` and paste the contents of the `.log` file here:**
https://0bin.net/paste/9jSydr9u#-wgXMfCZAPBxXYI7XC6EZ7I1GFYbS2mXW4HFb+Em1RY

PyGamer0:
To me the font looks brighter in neovide (it doesn't look like the weight of the font has changed). This is probably due to the way neovide handles fonts.

kabouzeid:
The font is definitely more bold too. I'm also experiencing this issue.

Kethku:
Its possible this is a font selection issue on linux. Though, like @PyGamer0, I'm not convinced this isn't just a subtle change in the way the font is rendered. As such I'm going to change the name of this issue unless someone can show a side by side comparison which shows the bold variant of a font which matches the font rendered in neovide.

kabouzeid:
The font is only rendered more bold, it's not actually a different font. For example my icon font is rendered considerably more bold and it certainly doesn't have a bold font variant.

phdpe:
> The font is only rendered more bold, it's not actually a different font. For example my icon font is rendered considerably more bold and it certainly doesn't have a bold font variant.

Yes, I just installed Neovide on Windows and noticed the difference too. I tried to set `guifont` with `w` or `l` but neither works...

rmccord7:
I just installed neovide on windows as well and I see the same thing.

gmr458:
For some reason Neovide renders fonts in a way that makes them look thicker than they are, here's a comparison:
*open each image in different tabs to clearly see the difference

Alacritty:
[...]

Issue #1102, opened 2021-12-20
Colors more vivid in neovide than elsewhere


JeWe37:
**Describe the bug**
Colors with the same color scheme appear differently as in terminal problems running raw neovim.

**To Reproduce**

The problem is most visible with themes that are not too vibrant by default(but as far as I can tell it is still present everywhere). In the screenshots I am using the shusia option of the sonokai colorscheme.

**Expected behavior**
The colors should match those of terminal applications

**Screenshots**
![image](https://user-images.githubusercontent.com/26010054/146772953-687fa0cf-453f-4190-acc5-da413bf1b7da.png)

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Neovide Version 0.8.0
 - Neovim Version 0.6.0

**Additional context**
I previously used the older Vulkan version of neovide where this issue did not appear. As is, it makes it hard to properly match color schemes and at least in my eyes, the more vivid colors stand out far too much.

LoipesMas:
I'm not sure, but I think it might be caused by antialiasing.
![image](https://user-images.githubusercontent.com/46327403/146832454-6c3eadd3-32c7-4648-9465-85d044fcb33e.png)
Neovide on the left, ST on the right.
From what I've seen, the color in the middle of the character is the same, but edges look differently. In Neovide it looks like rgb split, while in ST (and Konsole too, I think) the edges are just less opaque (whidh is how it should be done).

LoipesMas:
Indeed, changing font edging from `Edging::SubpixelAntiAlias` to `Edging::AntiAlias` fixes that.
Not sure if neovide is doing something wrong, thus making subpixel antialiasing look bad, or if it's just bad.
![image](https://user-images.githubusercontent.com/46327403/146834988-a3a3db84-96b6-46ce-837d-fdf18a24a34e.png)
Subpixel on the right (current main), normal antialias on the left (after I changed it)

JeWe37:
Right, yes, I just tried it out here too by just trying a solid block where it is more clearly visible, and that got me the same result. The question is

LoipesMas:
I think some people reported that the fonts look "weird" and this might be the cause. I'm pretty sure non-subpixel is better (or at least more people would prefer it), but I might be wrong. If there isn't a big majority, maybe we could do an option to set which one to use.

JeWe37:
[...]

PR #2510, opened 2024-04-28
feat: Configurable contrast and gamma

fredizzimo:
A recent Skia update made it possible to configure the gamma and the contrast of the rendered text. So this makes that configurable from the Neovide side. The defaults have been chosen to give readable text on all backgrounds and an accurate color representation. But it can be changed to be similar to your favourite terminal for example.

Here are some screenshots for comparison (make sure to open the images in full size)
**Neovide main (with fractional font size support (https://github.com/neovide/neovide/pull/2500)**
Neovide master
**This PR**
Neovide new
**Alacritty**
Alacritty
**Wezterm**
wezterm
**Kitty**
kitty
**Foot**
Foot
**Neovide emulating alacritty (gamma 0.8, contrast 0.1)**
Neovide alacritty

As you can see different terminals render the text very differently. And IMO Neovide has always been among the better ones, especially when you take into account different colors and backgrounds, and that further improves it.

The default hinting has also been changed from full to slight. Full makes the text too bold, and sometimes distorts the text too much.

Closes these for good
* https://github.com/neovide/neovide/issues/1231
* https://github.com/neovide/neovide/issues/1102
* And let you adjust the gamma to be equal to Chromium https://github.com/neovide/neovide/issues/2398, I did not verify the exact settings.

- No

## Notes
* This is best tested together with https://github.com/neovide/neovide/pull/2500, you can merge yourself or use my branch here https://github.com/fredizzimo/neovide/tree/fsundvik/fractional-fonts2-gamma


fredizzimo:
For a technical overview of what this actually does, see:
[The Raster Tragedy in Skia](https://skia.org/docs/dev/design/raster_tragedy/)

fredizzimo:
I think the default hinting needs to be changed to `none`. `slight` still causes some of the fonts and sizes to look wrong. I'm not sure if it's visible with just this PR, but together with #2500 it does.

On Linux, Wayland at least on my monitor, I often prefer slight due to the slightly improved quality, but I think none is OK on that as well.

fredizzimo:
I changed the default hinting to `none`

Issue #2503, opened 2024-04-24
Cursor is offset from click


9mm:
> I can confirm that it's the padding that's causing the issue. The pixel_region function does not take that into account. If you have time, you can create a quick issue so that we don't forget it. I will fix it, maybe already tomorrow, but I think I will go to bed now.

From this commit: https://github.com/neovide/neovide/commit/776ad1db5251aacf19a68580e044e92071bfb37c

fredizzimo:
This should be fixed by
* #2511

PR #2511, opened 2024-04-28
fix: Cursor offset

fredizzimo:
* Fixes a regression caused by https://github.com/neovide/neovide/pull/2311. The absolute position of the mouse was returned instead of the relative one.
* A small refactor to the mouse manager, pass a struct instead of individual parameters

* Fixes https://github.com/neovide/neovide/issues/2503

- No

### DIFF
--- a/src/window/mouse_manager.rs
+++ b/src/window/mouse_manager.rs
@@ -48,6 +48,13 @@ struct DragDetails {
     button: MouseButton,
 }
 
+struct EditorState<'a> {
+    grid_scale: &'a GridScale,
+    window_regions: &'a Vec<WindowDrawDetails>,
+    window: &'a Window,
+    keyboard_manager: &'a KeyboardManager,
+}
+
 #[derive(Debug)]
 struct TouchTrace {
     start_time: Instant,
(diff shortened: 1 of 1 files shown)


9mm:
I checked on this branch and it still occurs, but now its only 1 line off instead of 2

fredizzimo:
Hm strange, I just tested all top paddings from 0-20, and all of them work. Maybe this is some macOS specific issue? Maybe affected by the `--frame` parameter?

9mm:
hmm.. i have spent tons of hours coding today so my mind is tired, but as a quick test i tried...

```
neovide git:(pr/2511) NEOVIDE_FRAME=full NEOVIDE_TITLE_HIDDEN=0 cargo run
```

which i think are the defaults and i got the same effect

these rae the current ones i use

```
export NEOVIDE_FRAME=buttonless
export NEOVIDE_TITLE_HIDDEN=1
export NEOVIDE_VSYNC=0
export NEOVIDE_FORK=1
```

9mm:
i will eat some food and poke around

9mm:
I tried a bunch of different options and couldnt seem to make it budge ;/ any ideas of what i should check?

9mm:
Ah, I have this:

if i comment these out, then it works right:

```
  -- padding
  -- vim.g.neovide_padding_top    = 20
  -- vim.g.neovide_padding_bottom = 20
  -- vim.g.neovide_padding_right  = 20
  -- vim.g.neovide_padding_left   = 20
```

fredizzimo:
Hm.. this PR is supposed to fix the case with the padding. It's broken on main for me and this fixes it.

But I just noticed another bug, the offset gets wrong with splits.

fredizzimo:
Ah, that gave me a clue. If I open a tabbar, then the offset gets wrong in the vertical direction.

fredizzimo:
I went ahead and merged this, since it's a pretty bad regression.

PR #2526, opened 2024-05-07
fix: respect floating shadow setting in layer rendering

sid-6581:
- Fix

- No

Floating layer rendering rendered the floating shadow even if disabled in settings. This PR checks the setting before rendering the shadow.


fredizzimo:
Thank you!

Issue #2502, opened 2024-04-24
More floating window rendering issues


sid-6581:
In `mini.files` the help window is transparent:

![image](https://github.com/neovide/neovide/assets/28602336/de5db985-ad8a-4a4e-9d5c-71007ddb37c0)

In `neo-tree` the preview window isn't transparent, but the contents are off by one, where the first line is rendered one line too high, and the bottom content line is blank:

![image](https://github.com/neovide/neovide/assets/28602336/4a4b41d2-331b-4e7c-a40f-458744f19efc)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Neovide Version: Latest commit
 - Neovim Version: Nightly

fredizzimo:
I'm commenting on the first issue first, and will investigate the second one next.
The windows have z-index 99 and 100 respectively, and with:
* https://github.com/neovide/neovide/pull/2467
It treats those as one layer, so they become transparent with respect to each other. This was something I was afraid would happen when I made this comment:
* https://github.com/neovide/neovide/pull/2442#discussion_r1545492883
And we had more discussion about it on Discord here https://discordapp.com/channels/896132176091955261/896263985614168125/1225611427495284776

There are basically two options, I think.
1. Standardize that consecutive z-indices are special around the whole nvim echo system. The issue can be fixed by using z-index 100 and 102 for example.
2. Properly experiment, if we actually need to combine different layers, or if it's enough to use a single layer while still maintaining compatibility with the terminal mode rendering. This probably also requires some adaptation on the plugin side, but if it works, I think it's a much nicer option than number 1.

@Kethku

fredizzimo:
For the second issue, I'm not 100% sure yet, but almost. The preview window consists of two subwindows, the first one has this anchor
> TRACE [neovide::channel_utils] neovim_handler WindowFloatPosition { grid: 10, anchor: NorthWest, anchor_grid: 1, anchor_row: -1.0, anchor_column: 40.0, focusable: false, sort_order: Some(39) }

While the second one has this:
> TRACE [neovide::channel_utils] neovim_handler WindowFloatPosition { grid: 11, anchor: NorthWest, anchor_grid: 10, anchor_row: 1.0, anchor_column: 1.0, focusable: true, sort_order: Some(40) }

So, the second one is attached to the first one, but the first one is actually anchored outside the screen. Neovim does not reposition floats that are outside the grid, so we are forced to do it ourselves:
* https://github.com/neovim/neovim/issues/24801

[...]

PR #2527, opened 2024-05-07
fix: don't group floating windows unless one is fully contained in the other

sid-6581:
- Fix

- No

This is an attempt to make it less likely that floating windows with consecutive z-indexes are rendered together if they shouldn't be rendered together, like the first screenshot in #2502.

Instead of checking if the window rects intersect, I check if one is contained in the other. This isn't foolproof, because we could still have windows with consecutive indexes that should not be rendered together (like the first screenshot in #2502, but in a bigger directory), so the whole window grouping thing needs a lot more thought, and should probably be able to be disabled entirely by an option.

Still, I can't think of any time it would make sense to render two windows with consecutive z-indexes together if one isn't fully contained in the other.

### DIFF
--- a/src/renderer/rendered_layer.rs
+++ b/src/renderer/rendered_layer.rs
@@ -3,7 +3,7 @@ use skia_safe::{
     canvas::SaveLayerRec,
     image_filters::blur,
     utils::shadow_utils::{draw_shadow, ShadowFlags},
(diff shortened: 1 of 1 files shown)


fredizzimo (approved):
I think this looks good, but it's better that @Kethku double checks, since she knows the details better.

fredizzimo:
Thanks, I decided to merge this, since it's strictly better than the previous behaviour, although not perfect.

PR #2528, opened 2024-05-07
fix: save grid size for maximized windows

sid-6581:
- Fix

- No

We're currently saving the grid size for non-maximized windows only. This PR also saves it for maximized windows, because it has the same benefit as for non-maximized windows.


fredizzimo:
Thank you!

Issue #2522, opened 2024-05-05
[Bug] `Alt+Enter` mapping behaves incorrectly in the Neovide binaries built from latest master


pidgeon777:
**Describe the bug**
In my LazyVim, setup, the `Alt + Enter` combination is assigned to a specific action.

In the latest stable version of Neovide, pressing `Alt + Enter` will trigger that action, without any side effect.

With the release binary built from the latest master version of Neovide, `Alt + Enter` leads Neovide to perform the following unwanted actions very quickly:

1. Turn on fullscreen mode.
2. Turn off fullscreen mode.
3. Lose focus on the window.

**To Reproduce**
1. Start Neovide with the following arguments:
- `--frame none`
- `--log`
2. Press `Alt + Enter` in Neovide.
3. Observe the behavior.
4. Regain focus on Neovide (if necessary).
5. Press `q` to quit Neovide.

**Expected behavior**
I would expect Neovide not to flash and/or lose focus when pressing `Alt + Enter`, and simply execute the associated Neovim action, exactly like the stable version.

**Desktop (please complete the following information):**
 - OS: Windows 11 Pro
 - Neovide Version: 0.12.2 (built from the latest master version)
 - Neovim Version: NVIM v0.10.0-dev-3057+g9e2f378b6

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
This is the log of the stable Neovide release (no issue), starting from the moment `Alt + Enter` is pressed:

```log
TRACE [neovide::window::keyboard_manager] KeyEvent {
    physical_key: Code(
        Enter,
    ),
    logical_key: Named(
        Enter,
    ),
    text: None,
    location: Standard,
    state: Released,
    repeat: false,
    platform_specific: KeyEventExtra {
        text_with_all_modifers: None,
        key_without_modifiers: Named(
            Enter,
        ),
    },
}
TRACE [neovide::window::keyboard_manager] Modifiers { state: ModifiersState(0x0), pressed_mods: ModifiersKeys(0x0) }
TRACE [neovide::window::keyboard_manager] KeyEvent {
    physical_key: Code(
        AltLeft,
    ),
    logical_key: Named(
        Alt,
    ),
    text: None,
    location: Left,
    state: Released,
    repeat: false,
    platform_specific: KeyEventExtra {
        text_with_all_modifers: None,
        key_without_modifiers: Named(
            Alt,
        ),
    },
}
TRACE [neovide::window::keyboard_manager] KeyEvent {
[...]

Issue #2517, opened 2024-05-03
[Bug] `Using the D3D12 SDKLayers dll requires that the latest SDKLayers for Windows 10 is installed` error when starting `neovide`


pidgeon777:
**Describe the bug**

When building and running the latest `neovide` sources with `cargo build --release` (`master` branch), the resulting `neovide.exe` binary crashes.

Here is the log:

```log
C:\Work\Repositories\GitHub\neovide\neovide\target\release\deps>gdb neovide.exe
GNU gdb (GDB) 14.1
Copyright (C) 2023 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-w64-mingw32".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from neovide.exe...
(No debugging symbols found in neovide.exe)
(gdb) run
Starting program: C:\Work\Repositories\GitHub\neovide\neovide\target\release\deps\neovide.exe 
[New Thread 226364.0x37404]
[New Thread 226364.0x37438]
[New Thread 226364.0x37a88]
warning: Can not parse XML library list; XML support was disabled at compile time
[New Thread 226364.0x37a70]
[New Thread 226364.0x37a74]
[Thread 226364.0x37a70 exited with code 0]
[Thread 226364.0x37a74 exited with code 0]
[New Thread 226364.0x37b20]
[New Thread 226364.0x37b14]
[New Thread 226364.0x37b3c]
[New Thread 226364.0x37b40]
[New Thread 226364.0x37b30]
[New Thread 226364.0x37b34]
[New Thread 226364.0x37b54]
[New Thread 226364.0x37ad4]
[New Thread 226364.0x37af0]
[New Thread 226364.0x37b60]
[New Thread 226364.0x37b18]
[New Thread 226364.0x37b1c]
[New Thread 226364.0x37b04]
[New Thread 226364.0x37b28]
[New Thread 226364.0x37b48]
[New Thread 226364.0x37b0c]
[...]

Issue #2465, opened 2024-04-04
neovide_remember_window_size and neovide_transparency options stopped working after updating to the latest commit


OrangeLightning219:
**Describe the bug**
After not updating for a long time I decided to update neovide to the latest version. After the update the `neovide_remember_window_size` and `neovide_transparency` stopped working. When launching neovide the window opens very small and messes up all my splits and the background is fully opaque.

**To Reproduce**
1. Open neovide
2. Close the window
3. Open neovide again

**Expected behavior**
Neovide remebers previous window size and the background is transparent.

**Screenshots**
Here's a short video showing the problem:
https://github.com/neovide/neovide/assets/22556400/7961b616-7018-4089-9af3-b9995681c8c6

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Neovide Version built from source from [this commit](https://github.com/neovide/neovide/commit/bcd913496924c0cfd5a99216dfb2abef2d25dcc3)
 - Neovim Version v0.10.0-dev-2210+g38bb0e1da

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
I found these two lines in the log:
`TRACE [neovide::settings] Initial value load failed for transparency: Error processing request: 1 - 'Key not found: neovide_transparency')`
`TRACE [neovide::settings] Initial value load failed for remember_window_size: Error processing request: 1 - 'Key not found: neovide_remember_window_size')`
Here's the full file: [neovide_rCURRENT.log](https://github.com/neovide/neovide/files/14863791/neovide_rCURRENT.log)

**Additional context**
My neovim config: https://github.com/OrangeLightning219/dotfiles/tree/master/nvim

fredizzimo:
The log file indeed shows that the settings are not updated. The initial value load one is normal, but it should be updated after that.

Does it work if you set it from inside Neovim with `:let g:neovide_transparency=0.5`?

OrangeLightning219:
If I set the transparency from inside neovim the background just gets darker. With transparency set to 0 it's pure black.

fredizzimo:
That's probably a different bug actually, with the new D3D rendering backend.

But, since changing the values directly has an effect, it seems to indicate that there's something wrong with our settings reading during the startup.

raggi:
It doesn't appear to be working with `--opengl` either

OrangeLightning219:
When running with `--opengl` the transparency works for me, but the window size is still broken.

Kethku:
[...]

Issue #2471, opened 2024-04-06
[RFC] Proposal: handling of global transparency


fredizzimo:
We currently have several transparency issues, here are a few:
* https://github.com/neovide/neovide/issues/2375
* https://github.com/neovide/neovide/issues/2318
* https://github.com/neovide/neovide/issues/2275
* https://github.com/neovide/neovide/issues/1508

This is unrelated to this proposal, but should be fixed nevertheless.
* https://github.com/neovide/neovide/issues/2327

The main issue is that neovide transparency applies to all background colors as they are defined in Neovide, so some elements that should not be transparent, like the visual mode selections becomes transparent. The other equally important issue is that some elements should be transparent even if they are rendered as foreground, like the statusline elements.

The third issue is that the default blend mode for floating windows is not always the desirable one.

In order to fix all these issues, we need more customization than what's currently possible. We have one open proposal on the Neovim side.
* https://github.com/neovim/neovim/issues/24472

But after some thinking, I think it's perhaps too complex, and I don't think we can get it into the upcoming 0.10 release either. There are also numerous design decisions that need to be correctly taken and can't easily be fixed if it does not work in practice. So, after some thinking I came up with an alternative solution.

First, let's fix the obvious naming problem, `neovide_transparency`, should really be `neovide_opacity`, since `1` means fully opaque, while `0` means fully transparent. So, let's alias these parameters to each other, and deprecate the old one, so that it still works, but all documentation only refers to `neovide_opacity`.

Neovim always sends both the GUI and terminal colors, so we can (ab)use the terminal color index to control the transparency. Which is done with a new function.

`neovide_set_transparent_color(color_index, opts)`

The `color_index` is the terminal ansi color number, and opts is table with all optional fields as described below

The following fields describes how the opacity is calculated for all non-floating window colors:
* `base_opacity`  - The base opacity, defaults to `0`.
* `opacity_multiplier` - The multipler specified here is multiplied by the global `neovide_opacity`, defaults to `1.0`
* `global_disable` - When set to `true`, the color is always fully opaque when `neovide_opacity` is set to `1.0`. Defaults to `false`
* `applies_to_foreground` - Applies the transparency also to Neovim foreground colors, defaults to `true`

[...]

PR #2529, opened 2024-05-07
fix: Use composition to fix window transparency when using D3D

fredizzimo:
In order to render transparent window backgrounds on Windows with D3D, the drawing is now done using Window's composition engine, see https://learn.microsoft.com/en-us/archive/msdn-magazine/2014/june/windows-with-c-high-performance-window-layering-using-the-windows-composition-engine. This also fixes, the alt-enter hijacking, since the swapchain is no longer bound to the Window. 

Additionally, this PR disables the D3D debug layer by default, since it's not supported on some systems without additional software.

* Fixes https://github.com/neovide/neovide/issues/2517
* Fixes partially https://github.com/neovide/neovide/issues/2465
* Fixes https://github.com/neovide/neovide/issues/2522

- No

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ tracy-context-switch-tracing = ["tracy-client-sys?/context-switch-tracing"]
 tracy-sampling = ["tracy-client-sys?/sampling"]
 tracy-code-transfer = ["tracy-client-sys?/code-transfer"]
 tracy-callstack-inlines = ["tracy-client-sys?/callstack-inlines"]
+d3d_debug = [] # Enable the D3D debug layer
 
 [dependencies]
 anyhow = { version = "1.0.75", features = ["backtrace"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,6 +89,7 @@ skia-safe = { version = "0.73.0", features = ["gl", "d3d", "textlayout"] }
 windows = { version = "0.56.0", features = [
     "Win32_Graphics_Direct3D",
     "Win32_Graphics_Direct3D12",
+    "Win32_Graphics_DirectComposition",
     "Win32_Graphics_Dwm",
     "Win32_Graphics_Dxgi",
     "Win32_Graphics_Dxgi_Common",
--- a/src/renderer/d3d.rs
+++ b/src/renderer/d3d.rs
@@ -11,10 +11,12 @@ use windows::core::{Interface, Result, PCWSTR};
 use windows::Win32::Foundation::{CloseHandle, HANDLE, HWND};
 use windows::Win32::Graphics::Direct3D::D3D_FEATURE_LEVEL_11_0;
 use windows::Win32::Graphics::Direct3D12::{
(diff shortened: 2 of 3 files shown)


pidgeon777:
Great! 👏

If I understand correctly, in the latest version it is possible to run Neovide also with `--opengl`.

I would like to ask which would be the differences on Windows 11 running Neovide either with `--opengl` or in D3D12 mode (performances, features etc.).

fredizzimo:
The difference is mainly that D3D is smoother on most systems, it should be on all, but there might still be some bugs in the implementation, so `--opengl` is still supported. It's also more compatible, especially with older hardware.

pidgeon777:
If I understand correctly it is also now possible with D3D12 to set Neovide transparency on Windows? If yes, how could this be achieved? Just for a quick test.

fredizzimo:
You can either download the artifacts from the CI https://github.com/neovide/neovide/actions/runs/8991931441, or build this branch manually 
image

pidgeon777:
Thanks for the quick reply, I downloaded the artifact from your link here:

https://github.com/neovide/neovide/actions/runs/8991931441

Strangely the `Alt + Enter` issue still persists, also setting `vim.g.neovide_transparency = 0.0` renders makes the background darker:

image

This is when it is set to `1.0`:

image

I'm running:

`neovide.exe --frame none --log`

fredizzimo:
That's very strange, it definitely fixed both issues for me, maybe it's somehow running the wrong exe? I think powershell by default don't run exes in the current directory, unless you give the path like `.\neovide`

pidgeon777:
I compiled from sources Neovide and it created 4 binaries (debug, release, deps etc.).

As you suggested, it was running the previous version.

image

Even transparency now works fine. My congratulations for the great job! 👍

pidgeon777:
I found a little bug in my basic LazyVim configuration (noice enabled by default), this is the Lazy panel:

image

When I try to enter the diff subwindow for an update (press `d`), I get this:

image

The diff buffer is focused and I can move the cursor in it correctly, but the previous buffer is still displayed in the background.

When I press `q` to exit the diff window, the lazy buffer is correctly focused.

fredizzimo:
@pidgeon777, are you using `g:neovide_transparency=0`, that's not really well supported, since it applies the transparency to all background colors, See the discussions here for more information:
* https://github.com/neovide/neovide/issues/2471

Another possible cause could be that Neovim renders several non-floating windows on top of each other and does not clear the underlying window automatically. Essentially the same cause that causing this:
* https://github.com/neovide/neovide/issues/2333

pidgeon777:
I am currently experiencing some confusion regarding the transparency settings in Neovide.

Based on my observations, it appears that in the previous commit, setting the `g:neovide_transparency` parameter to `0` resulted in only the Neovide background becoming transparent, while the code remained opaque. However, in the current version, it seems that all graphical elements of Neovide, including the code, are affected by this setting.

Consequently, I am seeking guidance on how to adjust the settings such that only the Neovide background is transparent, while the rest of the Neovide graphical elements remain opaque.

PR #2545, opened 2024-05-14
docs: Fix reversed autocmds for setting input_ime

cpmsmith:
The sample Vimscript autocmds for setting `neovide_input_ime` are reversed on the `Cmdline` events, compared to the `Insert` events, and the Lua sample.

- Documentation (fix)

- No


fredizzimo:
Thank you. It was indeed reversed. I think I borrowed the original lua snippet (which is probably needlessly complex) from somewhere, and then did a quick translation to vimscript and did a mistake on the way.

Issue #2549, opened 2024-05-16
Crash with neovim 0.10: Cannot parse api info: `api_prerelase` is Nil (not Boolean)


allvpv:
**Describe the bug**
Neovide encounters an error when attempting to be run with Neovim 0.10, as it fails to parse API information due to the api_prerelase key being Nil. The Neovide assumes it to be Boolean.

**To Reproduce**
1. Download or compile the latest Neovim 0.10 version.
2. Attempt to run neovide with `--neovim-bin` pointing to the aforementioned Neovim release.
3. Observe crash.

**Expected behavior**
  A check for Nil in `api_info.rs:204`.  After recompilation, Neovide works as expected. (Or maybe that's a Neovim bug).

**Desktop (please complete the following information):**
 - Fedora 40, Linux 6.8.9, aarch64

```
TRACE [neovide] Neovide version: 0.12.2
DEBUG [neovide::bridge::command] Starting neovim with: Command { std: "/home/allvpv/.local/bin/nvim" "--embed" "-p", 
kill_on_drop: false }
ERROR [neovide::error_handling] Neovide just crashed :(
This is the error that caused the crash. In case you don't know what to do with this, please feel free to report this
 on https://github.com/neovide/neovide/issues!

Failed to parse Neovim api information

Caused by:
    nil
```

fredizzimo:
I'm fixing this right now, and also preparing a release of Neovide. 

I'm not sure why this would be Nil, the documentation does not exactly say so, it but they check it like a boolean
https://github.com/neovim/neovim/blob/a664246171569209698c0b17b1d7af831f6603d2/runtime/doc/api.txt#L180-L181,

And here's it's also set to false:
https://github.com/neovim/neovim/blob/a664246171569209698c0b17b1d7af831f6603d2/CMakeLists.txt#L151

And their lua generator sets it to that here https://github.com/neovim/neovim/blob/a664246171569209698c0b17b1d7af831f6603d2/src/nvim/generators/nvim_version.lua.in#L8

allvpv:
Yeah, in the previous stable version it was just `false`, so this is probably a Neovim issue somewhere in the build process.

allvpv:
OK, maybe you've saw this already, but here are the lines to blame.

```
for _, item in ipairs(version) do
  -- NB: all items are mandatory. But any error will be less confusing
  -- with placeholder vim.NIL (than invalid mpack data)
  put(item[1], item[2] or vim.NIL)
end


[...]

PR #2550, opened 2024-05-16
fix!: require Neovim 0.10.0

fredizzimo:
Neovim 0.10 is now required for stability and code maintenance reasons. Our workaround for older versions of Neovim has also been removed.

Also
* Fixes https://github.com/neovide/neovide/issues/2549

- Yes, Neovim 0.9.4 was enough before.


9mm:
deleting code is fun 🏎️

Issue #2556, opened 2024-05-17
Neovide 0.13.0 panics on Windows 11


myth0genesis:
**Describe the bug**
After installing 0.13.0 with both the .msi installer and building from source, Neovide panics as soon as I attempt to run the executable

**To Reproduce**
Via installer:
1. Download Neovide 0.13.0 installer for Windows from neovide.dev
2. Run Neovide installer
3. Run Neovide executable

Via building from source:
1. Updated Rust via `rustup update`
  (CMake version: 3.29.1 already installed)
  (LLVM version: 18.1.4 already installed)
2. Ran command
  ```cargo install --git https://github.com/neovide/neovide.git```
3. Run Neovide executable

**Expected behavior**
Neovide executable runs.

**Desktop:**
 - OS: Windows 11 Home Version 10.0.22631 Build 22631
 - Neovide Version 0.13.0
 - Neovim Version 0.10.0
 - NVidia RTX 3070ti GPU
 - NVidia Game Ready drivers are latest version (552.44)

Contents of log:
2024-05-17 18:05:11 - Neovide panicked with the message 'Failed to create DXGI factory: Error { code: HRESULT(0x887A0001), message: "The application made a call that is invalid. Either the parameters of the call or the state of some object was incorrect.\r\nEnable the D3D debug layer in order to see details via debug messages." }'. (File: src\renderer\d3d.rs; Line: 112, Column: 59)
   0: \<unknown\>
   1: \<unknown\>
   2: \<unknown\>
   3: \<unknown\>
   4: \<unknown\>
   5: \<unknown\>
   6: \<unknown\>
   7: \<unknown\>
   8: \<unknown\>
   9: \<unknown\>
  10: \<unknown\>
  11: \<unknown\>
  12: \<unknown\>
  13: BaseThreadInitThunk
  14: RtlUserThreadStart

**Additional context**
I'm wary of updating CMake and LLVM frequently because the build ecosystem for my job breaks pretty much every time I do it, so that's why CMake and LLVM on my system aren't the absolute newest versions available (probably a skill issue on my part, but I just do what works for me).
Neovim 0.10.0 currently runs as expected on my system. I'm also using LazyVim, so the expected behavior in my specific case is that the dashboard comes up when I run Neovide.
It should also be noted that not a single error came up when building with Rust.

fredizzimo:
If you are already building from source, can you run
`cargo run --features d3d_debug` if that does not work,

Then it looks like this line is wrong
[...]

PR #2559, opened 2024-05-18
fix: Require only DXGI 1.2 when not debugging

fredizzimo:
For some reason it appears like dxgi 1.4 is unavailable on some Windows 10 and Windows 11 machines. I have no idea why, since it should be bundled with Windows since the very first 10 release.

This relaxes those requirements, and also don't pass `DXGI_CREATE_FACTORY_DEBUG`, since that's another failure point.

* fixes https://github.com/neovide/neovide/issues/2556

- No


fredizzimo:
I' merged this one, as it was confirmed to fix the problem

Issue #2554, opened 2024-05-17
Neovide 0.13.0 on macOS launches behind other windows.


humblehacker:
**Describe the bug**
Neovide 0.13.0 on macOS launches behind other windows.

**To Reproduce**
1. Start your terminal (I'm using iTerm)
2. Install Neovide 0.13.0 from homebrew
    `brew install neovide`
3. Start Neovide
    `neovide`

**Expected behavior**
Neovide launches in the foreground.

**Desktop (please complete the following information):**
 - OS: macOS 14.4.1 (23E224)
 - Neovide Version: 0.13.0
 - Neovim Version: 0.10.0

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/15351407/neovide_rCURRENT.log)

**Additional context**
I tried launching both with `--fork` and `--no-fork`. Both options launched with the window in the background.

I also reinstalled Neovide 0.12.2 from homebrew `brew install --cask neovide`, and Neovide correctly launches in a foreground window.

humblehacker:
Seems this might be an issue with the homebrew formula. I uninstalled (`brew uninstall neovide`), downloaded and installed from the dmg from the GitHub release page, and launched from the terminal using a symlink to the binary in the app bundle. It correctly launches in a foreground window (albeit with a strange opening animation that shows only the titlebar expanding to full screen, with the rest of the window appearing only when the animation is complete).

humblehacker:
The homebrew issue may be a red herring. The problem seems intermittent. I've been able to reproduce the problem with both the homebrew binary and the one found here.

fredizzimo:
This is opposite to this bug, which I think we also has seen in Neovide
* https://github.com/rust-windowing/winit/issues/2051

I expect both be fixed when we refactor our window creation to happen at a different place and move to winit 0.30, which should happen pretty soon.

humblehacker:
Thanks for the info! I'll stop investigating and revert to 0.12.2 until that's ready.

fredizzimo:
I have a potential fix for this here:
* https://github.com/neovide/neovide/issues/915

It would be good if someone could test it.

bboles:
I am still experiencing this issue with neovide version 0.13.1.  I'm on MacOS 14.5 and neovim version 0.10.0.

Issue #2330, opened 2024-01-31
neovide does not focus on launch from terminal (macOS, iTerm2)


halostatue:
**Describe the bug**

I launched `neovide` from iTerm2 and it launched *behind* the iTerm2 window and all the way at the end of the app switcher. (I had been doing something where I had set neovide to be on the right side of the screen the same as my typical iTerm2 window, and I kept it here because it shows the behaviour fairly well.)

Even adding `vim.cmd("NeovideFocus")` to the end of `init.lua` is not fixing the issue.

**To Reproduce**

1. Type `neovide` in iTerm2 (this probably happens with Terminal.app as well, but I currently use iTerm2).
2. Observe that the window opens *behind* the iTerm2 window.
3. Observe on ⌘-Tab that Neovide's icon is at the *end* of the app switcher.

**Expected behavior**

Neovide should take focus (especially if `vim.cmd("NeovideFocus")` is in `init.lua` or `NeovideFocus` is in `init.vim`) from iTerm2 and be positioned in front of iTerm2.

**Screenshots**

![CleanShot 2024-01-30 at 23 41 08@2x](https://github.com/neovide/neovide/assets/11361/4bce517e-44cf-433a-aae9-5ba86c6b63d7)

**Desktop (please complete the following information):**
 - OS: macOS 14.3 23D56 arm64
 - Neovide 0.12.2
 - Neovim 0.9.5

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/14107279/neovide_rCURRENT.log)

**Additional context**

n/a

D00mch:
related https://github.com/neovide/neovide/issues/2324

wSedlacek:
I noticed this issue start to occur after updating when using Raycast.
https://www.raycast.com/

After updating Neovide when Neovide is launch from Raycast it starts in the background.
I did not notice this behavior on prior versions.

thraizz:
For me, this is also the case. Launching from Finder results in the same behavior, even with all plugins disabled. Scheduling `NeovideFocus` event does not help for me sadly.

brnt:
Seeing same behavior from iTerm2 as described above.

Also note that when launching from Spotlight, the neovide window is focused, but is appended to the _end_ of the cmd-tab switcher (same as described above when launching from a terminal), which is the opposite of most macOS applications. The focus+stacking issue is likely separate from the cmd-tab issue.

[...]

Issue #2142, opened 2023-11-20
White flash during Neovide startup on Microsoft Windows


harg:
**Describe the bug**
On Neovide Launch, an empty window with white background is displayed just before rendering the correct colorscheme background, causing an annoying white flash.

**To Reproduce**
1. Launch Neovide on Microsoft  Windows

**Expected behavior**
On Neovide Launch, display the neovide window with the colorscheme background (or black background)

**Desktop :**
 - OS: Microsoft Windows 10
 - Neovide latest version
 - Neovim latest version

sid-6581:
I see that as well, I don't think the surface is being filled with anything before the window is drawn the first time. If nobody else fixes it, I have it on my list to look at when I get some free time.

fredizzimo:
I forgot to comment about this issue, but I think it's mostly because this wint function does not use a single WinApi call to set the window state 

https://github.com/rust-windowing/winit/blob/17296e98783a19c2f0a552340682e3de2b863de8/src/platform_impl/windows/window_state.rs#L305-L434

So it's unable to for example resize the window and show it at the same time. If you just set the size, it's temporary shown then resized and hidden again, and if you show it, then it's temporarily shown with the wrong size. And then it's not a all sure that we happen to render anything between those actions. Especially when starting a maximised window. 

The winapi function `SetWindowPos` can do everything at once, so there's no reason for the function to be written that way.

fredizzimo:
I forgot to add:
I don't remember what the issue was that prevented me from doing this, but I think it's resolved now with the way we calculate the window size before creating the window. I think it would be better to create the window at the time it's shown, rather than creating it as hidden, since I think that codepath is atomic. 

Winit also has some plans for only allowing creation of windows inside the event loop, so that would support that flow a lot better. And there was also some issue, that I don't remember right now, that would be solved, or solvable by moving the creation inside the loop. So those would be other good reasons for moving the creation.

fredizzimo:
The issue was fixed in some of the 0.12.x versions of Neovide.

sid-6581:
@fredizzimo I don't think this should be closed yet, I'm still seeing the same issue with both Direct3D and OpenGL rendering on Windows 11.

fredizzimo:
Ok, I will re-open. I see some minor flashing when launching in full screen mode, but not nearly as bad as it was when the issue was reported.

[...]

PR #2562, opened 2024-05-18
fix: Delay the window creation until it's time to draw

fredizzimo:
Instead of creating a hidden window and then potentially resize and show it, we now create the windows inside the event loop with the correct size on demand. 

This fixes a few macOS issues, and also some flickering and bad animation when the window is created. It also probably fixes some issues with tiling window managers

* Fixes https://github.com/neovide/neovide/issues/2142
* Fixes https://github.com/neovide/neovide/issues/2554
* Fixes https://github.com/neovide/neovide/issues/2330
* Fixes https://github.com/neovide/neovide/issues/2087
* Fixes https://github.com/neovide/neovide/issues/1888
* Fixes https://github.com/neovide/neovide/issues/1806
* Fixes https://github.com/neovide/neovide/issues/1783
* Fixes https://github.com/neovide/neovide/issues/915

- No

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -60,8 +60,7 @@ use renderer::{cursor_renderer::CursorSettings, RendererSettings};
 #[cfg_attr(target_os = "windows", allow(unused_imports))]
 use settings::SETTINGS;
 use window::{
-    create_event_loop, create_window, determine_window_size, main_loop, UserEvent, WindowSettings,
-    WindowSize,
+    create_event_loop, determine_window_size, main_loop, UserEvent, WindowSettings, WindowSize,
 };
 
 pub use channel_utils::*;
@@ -97,8 +96,7 @@ fn main() -> NeovideExitCode {
         Err(err) => handle_startup_errors(err, event_loop).into(),
         Ok((window_size, font_settings, _runtime)) => {
             clipboard::init(&event_loop);
-            let window = create_window(&event_loop, &window_size);
-            main_loop(window, window_size, font_settings, event_loop).into()
+            main_loop(window_size, font_settings, event_loop).into()
         }
     }
 }
--- a/src/renderer/d3d.rs
+++ b/src/renderer/d3d.rs
@@ -40,9 +40,14 @@ use winit::{
 use super::{vsync::VSyncWinSwapChain, RendererSettings, SkiaRenderer, VSync};
 #[cfg(feature = "gpu_profiling")]
 use crate::profiling::{d3d::create_d3d_gpu_context, GpuCtx};
-use crate::{profiling::tracy_gpu_zone, settings::SETTINGS, window::UserEvent};
+use crate::{
+    profiling::{tracy_gpu_zone, tracy_zone},
+    settings::SETTINGS,
+    window::UserEvent,
+};
 
 fn get_hardware_adapter(factory: &IDXGIFactory4) -> Result<IDXGIAdapter1> {
+    tracy_zone!("get_hardware_adapter");
     for i in 0.. {
         let adapter = unsafe { factory.EnumAdapters1(i)? };
         let mut desc = Default::default();
@@ -97,6 +102,7 @@ pub struct D3DSkiaRenderer {
 
 impl D3DSkiaRenderer {
     pub fn new(window: Window) -> Self {
+        tracy_zone!("D3DSkiaRenderer::new");
         #[cfg(feature = "d3d_debug")]
         unsafe {
             let mut debug_controller: Option<ID3D12Debug> = None;
@@ -117,6 +123,7 @@ impl D3DSkiaRenderer {
 
         let mut device: Option<ID3D12Device> = None;
         unsafe {
+            tracy_zone!("create_device");
             D3D12CreateDevice(&adapter, D3D_FEATURE_LEVEL_11_0, &mut device)
                 .expect("Failed to create a Direct3D 12 device");
         }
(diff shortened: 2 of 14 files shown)


fredizzimo:
Sorry, this still need some fixing on macOS

sid-6581:
With regards to the white flickering, oddly enough I'm not experiencing this on my home system with a 4090, but I am experiencing it constantly on my work system with a 3070Ti. I'll check it on my work system on Monday.

What I am seeing otherwise on this PR on my home system with a 4090, is that a window frame is created with the wrong size initially, then it flickers away and is resized to the correct size. The wrong size window frame is empty (meaning nothing is drawn inside the frame at all, it still just shows the apps underneath it), and after resizing, the frame is still hollow for a split second before the contents are drawn.

I'll try to debug it and see if it's an easy fix on Windows.

fredizzimo:
@sid-6581, I think I might know the reason. It's not taking the scale factor into account before the window is created. I'm actually fixing that now, since I noticed during the macOS fixing.

fredizzimo:
The monitor handle has a scale factor, https://rust-windowing.github.io/winit/winit/monitor/struct.MonitorHandle.html#method.scale_factor, but I don't know how to get the monitor where the Window is going to be created.

fredizzimo:
@sid-6581 Alacritty seem to resize the window immediately after it's created, so maybe we can get away with that?

https://github.com/alacritty/alacritty/blob/38fed9a7c233e11e5f62433298235281fc3de885/alacritty/src/display/mod.rs#L415-L418

Could you test by, calling`self.renderer.grid_renderer.handle_scale_factor` with the correct scale factor immediately after the window is crated.

Then do the size calculations, as before, and finally call `window.set_inner_size()`

fredizzimo:
@sid-6581, I made those changes, it does not seem to have any bad effects on Wayland at least

sid-6581:
@fredizzimo now it is actually worse for some reason. It still creates the big empty frame, then it resizes it down to a frame that is too small (but in the correct position), before it finally resizes it to the correct size. So one extra wrong sized frame in the middle compared to previously.

fredizzimo:
Strange, I guess the first thing to check is that neither of these are resizes do anything after the initial creation

https://github.com/neovide/neovide/blob/95fc3a1217a4930909d403d0df478e8ba2754e69/src/window/window_wrapper.rs#L614-L629

fredizzimo:
On wayland everything is smooth with just one nice animation to the final size. I tested both without scaling and with scaling, and the following cases

1. Starting normally, letting the window resize to the previous size
2. Starting with `--grid 100x50`
3. Starting with `--size 500x500`
4. Starting with `--grid` and set `lines` and `columns` in my `init.lua`
5. Starting with `--maximized`

sid-6581:
@fredizzimo these changes fixed things on Windows for me (I have to confirm the white flashing on my work computer still):

image

Does that mess anything up on Wayland?

fredizzimo:
@sid-6581, it seems to work perfectly, I pushed the changes

fredizzimo:
Actually, I managed to test it on Windows through remote desktop. And I observed the following with the new changes.
1. With D3D the window is transparent with a frame for a while before it's rendered
2. With `--opengl` there's a quite bad white flash.

I guess we need to either render sooner after, or first clear the window with the Neovim background color and for D3D composition only enable it after the first frame has been renderered.

fredizzimo:
I have not profiled it, but I guess it's loading the fonts, and therefore the first frame is delayed too much.

I don't think that can be avoided when the scale factor changes, or even when the guifont is updated right before the window is shown.

fredizzimo:
I did some profiling, and the loading of the font is fast, due to the caching, but the whole create window action takes a long time. And the first frame render also takes 140ms. So, I think the only option we have is to set the window color when creating the window to the background color.

image

fredizzimo:
I think I have resolved the flickering issues now on D3D. This also works well in Wayland.

On Windows OpenGL there's still a white flash, but I don't think there's much to do about it, I think we need lower-level access to the windows messages than winit provides to do that, for example paint inside the actual `WM_PAINT` message.

I have not tried X11 or macOS and how this interacts with tiling window managers. So, it would be nice if someone could test that.

fredizzimo:
As a side-note, it would be nice to have some splash screen or even just a mouse cursor indicating that Neovide is starting on Windows, especially when using WSL, when it can take 10s of seconds, especially when there's a cold start of WSL

But even just launching the native version is quite slow:
image

Most of the time is spent enumerating adapters and creating the D3D device. I'm not sure what happens before that, but on Linux the launching is al

On Linux on the other hand the launching is much faster, but I think it could still be optimized a bit if the font loading was not a blocking operation.

sid-6581:
@fredizzimo I can confirm that everything looks great right now on my home Windows 11 system. No empty frames with Direct3D. As you mentioned, there is a white flash with OpenGL, and from what I remember that is caused by the long initialization without processing `WM_PAINT` messages.

falcucci:
I've at least done some changes and committed here to compile on macOS for now.

I also have been through every issue mentioned here related to macOS and all of them seems to be gone.

fredizzimo:
I'm going to merge this now, to get a bit more testing on main before we do a release.

sid-6581:
@fredizzimo I'm back on my work computer, and I can confirm that this also fixes the white flash on it, with Direct3D.

I'll also check Direct3D on my laptops, because I think it wasn't picking the right adapter before, so it was really slow. Hopefully some of the latest changes help with that, but if not I'll dig into the reason.

PR #2564, opened 2024-05-18
fix: Fix underline rendering

### LINKED ISSUE #2561: Underline renders way too low
**Describe the bug**
The underline underneath underlined text renders way too low, to the point of being inside the cell below.

**To Reproduce**
View any underlined text - easiest is probably to open the README in the neovide repo.

The font I'm using is JetBrains Mono @ 11.2pt, but it seems to be happening to some extent with any font and at any font size.
This is my `config.toml`:
```toml
[font]
normal = ["JetBrains Mono", "JetBrainsMono NF"]
size = 11.2
# needed after #2500, to keep the same spacing (i.e. only in the last screenshot)
width = -0.44
```

I did a `git bisect`, which found 6b04db543f68d03a1cd6e1ad4b3370732f9e1a0a (#2484) to be the offending commit.

**Screenshots**

These screenshots are on macOS, but I'm getting the same issue on Linux.

On f9848f09c9a1cd096bff17a799e96c5759003086
f9848f0
On 6b04db543f68d03a1cd6e1ad4b3370732f9e1a0a
6b04db5
On latest commit (818ff73f410dd6a1170c8a6414e8b9f33d46d50b, v0.13.0)
818ff73

 - OS: macOS 14.4.1, Arch Linux
 - Neovide Version 0.13.0
 - Neovim Version 0.10.0

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/15363188/neovide_rCURRENT.log)

AThePeanut4:
- Fix

Figured fixing #2561 would be fairly straightforward, so I took a stab at it. The core issue was the sign flip in `src/renderer/grid_renderer.rs:167` - `self.shaper.underline_position()` returns a negative number so subtracting it from the final underline position was offsetting the underline in the wrong direction. From what I can tell, the reason why it wasn't causing an issue before 6b04db543f68d03a1cd6e1ad4b3370732f9e1a0a was because `self.shaper.underline_position()` was usually just returning zero due to lost precision from all the float -> int conversions everywhere.

I also had to tweak the positioning of the other underline types, since those were based off of the previous incorrect positioning. I enabled anti-aliasing for all underline types because without it the bottom line of the double underline was appearing super thick for me. I couldn't see any negative effects of that change on any of the other underline types.  

- No

Fixes #2561

### DIFF
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -214,6 +214,10 @@ impl CachingShaper {
         metrics.ascent - metrics.underline_offset
     }
 
+    pub fn stroke_size(&mut self) -> f32 {
+        self.metrics().stroke_size
+    }
+
     pub fn y_adjustment(&mut self) -> f32 {
         let metrics = self.metrics();
         metrics.ascent + metrics.leading + self.linespace / 2.
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -210,7 +210,8 @@ impl CachingShaper {
     }
 
     pub fn underline_position(&mut self) -> f32 {
-        self.metrics().underline_offset
+        let metrics = self.metrics();
+        metrics.ascent - metrics.underline_offset
     }
 
     pub fn y_adjustment(&mut self) -> f32 {
--- a/src/renderer/grid_renderer.rs
+++ b/src/renderer/grid_renderer.rs
@@ -164,9 +164,9 @@ impl GridRenderer {
         let region = self.compute_text_region(clip_position, cell_width + 2);
 
         if let Some(underline_style) = style.underline {
-            let line_position = self.grid_scale.0.height + self.shaper.underline_position();
-            let p1 = pos + PixelVec::new(0.0, line_position);
-            let p2 = pos + PixelVec::new(width, line_position);
+            let underline_position = self.shaper.underline_position();
+            let p1 = pos + PixelVec::new(0.0, underline_position);
+            let p2 = pos + PixelVec::new(width, underline_position);
 
             self.draw_underline(canvas, style, underline_style, p1, p2);
             drawn = true;
--- a/src/renderer/grid_renderer.rs
+++ b/src/renderer/grid_renderer.rs
@@ -164,11 +164,12 @@ impl GridRenderer {
         let region = self.compute_text_region(clip_position, cell_width + 2);
 
         if let Some(underline_style) = style.underline {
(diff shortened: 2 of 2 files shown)


fredizzimo:
I haven't tested this yet, but is the line thickness consistent with this change? Also when scrolling around? You could try different fractional font sizes.

The reason why I'm asking, is that during the development of the fractional grid support, I noticed some effects like that, where the line thickness was changing depending on the scroll position.

Therefore, I think it might be good to round the underline position to an integer pixel position. The fonts already have `baseline_snap` on so the offset should still be the same https://github.com/neovide/neovide/blob/818ff73f410dd6a1170c8a6414e8b9f33d46d50b/src/renderer/fonts/font_loader.rs#L28

AThePeanut4:
I usually use fractional font sizes, and I'm not seeing any issues there. The line thickness is consistent with the latest commit - the only thing that changes is the vertical positioning of the underline. Underlines are thinner than they were prior to 6b04db543f68d03a1cd6e1ad4b3370732f9e1a0a, but if I switch to the commit before that and enable anti-aliasing, the lines become as thin as they are now.

I think anti-aliasing being disabled is probably also the reason for why the underline thickness was changing depending scroll position. Scrolling doesn't appear to affect the thickness for me on this branch.

So yeah given that I have enabled anti-aliasing for all underline types, I don't think rounding the position is really necessary. But I don't really mind either way, it's an easy enough change.

I've been developing and testing this only on my Mac - I'll go test it out on my Linux machine and see if I notice any problems.

AThePeanut4:
I'm not encountering any problems on my Linux machine (Wayland/Hyprland).

I've also added a fix for an unrelated issue with the undercurl rendering, where it would quite frequently render one too many squiggles (thus extending into the next cell) because of floating-point inaccuracies in the while condition.

fredizzimo:
I think the underline is too high now, the right one is rendered with LibreOffice
image

I also think that we should use this for the thickness, maybe scaled by `neovide_underline_stroke_scale` 
```rust
/// Recommended thickness of an underline or strikeout stroke.
pub stroke_size: f32,
```

fredizzimo:
It looks better now, I still think it would be better to not use antialiasing though, round the sizes and render at exact pixel coordinates. Here's a comparison again between LibreOffice and the current version

image

AThePeanut4:
@fredizzimo Yeah the anti-aliasing is just making the lines too thick which messes with everything. It looks much better now.

fredizzimo:
Thank you, this is good now!

fredizzimo:
This also greatly improved the undercurl rendering, now text with undercurls is actually readable.

Issue #2567, opened 2024-05-19
When entering Chinese punctuation (such as 。 ， ！ ？ ¥) on the Mac platform, it will become English punctuation.


HughLink:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. macOS 11.4]
 - Neovide Version [e.g. 0.6.0]
 - Neovim Version [e.g. 0.5.0]

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

**Additional context**

fredizzimo:
I see some problems in KDE Plasma 6 Wayland with fcitx5 too.

It looks like `neovide_input_ime` now defaults to false for some reason. Can you try to set it after Neovim has started, and see if that fixes the problem?
https://neovide.dev/configuration.html#ime

Meanwhile I will try to fix the defaults.

fredizzimo:
I hope this fixes it:
* https://github.com/neovide/neovide/pull/2571

HughLink:
Thank you, but setting neovide_input_ime = true did not solve the problem

as follows:
Use nvim in Neovide to enter Chinese punctuation as English punctuation
Use nvim in iTerm2 and enter Chinese punctuation as Chinese punctuation.

@fredizzimo

HughLink:
> I hope this fixes it:
>
> * [fix: Broken IME enable and popup position #2571](https://github.com/neovide/neovide/pull/2571)

I have also compiled and used this version, which fixed the position problem👍, but did not fix the Chinese punctuation problem.

fredizzimo:
Ok, can you run `neovide --log` then type something with that punctuation and then close neovide. Then attach the log file here. I don't have a macOS machine myself so I have to rely on the log file to debug it.

fredizzimo:
I think this might be the same bug as
* https://github.com/neovide/neovide/issues/2014, which has been reported to winit  https://github.com/rust-windowing/winit/issues/3095

But we need a log file in order to confirm. @falcucci was able to repeat it so I think he can provide one.

fredizzimo:
Related issues for Alacritty:
* https://github.com/alacritty/alacritty/issues/7763
* https://github.com/alacritty/alacritty/issues/7122

fredizzimo:
[...]

Issue #2566, opened 2024-05-19
On the Mac platform, when inputting using a non-English input method (an input method with an input box), the input box is in the wrong position and is fixed in the upper left corner of the page.


HughLink:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. macOS 11.4]
 - Neovide Version [e.g. 0.6.0]
 - Neovim Version [e.g. 0.5.0]

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

**Additional context**

fredizzimo:
I can see this problem in KDE Plasma 6 Wayland with fcitx 5 too. I will fix it

fredizzimo:
This should fix it:
* https://github.com/neovide/neovide/pull/2571

Issue #2178, opened 2023-12-14
ime popup position is incorrect 


colorsakura:
**Describe the bug**
neovide: 0.11.2.r51.gb6374265-1
![image](https://github.com/neovide/neovide/assets/32534829/9053f59f-094f-47b7-9a1d-fbf8ff9e1147)

alacritty: git version
![image](https://github.com/neovide/neovide/assets/32534829/3b42a631-d9f4-4e53-bbf4-3b1a39d099ce)

alacritty fix this bug: https://github.com/alacritty/alacritty/pull/7389

**To Reproduce**
1. open neovide
2. use fcitx5 im

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [Archlinux]
 - Neovide Version [0.11.2.r51.gb6374265-1]
 - Neovim Version [master branch]

**Please run `neovide --log` and paste the contents of the `.log` file created in the current 
log:
https://paste.rs/6vHqe

**Additional context**

PR #2571, opened 2024-05-19
fix: Broken IME enable and popup position

fredizzimo:
The IME enable state, is now called before each event is processed. For some reason it does not apply properly otherwise. That's also what was done before Neovide 0.13. 

The position is also fixed, it used the grid position instead of the physical position before. I also fixed the one line offset bug, and a bug where the position is incorrect when starting the editing.

* Fixes https://github.com/neovide/neovide/issues/2567
* Fixes https://github.com/neovide/neovide/issues/2566
* Fixes https://github.com/neovide/neovide/issues/2178

- No


fredizzimo:
I went ahead and merged this, so we can prepare a release.

fredizzimo:
It looks like these fixes it causes a lot of crashes in `fcitx5`, I will try to find a better fix.

Issue #2574, opened 2024-05-20
Float background transparent


VICTORVICKIE:
**Describe the bug**
Float background becomes transparent
![image](https://github.com/neovide/neovide/assets/45727291/c2de0b79-e131-4091-9c32-a91511f7fd37)

**To Reproduce**
1. `neovide -- -nu minimal.lua`
2. run command `:Neogit` in a git repo
3. press `?`

**Expected behavior**
The background should be filled as shown (neovide 0.12.2 at image)

**Desktop (please complete the following information):**
 - Microsoft Windows 23H2
 - Neovide Version 0.13.0
 - Neovim Version 0.10.0

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide-0.12.2_rCURRENT.txt](https://github.com/neovide/neovide/files/15377004/neovide-0.12.2_rCURRENT.txt)
[neovide-0.13.0_rCURRENT.txt](https://github.com/neovide/neovide/files/15377024/neovide-0.13.0_rCURRENT.txt)

**Additional context (minimal.lua)** 
[minimal.lua.txt](https://github.com/neovide/neovide/files/15377029/minimal.lua.txt)

badosu:
Oh, I was about to file this issue. Seeing this with LazyVim and the Lazy popup screen (e.g. opening log tab and opening a diff with d)

fredizzimo:
There's a PR up now that fixes the problem
#2580

PR #2578, opened 2024-05-21
fix: group floating window

### LINKED ISSUE #2524: Z-index layers are not being merged correctly.
**Describe the bug**
Z-index layers are not being merged correctly causing shadows to show to show for every window layer.

**To Reproduce**
1. open neovide.
2. make sure shadows are enabled.
3. open telescope or any multi-layer plugin.

**Expected behavior**
similar z-index layers get merged in to a single window.

**Screenshots**
>https://github.com/neovide/neovide/assets/6450181/79b7b8f6-f570-48f0-8ba3-8028be396117

 - OS: [Windows 10]
 - Neovide Version [Main]
 - Neovim Version [v0.10.0-dev-3080+gefb44e0ca]

[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/15238198/neovide_rCURRENT.log)

**Additional context**

TwIStOy:
- Fix

- No

close #2524

### DIFF
--- a/src/renderer/rendered_layer.rs
+++ b/src/renderer/rendered_layer.rs
@@ -202,6 +202,7 @@ pub fn group_windows(
     for i in 0..windows.len() {
         let _ = get_window_group(&mut windows, i);
     }
+    windows.sort_by(|a, b| a.group.cmp(&b.group));
     windows
         .into_iter()
         .group_by(|window| window.group)


fredizzimo:
Thanks, I will look into this in the evening. It's possible that we will also put the grouping across layers as an experimental feature toggle, to avoid issues like:
* https://github.com/neovide/neovide/issues/2574

I personally think that floating windows should never be grouped across layers, even if some plugins rely on that currently. But it should be possible to write all those UIs using a single layer, which IMO is a better than our unannounced convention of grouping all consecutive layers together.  That way only a few plugins need to adopt, maybe only nui and telescope.

fredizzimo (approved):
Yes, sorting is needed here, otherwise `group_by` will create multiple groups for the same group id. That explains the random behaviour of the bug, since the order of the window depend on the window size and maybe even random factors.

PR #2583, opened 2024-05-21
docs: Update list of currently available settings

ekorchmar:
Not all of the available settings are currently listed, as documentation suggests.
* Fork: important, as recent release changed the default behaviour and command line option is worded differently; had to dive into the code to find it.
* Theme: another recently added option.
* Font: provided a (default) example to better explain the expected format.

- Documentation 

- No

### DIFF
--- a/website/docs/config-file.md
+++ b/website/docs/config-file.md
@@ -30,11 +30,17 @@ no-multigrid = false
 vsync = true
 maximized = false
 srgb = false
+fork = false
 idle = true
 neovim-bin = "/usr/bin/nvim" # in reality found dynamically on $PATH if unset
 frame = "full"
+theme = "auto"
 title-hidden = true
 tabs = true
+
+[font]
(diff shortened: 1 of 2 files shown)


website/docs/config-file.md
```diff
@@ -30,11 +30,17 @@ no-multigrid = false
 vsync = true
 maximized = false
 srgb = false
+fork = false
 idle = true
 neovim-bin = "/usr/bin/nvim" # in reality found dynamically on $PATH if unset
 frame = "full"
+theme = "auto"
 title-hidden = true
 tabs = true
```
fredizzimo, line 39: Could you sort this and the settings in `config.rs` alphabetically. That way the order is less random, and it's easier to compare.

ekorchmar: Done in ab188aa. Local `cargo test` also ran successfully after the change. Can never be too safe

Issue #2407, opened 2024-03-07
feat: launch neovide as `--cwd $HOME` if not given directory path on macOS


falcucci:
**Is your feature request related to a problem? Please describe.**
Usually users who doesnt know neovide or it's having a trial, launches their apps from the Dock or Finder,
making it to always being opened as root path / and not the user $HOME path.

**Describe the solution you'd like**
https://github.com/neovide/neovide/pull/2405 fixes that.

**Describe alternatives you've considered**
Usually macOS users who get into neovide as their daily editor finds a trick by adding this setup on their own neovim setup, but it can generates complains, once it shouldn't open as root path as default.

polachok:
Can't we just do something like this 🤔 

```diff
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -182,6 +182,15 @@ fn platform_which(bin: &str) -> Option<String> {

 #[cfg(target_os = "macos")]
 fn nvim_cmd_impl(bin: String, mut args: Vec<String>) -> TokioCommand {
+    use std::io::IsTerminal;
+
+    let home_dir = {
+        objc2::rc::autoreleasepool(|pool| unsafe {
+            let home_dir = icrate::Foundation::NSHomeDirectory();
+            home_dir.as_ref().as_str(pool).to_owned()
+        })
+    };
+
     let shell = env::var("SHELL").unwrap_or_else(|_| "/bin/sh".to_string());
     args.insert(0, bin);
     let args = match shlex::try_join(args.iter().map(String::as_str)) {
@@ -194,6 +203,9 @@ fn nvim_cmd_impl(bin: String, mut args: Vec<String>) -> TokioCommand {
     }
     cmd.arg("-c");
     cmd.arg(&args);
+    if !std::io::stdin().is_terminal() {
+        cmd.current_dir(home_dir);
+    }
     cmd
 }
 ```

falcucci:
@polachok actually not, this will override in case the tty session is launched w a startup directory (e.g Neohub) and would break its behavior, so it must respect the path.

It also overrides the --cmd/-c command which must work properly as before in neovide such as `neovide -- -- -c "cd /project"`

your trial (which actually was my first implementation) just works in case neovide is launched from Finder and the Dock itself.

polachok:
This works from Finder, Dock and Neohub and I don't see how it may affect -c "cd /project" in any way. But yeah, whatever.

```diff
--- a/src/bridge/command.rs
[...]

PR #2587, opened 2024-05-23
fix: Don't use login shell on macOS if `$TERM` is set

### LINKED ISSUE #2584: `$PATH` is modified when launching Neovide from interactive shell
**Describe the bug**
Apologies in advance for an annoying situation 😅. I believe this is caused by the changes in #2405.

#2405 fills in environment variables when Neovide isn't run from a login shell, but when it _is_ run from a login shell in which env variables have been changed, e.g. a Python virtualenv has been activated, it re-runs the default initialization, which can e.g. prevent Pyright from correctly detecting the virtualenv.

**To Reproduce**
Very minimal repro on macOS:
1. In an interactive shell run `PATH="/foobar:$PATH" neovide`
2. In the Neovide process, run the ex command `:echo $PATH`
   - Note that default entries, e.g. `/usr/bin`, are before `/foobar` in the PATH

More real-world scenario:
1. Create a python venv in a given directory, e.g. `python -m venv ./.venv`
2. Activate the venv: `source .venv/bin/activate`
3. Run `neovide`
4. Run ex command `:checkhealth provider.python`
   - Observe in the _Python virtualenv_ section:

         Python virtualenv
         - WARNING $VIRTUAL_ENV is set to: /my-project-directory/.venv
           And its bin directory contains: python, python3, python3.11
           But $PATH yields this python3 executable: /opt/homebrew/opt/python@3.12/bin/python3.12
           And $PATH in subshells yields this python3 executable: /opt/homebrew/opt/python@3.12/bin/python3.12
           And $PATH yields this python3.11 executable: /opt/homebrew/opt/python@3.11/bin/python3.11
           And $PATH in subshells yields this python3.11 executable: /opt/homebrew/opt/python@3.11/bin/python3.11
           So invoking Python may lead to unexpected results.
           - ADVICE:
             - $PATH ambiguities arise if the virtualenv is not properly activated prior to launching Nvim. Close Nvim, activate the virtualenv, check that invoking Python from the command line launches the correct one, then relaunch Nvim.
             - $PATH ambiguities in subshells typically are caused by your shell config overriding the $PATH previously set by the virtualenv. Either prevent them from doing so, or use this workaround: https://vi.stackexchange.com/a/34996

**Expected behavior**
I _think_ that if Neovide is being run from an interactive/login shell, it makes sense not to do all the extra login shell setup, as it will already have been set up by the existing shell. However it's very possible I'm missing something and that would cause problems.

**Desktop:**
 - OS: macOS 14.4.1
 - Neovide Version: v0.13.0
 - Neovim Version: v0.10.0

**Additional context**
For the virtualenv/Pyright case specifically, this can be worked around by running `:PyrightSetPythonPath ./.venv/bin/python`

Strictly speaking I could change my shell configuration to, before modifying the `PATH`, check whether the entries it's going to add are already present, and that would fix this for some cases.

cpmsmith:
If we assume that having `$TERM` set means the user is in an interactive terminal session, we can simply run the nvim command directly instead of wrapping it in `/usr/bin/login`, preserving their environment variables. Testing locally, this does work: the simple `PATH="/foobar:$PATH" neovide` test I used in the bug report works again.

Note that this is different from the [pre-0.13 behaviour](https://github.com/neovide/neovide/blob/71e9fa49fcb393662ee03e42968724b91cde8e62/src/bridge/command.rs#L192-L194), which would still spawn a shell but just not provide `-l`. This just executes `nvim` directly. That's what I think @fredizzimo is suggesting in [this comment](https://github.com/neovide/neovide/issues/2584#issuecomment-2125654862), but I don't know if spawning a shell has effects other than setting environment variables that we'd still need.

I did notice when testing locally that if I launch the app bundle (a release build) from Finder, its cwd is `/`, which would seem to be a regression of #2407, however it does the same thing on `main`, so I don't think it's related to my change.

Fixes #2584

- Fix

- No

### DIFF
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -49,6 +49,16 @@ fn build_login_cmd_args(command: &str, args: &[&str]) -> (String, Vec<String>) {
 
     use crate::error_handling::ResultPanicExplanation;
 
+    // If $TERM is set, we assume user is running from a terminal, and we shouldn't
+    // re-initialize the environment.
+    // See https://github.com/neovide/neovide/issues/2584
+    if env::var_os("TERM").is_some() {
+        return (
+            command.to_string(),
+            args.iter().map(|s| s.to_string()).collect(),
+        );
+    }
+
     let user = env::var("USER").unwrap_or_explained_panic("USER environment variable not found");
     let shell = env::var("SHELL").unwrap_or_else(|_| "/bin/zsh".to_string());
 


fredizzimo:
@cpmsmith, thank you. 

Yes, I don't think that spawning a shell when you are already inside one is needed.  We don't do it on Linux either.

And we reverted the part that sets the startup directory, as that caused other problems. So that's why you don't see the effect of that.

@9mm, Could, you test if this fixes your problems? @falcucci, could you test that launching from the finder still works as you expect?

9mm:
Will try this one out tomorrow, thanks!

falcucci:
Reproduced and working normally from Finder.

I believe we can merge this.

falcucci (approved):

9mm:
Amazing.. I think that works and it also solves the problem of /usr/bin appearing at the front of my path requiring tons of hacks to duplicate other paths in front of that just so all my system binaries arent used

PR #2596, opened 2024-05-26
Update snippet for dynamic scale factor change

kdarkhan:
Based on my testing, the current version of the snippet does not change scaling until after the next update. To see the scale change, I usually need move cursor after every scale change.

The proposed solution makes it more intuitive as the update happens immediately.

Do note that I haven't tested the Vimscript version but hope it works as I am not very
versed in it. I verified that Lua version works as expected.

- Documentation 

- No

### DIFF
--- a/website/docs/faq.md
+++ b/website/docs/faq.md
@@ -48,6 +48,7 @@ VimScript:
 let g:neovide_scale_factor=1.0
 function! ChangeScaleFactor(delta)
   let g:neovide_scale_factor = g:neovide_scale_factor * a:delta
+  redraw!
 endfunction
 nnoremap <expr><C-=> ChangeScaleFactor(1.25)
 nnoremap <expr><C--> ChangeScaleFactor(1/1.25)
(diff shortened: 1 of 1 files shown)


xzbdmw:
This one works immediately
`keymap({ "n", "v" }, "<D-=>", ":lua vim.g.neovide_scale_factor = vim.g.neovide_scale_factor + 0.1<CR>")`

kdarkhan:
@xzbdmw I assume your version works because it calls an `ex` mode and that probably forces a UI redraw.

fredizzimo:
I would rather see that this is properly fixed rather than documenting a workaround. It should be quite easy to fix by handling it here:
https://github.com/neovide/neovide/blob/4d567d726ae8c2e89c6f4aa2a652766955bd0746/src/window/window_wrapper.rs#L197 and moving this code to the handler.

https://github.com/neovide/neovide/blob/4d567d726ae8c2e89c6f4aa2a652766955bd0746/src/renderer/mod.rs#L396-L402

kdarkhan:
I am not very familiar with the codebase to make this contribution. Feel free to take over. I am running Neovide on Linux btw if that's needed for reproducing the issue.

fredizzimo:
I'm closing this, since this was merged:
* https://github.com/neovide/neovide/pull/2610

kdarkhan:
Thanks for a quick turnaround.

Issue #2572, opened 2024-05-20
neovide --version inside script fails with Neovide panicked with the message 'Trying to retrieve a settings object that doesn't exist: {:?}'.


edi9999:
**Describe the bug**

**To Reproduce**

Run : 

```sh
 a="$(neovide --version)"; echo "$a"
```

Results in : 

```
xkbcommon: ERROR: couldn't find a Compose file for locale "en_US.utf-8" (mapped to "en_US.utf-8")
Neovide panicked with the message 'Trying to retrieve a settings object that doesn't exist: {:?}'. (File: src/settings/mod.rs; Line: 98, Column: 14)
This is a bug and we would love for it to be reported to https://github.com/neovide/neovide/issues

Backtrace saved to neovide_backtraces.log!
```

**Expected behavior**

Should return 0.13.0

Note that the following works fine (when not having a variable) : 

```
neovide --version
```

Outputs : `neovide 0.13.0`

**Desktop (please complete the following information):**
 - OS: Ubuntu 22
 - Neovide Version 0.13.0
 - Neovim Version 0.10.0

**Additional context**

In your tests, I would test the following cases which all should work : 

```
neovide --version
neovide --version | tee /tmp/log
a="$(neovide --version)"; echo "$a"
bash -c 'neovide --version'
sh -c 'neovide --version'
a="$(bash -c 'neovide --version')"; echo "$a"
```

fredizzimo:
libxbcommon unfortunately does not support non-utf8 locales, see https://github.com/xkbcommon/libxkbcommon/issues/76

Make sure you have an utf-8 locale https://help.ubuntu.com/community/Locale

edi9999:
Sorry for some of the noise, I don't care about the first line : 

```
xkbcommon: ERROR: couldn't find a Compose file for locale "en_US.utf-8" (mapped to "en_US.utf-8")
```

When I run : 

```
neovide --version
```

It gives me 

```
xkbcommon: ERROR: couldn't find a Compose file for locale "en_US.utf-8" (mapped to "en_US.utf-8")
neovide 0.13.0
```

However, the bug is when run in a subshell, ie : 

```
v="$(neovide --version)"
```

It outputs : 

```
xkbcommon: ERROR: couldn't find a Compose file for locale "en_US.utf-8" (mapped to "en_US.utf-8")
[...]

Issue #2204, opened 2023-12-25
neovide --version creates a window instead of just returning


edi9999:
```
#!/usr/bin/env bash

set -euo pipefail

source /usr/bin/env_parallel.bash

install_one() {
	vim --version
	neovide --version
}
env_parallel install_one <<<"foo"
```

With following script, 

vim --version responds with its version (9.0 on my computer), but neovide --version does'nt respond in the terminal, instead, it opens up a new window.

It does that only when run inside "env_parallel", it works ok if I just run install_one.

fredizzimo:
In order to ensure that errors and output is always displayed we now always display a GUI windows if a terminal is not detected.

We didn't consider scripting usage, but maybe we should, and also detect piped output.

edi9999:
How can I tell neovide that it is indeed in a terminal ?

Maybe simply setting an environment variable ??

Setting TERM=something should probably be a good way ??

edi9999:
In my term , I have TERM=screen-256color

fredizzimo:
On the Neovide side, we use this https://github.com/neovide/neovide/blob/af9869bd6d8d3434b187e29f85032ee74e40b536/src/error_handling.rs#L104

I  think checking for `TERM`, instead of that would indeed be a good fix in this case, at least on Linux, since that would handle all the cases where it's started from a terminal, even when the stdout is piped. And probably still show the GUI when launched in other ways. 

But there might still be some edge cases that works wrong.

edi9999:
Would people really run `neovide --version` from outside of a terminal ??

How would they run it ?

PR #2621, opened 2024-06-01
fix: Neovide --version and Neovide --help

fredizzimo:
1. Fixes crash when the GUI error message window is shown and the command line parsing fails.
2. Fixes yanking from the GUI error message window, when it's shown before the clipboard has been initialized
3. The errors/messages are always printed to stdout/stdout when they originate from the command line parsing. After that they are either shown in the console or the GUI window depending on if a TTY is attached or not. 

* Fixes https://github.com/neovide/neovide/issues/2572
* Fixes https://github.com/neovide/neovide/issues/2204

- No


Kethku:
lgtm

Issue #2599, opened 2024-05-27
get neovide version in vimscript


cizmazia:
**Is your feature request related to a problem? Please describe.**

I have not found a way to determine the version of the running neovide process.

**Describe the solution you'd like**

Make the neovide version available via a read-only variable `g:neovide_version`.

**Describe alternatives you've considered**

```vim
echo execute('!neovide --version')
```

Panicks with the message 'Trying to retrieve a settings object that doesn't exist: {:?}'. (File: src/settings/mod.rs; Line: 98, Column: 14)

Possibly related #2572.

PR #2622, opened 2024-06-01
feat: Provide a neovide_version variable

fredizzimo:
This adds a new variable `neovide_version`. The channel information is also extended to contain the patch version.

* Fixes https://github.com/neovide/neovide/issues/2599

- No


Kethku:
lgtm

Issue #2576, opened 2024-05-20
Issues with --no-multigrid


CipherJay:
I have been using an alias to run `neovide --no-multigrid` because I found that this makes the background in Telescope and other floating windows look the same as they do in neovim. As of the latest update to version 13, a few things are happening differently.

I have my alias set like this:

`alias vim = 'neovide --no-multigrid"`
 
This used to open up Neovide and the corresponding file passed to it in a window that was not dependent on the terminal. Now, it hangs in the terminal until I close Neovide and it provides me some new errors:

```
ERROR [neovide::renderer] WindowDrawCommand: Viewport { scroll_delta: 0.0 } sentfor uninitialized grid 2
ERROR [neovide::renderer] WindowDrawCommand: Viewport { scroll_delta: 0.0 } sentfor uninitialized grid 2
ERROR [neovide::renderer] WindowDrawCommand: Viewport { scroll_delta: 0.0 } sentfor uninitialized grid 2
ERROR [neovide::renderer] WindowDrawCommand: Viewport { scroll_delta: 0.0 } sentfor uninitialized grid 2
```
I am currently running Neovide on Arch Linux (Zen) Kernel 6.8.9 with latest nvidia-dkms, running Hyprland 0.40.

Here's the current log when I run  `neovide --no-multigrid --log`:

https://gist.github.com/CipherJay/1a6d2e34236bfaaf82b6973ac923c44c

Please let me know if I am doing something wrong or if there is a correct way to achieve what I am trying to with this new version. I'm just looking to have it look as it does on neovim with a transparent background like this:

![image](https://github.com/neovide/neovide/assets/116757358/36858a16-4e4b-4db3-a091-5dc5d457fb1c)

Thanks for your help and for making the best gui for Neovim.

HeftyFair:
same error on macos.

fredizzimo:
It looks like a Neovim bug, grid 2 should never exist when multigrid is not enabled. But it does no harm, so we should probably change it to a warning instead so that it's not printed to the console by default.

mbway:
> What is the correct way for me to disable this warning

for now, instead of your alias: `alias vim = 'neovide --no-multigrid"` you could try:
```
vim() {
    neovide --no-multigrid "$@" 2>/dev/null
}
```
which works for me

hanspinckaers:
Maybe it helps to add NEOVIDE_FORK=1?

`alias vim = 'NEOVIDE_FORK=1 neovide --no-multigrid'`

PR #2623, opened 2024-06-01
fix: Remove some logspam

fredizzimo:
* Remove the error log for WindowCommand sent to uninitialized grid when multigrid is disabled. 
  * Fixes https://github.com/neovide/neovide/issues/2576
* Remove unnecessary `prepare_lines` message 

- No


Kethku:
looks good to me

PR #2625, opened 2024-06-02
chore: Remove obsolete `--no-fork` arguments

last-partizan:
I noticed that my neovide call looks like this in the process list:

```
> ps ax | grep -E '[n]eovide'
 134800 pts/2    Sl     0:00 /usr/bin/neovide --no-fork --fork
```

I'm using `--fork` argument, why there's --no-fork?

@fredizzimo probably you forgot to remove this?

I removed this arg, and it works as expected. Also, cleaned up docs.

- Fix

- No

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -246,7 +246,6 @@ fn maybe_disown() {
             .stdin(process::Stdio::null())
             .stdout(process::Stdio::null())
             .stderr(process::Stdio::null())
-            .arg("--no-fork")
             .args(env::args().skip(1))
             .spawn()
             .is_ok());
--- a/website/docs/editing-with-external-tools.md
+++ b/website/docs/editing-with-external-tools.md
@@ -11,7 +11,7 @@ your own responsibility._
 In your configuration file:
 
 ```yaml
-editor: "neovide --no-fork"
+editor: "neovide"
 ```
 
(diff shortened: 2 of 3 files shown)


src/main.rs
```diff
@@ -246,7 +246,6 @@ fn maybe_disown() {
             .stdin(process::Stdio::null())
             .stdout(process::Stdio::null())
             .stderr(process::Stdio::null())
-            .arg("--no-fork")
```
fredizzimo, line 249: Hm.. I think we still want to force `--no-fork` here, in case the users have set the environment variables and the configuration file to fork. Otherwise, there's a risk of getting a recursion where it forks infinitely.

I'm not sure where the duplicate elements come from though.

last-partizan: Duplicate option comes from this line.

I start neovide with `--fork`, and it then restart itself with `--no-fork` + old args.

I tested setting NEOVIDE_FORK variable (to 0 and 1), and both cases works as expected. Setting a variable also works fine.

I think, `is_tty()` returns false for forked process, and we don't need explicit `--no-fork` here.

fredizzimo: Ah, yes, that makes sense

fredizzimo:
Thank you!

PR #2626, opened 2024-06-02
fix: Unset ARGV0 on startup

### LINKED ISSUE #2477: In neovide appimage terminal[:term] mode rustc not working properly Shell:zsh
**Describe the bug**
❯ rustc --version

error: unknown proxy name: 'neovide'; valid proxy names are 'rustc', 'rustdoc', 'cargo', 'rust-lldb', 'rust-gdb', 'rust-gdbgui', 'rls', 'cargo-clippy', 'clippy-driver', 'cargo-miri', 'rust-analyz

er', 'rustfmt', 'cargo-fmt'.

In neovide appimage only 

**To Reproduce**
1. Open Neovide appimage
2. go to terminal mode by using :term command
3. Then type rustc --version or rustup --version

**Desktop :**
 - OS:  AntiX 23
 - Neovide Version 0.12.2
 - Neovim Version v0.10.0-dev-2637+g400ef8aaa

agraven:
This should close #2477 with child processes run from zsh having the wrong argv[0].

Strictly speaking this could be put behind an appimage feature flag which then AppImage could be built with, but for something this small it didn't really feel necessary. I figured it would be best to remove the env variable the same place all the other env manipulation is done, but if somewhere else makes more sense I'm happy to move it.

I have yet to actually build and test an appimage with this change.

- Fix

- No

### DIFF
--- a/src/settings/config.rs
+++ b/src/settings/config.rs
@@ -76,6 +76,8 @@ impl Config {
     }
 
     fn write_to_env(&self) {
+        // This variable is set by the AppImage runtime and causes probles for child processes
+        env::remove_var("ARGV0");
         if let Some(wsl) = self.wsl {
             env::set_var("NEOVIDE_WSL", wsl.to_string());
         }


fredizzimo:
I tried the Appimage from the artifacts, but unfortunately it does not seem like it fixes the bug, and the environment variable `ARGV0` is still set inside the spawned Neovim instance.

I'm not sure why. The code seems to be correct. 

Another approach would be to remove it from the environment of the spawned Neovim instance only. In that case I think you need to copy the existing environment and remove it from there and pass it to the `TokioCommand` in `command.rs`.

fredizzimo:
Perhaps it's not the best place, that code is only called when the config file load fails.

https://github.com/neovide/neovide/blob/1db63d93b18cc1eb5ac5b3e67b3ca63505f10800/src/settings/config.rs#L67

But this could also mean that the config file is not working from the AppImage for some reason.

fredizzimo:
I made another fix here and verified that it works:
* https://github.com/neovide/neovide/pull/2691

So, closing this one.

Issue #2474, opened 2024-04-08
Cannot enable font features for CascadiaCode-Italic


mathletedev:
**Describe the bug**
When adding cursive option to `font.features`, text remains non-cursive

**To Reproduce**
`config.toml`:
```toml
[font]
normal = [ "Cascadia Code" ]
size = 16

[font.features]
# I have also tried CascadiaCode = [ "+ss01" ]
CascadiaCode-Italic = [ "+ss01"  ]
```

**Expected behavior**
Setting these options in the Kitty terminal
```
font_family Cascadia Code
font_size 12
font_features CascadiaCode-Italic +ss01
```
produces cursive when text is italic

**Screenshots**
Kitty + Neovim:
![image](https://github.com/neovide/neovide/assets/63214683/f8d14ef7-d742-4c0e-b1a7-25875acbe47b)

Neovide:
![image](https://github.com/neovide/neovide/assets/63214683/2286cb47-df6f-4ed4-933c-a66dac38d0f1)

**Desktop (please complete the following information):**
 - OS: NixOS 24.05
 - Neovide Version: 0.12.2
 - Neovim Version: 0.9.5

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
Log file too large, will attach if needed

**Additional context**

iwinux:
This is working for me:

```toml
[font]
normal = ["Cascadia Code PL"]
size = 16

[font.features]
# simple compiler & simulator
# -> => != <> >= 0 1 2 3
"Cascadia Code PL" = ["+ss01", "+ss02", "+ss04", "+ss05", "+ss06", "+zero", "+onum"]
```

<img width="859" alt="image" src="https://github.com/neovide/neovide/assets/98864/7cd95763-a947-45b0-966e-a743be8dcc81">

mathletedev:
Ah, I was missing a space in the font name; it works now. Thank you so much!

PR #2639, opened 2024-06-06
Fix wrong toml format and improve font fallback logic

### LINKED ISSUE #2632: Config file not working
**Describe the bug**
I try to use this config file in `$HOME/.config/neovide/config.toml`
```toml
[font]
    normal = [] # Will use the bundled Fira Code Nerd Font by default
    size = 14.0

fork = false
frame = "buttonless"
idle = true
maximized = false
neovim-bin = "/usr/bin/nvim" # in reality found dynamically on $PATH if unset
no-multigrid = false
srgb = false
tabs = true
theme = "auto"
title-hidden = true
vsync = true
wsl = false
```
but it doesn't work, like if I set frame=buttonles on command line it will be buttonless, but on config file nothing happened. And I try to change other parameters, like font size but nothing happened either.

**To Reproduce**
1. run `brew install --cask neovide`
2. run `sudo launchctl config user path "$(brew --prefix)/bin:${PATH}"`
3. run `mkdir ~/.config/neovide/`
4. run `nvim ~/.config/neovide/config.toml` and paste the content of config
5. run `neovide`

 - OS: [macOS 14.4.1]
 - Neovide Version [0.13.1]
 - Neovim Version [0.10.0]

[neovide_rCURRENT.log](https://github.com/user-attachments/files/15569765/neovide_rCURRENT.log)

xzbdmw:
- Fix #2632 
- Documentation

- No

### DIFF
--- a/src/renderer/fonts/font_options.rs
+++ b/src/renderer/fonts/font_options.rs
@@ -204,7 +204,12 @@ impl FontOptions {
             .map(|fonts| {
                 fonts
                     .iter()
-                    .flat_map(|font| font.fallback(&self.normal))
+                    .filter(|font| font.family.is_some())
+                    .map(|font| FontDescription {
+                        family: font.family.clone().unwrap(),
+                        style: None,
+                    })
+                    .chain(self.normal.iter().cloned())
                     .collect()
             })
             .unwrap_or_else(|| self.normal.clone());
@@ -212,7 +217,7 @@ impl FontOptions {
         fonts
             .into_iter()
             .map(|font| FontDescription {
-                style: font.style.or_else(|| style.name().map(str::to_string)),
+                style: style.name().map(str::to_string),
                 ..font
             })
             .collect()
@@ -382,25 +387,6 @@ impl FontDescription {
             FontStyle::default()
         };
         (self.family.as_str(), style)
-    }
-}
-
-impl SecondaryFontDescription {
-    pub fn fallback(&self, primary: &[FontDescription]) -> Vec<FontDescription> {
-        if let Some(family) = &self.family {
-            vec![FontDescription {
-                family: family.clone(),
-                style: self.style.clone(),
-            }]
-        } else {
-            primary
-                .iter()
-                .map(|font| FontDescription {
-                    family: font.family.clone(),
-                    style: self.style.clone(),
-                })
-                .collect()
-        }
     }
 }
 
--- a/website/docs/config-file.md
+++ b/website/docs/config-file.md
@@ -96,7 +96,7 @@ normal = ["MonoLisa Nerd Font"]
 size = 18
 
 [font.features]
-MonoLisa = [ "+ss01", "+ss07", "+ss11", "-calt", "+ss09", "+ss02", "+ss14", "+ss16", "+ss17" ]
+"MonoLisa Nerd Font" = [ "+ss01", "+ss07", "+ss11", "-calt", "+ss09", "+ss02", "+ss14" ]
 ```
 
 Specify font weight:
--- a/website/docs/config-file.md
+++ b/website/docs/config-file.md
@@ -25,10 +25,6 @@ There are two types of settings:
 Settings currently available in the config file with default values:
 
 ```toml
-[font]
-    normal = [] # Will use the bundled Fira Code Nerd Font by default
-    size = 14.0
-
 fork = false
 frame = "full"
 idle = true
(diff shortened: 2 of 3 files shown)


website/docs/config-file.md
```diff
@@ -98,3 +98,20 @@ size = 18
...
+```toml
+[font]
+size = 19
+hinting = "full"
+edging = "antialias"
+
+[[font.normal]]
+family = "JetBrainsMono Nerd Font Propo"
+style = "W400"
+
```
fredizzimo, line 113: I think we could add a second element here to demonstrate fallback fonts:
```toml
[[font.normal]]
family = "Noto Sans CJK SC"
style = "Normal"
```

I thought that would be enough, but it may not be actually. Checking the code, it looks like it fallbacks to normal only when a style is not found, and not just for finding fallbacks. So, it's possible that the fallbacks have to be duplicated for all weights. 

That needs some testing to determine if it's the case or not. But I think fallbacks like that is something we do want to support to avoid duplication.

xzbdmw: I'll trying to modify the fallback logic and add some test.

xzbdmw: To be clear, do you mean
```toml
[[font.normal]]
family = "not found1"
style = "W400"

[[font.normal]]
family = "menlo"

[[font.bold]]
family = "not found2" #
style = "W600"
```

In this situation, bold font should fallback to menlo instead of builtin fira code(current behaviour).

fredizzimo: Yes, I think that behaviour makes the most sense, since missing style elements falls back to the normal style, so the fallbacks should work the same way. 

In this case, the full fallback list for bold would be `not found2`, `not found1`, `menlo`, in that order.

xzbdmw:
@fredizzimo How can I add test, I don't find font other than fira code to test fallback
To be clear, here are some examples:

### before this change:
1. 
```toml 
[[font.normal]]
family = "not found1"
style = "W400"

[[font.normal]]
family = "menlo"
style = "W400"

[[font.bold]]
family = "asdasd"
style = "W600"
```
This will fallback to fira code, because in previous version, if `[[font.bold]]`'s `family` is specified, it simply don't add `[[font.normal]]` as a fallback, but it will fail at swash process, then it will fallback to fira code.

2. 
```toml 
[[font.normal]]
family = "not found1"
style = "W400"

[[font.normal]]
family = "menlo"
style = "W400"

[[font.bold]]
style = "W600"
```
This will add menlo to fallback list, because font family is not specifed.

### After this change
1.1 will always fallback to `[[font.normal]]`, 1.2 behaves the same.

fredizzimo:
I think you should be able to create similar tests as the ones in `font.rs`, that is to test the actual font config instead of the whole integration with Swash, which is much harder, and don't necessary here IMO.

xzbdmw:
@fredizzimo Added two test.

src/renderer/fonts/font_options.rs
```diff
@@ -204,7 +204,12 @@ impl FontOptions {
             .map(|fonts| {
                 fonts
                     .iter()
-                    .flat_map(|font| font.fallback(&self.normal))
+                    .filter(|font| font.family.is_some())
+                    .map(|font| FontDescription {
+                        family: font.family.clone().unwrap(),
+                        style: None,
+                    })
+                    .chain(self.normal.iter().cloned())
```
xzbdmw, line 212: Should we respect users normal weight?
eg.
```toml
[[font.normal]]
family = "not found1"
style = "W400"

[[font.normal]]
family = "menlo"
style = "italic"

[[font.bold]]
family = "something"
style = "W600"
```
Should this force normal's weight to Bold or use italic?

fredizzimo: I think we should force the weight to bold. That's more simple, and people who need special fallbacks and weights can always manually define fallbacks for all styles.

fredizzimo:
@xzbdmw, sorry for taking so long with this. I'm on vaccation with only an ipad available and the CI is broken (not your fault) so I didn't merge.

I'm not sure when someone will have time to fix that (my suggested fix is described here https://github.com/neovide/neovide/pull/2641#issuecomment-2156021707). But until that's done there's perhaps another minor fix we could include in this PR, the fallback example is wrong, see this comment https://github.com/neovide/neovide/issues/2474#issuecomment-2182972671

xzbdmw:
Ah, I see, should unfounded font settings print a warning?

fredizzimo:
@xzbdmw, just fixing the documentation is good enough for now. Unless you also want to add a warning.

To me a warning makes sense, but we have various reports of people not even wanting warnings for completely missing fonts, so it could cause some complaints. But just in terms of debugging issuses, I think it's worthwhile.

fredizzimo:
The CI has now finally been fixed, can you rebase on top of main?

xzbdmw:
Done!

fredizzimo:
Thank you!

PR #2640, opened 2024-06-06
style: simplify string formatting

hamirmahal:
- Codestyle

- No


Kethku:
@hamirmahal this fails on windows. Can you look at the action failure?

PR #2641, opened 2024-06-07
chore: Fix reviewdog installation on Windows

fredizzimo:
The reviewdog installation started to fail after the 20240603.1.1 windows image update

- No

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-latest, macOS-12, ubuntu-latest]
+        os: [win22/20240514.3, macOS-12, ubuntu-latest]
         toolchain: [stable, nightly]
     runs-on: ${{ matrix.os }}
 


fredizzimo:
This approach did not work. I think the best we can do is to either remove reviewdog completely and do a standard `cargo clippy` command. The reviewdog is quite confusing anyway, since you get duplicates for each configuration. It's also not a hard failure.

In order to do this, I think we need to lock the version of rust, at least for the stable branch, so that we don't suddenly start to get new errors when the version changes. In the nightly branch, all clippy errors could simply be just warnings. and not fail the build.

Alternatively, we could report a bug here https://github.com/reviewdog/action-setup, and disable reviewdog completely until it's fixed. Note I'm not 100% sure that it was the 20240603.1.1 version of the window image that broke it, but it very much looks like it by checking the history of failed and successful CI runs. I don't see anything that stands out in the changed list https://github.com/actions/runner-images/releases/tag/win22%2F20240603.1. It's also of course possible that it's a bug in the image itself and will be changed in the next image release.

fredizzimo:
This was replaced by:
* https://github.com/neovide/neovide/pull/2685

PR #2643, opened 2024-06-08
Allow for the cursor icon to be configured on startup

Jay-Madden:

- Feature

I also opened an issue to discuss if this is a feature that is wanted at all here: https://github.com/neovide/neovide/issues/2644

But i figured an initial implementation couldnt hurt as well. 

## To Do
this PR still needs docs updates, but I will do those after there is a signal this feature is wanted at all.

No, if no config is found the fallback remains the default value

### DIFF
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -67,6 +67,10 @@ pub struct CmdLineSettings {
     #[arg(long = "title-hidden", env = "NEOVIDE_TITLE_HIDDEN", value_parser = FalseyValueParser::new())]
     pub title_hidden: bool,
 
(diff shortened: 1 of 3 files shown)


fredizzimo:
Thanks, I think we will wait a bit still with this. We don't want to introduce an option that we have to deprecate later. 

We might not have to wait for Neovim support for additional cursor shapes, but we still want the option to be flexible and extendible enough for the future. So, we are going to start working on a more extensible way of configuring Neovide shortly.

falcucci:
@Jay-Madden closing in prol of #2834

sorry about not realizing this PR before,

thank you for your contribution!

PR #2607, opened 2024-05-29
fix: Position underline correctly with linespace >0

cpmsmith:
Fixes neovide/neovide#2600

## Before
image
image

## After
image
image

## What kind of change does this PR introduce?
- Fix

## Did this PR introduce a breaking change?
- No

Issue #1774, opened 2023-02-20
Vertical lines issue


alloc33:
I'm not sure is it line height or something else, but with JetBrains mono font I have this kind of empty spaces. 
Everything is ok in Iterm2 for example. What is the problem and is it possible to fix that?
<img width="990" alt="Screenshot 2023-02-20 at 1 02 42 PM" src="https://user-images.githubusercontent.com/45129790/220013544-90541863-1b63-4321-9123-8a838f15ece1.png">
<img width="721" alt="Screenshot 2023-02-20 at 12 54 27 PM" src="https://user-images.githubusercontent.com/45129790/220013307-25a97e22-23f8-402c-b766-2fd656bc391d.png">

hsheikhali1:
Any updates on this? I'd love to increase the line height but have the actual vertical lines render properly here.

fredizzimo:
With custom additional spacing there will always be gaps, when rendered correctly according to the font. And even without, the font width and height need to match exact pixels, or there will be gaps. 

We need custom box character drawing to fix the issue, see:
* https://github.com/neovide/neovide/issues/2491

PR #2647, opened 2024-06-09
feat: Add underline_offset option for adjusting underline y pos

bravo1003:

- Feature

- No, default offset is 0 which is basically the same as not setting any offset

## ToDo
Let me know if this feature makes sense, I can update doc accordingly
Where this option should be in `pub fn draw_foreground` or `pub fn draw_underline`?


src/renderer/grid_renderer.rs
```diff
@@ -166,8 +166,9 @@ impl GridRenderer {
         if let Some(underline_style) = style.underline {
             let stroke_size = self.shaper.stroke_size();
             let underline_position = self.shaper.underline_position();
-            let p1 = pos + PixelVec::new(0.0, underline_position);
-            let p2 = pos + PixelVec::new(width, underline_position);
+            let underline_offset = SETTINGS.get::<RendererSettings>().underline_offset;
```
cpmsmith, line 169: I'm inclined to say this should be inside [the implementation of `underline_position`](https://github.com/neovide/neovide/blob/1db63d93b18cc1eb5ac5b3e67b3ca63505f10800/src/renderer/fonts/caching_shaper.rs#L213), and probably expressed as a fraction of `current_size`, so it scales with font size/screen DPI.

fredizzimo:
I would prefer if the automatic setting was good enough for all cases and we wouldn't need to expose a setting for this.

Maybe, there's still something wrong with our offset calculation? Do you have some screenshots how it looks by default, and with the adjustments you prefer? And also comparisons to other editors?

We also merged this related fix last week, maybe the issues you were seeing was caused by that?
* https://github.com/neovide/neovide/pull/2607

bravo1003:
> I would prefer if the automatic setting was good enough for all cases and we wouldn't need to expose a setting for this.
> 
> Maybe, there's still something wrong with our offset calculation? Do you have some screenshots how it looks by default, and with the adjustments you prefer? And also comparisons to other editors?
> 
> We also merged this related fix last week, maybe the issues you were seeing was caused by that?
> 
>     * [fix: Position underline correctly with linespace >0 #2607](https://github.com/neovide/neovide/pull/2607)

@fredizzimo 
I don't think there's something wrong with the current calculation.
This is just an minor enhancement that I think would be good to have.
As you can see, here I'm using rainbow delimiter nvim and indent-blankline. And with this I was able to tweak the underline pos to my liking.
By default, Wezterm and Kitty also doesn't look too nice either with default setting, but they offer `underline_position` setting for users to tweak it. Hence this PR is to match that(Plus I don't like my udnerline too close either). 
Of course, this is something really minor, so if you don't think this makes much sense, we can keep the current behavior. :p

# Without offset
screenshot_2024-06-11_12-47-26_all

# With offset
screenshot_2024-06-11_12-44-58_all

fredizzimo:
To me it does look like there's something wrong with the underline position in that screenshot, I don't think it should ever overlap with the bottom of `z` for example.

The strange thing is that it seems to have reverted to the behaviour before the fix, see this message:
https://github.com/neovide/neovide/pull/2564#issuecomment-2119200428

That's why I would like to see comparisons with some other program, preferably a real text editor, which renders the fonts as designed, I used LibreOffice in my example.

I only have an ipad available myself for a month, so I can't do any comparisons myself.

bravo1003:
Oh, you're actually right!
I use the same font under LibreOffice and it renders underline as such
OS: Ubuntu 22.04
Font used: FantasqueSansM Nerd Font

screenshot_2024-06-11_13-37-11_all

For comparison(default neovide default with no offset)
screenshot_2024-06-11_13-34-49_all

Quick comparison of another font: JetBrainsMono Nerd Font"
And it also renders a lot better. 
screenshot_2024-06-11_13-46-12_all

screenshot_2024-06-11_13-43-20_all

So I think this might be issue with handling specific font inside neovide.

fredizzimo:
It's a little bit strange, because our position should actually be lower than the designed position with the current code.

https://github.com/neovide/neovide/blob/1db63d93b18cc1eb5ac5b3e67b3ca63505f10800/src/renderer/grid_renderer.rs#L251

That should actually be 
 `let p1 = (p1.x.round(), (p1.y - stroke_width / 2.).round());` aligning the line middle with the font underline position.

That probably means that we interpret the font underline position wrong. Maybe the baseline snap is not take. into account or something.

fredizzimo:
I'm not able to test this myself, but I believe that this should take the `linespace` into account, which means it could be `self.y_adjustment() - metrics.underline_offset`

https://github.com/neovide/neovide/blob/1db63d93b18cc1eb5ac5b3e67b3ca63505f10800/src/renderer/fonts/caching_shaper.rs#L215

The code that I mentioned in my previous message also likely needs to be updated.

bravo1003:
Keep in mind, the same font seems to have problem not only in neovide, but in WezTerm as well.

screenshot_2024-06-11_15-34-07_all

I can do some digging around the code you mentioned above.
Basically, my guessing is that there should be some special logic handling the underline position that we did not taken into account of. I'll see what I can find when I have some spare time.

cpmsmith:
> I would prefer if the automatic setting was good enough for all cases and we wouldn't need to expose a setting for this.
>
> Maybe, there's still something wrong with our offset calculation?

I will say I was looking into this myself in connection with #2607, because on my setup, with [Input Mono](https://input.djr.com), the underline ends up very close to the text:

image

According to [FontDrop](https://fontdrop.info/), Input has an `underlinePosition` of 0 (found under Data → PostScript table), so this appears "correct", or at least to-spec. Fantasque Sans Mono appears to have an `underlinePosition` of 1, which compared to the cap height of 1281 is practically zero.

I think people may expect renderers not to respect this value, though. I noticed that in CSS, there's a [`text-underline-position`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-underline-position) property, and it has a `from-font` value which is _not_ the default. The default is `auto`,  i.e. ["the user agent may use any algorithm to determine the underline’s position"](https://www.w3.org/TR/css-text-decor-3/#underline-auto). This is what the two values look like in Firefox for Input Mono:

image

fredizzimo:
I just noticed that I wrote `linespace` instead of `leading`(linegap) in my previous message. See https://freetype.org/freetype2/docs/glyphs/glyphs-3.html for the meanings.

The underline is relative to the baseline, so to get the underline position from the top of the font box that needs to be included.

Also note that `draw_text_blob` uses the y position of the baseline, that's why `y_adjustment` is added to the position, when drawing the actual text.

fredizzimo:
Another thing that I noticed is 
https://github.com/neovide/neovide/blob/1db63d93b18cc1eb5ac5b3e67b3ca63505f10800/src/renderer/fonts/caching_shaper.rs#L224

Should probably be `metrics.ascent + metrics.leading / 2. + self.linespace / 2.` to distribute the font linespace equally, just like our own additional `linespace`. That might even fix some issues like this https://github.com/neovide/neovide/issues/1774

src/renderer/grid_renderer.rs
fredizzimo: I agree with both of these suggestions.

fredizzimo (commented):
Can you also add some documentation?

src/renderer/grid_renderer.rs
fredizzimo: Additionally, I think that it makes sense to ignore the font `underline_offset` offset completely if this is set, so that it completely overrides it.

fredizzimo:
Actually, when thinking more about this, I think we should only support this setting in the font config file.

https://neovide.dev/config-file.html#font, so that it can be specified per font and style.

kmoschcau:
Hi @fredizzimo, sorry to necro this thread, but are there still plans to add an underline offset as a feature? Maybe as a new PR.

fredizzimo:
Yes, I think so, if someone has time to do it, I don't think the core team has time for a while. I personally, don't even have access to a development pc right now.

First, I think the automatic calculation should be fixed accrording to the comments here, and then, if still needed add a setting.

In particular, when the font specifies an offset of 0, then the line thickness should be used. But I also think there are some potential rounding problems.

Issue #2669, opened 2024-06-26
<C-Space> does nothing


fredizzimo:
### Discussed in https://github.com/neovide/neovide/discussions/2668

<div type='discussions-op-text'>

<sup>Originally posted by **Geobert** June 25, 2024</sup>
Hi, 

I’m beginning my journey with nvim through neovide, on Windows (for now).

I’m using `kickstart` and it has `<C-Space>` mapped to trigger `cmp` but it doesn’t work.
I added another map to test the command and it works, so something is messing around with C-space and I can’t see what :-/
</div>

fredizzimo:
The log shows that `<C-Space>` is received by winit, but for some reason not sent to Neovim

[neovide_rCURRENT.log](https://github.com/user-attachments/files/15987217/neovide_rCURRENT.4.log)

fredizzimo:
I won't be able to fix this myself until earliest the end of next week, but I think the problem is this condition
https://github.com/neovide/neovide/blob/24fcfc94c30775a6ff08277e76d266d2a1bde62c/src/window/keyboard_manager.rs#L301-L308

`None` should also generate `space`, in addition to `" "`, while any other text should not.

Geobert:
I’ll see what I can do as I’m a Rustacean myself :)

PR #2672, opened 2024-06-26
fix: C-Space was not mappable on Windows

Geobert:
Fixes https://github.com/neovide/neovide/issues/2669

- Fix

- No

I had this keymap that wasn’t working before the change and works with it
```
vim.keymap.set('n', '<C-Space>', '<cmd>echomsg "Test!"<CR>')
```

### DIFF
--- a/src/window/keyboard_manager.rs
+++ b/src/window/keyboard_manager.rs
@@ -301,7 +301,7 @@ fn get_special_key(key_event: &KeyEvent) -> Option<&str> {
         NamedKey::Space => {
             // Space can finish a dead key sequence, so treat space as a special key only when
             // that doesn't happen.
-            if key_event.text == Some(" ".into()) || key_event.text == None {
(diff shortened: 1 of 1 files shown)


Geobert:
Not sure what happend here, but I think my change has nothing to do with it 😅

fredizzimo:
Yes, the Window CI builds have been broken for a while, and I think we won't merge that until it's fixed. I made a comment here https://github.com/neovide/neovide/pull/2641#issuecomment-2156021707, but since I'm on vaccation, I haven't been able to do anything about it myself.

There is a problem with your code though, if you check the files tab, `cargo clippy` does not like the `== None` check. That's also the other reason I want to get rid of reviewdog, the output is confusing and the errors hard to spot.

Geobert:
Sorry about that :)

fredizzimo:
The CI has finally been fixed now, can you rebase on top of main?

fredizzimo:
Thank you!

PR #2687, opened 2024-07-08
fix: A smal tweak to how the baseline offset and underline position is calculated

fredizzimo:
* Rename `y_adjustment` to `baseline_offset`
* The `leading/linegap` is now distributed equally on top and bottom, like for the `linespace` setting
* The underline offset is relative to the `baseline_offset`

- No


fredizzimo:
Note, I have not found a single font with a `leading` set, but I believe this way of calculating it makes more sense.

falcucci:
Reproduced and works fine.

PR #2691, opened 2024-07-09
fix: Unset ARGV0 to fix AppImage launch

fredizzimo:
Unset `ARGV0` to fix AppImage

* fixes https://github.com/neovide/neovide/issues/2477

- No


src/main.rs
Kethku, line 1: probles should be problems

PR #2692, opened 2024-07-09
fix: terminal file handler and UI command to use tabs based on settings 

falcucci:
- Fix

- No


alexventuraio:
I'm not sure if this is something on my NeoVim config but I still have the same issue when opening my project folder with `neovide .` even in the lates `0.13.2` version in MacOS Sonoma 14.5 (23F79)

Issue #2613, opened 2024-05-31
System clipboard not work when using with Astronvim


Crispy13:
No "+", "*" option in register with neovide.

fredizzimo:
Which platform, and do you have the clipboard provider tool installed https://neovim.io/doc/user/provider.html#clipboard?

You should be able to verify with `:checkhealth`

Crispy13:
Thanks for response

I use neovide with the following:
Astonvim : running on remote centos server with `--headless` 
neovide: running on the windows client

both are x86_64.

What I've found was
1. `set clipboard=""` with https://docs.astronvim.com/recipes/neovide/  and `set clipboard+=unnamedplus` after entering the neovim server makes clipboard work. `set clipboard=""` was enough to make "+","*" registers work. but `y` content not transferred to "+" "*".
2. When not doing 1, `:checkhealth` showed clipboard provider is "neovide".

fredizzimo:
Hm, `unnamedplus` should be the default in astronvim  https://github.com/AstroNvim/AstroNvim/blob/e1e08e29d2df6a2be92a012edba52a1b7f7eaa05/lua/astronvim/plugins/_astrocore_options.lua#L8

Are you saying that isn't the case? Or is there some more problems even after it's set. Note that the Neovide clipboard provider only synchronizes the Windows clipboard, not the one running on the server. If you want to use that one, you can set `vim.g.clipboard=false`

I can see a potential problem with line endings, since the clipboard runs on the Window side, and the server on the linux side. That's something we don't take into account right now.

fredizzimo:
`--headless` is likely the problem here, try running with just `--listen`. Headless does not wait for the UI (Neovide) to connect before running the init files, while just `--listen` does.

Crispy13:
Thanks for response.

I commented the lines of `set clipboard=""` and `set clipboard+=unnamedplus`, then launch nvim with `nvim --listen 0.0.0.0:12345` but clipboard did not work. No +, * registry.

Also without `--headless` nvim instance ran in the remote server and UI of neovide on my windows pc broken after some works.

register window was not closed.

fredizzimo:
I checked the remote clipboard on a Windows machine. It works, but it looks like there are some performance issues. I'm not sure if that is related to your problem or not, probably not.

[...]

PR #2696, opened 2024-07-10
fix: Fix the remote clipboard support

fredizzimo:
* The clipboard provider is now properly reloaded after setting up the clipboard. 
* This also adds support for the selection register `*` on Linux

* Fixes https://github.com/neovide/neovide/issues/2613

- No

### DIFF
--- a/src/clipboard.rs
+++ b/src/clipboard.rs
@@ -2,7 +2,10 @@ use std::error::Error;
 use std::sync::OnceLock;
 
 #[cfg(target_os = "linux")]
-use copypasta::wayland_clipboard;
+use copypasta::{
+    wayland_clipboard,
+    x11_clipboard::{Primary as X11SelectionClipboard, X11ClipboardContext},
+};
 use copypasta::{ClipboardContext, ClipboardProvider};
 use parking_lot::Mutex;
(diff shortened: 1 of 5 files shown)


src/clipboard.rs
```diff
@@ -14,24 +17,60 @@ use crate::window::UserEvent;
...
+
+static CLIPBOARD: OnceLock<Mutex<Clipboard>> = OnceLock::new();
 
 pub fn init(event_loop: &EventLoop<UserEvent>) {
     CLIPBOARD
         .set(Mutex::new(match event_loop.raw_display_handle() {
             #[cfg(target_os = "linux")]
             RawDisplayHandle::Wayland(WaylandDisplayHandle { display, .. }) => unsafe {
-                Box::new(wayland_clipboard::create_clipboards_from_external(display).1)
+                let clipboards = wayland_clipboard::create_clipboards_from_external(display);
```
Kethku, line 33: nit: could we destructure clipboards into more reasonable names rather than using clipboards.0 and clipboards.1 especially since they are passed in reverse order?

Kethku:
minor comment. take it or leave it. merge when ready

falcucci (approved):

PR #2698, opened 2024-07-11
chore: Update to winit 0.30.3

fredizzimo:
NOTE: Includes the 0.13.2 version bump, so don't merge yet until the release is out.

Also 
* Glutin 0.32,
* Glutin-winit 0.5.0 
* raw-window-handle 0.6.0

- No

### DIFF
--- a/src/window/update_loop.rs
+++ b/src/window/update_loop.rs
@@ -6,6 +6,9 @@ use winit::{
     event_loop::{ActiveEventLoop, ControlFlow, EventLoopProxy},
 };
 
+#[cfg(target_os = "macos")]
+use icrate::Foundation::MainThreadMarker;
+
 use super::{save_window_size, CmdLineSettings, UserEvent, WindowSettings, WinitWindowWrapper};
 use crate::{
     profiling::{tracy_plot, tracy_zone},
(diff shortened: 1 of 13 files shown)


src/window/update_loop.rs
```diff
@@ -269,61 +318,89 @@ impl UpdateLoop {
...
+    fn resumed(&mut self, event_loop: &ActiveEventLoop) {
+        tracy_zone!("resumed");
+        self.create_window_allowed = true;
+        self.schedule_next_event(event_loop);
+    }
 
-        ControlFlow::WaitUntil(self.get_event_deadline())
+    fn exiting(&mut self, event_loop: &ActiveEventLoop) {
+        tracy_zone!("exiting");
+        self.window_wrapper.exit();
```
fredizzimo, line 403: I don't like this this, but the window and skia has to be destroyed before the event loop finishes to avoid this crash:
* https://github.com/neovide/neovide/issues/2223

It would be nice if the event loop took ovnership of the `UpdateLoop` and destroyed it automatically. But unfortunately `event_loop.run_app` takes a reference.

fredizzimo:
Switching this to draft, since we probably want to make yet another 0.13 release with some font fixes as reported here 
* https://github.com/neovide/neovide/issues/2707

Additionally, the glamour update should be merged first, to follow the same order I'm updating the Vide integration.
* https://github.com/neovide/neovide/pull/2708

falcucci (approved):
No issues at all for me, for now.

PR #2700, opened 2024-07-12
feat: update deployment scripts for macOS .dmg creation

falcucci:
- Feature

- No

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-latest, macOS-12, ubuntu-latest, macOS-14]
+        os: [windows-latest, macOS-12, ubuntu-latest]
         toolchain: [stable, nightly]
     runs-on: ${{ matrix.os }}
 
@@ -86,7 +86,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-latest, macOS-12, ubuntu-latest, macOS-14]
+        os: [windows-latest, macOS-12, ubuntu-latest]
         toolchain: [stable, nightly]
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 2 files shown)


fredizzimo:
Thank you!

9mm:
hey dont forget about the `rm` command that uses a env variable, might want to audit that to make sure its good. i dont think its actually used in the build script (its in run/make-icns script) just dont want any future issues emanating from code i originated

falcucci:
@9mm I am not sure if I get it, would be better if you point out in the code where it would be.

I don't see any need to run a `rm` in the CI workflow if that's what you mean, but from running it locally maybe it would make sense for testing.

9mm:
I made a commit that added 2 files to macos-build folder, one of which you were editing in this PR:

https://github.com/neovide/neovide/tree/main/macos-builder

The other file, make-icns is used to generate proper retina backgrounds etc, but it includes an `rm` command with a path environemnt variable in it. I mentioned it before that someone should audit that as im not an "idiomatic" bash expert.

whether that file is used in the build process or not is up to you, but i suspect it doesnt need to be unless those images are edited (so just local dev stuff), i still wanted to point it out though as @fredizzimo mentioned usually you dont want env variables in rm commands, but then also said maybe its ok because the variable is in the same file

fredizzimo:
It should be fine in this case, it's more like a local variable of the script, and there are no dependencies to outside environment variables.

The script also sets `-set e`, so if the path can't be created, for example if it exists, then it will exit with an error, so there's no risk of deleting existing directories.

ThomChroma:
Since 0.13.2 the Linux tar.gz files are also zipped. I don't know how the deployment scripts works but could one of these commits be the cause? Or has something else changed in $GITHUB_ENV?

fredizzimo:
@ThomChroma, sorry about that. Our release process is manual at the moment, and I simply made a mistake by not unpacking the zip files. The build artifacts are always zipped, so they need manual uncompress. 

I have now uploaded the uncompressed release for 0.13.3, but left the compressed one in case someone is relying on it.

PR #2708, opened 2024-07-14
chore: Switch to glamour instead of euclid

fredizzimo:
Use Glamour instead of Euclid for Vide compatibility

- No

### DIFF
--- a/src/renderer/rendered_window.rs
+++ b/src/renderer/rendered_window.rs
@@ -141,9 +141,9 @@ impl RenderedWindow {
             scroll_delta: 0,
             viewport_margins: ViewportMargins { top: 0, bottom: 0 },
 
-            grid_start_position: grid_position.cast(),
-            grid_current_position: grid_position.cast(),
-            grid_destination: grid_position.cast(),
+            grid_start_position: grid_position.try_cast().unwrap(),
+            grid_current_position: grid_position.try_cast().unwrap(),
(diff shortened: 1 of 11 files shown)


src/renderer/rendered_window.rs
```diff
@@ -146,7 +148,7 @@ impl RenderedWindow {
             return destination;
         }
 
-        let mut grid_size: GridSize<f32> = self.grid_size.cast();
+        let mut grid_size: GridSize<f32> = self.grid_size.try_cast().unwrap();
```
Kethku, line 151: is there no cast function that panics? this pattern seems odd

Kethku:
lgtm aside from the slightly odd try_cast().unwrap() pattern. merge when ready

src/renderer/rendered_window.rs
fredizzimo: To my it looks pretty normal, it uses this under the hood https://docs.rs/num/latest/num/trait.NumCast.html#. The built in `TryInto/TryFrom` are also similar, but they return a result instead.

There is a non-failing cast for converting just the units, https://docs.rs/glamour/latest/glamour/struct.Point2.html#method.cast, but it does not apply here, since we are converting from int to float.

There's also `into` which can convert between vector and point for example.

As far as I know there are no non-failing cast, similar to `as` unless you convert it to something else first.

PR #2710, opened 2024-07-14
fix: use pre-defined font weights only if user does not specify one

### LINKED ISSUE #2707: Font style config for normal style doesn't work in 0.13.2(was working in 0.13.1)
**Describe the bug**
Font style config for normal font in the `config.toml` file doesn't work in 0.13.2. My config file looks like this:

```
fork = true

[font]
size = 16
hinting = "full"
edging = "antialias"

[[font.normal]]
family = "FiraCode Nerd Font Mono"
style = "Medium"
```

But in Neovide the normal text's font weight is always `Normal`, changing the style config to other values(like `Bold`) also won't work.

**To Reproduce**
1. Set normal font's style to a value other than `Normal`
2. Open neovide

**Expected behavior**
Normal text in Neovide should apply font style defined in the config file

**Screenshots**
Below is a comparasion of what it looks like in 0.13.1(left) and 0.13.2(right):

 - OS: WIndows 11
 - Neovide Version: 0.13.2
 - Neovim Version: 0.10.0

[neovide_r2024-07-14_15-56-32.log](https://github.com/user-attachments/files/16216179/neovide_r2024-07-14_15-56-32.log)

**Additional context**
It looks like the issue is related to [this line](https://github.com/neovide/neovide/blob/f56145238570e9e3134a8d7d43144690661f62d0/src/renderer/fonts/font_options.rs#L220C1-L220C57). If no other styles are applied, the style is set to `None` istead of the one defined by the normal font style.

xzbdmw:
- Fix #2707 
This reverts some change from my previous pr, that was too naive :smile:
- No

### DIFF
--- a/src/renderer/fonts/font_options.rs
+++ b/src/renderer/fonts/font_options.rs
@@ -192,6 +192,10 @@ impl FontOptions {
     }
 
     pub fn font_list(&self, style: CoarseStyle) -> Vec<FontDescription> {
+        if style == CoarseStyle::default() {
+            return self.normal.clone();
+        }
+
         let fonts = match (style.bold, style.italic) {
             (true, true) => &self.bold_italic,
             (true, false) => &self.bold,
@@ -199,7 +203,13 @@ impl FontOptions {
             (false, false) => &None,
         };
 
-        let fonts = fonts
+        let normal_fallback = self.normal.iter().map(|font| FontDescription {
+            // use current requested font style instead of normal
+            style: style.name().map(str::to_string),
+            family: font.family.clone(),
+        });
+
+        fonts
             .as_ref()
             .map(|fonts| {
                 fonts
@@ -207,20 +217,12 @@ impl FontOptions {
                     .filter(|font| font.family.is_some())
                     .map(|font| FontDescription {
                         family: font.family.clone().unwrap(),
-                        style: None,
+                        style: font.style.clone(),
                     })
-                    .chain(self.normal.iter().cloned())
+                    .chain(normal_fallback.clone())
                     .collect()
             })
-            .unwrap_or_else(|| self.normal.clone());
-
-        fonts
-            .into_iter()
-            .map(|font| FontDescription {
-                style: style.name().map(str::to_string),
-                ..font
-            })
-            .collect()
+            .unwrap_or_else(|| normal_fallback.collect())
     }
 
     pub fn possible_fonts(&self) -> Vec<FontDescription> {
--- a/src/renderer/fonts/font_options.rs
+++ b/src/renderer/fonts/font_options.rs
@@ -217,7 +217,10 @@ impl FontOptions {
                     .filter(|font| font.family.is_some())
                     .map(|font| FontDescription {
                         family: font.family.clone().unwrap(),
-                        style: font.style.clone(),
+                        style: font
+                            .style
+                            .clone()
+                            .or_else(|| style.name().map(str::to_string)),
                     })
                     .chain(normal_fallback.clone())
                     .collect()


fredizzimo:
The main weight appears to be working now, but it looks like the weights for the fallbacks are not correct, consider the following config:
```toml
[font]
size = 10.5
hinting = "full"

[[font.normal]]
family = "FiraCode Nerd Font"
style = "W500"

[[font.normal]]
family = "Noto Sans Mono CJK SC"
style = "Normal"

[[font.italic]]
family = "FiraCode Nerd Font"
style = "Italic"

[[font.bold]]
family = "FiraCode Nerd Font"
style = "Bold"

[[font.bold_italic]]
family = "FiraCode Nerd Font"
style = "Bold Italic"
```

And a markdown file
Normal 测试字体

*Italic 测试字体*

**Bold 测试字体**

In Neovide it displays the Chinese text as normal, but I think it should apply bold and italic.

image


Note that changing the Noto Sans to:

```toml
[[font.normal]]
family = "Noto Sans Mono CJK SC"
style = "Bold"
```
Changes it to bold everywhere. But bold should only apply to the Normal weight


Actually, a config that looks like this without any fallbacks does not work either
```toml
[font]
size = 10.5
hinting = "full"

[[font.normal]]
family = "FiraCode Nerd Font"
style = "Bold"
```

image


That makes all variations bold, but I think the style should only modify the normal style. For example, someone might want to tweak the weight a little bit.

xzbdmw:
I see, I forget why I made change in previous pr, that’s for this situation, I’ll add more test.

fredizzimo:
Another thing that I noticed, and I'm not sure if it's related to your previous changes or not, but I don't remember seeing that before, is that Neovide becomes very slow when you set the primary font to a non-existing one, and it falls back to the default embedded Nerd Font.

xzbdmw:
> And a markdown file
Normal 测试字体

Italic 测试字体

Bold 测试字体

In Neovide it displays the Chinese text as normal, but I think it should apply bold and italic

It seems can't find other Chinese characters no matter what I specify

fredizzimo:
Hm, for me this PR with the following markdown file and config displays this

image
```toml
[font]
size = 10.5
hinting = "full"

[[font.normal]]
family = "FiraCode Nerd Font"
style = "Normal"

[[font.normal]]
family = "Noto Sans Mono CJK SC"
style = "Bold Italic"

[[font.italic]]
family = "FiraCode Nerd Font"
style = "Italic"

[[font.bold]]
family = "FiraCode Nerd Font"
style = "Bold"

[[font.bold_italic]]
family = "FiraCode Nerd Font"
style = "Bold Italic"
```
```markdown
Normal 测试字体

*Italic 测试字体*

**Bold 测试字体**
```

So, at least for me it can display 'Bold Italic`, or any other styles, compare to
Normal
image

Bold:
image

xzbdmw:
@fredizzimo Now these examples should all be correct, and also respect user-specified one.

src/renderer/fonts/font_options.rs
```diff
@@ -192,35 +192,37 @@ impl FontOptions {
     }
 
     pub fn font_list(&self, style: CoarseStyle) -> Vec<FontDescription> {
+        if style == CoarseStyle::default() {
+            return self.normal.clone();
+        }
```
xzbdmw, line 197: The early return looks ok to me, this change makes sure 1)we respect user settings of normal 2)fallback to normal always applys the corresponding font style

xzbdmw:
> That makes all variations bold, but I think the style should only modify the normal style. For example, someone might want to tweak the weight a little bit.

This is also fixed, in 13.1 it's the same behaviour.

fredizzimo:
Thank you! All the tests I can think of work correctly now.

Issue #2609, opened 2024-05-30
Neovide Opening File Opens Same File Twice


lycheeje11y:
When I open a file like "neovide main.rs", the file opens, but it opens the file twice in two different tabs. How do I open the file only once? Am I missing something here?

s4Dt0y:
As a side note, one workaround is to use the add the --no-tabs flag.

fredizzimo:
I can't repeat this here, which platform are you running on?

carlmueller:
It’s happening to me on MacOS Sonoma.

s4Dt0y:
> I can't repeat this here, which platform are you running on?

I'm on mac sonoma as well...the context for this is trying to set up neovim with unity. I started with trying nvr, but gave up after a couple of days. So I instead tried Neovide which def works better out of the box, but opens up the files twice.

carlmueller:
The following workaround works for me.

`open -a neovide filename`

fredizzimo:
Very strange, maybe the commandline parser interprets the file as both a file to open and neovim arguments. Can you try with `neovide main.rs --`, forcing the neovim arguments to be empty?

Although, it should be the same on all platforms.

carlmueller:
`neovide test.txt --` still opens the file twice.

carlmueller:
`neovide test.txt --log` does open the file twice (while `neovide --log test.txt` does not).  I'm attaching the log file.
[neovide_rCURRENT.log](https://github.com/user-attachments/files/15568416/neovide_rCURRENT.log)

fredizzimo:
Here's the stupid design Apple has decided to implement
https://lists.apple.com/archives/cocoa-dev/2009/May/msg00480.html,

TLDR; there's no way to disable the built in parsing of the arguments, so we need to ignore the openfile requests for them, or maybe always only use openfile, and ignore the initial arguments. 

The problem is that I can't find any official documentation of this, and nothing about what the scheme it uses, if everything that looks like a file is passed to `openFile`, or if the file actually has to exist for example. And in both cases we need to be able to filter the list of arguments.

I guess that's something for the one actually fixing the bug to find out...

falcucci:
For ppl who wants this fixed the workaround would be disabling neovide `tabs` to make it appears as Buffer only using
```bash
neovide --no-tabs <file>
```

or even at our [config.toml](https://neovide.dev/config-file.html#available-settings) file.

```toml
tabs = false
```

fredizzimo:
[...]

Issue #2629, opened 2024-06-03
macOS: initial opening via `open -a "Neovide"` does not set arglist


chrisgrieser:
**Describe the bug**
When opening Neovide via `open -a "Neovide" {filename}`, the arglist (as well as related values such as `argv` or `argc`) are not set. 

This means that using `open -a "Neovide"`, there is no way to tell if Neovide was opened with an argument or not. This is relevant for configs, where you want to open a dashboard on startup only when opening Neovide without argument.

**To Reproduce**
1. `open -a "Neovide" {filename}`
2. run `:lua = vim.fn.argc()`. Result is `0`
3. in comparison, `neovide {filename}` results in an `argc` of `1`

**Expected behavior**
arglist should not be empty.

**Screenshots**

**Desktop (please complete the following information):**
```
NVIM v0.10.0
Build type: Release
LuaJIT 2.1.1716656478
macOS 14.4.1 (arm)
neovide 0.13.1
```
[neovide_rCURRENT.log](https://github.com/user-attachments/files/15535559/neovide_rCURRENT.log)

fredizzimo:
I'm not a macOS user, but I think 
`open -a Neovide --args {filename}` should do it. That will pass the arguments to Neovide instead of going through the finder. Or even `open -a Neovide --args -- {filename}`, which passes it directly to neovim.

I'm not sure if it can be fixed for the finder, since it uses the same functionality for drag and drop https://github.com/neovide/neovide/pull/2395/files.

fredizzimo:
When looking at a different issue, I saw this https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428385-applicationdidfinishlaunching.

So, we should be able to wait for that until the list of files is passed to neovim. Or perhaps even delay the launching of neovim until that point, so that we can properly pass the files as a command line.

* See https://github.com/neovide/neovide/issues/2609, as it's probably worth tackling both issues at once.

kandayin:
> open -a Neovide --args -- {filename}

this works! Would you happen to know how to properly pass this as an env var? eg

```bash
export EDITOR="open -a Neovide --args -- " <-- this still gives 2 tabs when $EDITOR is invoked
```

falcucci:
Using `open` as the application launcher, both ways already mentioned by @fredizzimo should work.

The only thing to notice is that on macOS it will always launch taking `/` as the root path for those specific cases, so I would update the suggestion adding the `$PWD` for neovide context:

```bash
[...]

PR #2715, opened 2024-07-17
Don't re-open files already present in args

singiamtel:
Fixes https://github.com/neovide/neovide/issues/2609.

My fix is quite ugly, I understand if this is not the path the maintainers want to take to fix this. I'm happy to implement better suggestions, I just want the bug gone

I think the regression was in fcf5e87b 

- Fix

- No

### DIFF
--- a/src/window/macos.rs
+++ b/src/window/macos.rs
@@ -426,7 +426,13 @@ pub fn register_file_handler() {
         autoreleasepool(|pool| {
             for file in files.iter() {
                 let path = file.as_str(pool).to_owned();
-                send_ui(ParallelCommand::FileDrop(path));
+                if !SETTINGS
+                    .get::<CmdLineSettings>()
+                    .neovim_args
+                    .contains(&path)
+                {
+                    send_ui(ParallelCommand::FileDrop(path));
+                }
             }
         });
     }


fredizzimo:
Thank you!

This solves part of the problem. But there's another related problem:
* https://github.com/neovide/neovide/issues/2629

It would be nice if we could delay the launch of Neovim until all the initial `handle_open_files` have been called, then merge the `neovim_args` and send the whole list as arguments to Neovim. If I recall correctly, there's another event that is sent, when all the initial `handle_open_files` have been handled. 

But note that this should only be done for the initial launch, `handle_open_files` should still be sent as `FileDrop` after Neovim has been launched.

Since, this is quite complex, I understand if you can't or don't want to implement this. But I would rather merge the complete solution, than a workaround which will have to be replaced later.

fredizzimo:
@falcucci, could you test this at some point? Since I can't do it myself.

singiamtel:
Hi @fredizzimo,

I've been trying to understand the events better but haven't made much progress. I think the bug is a race condition, adding this print is enough to break the repro in my local machine...

```diff
diff --git a/src/window/mod.rs b/src/window/mod.rs
index 663d5d3..13932a7 100644
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -290,6 +290,7 @@ pub fn main_loop(
     let res = event_loop.run(move |e, window_target| {
         #[cfg(target_os = "macos")]
         menu.ensure_menu_added(&e);
+        println!("{:?}", e);
 
         match e {
             Event::Resumed => {
```

So the proper fix is probably too much for me, specially as this is my first time working on this codebase. 

I'll fix the format error on the CI, feel free to either merge the workaround or close the PR

fredizzimo:
I think this can be closed, we merged this
* https://github.com/neovide/neovide/pull/2798

PR #2716, opened 2024-07-17
docs: consolidate macOS build instructions and scripts

falcucci:
- Codestyle
- Refactor
- Documentation

- No

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -189,10 +189,10 @@ jobs:
           elif [[ $RUNNER_OS == "macOS" ]]; then
             cd $GITHUB_WORKSPACE
             # create .dmg for x86_64-apple-darwin
-            ./macos-builder/run x86_64-apple-darwin
+            TARGET_ARCHITECTURE=x86_64-apple-darwin ./macos-builder/run
 
             # create .dmg for aarch64-apple-darwin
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -189,10 +189,10 @@ jobs:
           elif [[ $RUNNER_OS == "macOS" ]]; then
             cd $GITHUB_WORKSPACE
             # create .dmg for x86_64-apple-darwin
-            ./macos-builder/run x86_64-apple-darwin
+            TARGET_ARCHITECTURE=x86_64-apple-darwin ./macos-builder/run
```
fredizzimo, line 192: I wonder if we should leave this as before. Someone might already rely on these scripts, and it's an annoying breaking change. 

Or at least it will be if for example homebrew update to use the scripts from release 0.13.3, then they need to update them again when the next Neovide version is released.

falcucci: I'll check what I can do, maybe bring compatibility for both is also good.

falcucci: maybe it's simpler the other way so I'll probably make a rollback of how we call it

falcucci: done

fredizzimo:
Thanks, I think this is good now

PR #2719, opened 2024-07-19
Initial Vide support

fredizzimo:
**NOTE:** This targets the wgpu branch

This integrates the Vide WGPU based renderer, https://github.com/neovide/vide.

The basics work, it renders texts and backgrounds, but several things still need to be done. And that will be done in follow up PRs. Here are a few things missing:
* The font is hardcoded, and no user configuration can be done
* The cursor is not rendered at all
* The built-in profiler is not rendered
* Vsync is not fully working as it should
* The performance is very bad, mostly due to the fact that the text shaping is done every frame
* Blur and shadows are not implemented
* The transparency probably does not work
* The error window is disabled, since the rendering is not yet ported.

- Probably yes, but the target is the wgpu branch, so it will not break the main Neovide build


fredizzimo:
I'm going to merge this to make it easier to continue my work. It can be more properly reviewed later when we merge into main.

PR #2734, opened 2024-07-27
fix(wgpu)!: Load and use the fonts from the config instead of hardcoded ones

fredizzimo:
Load and use the fonts from the config instead of hardcoded ones

- Yes, the linespacing is now multiplied by the scale factor as it should be.


Kethku:
lgtm!

fredizzimo:
I'm merging this, since the CI failure is unrelated.

PR #2743, opened 2024-07-31
feat: Rounded floating windows (#1993)

### LINKED ISSUE #1993: Rounded floating windows (or maybe even full transparency masks)
### Is your feature request related to a problem? Please describe.

Floating windows are cool by itself. But they can get even cooler with borders, for example as with the border options of [nvim-cmp](https://github.com/hrsh7th/nvim-cmp).

However, with the text borders enabled, suddenly the rectangular aesthetic starts being more of an issue, as it's seen in the corners of the floating windows here:

Two overlaid floating and blurred windows.

### Describe the solution you'd like

The introduction of a new configuration option `g:neovide_floating_corner_radius`, which allows setting a radius in pixels. This is the radius of a rounded rectangle, which should represent the floating window. Maybe the corners could even be configurable individually.

Technically this could be implemented by rendering a rounded rectangle with the same size as the floating window in fully opaque white on a fully transparent surface, which creates a mask. Before compositing the windows/after finishing the render of the floating window itself, this mask is multiplied with the fully rendered image.

**Bonus:** On top of the rounding itself, it'd be insanely cool to detect if a window was bordered, and in such case configurably subtract the size of the floating window. For example, in the screenshot above it's very visible that the space _above_/_below_ the window is visibly larger than the space _left-hand_/_right-hand_ to the window, which stems from the borders being casual text characters. This subtraction can't be applied to all windows though, as other floating windows may not have text borders.

### Describe alternatives you've considered

~~An API that allows marking bordered windows directly through Neovim and making the GUI render the borders all on its own instead. This would require adding such an option on the Neovim side first though, most likely directly through the `{*config}` parameter on `nvim_open_win`/`nvim_win_set_config`, see the `:help` pages of both.~~ turns out this _is_ already offered through the `border` option on on `nvim_open_win`/`nvim_win_set_config`

jheroy:
- Feature

- No

Fix #1993


jheroy:
Some snapshot with `vim.g.neovide_floating_corner_radius = 10.0`
nvim-cmp
image
telescope:
image

Lsp document:
image

idelice:
@jheroy unrelated to your PR, but how did you achieve transparent autocompletion window, telescope etc.? all mine are black while the editor itself is transparent.

jheroy:
> @jheroy unrelated to your PR, but how did you achieve transparent autocompletion window, telescope etc.? all mine are black while the editor itself is transparent.

https://neovide.dev/faq.html?highlight=winblend#how-to-enable-floating-and-popupmenu-transparency

fredizzimo:
How does this look like with different border styles? I guess the radius can be turned to not clip the borders when using different font sizes.

src/renderer/rendered_layer.rs
```diff
@@ -211,3 +211,9 @@ fn build_silhouette(regions: &[PixelRect<f32>]) -> (Path, Rect) {
 
     (silhouette, bounding_rect)
 }
+
+fn rect_to_round_rect_path(rect: &Rect, settings: &RendererSettings) -> Path {
+    let mut path = Path::new();
+    path.add_round_rect(*rect, (settings.floating_corner_radius, settings.floating_corner_radius), None);
```
fredizzimo, line 217: I think we at least need to take the scale factor into account here, otherwise the corners will be rounded differently when dragging the window between screens of different scale factors.

But perhaps it should even be a percentage of the line height, so that it can deal with changing the font size properly?

jheroy: Yes, I can change the code to take into account the scale factor, but a percentage of the line height should be a better solution. I haven't written a single line of rust code before, though, so I'm not quite sure how to get the current line height correctly.

fredizzimo: The `grid_renderer::grid_scale` contains the current line height, already multiplied by the scale factor https://github.com/neovide/neovide/blob/000b75c0f3bcfb602967cc5c2e8a9d584be06a02/src/renderer/grid_renderer.rs#L22

And `PixelSize<f32>::new(1.0, 1.0) * floating_corner_radius * grid_scale` should give you a pixel size as a fraction of the grid size when `floating_corner_radius` is less than one.  This gives different values for the X and Y axis though, so it might be better to just use the height.

src/renderer/rendered_layer.rs
```diff
@@ -197,10 +197,10 @@ pub fn group_windows(
         .collect_vec()
 }
 
-fn build_silhouette(regions: &[PixelRect<f32>]) -> (Path, Rect) {
+fn build_silhouette(regions: &[PixelRect<f32>], settings: &RendererSettings, grid_scale: GridScale) -> (Path, Rect) {
     let silhouette = regions
         .iter()
-        .map(|r| Path::rect(to_skia_rect(r), None))
+        .map(|r| rect_to_round_rect_path(&to_skia_rect(r), settings, grid_scale))
```
fredizzimo, line 203: I'm worried about this change when there are multiple floating windows together, like in some of the telescope layouts for example. This could leave gaps in the middle of the silhouette. 

So, I think the rounding only should be done on the outside.

fredizzimo (commented):
This also needs some documentation.

src/renderer/rendered_layer.rs
jheroy: But I need it to work this way, because I like how different parts of the telescope look like separate areas, like this:
image

Maybe a configuration option could be added to control this effect.

jheroy: Yeah, I submitted new commit to fix this issue.

src/renderer/rendered_layer.rs
jheroy: But there’s another problem here: when a floating window has two rectangles that are nested, two shadows appear(eg. `folke/noice.nvim`'s cmdline popup):
image

However, this should be a new issue.

src/renderer/rendered_layer.rs
jheroy: But there’s another problem here: when a floating window has two rectangles that are nested, two shadows appear(eg. `folke/noice.nvim`'s cmdline popup):
image

However, this should be a new issue.

Kethku: I believe this is a known issue. Theres no standard way currently to tell that two floating windows should exist at the same z level. On most guis this is rendered flat so its not a problem. We don't have that luxury

fredizzimo: I just tested it, that does not happen to me with the latest version of the plugin.

fredizzimo:
Thank you! I just tested this, and it works as expected.

griwes:
I will try to do a minimal repro later if necessary, but for me this PR broke the case of multiple floating windows sitting one next to the other, where they don't form a perfect rectangle - at least with `winblend = 66` across the board, in the parts of the geometry that are missing to create a perfect triangle, I now get a black, uhh, "fill"? Other factors I imagine may have impact here is that I use a transparent primary bg color, `neovide_transparency` is 85; this also happens with the new setting being left at 0.

fredizzimo:
My bad, I was reviewing this too quickly today. It does indeed look like it only considers the bounding rect, and not the full silhouette. 

The previous implementation, as commented here https://github.com/neovide/neovide/pull/2743#discussion_r1715931946 had another problem, it considered each window individually and not the whole z-index.

PR #2755, opened 2024-08-10
chore(wgpu): Update tracy

fredizzimo:
* Update tracy to the latest version, so that it's compatible with Vide
* Fix the `--no-vsync` mode
* Remove a duplicated `tracy_frame` event (vide takes care of that)
* Compile some external dependencies including vide and wgpu with optimizations even in debug mode. The Debug mode is still quite unusable, but that makes it a lot better already.

- No

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -652,9 +652,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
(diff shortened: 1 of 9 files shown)


Kethku:
lgtm

PR #2765, opened 2024-08-13
fix: Force Neovide to exit, even if the IO streams are not closed

fredizzimo:
* fixes https://github.com/neovide/neovide/issues/2182 (but probably not completely)

Neovim through libuv spawns child processes that inherits all the handles. This means that the stdio and stderr handles are not properly closed, so the nvim-rs read will hang forever, waiting for more data to read. To deal with that the Tokio runtime is forced to shutdown after a timeout.

The timeout for waiting for the streams to close has also been reduced to avoid having a too long shutdown delay.

Also see these libuv issues:
* https://github.com/libuv/libuv/pull/3856
* https://github.com/libuv/libuv/issues/1490

- No

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -36,7 +36,7 @@ pub use ui_commands::{send_ui, start_ui_command_handler, ParallelCommand, Serial
 const NEOVIM_REQUIRED_VERSION: &str = "0.10.0";
 
 pub struct NeovimRuntime {
-    runtime: Runtime,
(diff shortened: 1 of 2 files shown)


fredizzimo:
@sid-6581. I'm sorry, I don't think this fixes the issue you are experiencing, since this PR assumes that we can properly detect that Neovim itself quits before forcing a shutdown.

falcucci (approved):

PR #2768, opened 2024-08-20
chore: update objc2 and remove icrate. also some refactor.

crupest:
This pr does following 2 major things:

- Upgrade `objc2` and remove `icrate`, since `icrate` is deprecated and migrated to `objc2-*`. Related structs and  functions are updated to the new ones.
- Try to move as much as possible macOS specific code into `macos.rs`.
  - Menu creation.
  - File handler.

MacOS part improvement
- Rename `vsync_macos` -> `vsync_macos_display_link`. So we can create another one like Metal vsync on macOS, just like how it is done on Windows.
- Getting `NSWindow` from winit `Window` logic is extracted out into a function, because it is used both in `macos.rs` and `macos_display_link.rs`.
- Main thread related improvement
  - Improve `MainThreadMarker` creation. It is now only created and checked after entering macOS-specific code from common codes, instead of being created in common code and passed to macOS-specific code. It will then be used until getting out from macOS-specific code.
  - Objective-C classes `TitlebarClickHandler` and `QuitHandler` declared in rust are now marked as `mutability::MainThreadOnly` as it should be. 
  - Titlebar height is now calculated during `MacosWindowFeature` initialization. This makes it more safe because it can be ensured to run on main thread.
- `unsafe` is added to function `handle_open_files`, which will be passed to Objective-C. Otherwise, there is an error when compiled on nightly rust.
- `Id` -> `Retained`, `NSString::from_str` -> `ns_string!`, new Mask style is used.
- `pub` in some places has been modified. Redundant `unsafe` is removed. `use` statements are moved to the top of the file.

Now there are only following border api between `MacosWindowFeature` and common codes:
- `from_winit_window`: creation of MacosWindowFeature.
- `handle_scale_factor_update`: update scale factor and re-calculate titlebar height.
- `handle_size_changed`: listen to fullscreen-toggled event and show/hide titlebar click handler.
- `extra_titlebar_height_in_pixels`: calculate proper top-padding.
- `handle_settings_changed`: settings change event handler.
- `ensure_menu_added`: create menus in WindowEvent::Focused, though I'm not sure if it is necessary to create it there rather than create at window creation. But I prefer to leaving it there.

- Dependency update.
- Codestyle
- Refactor

No

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -268,31 +268,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0d8c1fef690941d3e7788d328517591fecc684c084084702d6ff1641e993699a"
 
 [[package]]
-name = "block-sys"
-version = "0.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ae85a0696e7ea3b835a453750bf002770776609115e6d25c6d2ff28a8200f7e7"
-dependencies = [
- "objc-sys",
-]
-
-[[package]]
-name = "block2"
-version = "0.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "15b55663a85f33501257357e6421bb33e769d5c9ffb5ba0921c975a123e35e68"
-dependencies = [
- "block-sys",
- "objc2 0.4.1",
-]
-
-[[package]]
 name = "block2"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2c132eebf10f5cad5289222520a4a058514204aed6d791f1cf4fe8088b82d15f"
 dependencies = [
- "objc2 0.5.2",
+ "objc2",
 ]
 
 [[package]]
@@ -1008,7 +989,7 @@ dependencies = [
  "glutin_glx_sys",
  "glutin_wgl_sys",
  "libloading",
- "objc2 0.5.2",
+ "objc2",
  "objc2-app-kit",
  "objc2-foundation",
  "once_cell",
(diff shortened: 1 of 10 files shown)


fredizzimo:
I think the code changes are good. 

@falcucci, could you test that this works for you as well.

@crupest, do you know if this fixes the build failure on the nightly toolchain, as can be seen here:
* https://github.com/neovide/neovide/actions/runs/10498992865/job/29141264588?pr=2769

crupest:
@fredizzimo Thanks for review.

> @falcucci, could you test that this works for you as well.

Also thanks very much. Whatever problem there is, please report it to me and I'll try to fix it.

> @crupest, do you know if this fixes the build failure on the nightly toolchain, as can be seen here:
> https://github.com/neovide/neovide/actions/runs/10498992865/job/29141264588?pr=2769

I think yes. The problem is lack of `unsafe` of function `handle_open_files` at [about line 415 of `macos.rs`](https://github.com/neovide/neovide/blob/000b75c0f3bcfb602967cc5c2e8a9d584be06a02/src/window/macos.rs#L415). This function is declared as a C function and passed as a callback to objective-c. Stable version of Rust is happy about declaring it as non-unsafe, but maybe nightly version uses a stricter type check and treats it as an error.

Actually I noticed [the build failure](https://github.com/neovide/neovide/actions/runs/10498992865/job/29141264588?pr=2769) in my first commit of this pr, which is exactly the same problem. Then I looked into it and found the cause of this problem. I referred to [the usage example in docs of library `objc2`](https://docs.rs/objc2/latest/objc2/runtime/struct.ClassBuilder.html#example) and noticed the `unsafe` of function `init_with_number`. Looks like `unsafe` is necessary is if parameter `this` is declared as a mutable reference, but I didn't try to learn why. I think maybe original author of this piece of code in neovide missed it by accident.

falcucci:
> Rename vsync_macos -> vsync_macos_display_link. So we can create another one like Metal vsync on macOS, just like how it is done on Windows.

This is far more organized at least so LGTM but the Metal support would be covered by [vide](https://github.com/neovide/vide) framework. (whispering it here just to make you aware since you once started working on it in the past)

falcucci (approved):

falcucci:
I've tested the features included in these changes and it works fine.

crupest:
> > Rename vsync_macos -> vsync_macos_display_link. So we can create another one like Metal vsync on macOS, just like how it is done on Windows.
> 
> This is far more organized at least so LGTM but the Metal support would be covered by [vide](https://github.com/neovide/vide) framework. (whispering it here just to make you aware since you once started working on it in the past)

I think it is ok even if I or someone does not finish Metal renderer before vide starts to be used. This is just a preparation, and I don’t mind if it is not used in the end. Thanks for your reminding. I appreciate every work that makes neovide better.

PR #2798, opened 2024-09-20
disable AppKit's command line processing

### LINKED ISSUE #2609: Neovide Opening File Opens Same File Twice

fboundp:
By default, AppKit will helpfully collect any command line arguments that do not start with a dash and pass them to `application:openFiles:`. We do our own non-trivial command line parsing for our arguments and for Neovim, so we must disable this default behavior. (Otherwise, command lines like `neovide -- -c 'color blue'` will result in a modified buffer named "color blue".)

- Fix: Fixes #2609 https://github.com/neovide/neovide/issues/2629

- No

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -107,7 +107,7 @@ skia-safe = { version = "0.75.0", features = ["gl", "textlayout"] }
 
 [target.'cfg(target_os = "macos")'.dependencies]
 objc2 = "0.5.2"
-objc2-foundation = { version = "0.2.2" }
+objc2-foundation = { version = "0.2.2", features = [ "NSUserDefaults" ] }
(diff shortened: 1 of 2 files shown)


9mm:
Dude awesome... going to test now

9mm:
YES!!! This works so well. this was driving me insane. THANK YOU @fboundp !!

crupest:
What an amazing fix!

skiss9:
Ran into this also, spent an hour or so thinking something was wrong in my nvim config before finding this thread. Happy to see it's fixed!

PR #2799, opened 2024-09-20
on macOS, activate when launched

### LINKED ISSUE #2330: neovide does not focus on launch from terminal (macOS, iTerm2)

fboundp:
Generally, it is considered bad form for an application to un-cooperatively take the focus. However, for applications that are frequently launched from the terminal, they must strongly request focus because the terminal, in its ignorance, will not automatically yield.

- Fix: Fixes #2330 

- No

### DIFF
--- a/src/window/macos.rs
+++ b/src/window/macos.rs
@@ -288,10 +288,15 @@ impl MacosWindowFeature {
         }
     }
 
-    pub fn ensure_menu_added(&mut self) {
+    /// Create the application menu and grab initial focus.
+    pub fn ensure_app_initialized(&mut self) {
         let mtm = MainThreadMarker::new().expect("Menu must be created on the main thread");
         if self.menu.is_none() {
             self.menu = Some(Menu::new(mtm));
+            let app = NSApplication::sharedApplication(mtm);
+
+            #[allow(deprecated)]
+            app.activateIgnoringOtherApps(true)
         }
     }
 }
--- a/src/window/update_loop.rs
+++ b/src/window/update_loop.rs
@@ -337,7 +337,7 @@ impl ApplicationHandler<UserEvent> for UpdateLoop {
                     .macos_feature
                     .as_mut()
                     .expect("MacosWindowFeature should already be created here.")
-                    .ensure_menu_added();
+                    .ensure_app_initialized();
             }
             _ => {}
         }


falcucci (changes requested):
Unfortunately this didn't work to me, at macOS Sonoma.

I'll dig into it to help to fix this once there are a couple of people willing this fix.

*The focus issue only happens when launched exclusively from bundled neovide tho. It means that if you use the common build binary you are safe from this issue.*

fboundp:
> Unfortunately this didn't work to me, at macOS Sonoma.

Whoops, that’s because while teasing apart my local branch into PRs I cherry picked an earlier version. 🤦‍♂️
Force pushed a mixup.

falcucci (approved):
@fboundp thank you, that works great now!

fredizzimo:
There's something strange going on here if this fixes the issue. 

Winit is supposed to set `with_activate_ignoring_other_apps` to true by default
https://docs.rs/winit/latest/x86_64-apple-darwin/winit/platform/macos/trait.EventLoopBuilderExtMacOS.html#tymethod.with_activate_ignoring_other_apps

Which does this https://github.com/rust-windowing/winit/blob/114512bb329ac2da78e33dcd54c3f466f34da394/src/platform_impl/macos/app_state.rs#L118

As far as I can see, that does the exact same thing as this PR. And it would be much better if we could use Winit for this to reduce the maintainance burden.

Therefore, I think it would be worth investigating why it's not working, and possibly create a Winit bug report.

falcucci:
@fredizzimo that was indeed weird but once this issue started to bother a lot, I guess that getting this fixed as soon as possible, was better. I have done some tests using the same method `with_activate_ignoring_other_apps` that winit exposes but it didn't work.

falcucci:
I am checking if this is something that works for specific enabled features only.

fredizzimo:
It's probably some initialization order thing, maybe we initialize something at the wrong point, overriding stuff from winit.

fboundp:
Oh yeah, good find @fredizzimo. winit even appears to trigger it in the `applicationDidFinishLaunching:` notification that I couldn’t figure out how to line up with. You are probably onto something about the initialization ordering; it took me a couple of tries to find a spot in Neovide itself where the `activateIgnoringOtherApps` call was effective.

sdfgsdfgd:
hi is there going to be a release for this fix pls

falcucci:
@sdfgsdfgd yes, release is coming very soon.

PR #2800, opened 2024-09-20
add --chdir option

fboundp:
Change the current working directory to the given path during startup. This is only useful in a situation where the caller does not launch neovide directly itself, but does provide the arguments to be used for launching. For example, consider this macOS `open` command invocation:

  open -b com.neovide.neovide --args --chdir ${PWD} --log Makefile

Note that macOS launches applications with `/` as their current working directory. Without specifying `--chdir`, Neovide would attempt to create the log file in `/` and would attempt to open the non-existent file `/Makefile`.

- Feature
- Codestyle
- Refactor
- Documentation
- Other

- No


Kethku:
This needs to be documented here https://github.com/neovide/neovide/blob/main/website/docs/command-line-reference.md before it can be merged.

fboundp:
force pushed after applying rustfmt & clippy suggestions

fboundp:
Let’s merge this or reject this

fredizzimo:
I have some doubts about this. Neovim already provides a cd command which can be used like this `neovide -- --cmd "cd ~/proj`.  And that runs on the target neovim instance, so it works when spawning neovim inside WSL for example. It's also possible to open a directory directly, which will open `netrw` or plugins like `neotree`. This can be done simply by `neovide ~/proj`

So, the feature is already there, although not as ergonomic as it could be in all cases. But maybe it could be included if it was changed to perform the cd command through Neovim instead.

fboundp:
There are a couple of motivations that led me to add this to my local fork.

a) I fronted neovide with a script that attempts to emulate VIM command line processing with high fidelity yet launches neovide through `open`. That requires communicating the current working directory thru the launch of neovide, thus this `--chdir` option. Using `-c`/`--cmd` does not help here. For example:

```
# This opens the file ~/foo.txt in neovim, but the intent was to open /tmp/foo.txt.
cd ~
nvim --cmd 'cd /tmp' foo.txt

# This attempts to open ${PWD}/foo.txt in neovide, but results in opening new file /foo.txt.
open -n -a neovide --args -- --cmd "cd ${PWD}" foo.txt
```

Even if `-c`/`--cmd` could be used to accomplish this, it would still have a couple of shortcomings. First, neovim will only accept up to 10 such arguments, and thus not all command lines can be represented this way. Second, futzing with proper quoting of the path name for the command becomes necessary.

b) The intent is not only to influence neovim’s working directory, but also neovide’s. This is important wherever neovide can reference the current directory, e.g. writing logs with `--log` or relative pathnames in `--server` or `--neovim-bin`.

fredizzimo:
In that case you could spawn a subshell

`(cd /tmp && neovide foo.txt)`, which leave the current shell working directory where it is.

Or since you already use a script, which is already a subshell and therefore won't change the working directory.
```bash
#!/usr/bin/env bash
cd /tmp
neovide foo.txt
```

fboundp:
You may be missing that this is for situations where neovide will be spawned by a process that is not a child of your current process. The canonical (maybe only?) example of this is `open` on macOS.

It is pretty common for other utilities to provide the ability to set the current working directory at startup, as it is useful (simplifying) in many contexts. Some examples: `tar -C`, `make -C`, `git -C`, `emacs --chdir`, `mtree -p`.

fredizzimo:
I think we can accept this, but it has to apply to the started Neovim process instead of Neovide. This mostly applies to the WSL configuration, where Neovim is started on a different file system than Neovide, so the Neovide working directory is useless. We might also want to add support for launching remote Neovim instances, for example through ssh at some point, and the same will apply for them as well.

For Neovide itself, we will change were the `backtraces.log` is written here https://github.com/neovide/neovide/pull/2717, and I think a similar change should be done for the logs. So, after that, Neovide itself will not be dependent on a working directory.

fboundp:
Updated to apply working directory to the nvim process.

fredizzimo:
I will check this tomorrow, I think there's still something that needs to be done to get it to propagate to the WSL process, the launching is quite special and complicated.

falcucci (approved):

fredizzimo:
I finally tested this on windows with WSL. It works ok, but the paths need to refer to the Windows file system, not the WSL one. This is different than opening a file with `neovide --wsl /path/to/file`, which can accept both windows and WSL file system paths.

I think it's ok, to just document that `--chdir` always accepts windows paths, even when `--wsl` is given. The use case for this on other operating systems than macOS is minimal anyway. You can always open a terminal and go to the directory you want and neovide will launch in that directory. On Windows you can furthermore set the working directory in a gui shortcut, and on Linux you can specify the path in the desktop entry.

Or you could simply do `neovide /path/to/where/want/to/be` and it will set the working directory and open netrw or neotree and similar.

falcucci:
@fboundp thank you and sorry for the delay to merge this. 

we can also add the chdir as an option in the config file but it can be done in another pr.

I have also tested this on macOS and works great.

Issue #2807, opened 2024-09-22
There's no error if trying to specify font style overrides without a family


fredizzimo:
**Describe the bug**
This was originally reported here by @yavorski

The `config.toml` fontconfig does not load properly if the family name of the styles is missing

**To Reproduce**
1. Create an invalid font configuration like this:
```toml
[font]
size = 11.25
hinting = "full"
edging = "antialias"
allow_float_size = true

[[font.normal]]
style = "Bold"
family = "Intel One Mono"

[[font.bold]]
style = "Bold"

[[font.italic]]
slant = "Italic"
style = "Bold Italic"

[[font.bold_italic]]
slant = "Italic"
style = "Bold Italic"
````
2. Observe that the configuration is loaded without errors, and the styles are rendered wrong. For example the italic style is not rendered in bold.

**Expected behavior**
An error should be given and the configuration should fail to load. Alternatively, the font configuration should load and the styles apply to the primary font. In that case, it should still give errors, when you have multiple entries, like this, since the configuration is ambiguous. But to me it seems to be better to always give an error for consistency.

```toml
[font]
size = 11.25
hinting = "full"
edging = "antialias"
allow_float_size = true

[[font.normal]]
style = "Bold"
family = "Intel One Mono"

[[font.bold]]
style = "Bold"

[[font.italic]]
style = "Bold Italic"

[[font.bold_italic]]
style = "Bold Italic"

[[font.normal]]
style = "Bold"
family = "Jetbrains Mono NL"

[[font.bold]]
style = "Bold"

[[font.italic]]
style = "Bold Italic"

[[font.bold_italic]]
style = "Bold Italic"
````

Also, notice that no error message is given for invalid keys like `slant = "Italic"`. In this case italics is part of `style` and not a separate key.

**Desktop (please complete the following information):**
 - OS: any
 - Neovide Version 0.11.3
 - Neovim Version: any

**Additional context**
The correct font configuration should look like this
```toml
[font]
size = 11.25
hinting = "full"
edging = "antialias"
allow_float_size = true

[[font.normal]]
style = "Bold"
family = "Intel One Mono"

[[font.bold]]
style = "Bold"
family = "Intel One Mono"

[[font.italic]]
style = "Bold Italic"
family = "Intel One Mono"

[[font.bold_italic]]
style = "Bold Italic"
family = "Intel One Mono"
````

fredizzimo:
@yavorski, can you confirm that it works correctly after fixing your configuration?

[...]

PR #2812, opened 2024-09-23
fix: error if trying to specify font style overrides without a family

xzbdmw:
- Fix
Now user can specify font style alone in config file, that will fallback to use first normal font family. I'm not sure where to throw the error..
Fix https://github.com/neovide/neovide/issues/2807

- Yes, fix the breaking change before?


fredizzimo:
Thank you! It seems to work correctly.

PR #2816, opened 2024-09-29
Windows app icon to display in old alt+tab on Windows.

madcdevelop:
- Feature: Able to see the windows app icon in the old alt+tab dialog.

- No

### DIFF
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -148,6 +148,10 @@ pub fn create_window(event_loop: &ActiveEventLoop, maximized: bool, title: &str)
         .with_visible(false);
 
     #[cfg(target_os = "windows")]
+    let window_attributes = window_attributes
+        .with_taskbar_icon(Some(load_icon()));
+
+    #[cfg(target_os = "windows")]
     let window_attributes = if !cmd_line_settings.opengl {
         WindowAttributesExtWindows::with_no_redirection_bitmap(window_attributes, true)
     } else {


fredizzimo:
I think this is good when the rust formatting error is fixed.

src/window/mod.rs
```diff
@@ -147,6 +147,9 @@ pub fn create_window(event_loop: &ActiveEventLoop, maximized: bool, title: &str)
         .with_transparent(true)
         .with_visible(false);
 
+    #[cfg(target_os = "windows")]
+    let window_attributes = window_attributes.with_taskbar_icon(Some(load_icon()));
```
falcucci, line 151: I think we could avoid loading the icon twice since it is loaded at [L132](https://github.com/neovide/neovide/pull/2816/files#diff-3236d45aa6e68d88a718f7a9e523800f3a3da86bfb4114a11c320382149b4b24R132)

to achieve this we need to .clone() the icon only for windows specifically, so I would suggest something like this for readability would be enough:

```rust
let window_attributes = Window::default_attributes()
    .with_title(title)
    .with_maximized(maximized)
    .with_transparent(true)
    .with_visible(false);

#[cfg(target_family = "unix")]
let window_attributes = window_attributes.with_window_icon(Some(icon));

#[cfg(target_os = "windows")]
let window_attributes = window_attributes
    .with_window_icon(Some(icon.clone()))
    .with_taskbar_icon(Some(icon));
```

falcucci (changes requested):

falcucci (approved):

falcucci:
thank you @madcdevelop

PR #2821, opened 2024-10-03
Remove `lazy_static`

InfyniteHeap:
Remove `lazy_static` dependency and use `LazyLock` instead, which comes from the standard library. This is because the implementation comes from the standard library seems more reliable and stable than the one comes from a third-party crate.

No.


falcucci (approved):
good catch @InfyniteHeap

Issue #2657, opened 2024-06-16
Cant get neovide to open maximized on Ubuntu 24.04


bjorntrondsen:
**Describe the bug**
I cant get neovide to open maximized on Ubuntu 24.04 LTS, Gnome 46, X11 (nvim v0.10.0)
This works
`./neovide.AppImage --fork --size=4096x2048`
This does not
`./neovide.AppImage --fork --maximized`

Also tested v0.12.0 with the same result. Any ideas?

**To Reproduce**
1. Dowload latest neovide.AppImage
2. Launch using `neovide.AppImage --maximized`

**Expected behavior**
Either to launch maximized every time or remember the window size from last time

**Desktop:**
 - OS: Ubuntu 24.04 LTS (Gnome 46, X11)
 - Neovide Version 0.13.1
 - Neovim Version 0.10.0

Trace from neovide --maximized --log attached
[neovide_rCURRENT.log](https://github.com/user-attachments/files/15857758/neovide_rCURRENT.log)

fredizzimo:
Hm. It looks like creating maximized windows on X11 needs special handling. See:
* https://github.com/rust-windowing/winit/issues/2360

But we only want to do that on X11 to avoid unwanted animations on other platforms. I won't be able to fix it myself for at least three weeks, so if someone wants to try, feel free to do so.

PR #2824, opened 2024-10-04
fix: maximization on X11

murar8:
Took a stab at a fix for https://github.com/neovide/neovide/issues/2657

- Fix

- No

### DIFF
--- a/src/window/window_wrapper.rs
+++ b/src/window/window_wrapper.rs
@@ -76,6 +76,7 @@ pub struct WinitWindowWrapper {
     is_minimized: bool,
     ime_enabled: bool,
     ime_area: (dpi::PhysicalPosition<u32>, dpi::PhysicalSize<u32>),
+    needs_x11_maximization: bool,
     pub vsync: Option<VSync>,
     #[cfg(target_os = "macos")]
     pub macos_feature: Option<MacosWindowFeature>,
@@ -88,6 +89,7 @@ impl WinitWindowWrapper {
     ) -> Self {
         let saved_inner_size = Default::default();
         let renderer = Renderer::new(1.0, initial_font_settings);
+        let needs_x11_maximization = matches!(initial_window_size, WindowSize::Maximized);
 
         Self {
             skia_renderer: None,
@@ -109,6 +111,7 @@ impl WinitWindowWrapper {
             },
             initial_window_size,
             is_minimized: false,
+            needs_x11_maximization,
             vsync: None,
             ime_enabled: false,
             ime_area: Default::default(),
@@ -292,6 +295,9 @@ impl WinitWindowWrapper {
             send_ui(SerialCommand::Keyboard("<NOP>".into()));
 
             self.is_minimized = false;
+        } else if self.needs_x11_maximization {
+            self.x11_set_maximized();
+            self.needs_x11_maximization = false
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/window/window_wrapper.rs
```diff
@@ -411,6 +411,17 @@ impl WinitWindowWrapper {
         skia_renderer.swap_buffers();
         if self.ui_state == UIState::FirstFrame {
             skia_renderer.window().set_visible(true);
+            let is_maximized = matches!(self.initial_window_size, WindowSize::Maximized);
+            let is_x11 = matches!(
+                skia_renderer.window().window_handle().unwrap().as_raw(),
```
murar8, line 416: Please let me know if unwrapping the window handle is ok at this point, I guess it must be?

src/window/window_wrapper.rs
```diff
@@ -410,7 +410,17 @@ impl WinitWindowWrapper {
         }
         skia_renderer.swap_buffers();
         if self.ui_state == UIState::FirstFrame {
-            skia_renderer.window().set_visible(true);
+            let window = skia_renderer.window();
+            let handle = window.window_handle().unwrap().as_raw();
```
murar8, line 414: Please let me know if unwrapping the window handle is ok at this point, I guess it must be?

fredizzimo:
Is it not enough to call `window.set_maximized(true)` here after showing the window?https://github.com/neovide/neovide/blob/6d0aa1acbd9c5bb97093d67b7c4dc7f2ba22e058/src/window/window_wrapper.rs#L413

According to this, it should work https://github.com/rust-windowing/winit/issues/2360#issuecomment-1179818125

We also have to be careful to not cause any artifacts which https://github.com/neovide/neovide/pull/2562 tries to fix. See the discussion there. Although it might be impossible in the case of maximization.

src/window/update_loop.rs
```diff
@@ -325,6 +350,7 @@ impl ApplicationHandler<UserEvent> for UpdateLoop {
         match event {
             WindowEvent::RedrawRequested => {
                 self.redraw_requested();
+                self.maximize_x11();
```
murar8, line 353: Not sure if this would be a good place to trigger the check, please let me know if you think it should be moved somewhere else.

murar8:
> Is it not enough to call `window.set_maximized(true)` here after showing the window?
I'm not very knowledgeable about X11 internals, but that was my first attempt and it doesn't seem to work, looks like it needs a rendering cycle with the window visible first. Please check out the updated code.

> We also have to be careful to not cause any artifacts which #2562 tries to fix. See the discussion there. Although it might be impossible in the case of maximization.
Thanks, will check it out! Not sure if the artifacts you are talking about are easy to see with the naked eye but I attached a video below. I also made some changes to the logic because I found a bug, feel free to check it out again!
[Grabación de pantalla desde 2024-10-08 11-15-12.webm](https://github.com/user-attachments/assets/5079ebdf-26ae-45e3-97e8-f59bb38a79b1)

murar8:
Interestingly, when testing out this code:
```diff
diff --git a/src/window/window_wrapper.rs b/src/window/window_wrapper.rs
index 39b57a0..0e821dd 100644
--- a/src/window/window_wrapper.rs
+++ b/src/window/window_wrapper.rs
@@ -410,8 +410,16 @@ impl WinitWindowWrapper {
         }
         skia_renderer.swap_buffers();
         if self.ui_state == UIState::FirstFrame {
-            skia_renderer.window().set_visible(true);
+            let window = skia_renderer.window();
+            window.set_visible(true);
             self.ui_state = UIState::Showing;
+            let is_x11 = {
+                let handle = window.window_handle().unwrap().as_raw();
+                matches!(handle, RawWindowHandle::Xlib(_) | RawWindowHandle::Xcb(_))
+            };
+            if is_x11 {
+                window.set_maximized(true);
+            }
         }
         tracy_frame();
         tracy_gpu_collect();
```
Works only when `--frame=none`

fredizzimo:
Does the latest version work without glitches? In that case I think it's good enough.

murar8:
@fredizzimo It is usable but still a bit glitchy, if you would like a temporary workaround it will work, otherwise feel free to close for now, when I have some free time I will hack on it some more.

fredizzimo:
I'm putting this into draft status for now. I have to test it myself, and I also don't think I saw any problems on X11 on KDE Plasma before, so the original problem might be limited to some desktops/window managers only.

Issue #2665, opened 2024-06-23
Give Windows rounded corners and hide title bar


Parsifa1:
May I ask if there is some possibility to provide rounded corners and a way to hide the title bar for Windows, like Wezterm has provided:

![image](https://github.com/neovide/neovide/assets/31800073/7d543d00-38fe-4d72-9387-81bf1eff1597)

the "frame none" can hide th title-bar but can not provide rounded frame

thanks a lot!

fredizzimo:
I'm not sure if it will work out of the box or not but Winit 0.30 added this https://docs.rs/winit/latest/winit/platform/windows/enum.CornerPreference.html. But we sre still in 0.29, so it's not as easy as just calling that function.

fredizzimo:
Winit 0.30.3 has now been merged, so if someone wants to implement this, it should now be possible.

falcucci:
since you can use `frame = "none"` and the corners preference has been [merged](https://github.com/neovide/neovide/pull/2825), this is now available on nightly.

once winit also implements the title apis we can also sort this without managing the frame to fix that.

PR #2825, opened 2024-10-04
Implement corner preference

agraven:
This should be most of the initial work for implementing the corner preference windows window setting as described in  #2665. I haven't actually built this on a windows machine yet, so there are probably pieces missing and small mistakes I haven't noticed

- Feature

- No

### DIFF
--- a/src/window/settings.rs
+++ b/src/window/settings.rs
@@ -24,6 +24,8 @@ pub struct WindowSettings {
     pub padding_right: u32,
     pub padding_bottom: u32,
     pub theme: String,
+    #[cfg(target_os = "windows")]
+    pub corner_preference: CornerPreference,
     #[cfg(target_os = "macos")]
     pub input_macos_alt_is_meta: bool,
     #[cfg(target_os = "macos")]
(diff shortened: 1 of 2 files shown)


fredizzimo:
Thank you!

Can you rebase onto the latest main? That will fix the Clippy errors.

And before this can be merged the new options need to be documented. 

Finally, you need to apply the settings when the window is created here https://github.com/neovide/neovide/blob/71602b3cde694cd82731c5aa7838bbc182270f50/src/window/window_wrapper.rs#L432, so that they take effect immediately and works anywhere in the nvim configuration.

fredizzimo:
For the documentation can you also add the `**Unrleaased yet.` tag, like this https://github.com/neovide/neovide/blob/281a062949ede46cbc72ba4d6f3cf1390219d0a9/website/docs/configuration.md?plain=1#L335

falcucci:
@agraven what is the current status on this? let me know and I would be happy to merge when it's done.

agraven:
Not very far. I don't have convenient access to a Windows machine either anymore, so unless someone else wants to take over the PR I might suggest closing it.

falcucci:
since I got access to a windows machine, I could sort this one.

thank you @agraven for initiating it anyway!

PR #2827, opened 2024-10-05
feat: add title bar colours for windows

teehee567:
- Feature

- No


fredizzimo:
Thank you!

Can you rebase onto the latest main? That will fix the Clippy errors. After that you need to run  `cargo fmt`. to fix the formatting errors in the CI.

And before this can be merged the new options need to be documented. 

Finally, you need to apply the settings when the window is created here https://github.com/neovide/neovide/blob/71602b3cde694cd82731c5aa7838bbc182270f50/src/window/window_wrapper.rs#L432, so that they take effect immediately and works anywhere in the nvim configuration.

fredizzimo:
For the documentation can you also add the `**Unrleaased yet.` tag, like this https://github.com/neovide/neovide/blob/281a062949ede46cbc72ba4d6f3cf1390219d0a9/website/docs/configuration.md?plain=1#L335

fredizzimo:
Thanks, I think everything is good now, once the CI failures are fixed.

fredizzimo:
It looks like the rest of the errors are not related to your work. A new rust compiler has been released with new Clippy warnings.

We will fix those errors a bit later, and you can rebase after they are merged.

falcucci (approved):
took the freedom to rebase your changes @teehee567 

thank you for the contribution!

PR #2832, opened 2024-10-10
bug: make sure to load and set the app icon for any context

### LINKED ISSUE #2831: Neovide doesn't load the icon when launched from terminal using the not bundled binary
The icon not being loaded is quite annoying.

It just feels that something is wrong and we are just running a process instead of the app.

CleanShot 2024-10-10 at 19 13 41@2x

It's okay to have multiple icons in the dock for now since we do not support multi-window (yet)

falcucci:
- add a function to load the neovide icon
- set the application icon image on app launch

- Fix

- No

fixes #2831

### DIFF
--- a/src/window/macos.rs
+++ b/src/window/macos.rs
@@ -1,3 +1,5 @@
+use std::{os::raw::c_void, str};
+
 use objc2::{
     declare_class, msg_send, msg_send_id, mutability,
     rc::{autoreleasepool, Retained},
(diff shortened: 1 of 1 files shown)


fredizzimo:
I think the icon needs to be embedded and loaded from memory like we do on windows. Otherwise it won't work when launching from other paths.

falcucci:
@fredizzimo our current [image](https://crates.io/crates/image) crate doesn't support icns, so I am using the apple apis instead.

PR #2834, opened 2024-10-11
feat: add mouse cursor icon option

falcucci:
- Fix

- No


Kethku:
looks good to me! merge when ready

Issue #2771, opened 2024-08-22
Cursor color is linked to the wrong ('Normal') highlight group


n00bmind:
**Describe the bug**
As explained in the title, the color of the cursor seems to be linked to the 'Normal' highlight group instead of 'Cursor', meaning I cannot have the cursor be a different color to the non-highlighted text.
I found this previous bug https://github.com/neovide/neovide/issues/1611 which, amazingly, was closed without an actual fix being provided.

**To Reproduce**
1. Set some highlight group for the cursor i.e. `hi Cursor guifg=#000000 guibg=#B80000`
2. Type `:hi Cursor` in the command line to check that the new highlight group values were applied correctly
3. Observe the cursor color hasn't changed
4. Repeat a similar change but for the 'Normal' highlight group
5. Observe the cursor color now does change

**Expected behavior**
The cursor color should be linked to its dedicated highlight group ('Cursor')

**Screenshots**
![image](https://github.com/user-attachments/assets/c83fc696-2c8b-4fa2-b3de-bbc6c1666998)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Neovide Version 0.12.2
 - Neovim Version 0.9.5

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide_rCURRENT.log](https://github.com/user-attachments/files/16719754/neovide_rCURRENT.log)

fredizzimo:
If this is a bug, then it's a Neovim one, because this is the information it sends to us
```
[...]

PR #2843, opened 2024-10-17
feat: add cursor color settings

Cathyprime:
- Feature

- No

adds the ability to set colors using g:neovide_cursor_fg and g:neovide_cursor_bg, which can be used to make the cursor work with 'Cursor' highlight group, if left empty it defaults to default style

### DIFF
--- a/src/renderer/cursor_renderer/mod.rs
+++ b/src/renderer/cursor_renderer/mod.rs
@@ -34,8 +34,8 @@ pub struct CursorSettings {
     trail_size: f32,
     unfocused_outline_width: f32,
     smooth_blink: bool,
-    bg: String,
-    fg: String,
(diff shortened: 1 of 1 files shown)


fredizzimo:
Unfortunately, I don't think we can accept this, since the cursor color is already controlled by `guicursor` https://neovim.io/doc/user/options.html#'guicursor', and the highlight groups defined there.

Notice, that by default Neovim uses the `Normal` highlight group instead of `Cursor`, but that can be configured on the Neovim side.

Also, check out this issue
* https://github.com/neovide/neovide/issues/2771

Cathyprime:
Ohh I see I see, thank you for taking the time to review the pr and response anyway, I'll try to contribute more meaningful code next time 😅

PR #2844, opened 2024-10-18
internal: Use macOS-latest runners and bump MACOSX_DEPLOYMENT_TARGET to 13.0

falcucci:


.github/workflows/build.yml
```diff
@@ -24,6 +24,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  MACOSX_DEPLOYMENT_TARGET: 13.0
```
fredizzimo, line 27: I think we could support macOS Catalina. That's what the homebrew formula supports https://github.com/Homebrew/homebrew-core/blob/4e09c9524e8fca0ca023532cd7b723078d2bc009/Formula/n/neovide.rb

fredizzimo: Isn't that 10.15? BTW, this also needs to match https://github.com/neovide/neovide/blob/af35b8a5b48c3219766de6a2e1c48b405e4f3e15/extra/osx/Neovide.app/Contents/Info.plist#L26C20-L26C24.

falcucci: totally though it was 12.0, my bad haha

falcucci: seems that we were using [10.11](https://github.com/neovide/neovide/blob/8be5c7821f66cec924b58a45bef897c038c9a7ae/.github/workflows/build.yml?plain=1#L178) instead

PR #2857, opened 2024-10-25
fix: rounded corners rendering

lucobellic:
Hello,
I would like to submit a fix for the rendering of rounded corners.

`rounded_silhouette` rendering was performed differently compared to the previous `silhouette` in `build_silhouette` which would completely hide my focused buffer in certain circumstances (multi-grid is enabled). 

Here is an example with neo-tree and a split buffer, where the bottom one is entirely hidden.

Screenshot from 2024-10-25 23-02-47

I'm not sure what is causing this issue.
It might be caused by one of my plugins, but this behavior was not present before this feature.
Also the issue is not consistent so unfortunately I was not able to create a reproducible scenario.

However, initializing the rounded path directly with `RRect` as it was done previously resolve the issue and keep the nice rounded corners.

I may have missed the reason for not applying `reduce` with `PathOp::Union` to `rounded_silhouette`, so feel free to suggest any changes or edit this PR.

- Fix

- No

### DIFF
--- a/src/renderer/rendered_layer.rs
+++ b/src/renderer/rendered_layer.rs
@@ -3,7 +3,7 @@ use skia_safe::{
     canvas::SaveLayerRec,
     image_filters::blur,
     utils::shadow_utils::{draw_shadow, ShadowFlags},
-    BlendMode, Canvas, ClipOp, Color, Paint, Path, PathOp, Point3, Rect,
+    BlendMode, Canvas, ClipOp, Color, Paint, Path, PathOp, Point3, RRect, Rect,
 };
 
 use glamour::Intersection;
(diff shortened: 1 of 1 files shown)


fredizzimo:
Thank you, 

Unfortunately this implementation has the same bug I was talking about here https://github.com/neovide/neovide/pull/2743#discussion_r1715931946

For example the most simple case
image

I don't think those corners should be rounded. 

I think it's strictly better than what we have currently and fixes the case where rounded corners are disabled. So, we will likely merge this this fix rather than reverting https://github.com/neovide/neovide/pull/2743 until a proper solution is found.

fredizzimo:
Thank you and happy [Hacktoberfest](https://hacktoberfest.com/)!

There are some other situations where it actually looks better like, this, and we don't have enough information from Neovim to make the correct decision. So, I'm merging this

image

PR #2860, opened 2024-10-29
change backtrace file

ZeyadMoustafaKamal:
- Fix

- No

I don't want `neovide_backtrace.log` to be displayed when I open the file manager or when I `ls`. So I added `.` to make it hidden. Also most of the tools I have used put `.` before filenames when they put them in the home directory. Also if you don't want this, you may change it to be like `$XDG_STATE_HOME/neovide/neovida_backtrace.log` or something else for other platforms.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -70,7 +70,7 @@ use crate::settings::{load_last_window_settings, Config, FontSettings, Persisten
 
 pub use profiling::startup_profiler;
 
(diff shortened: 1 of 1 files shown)


fredizzimo:
Thanks for the contribution, but we already have a PR that will change the location of the backtraces file, so we will merge that instead once the conflicts are resolved
* https://github.com/neovide/neovide/pull/2717

Issue #2751, opened 2024-08-07
^M is added at the end of a line when copy-pasting in a remote session (win32 host, linux client)


damiandimanov:
**Describe the bug**
Running a neovim remote server on a Windows machine and connecting to it using neovide on a Linux machine causes a carriage return symbol to be shown at the end of pasted lines.

Caused by this line (and the files being `ff=dos` on the windows machine): https://github.com/neovide/neovide/blob/3960baad20d19e10c0857e07f87ddd0c65e27177/src/bridge/clipboard.rs#L17

**To Reproduce**
1. Run neovim headless on a windows machine
2. Run `neovide --server=...`
3. Use `set clipboard=unnamed,unnamedplus`
4. Press yyp

The newly pasted line has the ^M symbol at the end

**Expected behavior**
The same line should be pasted, i.e. without the ^M symbol

**Desktop:**
 - OS: ArchLinux
 - Neovide Version 0.13.3
 - Neovim Version 0.10.1

fredizzimo:
I haven't looked closer at this, but it looks like it tries to detect the buffer format here:
https://github.com/neovide/neovide/blob/3960baad20d19e10c0857e07f87ddd0c65e27177/src/bridge/handler.rs#L50-L57

And somehow that either fails, or the wrong assumption was made when the code was written and that the lines should never have Windows line endings at that point.

Is the buffer type you are pasting into using unix or dos line endings?

damiandcoherent:
Thank you for looking into it.

It happens even when you paste in the same buffer you copied the line from. The buffer uses dos line endings.

fredizzimo:
It looks like someone made the wrong assumption when writing that code or something. The `win32yank.exe` implementation in Neovim unconditionally changes the line endings to windows on yank, and linux on paste

https://github.com/neovim/neovim/blob/0d3f2c4904d552635da07a6e2b4e75520b514940/runtime/autoload/provider/clipboard.vim#L138-L142. 

And that works properly for both does and unix fileformats. 

In Neovide this is broken even in WSL, due to the remote clipboard being forced there as well. I have not noticed this myself though even if I use Neovim all the time, since I have had a slightly custom clipboard function all this time. And when I do clean configuration tests, they have all been in files with unix endings.

So, the fix would be to always change the line endings to unix on paste. For the yank behaviour I think we could have an option to either keep the line endings of the buffer or to convert to the OS native endings.

fredizzimo:
This was fixed by
[...]

PR #2869, opened 2024-11-10
fix: pasting ^M symbols from clipboard

damiandimanov:
Related: #2751

- Fix

- No


fredizzimo:
Thank you! 

I finally got time to test this properly. It works in all combinations I have managed to test, both when pasting from and to windows.

But there might be a tiny bug, when copying something from a buffer with unix line endings into windows, it's always converted to `\r\n`. Most windows programs do a conversion on paste, so it's not a huge issue, but it might be a problem in some case.

PR #2873, opened 2024-11-15
chore: update `skia-safe` version to `0.80.0`

falcucci:
- No

### DIFF
--- a/src/renderer/d3d.rs
+++ b/src/renderer/d3d.rs
@@ -7,6 +7,7 @@ use skia_safe::{
     surface::BackendSurfaceAccess,
     Canvas, ColorSpace, ColorType, PixelGeometry, Surface, SurfaceProps, SurfacePropsFlags,
 };
+use windows::core::{Interface, Result, PCWSTR};
 use windows::Win32::Graphics::DirectComposition::{
     DCompositionCreateDevice2, IDCompositionDevice, IDCompositionTarget, IDCompositionVisual,
 };
@@ -38,10 +39,6 @@ use windows::Win32::System::Threading::{CreateEventW, WaitForSingleObjectEx, INF
 use windows::Win32::{
     Foundation::{CloseHandle, HANDLE, HWND},
     Graphics::Dxgi::DXGI_PRESENT,
-};
-use windows::{
-    core::{Interface, Result, PCWSTR},
-    Win32::Graphics::Dxgi::IDXGIFactory4,
 };
 use winit::{
     event_loop::EventLoopProxy,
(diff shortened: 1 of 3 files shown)


fredizzimo:
This works on Linux, but still needs testing on Windows, especially as there are Windows specific changes, and I don't have a Windows computer available right now.

I first thought that there was a performance regression, since it only ran at 72 FPS instead of 144. But I also tested the main version, and that has the same problem. I'm not sure when that has happened, or if I have too much open right now.  Since, this is not specific to this update it should not stop us from merging though.

src/renderer/d3d.rs
```diff
@@ -120,7 +126,7 @@ impl D3DSkiaRenderer {
         };
 
         #[cfg(not(feature = "d3d_debug"))]
-        let dxgi_factory: IDXGIFactory2 =
+        let dxgi_factory: IDXGIFactory4 =
```
fredizzimo, line 129: I don't think we can change this. See https://github.com/neovide/neovide/pull/2559

falcucci: sure, i'll rollback to `IDXGIFactory2`

src/renderer/d3d.rs
```diff
@@ -435,7 +438,7 @@ impl SkiaRenderer for D3DSkiaRenderer {
                     size.width,
                     size.height,
                     DXGI_FORMAT_UNKNOWN,
-                    self.swap_chain_desc.Flags,
+                    DXGI_SWAP_CHAIN_FLAG::default(),
```
fredizzimo, line 441: I don't think this can be changed either, since it has to match the flags of the created swapchain, which is `Flags: DXGI_SWAP_CHAIN_FLAG_FRAME_LATENCY_WAITABLE_OBJECT.0 as u32`

src/renderer/d3d.rs
```diff
@@ -401,7 +404,7 @@ impl SkiaRenderer for D3DSkiaRenderer {
             self.gr_context.submit(Some(SyncCpu::No));
 
             tracy_gpu_zone!("present");
-            if self.swap_chain.Present(1, 0).is_ok() {
+            if self.swap_chain.Present(1, DXGI_PRESENT::default()).is_ok() {
```
fredizzimo, line 407: I would prefer if we are specific here with 0, since the default is not explicitly documented to mean something and could therefore change. `0` has a documented meaning here https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/dxgi-present

src/renderer/d3d.rs
falcucci: done

fredizzimo (approved):

Kethku:
I confirmed this runs on windows although I did have to update rust. I think given that the ci passed, i was just running on an old version

PR #2016, opened 2023-09-07
Release 0.11.2

fredizzimo:
This release adds support for Neovim 0.9.2.

## Features
* Add width to guifont settings by @fritsstegmann in https://github.com/neovide/neovide/pull/1227
  * https://neovide.dev/configuration.html#font
## Bug Fixes
* Fix disappearing characters on Neovim 0.9.2 and greater by @fredizzimo in https://github.com/neovide/neovide/pull/2012
* Add missing Enter special key by @fredizzimo in https://github.com/neovide/neovide/pull/1997
* Improve nvim detection by @fredizzimo in https://github.com/neovide/neovide/pull/1946
* Show intro message when the UI command `msg_intro` is received by @fredizzimo in https://github.com/neovide/neovide/pull/1985
  * The `shortmess+=I` option now works.
  * The message now closes properly when using dashboard-nvim.

## Other changes
* Better logging of keyboard events by @fredizzimo in https://github.com/neovide/neovide/pull/1962

## New Contributors
* @fritsstegmann made their first contribution in https://github.com/neovide/neovide/pull/1227

**Full Changelog**: https://github.com/neovide/neovide/compare/0.11.1...0.11.2

PR #2883, opened 2024-11-24
fix!: background opacity rendering

lucobellic:
From https://github.com/neovide/neovide/pull/2780#issuecomment-2448637058.
Update background opacity rendering for floating and normal windows.

This PR changes:
- Text background blend for normal windows, as it was in 0.11.2 #2016 (same as other terminals like wezterm, kitty, etc.).
- Floating background rendering using blend, multi-grid, floating window, and default background information.
  -  Set opaque text background for floating windows with multi-grid since in this case all style blends are set to winblend/pumblend.
  - Only use blend for floating windows background to ensure consistent behavior when blend is 0.

The behavior to compute the background opacity is defined and documented in the `GridRenderer::get_background_opacity()` function.
 
<details>
<summary>Expand for examples...</summary>

With `neovide_transparency=0.75`
- Left: old behavior
- Right: new behavior

`winblend = 50`
241124_22h26m11s_screenshot

`winblend = 100`
image

`winblend = 0`
image

</details>

I hope this PR simplifies the review for the remaining work on transparency. Once this fix is merged, I will update the corresponding PR #2780.

> [!NOTE]
> I might have missed testing some configurations. Feel free to submit them so I can correct them.

- Fix

- Yes
  - Blending is different
  - Update floating rendering


falcucci (approved):
looks great.

fredizzimo:
Sorry for the delay of testing this again, but I have been quite busy and it's not completely straightforward to test correctly.

Unfortunately, it does not quite work properly.
Save this to a file called `blend.lua`
```lua
local ns = vim.api.nvim_create_namespace("my highlights")
vim.api.nvim_set_hl(ns, "MyBlend", {blend=90})
vim.api.nvim_buf_set_lines(0, 0, 0, false, {
    "This text should be normal (respecting winblend)",
    "And this should respect the highglight group blending",
    "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque consectetur",
    "dolor eu dictum convallis, magna ante vehicula dolor, ac dictum sem nunc eget",
    "tortor. Proin placerat massa ut justo porttitor, et blandit ante dictum.",
    "Maecenas eget semper turpis, eu porttitor massa. Curabitur in arcu vulputate",
    "lacus eleifend dapibus. Aenean imperdiet et est non pellentesque. Quisque sed",
})
vim.api.nvim_buf_add_highlight(0, ns, "MyBlend", 1, 0, -1)
vim.api.nvim_open_win(0, true, {width=50, height=10, relative='win', row=0, col=10, border="single", title="Title", footer="Footer"})
vim.api.nvim_set_hl_ns(ns)
vim.o.winblend=20
```
And run `:luafile blend.lua

Observe that nothing is transparent
image

This is with Neovide 0.13.1, which also is wrong, according to the documentation the `blend` option should only apply if `winblend` is set to non-zero, so blending the background is wrong. But it would be useful to be able to do that. What's more wrong is that the line in the floating window with the blend set, goes through both windows, it should only apply to the floating window itself.
image

The terminal Neovim does it correctly
image

lucobellic:
@fredizzimo Thank you for the review and Merry Christmas.

The non-transparent background with multi-grid is expected with this implementation.

With multi-grid, all `style.blend` in `grid_renderer` are overwritten by the `winblend` value. 
The current implementation only applies blend on the default background color for floating windows and sets all other blends to opaque to be able to see the background, mainly when `winblend` is set to 100.

However, transparency does not work if the `NormalFloat` background is not the same as the default/Normal background and does not allow blending for other colors in floating windows.

To achieve a proper and simpler implementation (similar to the Alacritty screenshot), it would require not having `style.blend` overwritten by `winblend` (or being able to get this info in style metadata) and directly drawing the floating background window with `winblend` in `render_layer.rs`.

Nonetheless, I'm not sure if this is currently possible.

So feel free to close this PR since it doesn't fix the issue and only uses a workaround that does not respect the documentation (the same goes for #2780 using the same idea).

fredizzimo:
Ok, thanks. I will take a look, I think we do have enough information to render it according to the specification, but I might be wrong.

fredizzimo:
According to my testing, the style blend values are maintained, but they are reported even when they shouldn't in the case of `winblend` being set to 0. So, I opened a question for Neovim here
* https://github.com/neovim/neovim/discussions/31737

Issue #2884, opened 2024-11-25
Cmdline Transparency Issue


kkkk0524:
**Is your feature request related to a problem? Please describe.**
I am experiencing an issue with the command line not being transparent in Neovide(13.1). Despite the settings for transparency being configured, the command line remains opaque. Could you please assist me in resolving this issue?(win11)

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #2916, opened 2024-12-27
fix: Neovide transparency works for floating windows again

fredizzimo:
* Alternative implementation of https://github.com/neovide/neovide/pull/2883
* The `neovide_transparency` option now works for floating windows again, this was regressed in Neovide 0.13.2
* Simplified the transparency rendering so the transparency info is completely stored in the lines and the background clear. So there's no longer any skia layer based blending.

* Fixes https://github.com/neovide/neovide/issues/2884

NOTE:
This does not respect the documentation https://github.com/neovim/neovim/issues/24472#issuecomment-2563970117
> blend={integer}					[highlight-blend](https://neovim.io/doc/user/syntax.html#highlight-blend) [opacity](https://neovim.io/doc/user/syntax.html#opacity)
	Override the blend level for a highlight group within the popupmenu
	or floating windows. Only takes effect if ['pumblend'](https://neovim.io/doc/user/options.html#'pumblend') or ['winblend'](https://neovim.io/doc/user/options.html#'winblend')
	is set for the menu or window. See the help at the respective option.

Blend is applied also at the toplevel non-floating windows, so you can set parts of the UI transparent, for example by making the lualine transparent but keep the rest. 

It's also applied without the need of setting `winblend` or `pumblend`.

NOTE: you can't set highlight blend for the `Normal` highlight group, due to restriction in the UI protocol. If that was possible, then the `neovide_transparency` setting would no longer be needed, and you would have more control. But there are also some other bugs in the way currently.
See:
* https://github.com/neovim/neovim/discussions/31737
* https://github.com/neovim/neovim/issues/24472

@lucobellic, you probably want to take a look at this

- No

### DIFF
--- a/src/renderer/grid_renderer.rs
+++ b/src/renderer/grid_renderer.rs
@@ -145,7 +145,7 @@ impl GridRenderer {
 
         BackgroundInfo {
             custom_color,
-            transparent: alpha > 0.0,
+            transparent: alpha < 1.0,
         }
     }
 
--- a/src/renderer/grid_renderer.rs
+++ b/src/renderer/grid_renderer.rs
@@ -99,6 +99,7 @@ impl GridRenderer {
         grid_position: GridPos<i32>,
         cell_width: i32,
         style: &Option<Arc<Style>>,
+        opacity: f32,
     ) -> BackgroundInfo {
         tracy_zone!("draw_background");
         let debug = SETTINGS.get::<RendererSettings>().debug_renderer;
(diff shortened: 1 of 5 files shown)


lucobellic:
I have just tested it, and LGTM.

Removing the usage of skia blend drastically simplifies the transparency behavior.

The only drawback is probably for some users using both `neovide_transparency > 0` and `winblend = 0` to get an opaque background only for floating windows.

falcucci (approved):
works as described.

fredizzimo:
Thanks, I'm merging this to get a bit more testing before the release

PR #2917, opened 2024-12-29
fix: add `text_background_opacity` in the opaque validation

falcucci:
During our changes to [fix](https://github.com/neovide/neovide/pull/2780) the opacity we also re-introduced a bug showing the [shadow](https://github.com/neovide/neovide/pull/2272) artifacts we already had in the past.

AT this moment we should update it adding the new `text_background_opacity` in the validation.

### DIFF
--- a/src/window/macos.rs
+++ b/src/window/macos.rs
@@ -257,9 +257,8 @@ impl MacosWindowFeature {
         }
     }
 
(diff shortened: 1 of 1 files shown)


fredizzimo:
I'm merging this, but I think it can still break if some of the highlight groups of the root windows use a blend. 

To fix that, we need to check all root windows for `has_transparency()`

PR #2919, opened 2025-01-02
Use composition information from Neovim

fredizzimo:
This requires this corresponding PR to Neovim https://github.com/neovim/neovim/pull/31837 to enable the support, but works without.

This fixes a lot of multigrid issues like:
* Windows positioned slightly wrong
* Windows rendered in the wrong z-order
* We don't seem to have any issues open for this, but this also fixes most of the issues re-connecting to a remote nvim instance,
  *  https://github.com/neovim/neovim/issues/30581
  * ~https://github.com/neovim/neovim/issues/30582~ (not fixed)
  * ~https://github.com/neovim/neovim/issues/30583~ (not fixed)

NOTE: Part of this PR was split out into 
* https://github.com/neovide/neovide/pull/3111

- No

### DIFF
--- a/src/bridge/events.rs
+++ b/src/bridge/events.rs
@@ -210,6 +210,9 @@ pub enum RedrawEvent {
         #[allow(unused)]
         focusable: bool,
         z_index: u64,
+        comp_index: Option<u64>,
+        screen_row: Option<u64>,
+        screen_col: Option<u64>,
     },
     #[allow(unused)]
     WindowExternalPosition {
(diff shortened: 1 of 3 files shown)


fredizzimo:
There are currently a few problems that I have noticed (I'm not sure if they are on the Neovim or Neovide side)
* It's not possible to drag splits
* The visual mode mouse drag does not refresh while moving the mouse
* There are some weird smooth scrolling arttifacts

fredizzimo:
I took it out of draft status, the Neovim PR is up, and I have verified all fixes.

But I think we should wait with merging this until after the release.

fredizzimo:
Neovim nightly now supports this, and it works correctly without the support as well, so I merged it.

PR #2920, opened 2025-01-02
chore: release 0.14.0

falcucci:


fredizzimo:
I think there are a few `unrleased yet`  tags in the documentation that need to be updated still.

PR #2921, opened 2025-01-03
The window should be redrawn after scrolling with the touch screen

yyy33:
- Fix

- No

### DIFF
--- a/src/window/mouse_manager.rs
+++ b/src/window/mouse_manager.rs
@@ -380,6 +380,8 @@ impl MouseManager {
                 }
             }
         }
+
+        editor_state.window.request_redraw();
     }
 
     pub fn handle_event(


fredizzimo:
I'm sorry, I'm not sure what the actual issue is. Maybe you know @falcucci?

The scroll command should always be sent to Neovim, and then when Neovide receives the response, the screen should update. So, no additional redraw should be needed, and if it is, it should do nothing, and there's a bug somewhere.

falcucci:
@yyy33 could you explain what is the unexpected behavior you are testifying?

yyy33:
Sorry for not explaining this clearly, I'm trying to port neovide to android,currently running neovim in termux via tcp connection for now and have got it running but not at a fully usable stage yet, still a lot of issues to be fixed.

I'm currently working on the touch screen issue, as the neovide doesn't respond when I scroll with my finger, I've found that I need to redraw after the scroll event is processed in order to get it to work properly, I assumed that this was a touch screen bug, or that the problem was elsewhere, and that redrawing after scrolling wasn't a good solution?

yyy33:
This is the effect of redrawing after scrolling, if you don't redraw after scrolling, neovide doesn't react at all

https://github.com/user-attachments/assets/b21c518d-1c21-47e3-8fd0-220770cfb103

yyy33:
> The scroll command should always be sent to Neovim,

My current understanding is that the touch event is handled and not sent to neovim, maybe I should read a little deeper to understand that.

fredizzimo:
An android port would be nice, but might be hard for our small team to maintain.

Winit should generate `handle_pixelscroll` events, which are buffered and sent as `handle_linescroll` and sent to neovim. Neovim then responds with the new contents of the buffer and how many lines were actually scrolled, which might be zero if you are already at the end or beginning for example.

It's not ideal, but it's very tricky to do exact pixelbased scrolling with this model, where we have to give the control to Neovim to get new content, we only know what's visible on the screen and the recent past for smooth scrolling.

One thing to note, you do need to set https://neovim.io/doc/user/options.html#'mousescroll' to 1, otherwise you need to scroll 3 lines at a time. And probably also speed up the https://neovide.dev/configuration.html#scroll-animation-length.

I do have some ideas to make that better, but I haven't had time to attempt anything myself, since it's not a priority for me personally. It has been discussed on our discord and I can share it here later, if you or someeone else wants to attempt.

PR #2928, opened 2025-01-04
fix: execute login using zsh shell configuration

falcucci:
## What kind of change does this PR introduce?
- Fix

## Did this PR introduce a breaking change? 
- No

fixes https://github.com/neovide/neovide/issues/2840

PR #2949, opened 2025-01-09
fix: --wsl error if using non-POSIX shell as login shell

### LINKED ISSUE #2941: "--wsl" option can fail if the WSL login shell is not POSIX compliant
**Describe the bug**
`--wsl` option can fail if the WSL login shell is not POSIX compliant. This occurs when setting `elvish` as the login shell, since it does not support the syntax `$SHELL`.

**To Reproduce**
1. Use `chsh` to set login shell as `elvish` within WSL
2. Launch `neovide --wsl`

**Expected behavior**
Neovide launches successfully

 - OS: Windows 11
 - Neovide Version 0.14.0
 - Neovim Version N/A

**Additional context**
I'd be happy to make a PR, changing `$SHELL` to `/bin/sh` is the simplest solution, since it guarantees a POSIX shell.  I'd have to check the code if there are any config options that might make this a breaking change. If so we could add a config option that allows specification of which shell to run.

ejrichards:
- Fixes #2941 

- No
- `nvim_cmd_impl` changed to use the syntax `wsl -- <bin> <args>` rather than `wsl $SHELL -lc <bin> <args>` which will still spawn the default shell.  The previous implementation actually spawns two shells.
- `create_platform_shell_command` changed to use `/bin/sh`, since non-POSIX shells might not have the `which` command

### DIFF
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -61,6 +61,7 @@ fn build_login_cmd_args(command: &str, args: &[&str]) -> (String, Vec<String>) {
 
     let user = env::var("USER").unwrap_or_explained_panic("USER environment variable not found");
     let shell = env::var("SHELL").unwrap_or_else(|_| "/bin/zsh".to_string());
+    let shell_name = shell.split('/').last().unwrap_or("zsh");
 
     let args = match shlex::try_join(args.iter().map(|s| (*s) as &str)) {
         Ok(args) => args,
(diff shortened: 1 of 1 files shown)


fredizzimo:
Thanks, but we can't change the shell to `sh`, becuase the user might have only configured their default shell and not `sh`. So, custom paths for example won't be set, which means that things like LSP won't work.

So, we need to find another way to do this, maybe bootstrap through another shell, like was done on macOS https://github.com/neovide/neovide/pull/2928.

PR #2950, opened 2025-01-09
fix: --wsl error if using non-POSIX shell as login shell

### LINKED ISSUE #2941: "--wsl" option can fail if the WSL login shell is not POSIX compliant

ejrichards:
Reopening #2949 , realized you need to specify `--shell-type login` to wsl.exe to not be a breaking change and accidentally closed the previous PR.

- Fixes #2941 

- No
- `nvim_cmd_impl` changed to use the syntax `wsl --shell-type login -- <bin> <args>` rather than `wsl $SHELL -lc <bin> <args>` which will still spawn the default shell.  The previous implementation actually spawns two shells.
- `create_platform_shell_command` changed to use `/bin/sh`, since non-POSIX shells might not have the `which` command

@fredizzimo 

> Thanks, but we can't change the shell to `sh`, becuase the user might have only configured their default shell and not `sh`. So, custom paths for example won't be set, which means that things like LSP won't work.
> 
> So, we need to find another way to do this, maybe bootstrap through another shell, like was done on macOS #2928.

It is only using `/bin/sh` for doing `which nvim` and `nvim -v`, the `wsl -- <bin>` syntax does spawn the user's default shell, for example, if I run `wsl -- sudo ps -ef` I can see:

```
236     233  0 00:08 pts/0    00:00:00 /usr/local/bin/elvish -c sudo ps -ef
```

When my default shell is `elvish`

### DIFF
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -99,8 +99,10 @@ fn create_platform_shell_command(command: &str, args: &[&str]) -> StdCommand {
     {
         if SETTINGS.get::<CmdLineSettings>().wsl {
             let mut result = StdCommand::new("wsl");
-            result.args(["$SHELL", "-lc"]);
-            result.arg(format!("{} {}", command, args.join(" ")));
+            result.args(["--shell-type", "login"]);
+            result.arg("--");
+            result.arg(command);
+            result.args(args);
 
             result.creation_flags(windows::Win32::System::Threading::CREATE_NO_WINDOW.0);
 
@@ -183,7 +185,7 @@ fn neovim_ok(bin: &str, args: &[String]) -> Result<bool> {
     if unexpected_output {
         let error_message = create_error_message(bin, &stdout, non_matching_stderr_lines, is_wsl);
         let command = if is_wsl {
-            "wsl '$SHELL' -lc '{bin} -v'"
+            "wsl --shell-type login -- {bin} -v"
         } else {
             "$SHELL -lc '{bin} -v'"
         };
@@ -260,11 +262,13 @@ fn nvim_cmd_impl(bin: String, args: Vec<String>) -> TokioCommand {
 }
 
 #[cfg(not(target_os = "macos"))]
-fn nvim_cmd_impl(bin: String, mut args: Vec<String>) -> TokioCommand {
+fn nvim_cmd_impl(bin: String, args: Vec<String>) -> TokioCommand {
     if cfg!(target_os = "windows") && SETTINGS.get::<CmdLineSettings>().wsl {
-        args.insert(0, bin);
         let mut cmd = TokioCommand::new("wsl");
-        cmd.args(["$SHELL", "-lc", &args.join(" ")]);
+        cmd.args(["--shell-type", "login"]);
+        cmd.arg("--");
+        cmd.arg(bin);
+        cmd.args(args);
         cmd
     } else {
         let mut cmd = TokioCommand::new(bin);
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -180,14 +180,14 @@ fn neovim_ok(bin: &str, args: &[String]) -> Result<bool> {
 
     let unexpected_output = !output.status.success()
         || !stdout.starts_with("NVIM v")
-        || non_matching_stderr_lines.len() != stderr.lines().count();
+        || !non_matching_stderr_lines.is_empty();
 
     if unexpected_output {
         let error_message = create_error_message(bin, &stdout, non_matching_stderr_lines, is_wsl);
         let command = if is_wsl {
-            "wsl --shell-type login -- {bin} -v"
+            format!("wsl --shell-type login -- {bin} -v")
         } else {
-            "$SHELL -lc '{bin} -v'"
+            format!("$SHELL -lc '{bin} -v'")
         };
 
         bail!("{error_message}{command}")


fredizzimo:
Yes, this looks much better, the `--shell-type` parameter didn't exist when the login shell launch was initially implemented here https://github.com/neovide/neovide/pull/1118. So, that's why the previous workaround had to be used.

> It is only using /bin/sh for doing which nvim and nvim -v, the wsl -- <bin> syntax does spawn the user's default shell, for example, if I run wsl -- sudo ps -ef I can see:

Can we change that to use the login shell as well? We want the path to be set, for example if `nvim` has been installed to `~/.local/bin` and added to the path by the user.

ejrichards:
Part of the issue originally was that `elvish` doesn't have the `which` command.

I'll double check tomorrow if the config parameter that sets the nvim path works through WSL to skip the `which` call. Then the concession can be on the smaller group of people like me that decide to use a more obscure shell.

Will update tomorrow.

ejrichards:
Updated to use `wsl --shell-type login -- <bin> <args>` in all cases

fredizzimo (approved):
Thanks, this looks good now, but someone else need to test before we merge.

src/bridge/command.rs
```diff
@@ -183,7 +185,7 @@ fn neovim_ok(bin: &str, args: &[String]) -> Result<bool> {
     if unexpected_output {
         let error_message = create_error_message(bin, &stdout, non_matching_stderr_lines, is_wsl);
         let command = if is_wsl {
-            "wsl '$SHELL' -lc '{bin} -v'"
+            "wsl --shell-type login -- {bin} -v"
```
fredizzimo, line 188: This fails in the CI on the nightly clippy check 
https://github.com/neovide/neovide/actions/runs/12694686463/job/35407023838?pr=2950#step:6:456

```
error: this looks like a formatting argument but it is not part of a formatting macro
   --> src\bridge\command.rs:188:40
    |
188 |             "wsl --shell-type login -- {bin} -v"
    |                                        ^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#literal_string_with_formatting_args
    = note: `-D clippy::literal-string-with-formatting-args` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::literal_string_with_formatting_args)]`

error: this looks like a formatting argument but it is not part of a formatting macro
   --> src\bridge\command.rs:190:26
    |
190 |             "$SHELL -lc '{bin} -v'"
    |                          ^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#literal_string_with_formatting_args
```

It's not really your fault, it has been broken before too, but I have only seen the linter complain now.  I think this pr https://github.com/neovide/neovide/pull/2405 broke it. It doesn't really work as expected either, so it would ne good if it could be fixed. But it can be done separately by someone else too.

The message is supposed to show which command to run when there's a problem. Previously it showed `Check that your shell doesn't output anything extra when running` followed by the command, and now just `Please check your shell configuration` followed by the wrongly formatted command.

fredizzimo: Sorry for adding more to this, that's not your fault, but I'm on vaccation and can't do any fixing myself. This check is also wrong
```rust
    let unexpected_output = !output.status.success()
        || !stdout.starts_with("NVIM v")
        || non_matching_stderr_lines.len() != stderr.lines().count();
```
It should be 
```rust
    let unexpected_output = !output.status.success()
        || !stdout.starts_with("NVIM v")
        || non_matching_stderr_lines.len()
```
So that it fails on any stderr output excpect the ignored "screen size is bogus" lines. But the output still prints the full stderr.

ejrichards: Could switch to something like `<bin>` in the meantime?

fredizzimo: You can use `format!("wsl --shell-type login -- {bin} -v")`

fredizzimo: Actually, leave this check for now at least, I don't understand how it even works currently, but somehow it seems to do the job, so better not touch it, unless you can figure it out.

```rust
    let unexpected_output = !output.status.success()
        || !stdout.starts_with("NVIM v")
        || non_matching_stderr_lines.len() != stderr.lines().count();
```

ejrichards: Just tested it, looks like it is doing the opposite of what it should be doing based on #2374, will add a fix shortly.

ejrichards:
Tested with: `let stderr = "your 1x123123 screen size is bogus. expect trouble";`
Launches successfully

Tested with: `let stderr = "your 1x123123 screen size is bogus. expect trouble\nasdf";`
```
ERROR: Unexpected output from neovim binary:
        /usr/bin/nvim -v
stdout: NVIM v0.10.3
Build type: RelWithDebInfo
LuaJIT 2.1.1713484068
Run "nvim -V1 -v" for more info

stderr: asdf


Please check your WSL configuration.
wsl --shell-type login -- /usr/bin/nvim -v
```

fredizzimo:
Can you check the final case as well, `let stderr = "some error";? I don't see any problems, but it's worth checking.

And the output could be
```
Please check your WSL configuration. Only the output from "nvim -v" should be output when you run:
wsl --shell-type login -- /usr/bin/nvim -v
```

ejrichards:
Sure, looks good, using `let stderr = "some error";`:

```
ERROR: Unexpected output from neovim binary:
        /usr/bin/nvim -v
stdout: NVIM v0.10.3
Build type: RelWithDebInfo
LuaJIT 2.1.1713484068
Run "nvim -V1 -v" for more info

stderr: some error


Please check your WSL configuration.
wsl --shell-type login -- /usr/bin/nvim -v
```

fredizzimo:
Thank you!

PR #2967, opened 2025-01-16
docs: clarify link includes details about config file settings

wulfgarpro:
- Documentation

- No

### DIFF
--- a/website/docs/config-file.md
+++ b/website/docs/config-file.md
@@ -45,8 +45,8 @@ normal = [] # Will use the bundled Fira Code Nerd Font by default
 size = 14.0
 ```
 
-Settings from environment variables can be found in [Command Line Reference](command-line-reference.md),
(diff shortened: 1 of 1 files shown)


fredizzimo:
Yes, this is much better, we didn't reciew the original text properly. I think the original text refers to this in `Settings priority`
> 1. Settings override these settings from the environment variables, but they can be overridden
   by command line arguments.
> 2. Runtime settings. These settings can be hot-reloaded in runtime.

And only the type 1. settings are described in the commandline reference. But nowhere does it say which setting is which. I think only the font is classified as "runtime" currently, so I think your update is good for now.

 Btw, I think the settings priority section could be reworded as well. Feel free to do so if you want.

fredizzimo:
Thank you! I merged this as it is. It can be further improved later.

Issue #2439, opened 2024-03-24
Factor out platform code into platform specific modules in bridge


Kethku:
Motivated by changes in https://github.com/neovide/neovide/pull/2405

PR #2985, opened 2025-01-26
feat: integrate window blur radius control

falcucci:

- Feature

- No

blur2


fredizzimo:
I would prefer to add support to winit for this instead. Our code gets too hard to maintain with a lot of platform specific code. And it has the potential to conflict with winit which also uses the same undocumented `CGSSetWindowBackgroundBlurRadius` API.
 
In winit this needs to be exposed https://github.com/rust-windowing/winit/blob/77f1c73f06634027c509967b067df071706f11d5/src/platform_impl/apple/appkit/window_delegate.rs#L947

falcucci:
@fredizzimo I am aware of it. this is something that I was already using. In case anyone would like to use, just use this branch for now.

src/window/window_wrapper.rs
```diff
@@ -574,7 +603,8 @@ impl WinitWindowWrapper {
             self.renderer.grid_renderer.grid_scale
         );
 
-        window.set_blur(window_blurred && transparency < 1.0);
+        self.apply_blur(window_blurred && transparency < 1.0, None);
+
```
fredizzimo, line 607: This breaks the blur support on Wayland, and possible other platforms that implement it in the future.

falcucci: Nope it doesn't. the `apply_blur` has its own definition for non mac_os use cases which make it working just as before.

falcucci:
> Our code gets too hard to maintain with a lot of platform specific code

@fredizzimo also regarding to this comment, this pull request isn't the root of the problem here.

we have already a [issue](https://github.com/neovide/neovide/issues/2439) for that

falcucci:
I'll be waiting the winit upstream to stabilize some changes on main to actually send this out there, seems they are changing and removing key features such as `rwh_06` and are not totally sync using glutin-winit upstream right now. I guess it will soon be updated and then update the changes there.

falcucci:
I'll leave this as draft to test the winit pr afterwards.

PR #2987, opened 2025-01-28
fix: separate text underline from baseline by at least 1 pixel

markstegeman:
In my experience, it's common for text underline to be separated from the baseline by a minimum of 1 pixel. I'd like to see the same behavior in neovide.

Problem:
- The underline_offset metric might be less than 1 pixel below the baseline (> -1)
- draw_underline rounds after adding stroke_width / 2 to the top position, which can lead to a position that is 1 pixel higher than intended

Solution:
- underline_offset.min(-1.)
- the top position and stroke_width / 2 should be rounded separately before being added together

Here's a comparison of underlined text in neovide before and after this change. The font is Iosevka.
underline-touching-baseline

- Fix

- No

### DIFF
--- a/src/renderer/grid_renderer.rs
+++ b/src/renderer/grid_renderer.rs
@@ -271,9 +271,12 @@ impl GridRenderer {
         let stroke_width = (stroke_size * underline_stroke_scale).max(1.).round();
 
         // offset y by width / 2 to align the *top* of the underline with p1 and p2
-        // also round to avoid aliasing issues
-        let p1 = (p1.x.round(), (p1.y + stroke_width / 2.).round());
-        let p2 = (p2.x.round(), (p2.y + stroke_width / 2.).round());
+        // offset is rounded up because the stroke asymmetry for odd values of stroke_width ends up
+        // on top
+        // round p1 and p2 to avoid aliasing issues
+        let offset = (stroke_width / 2.).ceil();
+        let p1 = (p1.x.round(), p1.y.round() + offset);
+        let p2 = (p2.x.round(), p2.y.round() + offset);
 
         underline_paint
             .set_color(style.special(&self.default_style.colors).to_color())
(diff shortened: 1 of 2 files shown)


src/renderer/grid_renderer.rs
```diff
@@ -271,9 +271,12 @@ impl GridRenderer {
         let stroke_width = (stroke_size * underline_stroke_scale).max(1.).round();
 
         // offset y by width / 2 to align the *top* of the underline with p1 and p2
-        // also round to avoid aliasing issues
-        let p1 = (p1.x.round(), (p1.y + stroke_width / 2.).round());
-        let p2 = (p2.x.round(), (p2.y + stroke_width / 2.).round());
+        // offset is rounded up because the stroke asymmetry for odd values of stroke_width ends up
+        // on top
+        // round p1 and p2 to avoid aliasing issues
+        let offset = (stroke_width / 2.).ceil();
```
fredizzimo, line 277: I think this now puts the line to far down in many csses. That's probably better, but I think for more control we could draw at fractional coordinates.

For example if we align the middle of the line to a pixel coordnate of 0.75, then for even widths there should be on more pixel below.  Odd widths could be rendered at 0.5, giving equal numbers of pixels above and below.

Antialiasing is off for the lines, so Skia should still draw at exact pixels.

markstegeman: Yes, you're right. I misunderstood how the lines are being drawn, because the existing comments talk about aliasing issues. I figured the positions given to draw_line should be in whole pixels.

I've been experimenting some more, and I think the proper behavior is:
- The underline offset and stroke width should be rounded to whole pixels
- The underline offset should be measured from the baseline position of the text, snapped to a whole pixel
- The position we call draw_line with should represent the middle of the line, so its y-coordinate should be whole for even stroke widths and end in .5 for odd stroke widths

fredizzimo:
Thank you! It looks good now, and I tested it quickly.

PR #2988, opened 2025-01-28
fix: build failure on case sensitive file system

danbee:
Normally on macOS the case here wouldn't matter, but if you happen to be running a case sensitive file system (as I am) then the build fails because it can't find the icon file.

- Fix

- No


falcucci (approved):
thank you!

PR #2993, opened 2025-01-31
cli-docs: add usage of frame on macOS

kindy:
- Documentation

- No

### DIFF
--- a/website/docs/command-line-reference.md
+++ b/website/docs/command-line-reference.md
@@ -6,6 +6,10 @@ normal vim variables.
 `$` in front of a word refers to it being an "environment variable" which is checked for, some
 settings only require it to be set in some way, some settings also use the contents.
 
+Note: On macOS, it's not easy to specify command line arguments when launching Apps, you can use
+[Neovide Config File](config-file.md) or `launchctl setenv NEOVIDE_FRAME transparent` to
+apply those setting.
+
 ## Information
 
 ### Version


fredizzimo:
I think a better way to do it is to use the config file https://neovide.dev/config-file.html.

The `launchctl` way of configuring, applies to all other environment variables as well, perhaps it could be documented at the to of the file instead? After this
> $ in front of a word refers to it being an "environment variable" which is checked for, some settings only require it to be set in some way, some settings also use the contents

It could also mention that most can be set from the config file.

Setting persistent environment variables, applies to all platforms, but I think Windows and Linux users know how to do that, without a special mention in the documentation, but maybe that's not the case for macOS users?

kindy:
> I think a better way to do it is to use the config file https://neovide.dev/config-file.html.

got it, I'll take a look config-file, thanks.

> The `launchctl` way of configuring, applies to all other environment variables as well, perhaps it could be documented at the to of the file instead?
> It could also mention that most can be set from the config file.

good idea.

> Setting persistent environment variables, applies to all platforms, but I think Windows and Linux users know how to do that, without a special mention in the documentation, but maybe that's not the case for macOS users?

on macOS, it's quite easy to set environment variables under terminal, but I don't know how to set environment variables for Apps before.

fredizzimo:
Thank you!

Issue #2969, opened 2025-01-18
Wrong zero width glyphs in some letters


ronisbr:
**Describe the bug**

Neovide draws some zero width glyphs wrong depending on the letter.

**To Reproduce**

Open Neovide and paste the following text

```
P̄ P̂ Ṗ
ā â ȧ
```

The output will look like:

<img width="60" alt="Image" src="https://github.com/user-attachments/assets/79f14997-291c-4cef-9567-1c6c2db4654f" />

Notice that the bar and hat looks correctly when the preceding letter is `a` but not when it is `P`. The over dot looks wrong in both cases.

**Expected behavior**

This is a screenshot of Neovim running in Ghostty, which renders the symbols correctly:

<img width="60" alt="Image" src="https://github.com/user-attachments/assets/f8a751d6-34a0-4b03-bc90-e6452d9eacde" />

**Desktop (please complete the following information):**
 - OS: macOS 15.2
 - Neovide Version 0.14.0
 - Neovim Version v0.10.3

fredizzimo:
I think this is font dependent, which font are you using? 

In any case it sounds like a duplicate of https://github.com/neovide/neovide/issues/2539, but I'm leaving this open for now, since from the user perspective the issue is different.

ronisbr:
Hi @fredizzimo !

I am using Jet Brains Mono in both Neovide and Ghostty.

ronisbr:
@fredizzimo 

I found another bug related to fonts. I am not sure if it is related to this issue and / or #2539 .

When I am using some fonts, like Monaco, que vertical lines are not continuous in Neovide. Here is what neovim looks like in Neovide using Monaco font:

<img width="807" alt="Image" src="https://github.com/user-attachments/assets/b34c7bc2-46c5-4d21-b9fe-5a731565356f" />

Here is the same configuration in Ghostty:

<img width="812" alt="Image" src="https://github.com/user-attachments/assets/c8a5608c-54b8-4a1f-a965-77e6ea048b4e" />

I tried changing `vim.o.linespace`  but it does not work even with very negative values.

fredizzimo:
That's probably https://github.com/neovide/neovide/issues/2491. I hsve not verified, but for the block characters it looks like they might not be avsilable in the font at all, so a fallback font is used, which has totally different proportions.

Using a nerd font improves that a bit, because it adds those to all fonts.

ronisbr:
[...]

Issue #2539, opened 2024-05-12
Ligatures are drawn at wrong character offset


valleydali:
**Describe the bug**
I am using the font Monaspace Neon 13 with all ligatures enabled.  This text:
```
a//a  b==b  c->c
```
should be displayed as follows (this screenshot is from TextEdit):
<img width="143" alt="Screenshot 2024-05-08 at 8 08 44 PM" src="https://github.com/neovide/neovide/assets/37566098/ea8fbcb5-e583-40c7-b175-a1eb32e903aa">

But Neovide displays it like so:
<img width="143" alt="Screenshot 2024-05-08 at 8 09 10 PM" src="https://github.com/neovide/neovide/assets/37566098/8bd0e811-b9ab-4237-8234-58b33e274474">

**To Reproduce**
1. Install the font. I installed it using Homebrew:
```
% brew tap homebrew/cask-fonts
% brew install --cask font-monaspace
```
2. Set up `config.toml` to use Monaspace Neon and enable ligatures. Here is my `config.toml`:
```
[font]
normal = ["Monaspace Neon"]
size = 13

[font.features]
"Monaspace Neon" = [ "+calt", "+liga", "+ss01", "+ss02", "+ss03", "+ss04", "+ss05", "+ss06", "+ss07", "+ss08", "+ss09" ]
```
3. Launch `neovide` and enter the test text: `a//a  b==b  c->c`

**Expected behavior**
The display should match the TextEdit screenshot above. Instead, as shown in the Neovide screenshot above, each ligature is drawn starting one character position to the left of where it should be drawn.

**Desktop (please complete the following information):**
 - OS: macOS 14.4.1 (Sonoma)
 - Neovide Version: 0.12.2, installed via Homebrew
 - Neovim Version: 0.9.5, installed via Homebrew

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/15284700/neovide_rCURRENT.log)

Kethku:
Yup that looks wrong. Let me take a look

Kethku:
Weirdly I cant reproduce this issue on current main. Can you try building from main?

Kethku:
actually i suspect this might be a bug with how swash handles ligatures from the most recent update to monaspace. Im running the initial release of the font which may explain why im not able to reproduce it.

[...]

PR #2998, opened 2025-02-05
fix: Glyph offsets within a single cluster

fredizzimo:
Use the shape information from Swash for positioning the individual glyphs of a cluster, so that characters consisting of multiple glyphs are rendered correctly. Note that each cluster is still aligned to the grid.

The y-offset was also reversed and is now fixed. This could be seen as accents being too low, especially on capital letters. I haven't seen a bug report for it though.

* Fixes https://github.com/neovide/neovide/issues/2969
* Fixes https://github.com/neovide/neovide/issues/2539

- No


fredizzimo:
I merged this so that we get more testing

PR #3009, opened 2025-02-11
feat: Use multiple cursor_vfx at the same time

zooeywm:
- Feature

vim.g.neovide_cursor_vfx_mode => vim.g.neovide_cursor_vfx_modes

For example: 

before: `vim.g.neovide_cursor_vfx_mode = "ripple"`

after: `vim.g.neovide_cursor_vfx_modes = { "railgun", "ripple" }`

https://github.com/user-attachments/assets/3c12c2f9-682c-49eb-8a70-fda9b26f903c


zooeywm:
Resolves #3009

fredizzimo:
The effect seems to be broken even on main for me, it seems to appear almost randomly. So, I think that needs to be fixed first, before being able to test this.

Also, I would like to avoid a breaking change by keeping the variable name, and let it accept either single string or an array.

zooeywm:
Thank you for your feedback! I understand the issue with the effect being broken on main, and I agree that it should be fixed first. Regarding the variable name, I also support keeping it unchanged to avoid a breaking change and allowing it to accept either a single string or an array.

However, I wanted to let you know that I’ve been quite busy recently, so I might not be able to address these issues immediately. Once my schedule clears up, I’ll prioritize working on the fixes and improvements.

Thank you for your patience and understanding!

fredizzimo:
I made a fix for the cursor vfx here https://github.com/neovide/neovide/pull/3052.

fredizzimo:
Another fix, but for the point highlight particles https://github.com/neovide/neovide/pull/3053

fredizzimo:
@zooeywm, the cursor effects on main are fixed now. So, when you have time you can rebase and fix the issues

zooeywm:
@fredizzimo done, please have a check

fredizzimo:
Thanks. 

I noticed one bug. It does not clear the previous effect when you try to switch modes at runtime. I think we need to get that fixed, otherwise it's hard to try out the effects. That also includes trying to turn off the effect by setting it to an empty string. Otherwise, this appears to work fine,

Can you also update the documentation to reflect that it can take an array of parameters?

Finally, the CI failed during linting, can you fix that?

src/renderer/cursor_renderer/mod.rs
```diff
@@ -55,9 +56,10 @@ impl Default for CursorSettings {
             trail_size: 0.7,
             unfocused_outline_width: 1.0 / 8.0,
             smooth_blink: false,
-            vfx_mode: cursor_vfx::VfxMode::Disabled,
+            vfx_mode: cursor_vfx::VfxModeList::default(),
             vfx_opacity: 200.0,
             vfx_particle_lifetime: 1.2,
+            vfx_particle_highlight_lifetime: -1.,
```
fredizzimo, line 62: I will let you judge a suitable default for these two. The default of 1.2 for the particle lifetime is likely too high, since the meaning was changed and even more so for the highlight effects.

fredizzimo (changes requested):

fredizzimo:
There's still a rustfmt CI error, but otherwise it looks good now. I think the lifetime parameters could be documented a bit more, but that's something that can be done later too.

fredizzimo:
I fixed the formatting and updated the documentation.

fredizzimo (approved):

fredizzimo:
Thank you!

PR #3010, opened 2025-02-11
feat(env): add NEOVIDE_CONFIG env var

amadaluzia:
- Feature

- No

## What does this PR add?
This PR allows Neovide to find its configuration via the `$NEOVIDE_CONFIG` environment variable. This allows for wrappers to be made without lying to Neovide about `$XDG_CONFIG_HOME` and any potential issues that could arise from that.


midirhee12:
Very cool

falcucci (changes requested):

src/settings/config.rs
```diff
@@ -28,7 +28,16 @@ fn neovide_config_dir() -> PathBuf {
 }
 
 pub fn config_path() -> PathBuf {
-    let mut config_path = neovide_config_dir();
+    let mut config_path = match env::var("NEOVIDE_CONFIG") {
+        Ok(val) => PathBuf::from(val),
```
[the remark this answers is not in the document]
falcucci: here is a tip for you:

```rust
pub fn config_path() -> PathBuf {
    let mut path = env::var("NEOVIDE_CONFIG_PATH")
        .ok()
        .map(PathBuf::from)
        .filter(|path| path.is_dir() && path.exists())
        .unwrap_or_else(|| {
            eprintln!("Not a valid config path! Using default value.");
            neovide_config_dir()
        });

    path.push(CONFIG_FILE);
    path
}
```

src/settings/config.rs
```diff
@@ -28,7 +28,16 @@ fn neovide_config_dir() -> PathBuf {
 }
 
 pub fn config_path() -> PathBuf {
-    let mut config_path = neovide_config_dir();
+    let mut config_path = match env::var("NEOVIDE_CONFIG") {
```
falcucci, line 31: `NEOVIDE_CONFIG` could confuse people to point it out to the file itself, rename it to `NEOVIDE_CONFIG_PATH` would make it explicit to avoid such confusion.

falcucci (dismissed):
some docs are missing as well, thanks

src/settings/config.rs
amadaluzia: I decided to leave `NEOVIDE_CONFIG` the same because I now changed it to point to the `configuration.toml`.

src/settings/config.rs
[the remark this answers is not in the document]
amadaluzia: Thanks for the tip! I am admittedly new to Rust so I thought this was very readable, I just didn't know you could do something like that. :smile: 

Checking to see if the directory also crossed my mind for some reason.

fredizzimo:
The latest version does not compile. I also think that it will always display an error message if the environment variable is not set.

amadaluzia:
> The latest version does not compile. I also think that it will always display an error message if the environment variable is not set.

I'll look at it by today or tomorrow.

website/docs/config-file.md
```diff
@@ -20,6 +20,8 @@ There are two types of settings:
 | macOS    | `$XDG_CONFIG_HOME/neovide/config.toml` or `$HOME/.config/neovide/config.toml` | `/Users/Alice/Library/Application Support/neovide/config.toml` |
 | Windows  | `{FOLDERID_RoamingAppData}/neovide/config.toml`                               | `C:\Users\Alice\AppData\Roaming/neovide/config.toml`           |
 
+You may use a different location by modifying the `$NEOVIDE_CONFIG` environment variable.
```
fredizzimo, line 23: Can you add that the variable should contain the full path to an existing file?

fredizzimo:
Thank you!

Issue #3012, opened 2025-02-13
Cannot map <A-S-0> on German Layout


uloco:
**Describe the bug**
the keymap doen't work on neovide, but it works in my terminal (iterm2).
I am using a German keyboard layout, the = sign is on 0 and activates with shift.

```lua
vim.keymap.set('n', '<A-S-0>', '<C-W><C-=>', {noremap = true})
```

**To Reproduce**
1. Use a german keyboard layout
2. Try to set the mapping above

**Expected behavior**
Keymap should work properly

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOS 15.3 (24D60)
 - Neovide Version 0.14.0
 - NVIM v0.10.3, Build type: Release, LuaJIT 2.1.1736781742

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

[neovide_rCURRENT.log](https://github.com/user-attachments/files/18782324/neovide_rCURRENT.log)

fredizzimo:
You should map it using the resulting character like this
`vim.keymap.set('n', '<A-=>', '<C-W><C-=>', {noremap = true})`

uloco:
This also does not work

uloco:
I will try it on neovide, but I used to have this map and it stopped working on neovim in the terminal after some update.

fredizzimo:
Another thing on macOS, you also need to set https://neovide.dev/configuration.html#macos-option-key-is-meta

uloco:
Unfortunately <A-=> neither works on neovim nor neovide. option-key-is-meta is set for the left meta key. <A-S-0> works on neovim but not on neovide.

fredizzimo:
I should have read the log, I  just couldn't believe that we have had this bug for so long without further reports.The logs show that window detects `M-=`, but we send `M-0`

The problematic code seems to be this https://github.com/neovide/neovide/blob/9ff21b850e9ace6cb342a20c93f263dc8f878dd4/src/window/keyboard_manager.rs#L174

It's supposed to handle some special case, but I'm not sure what. And I don't have a mac myself to test on.

@falcucci, could you check this out?

falcucci:
yep I can reproduce it and it sends `M-0` as you precisely identified @fredizzimo 

It seems that this should fix the problem tho?

```rust
#[cfg(target_os = "macos")]
if self.meta_is_pressed {
    return key_event
        .text_with_all_modifiers()
        .map(|text| self.format_key_text(text, false));
}
```

PR #3014, opened 2025-02-14
fix: remove redundant code for macos key event handling

falcucci:
- Fix

- No

fixes https://github.com/neovide/neovide/issues/3012


fredizzimo:
Let's merge this, and see if someone reports any new breakages

PR #3015, opened 2025-02-15
fix: Fix mouse manager rounding

fredizzimo:
The rounding done by the mouse manager was a bit inaccurate, which for example made it impossible to click the last column with some window sizes. This is now fixed

- No


1
falcucci, line 1: This can be deleted

falcucci (changes requested):
just a minor change

falcucci (approved):

Issue #2088, opened 2023-10-22
Support image.nvim / Kitty Graphics Protocol / image rendering in Neovide


Garoth:
Kitty, Konsole, WezTerm (and others) - have been exposing an image rendering API. Image.nvim provides a compelling implementation, rendering images in buffers, allowing for scrolling, images directly in markdown files, lua api, image filters

https://sw.kovidgoyal.net/kitty/graphics-protocol/
https://github.com/3rd/image.nvim

Imo this is exactly what Neovide is for. Sensible, non-intrusive extensions to Neovim's power. This is the kind of feature that's only possible when the terminal emulator and the editor are working together. There's already dozens of projects that support the Kitty Graphics Protocol approach, with countless applications - that make Neovim even more awesome

I may be personally convinced to help unless there's someone with  more time/passion available. Though I'd like to know the feasibility given the ongoing changes in the rendering engine

ZyX-II:
Neovide is not using a terminal and based on what I see `image.nvim` is directly writing to a terminal which will do nothing in the setup when neovim and neovide are connected via a socket. I do not say however it is impossible, just without image rendering support by neovim itself it would mean that image.nvim and neovide will both have to be modified — which is not a good thing because `image.nvim` is not the only project that displays images in neovim..

Garoth:
> Neovide is not using a terminal and based on what I see `image.nvim` is directly writing to a terminal which will do nothing in the setup when neovim and neovide are connected via a socket. I do not say however it is impossible, just without image rendering support by neovim itself it would mean that image.nvim and neovide will both have to be modified — which is not a good thing because `image.nvim` is not the only project that displays images in neovim..

I'm not sure what you're saying here. The graphics protocol works like terminal input, so you can stream it to whatever terminal emulator supports it, over the network or whatever. Neovide is a terminal emulator. Yes, there are several different attempts, and image.nvim supports two different methods. However Neovide supports 0, which is tremendously behind everyone. Even iTerm has their own version of this - but the Kitty Graphics Protocol is the most likely to win out. WezTerm supports pretty much every method. Neovim itself doesn't need to be modified, plugins can interact with these protocols directly - though I can see a future where they become standardized in core. `image.nvim` would not need to be modified if we support the protocol to handle the terminal imput & escape sequences

fredizzimo:
I actually started working on this a couple of months ago and will eventually pick it up again. 

[...]

PR #3039, opened 2025-03-05
feat: image rendering support

fredizzimo:
Initial image rendering support, to be used together with https://github.com/neovim/neovim/pull/31399.

* The backend registration is very hacky at the moment
* The loading is completely synchronous
* There is no error handling (it just crashes with wrong parameters)
* Partially implements https://github.com/neovide/neovide/issues/2088

The following
```lua
local filename = "assets/neovide-1024.png"
local img = vim.img.load({ filename = filename })
vim.img.show(img, { pos = { row = 10, col = 55 } })
vim.img.show(img, {
    pos = { row = 27, col = 83 },
    size = { width = 24, height = 19 },
    crop = { x = 200, y = 750, width = 160, height = 275 },
})
```
renders this
image

- No


fredizzimo:
I added basic support for some of the Kitty protocol. Well, parts of it works over RPC and not with terminal escape codes, but the unicode placeholders are part of the Neovim grid.

This works together with my fork and branch of snacks.nvim here https://github.com/fredizzimo/snacks.nvim/tree/neovide-image

It's still not ready for merging, more of the protocol needs to be implemented first. And I'm also planning to make the official Neovide image support use the kitty functions and share the code instead of using its own functions. But first I need to implement the absolute placement parts of the Kitty protocol.

PR #3043, opened 2025-03-08
Modified macOS Icon

austinlongmn:
I created (using an [online generator](https://www.candyicons.com/free-tools/app-icon-assets-generator)) a simple Neovide icon that looks more at home in the macOS ecosystem.

Apple provides guidelines for the way macOS icons should appear. Using the plain Neovide icon is OK, but it seems out of place against all the other apps' icons.

I created a new file `assets/neovide-macos-1024.png` and modified the references in `macos-builder/make-icns` to use the macOS icon instead.

Note that I had to delete `target/aarch64-apple-darwin` before running `macos-builder/run` because it seems to cache some parts of the bundle.

- Other

- No


austinlongmn:
Actually, I realized that the website I used to add padding to the image was for personal use only. I do not wish to create legal issues, so I will close this PR without merging.

Issue #3030, opened 2025-02-28
Neovide crash on mac os when using nushell as default shell


FractalDesigns:
**Describe the bug**
Neovide crash when opened from mac app but when launched from terminal it opens fine.

**To Reproduce**
1. Set nushell as default shell
2. open neovide from Applications folder
3. See error

**Expected behavior**
Expect to open Neovide without error.

![Image](https://github.com/user-attachments/assets/aaf00137-d204-4e32-bb04-ab1c016705b8)

**Screenshots**
I get following error :

![Image](https://github.com/user-attachments/assets/4f4b901b-1bae-4d11-b2c0-df19cff8a9fd)

**Desktop (please complete the following information):**
 - OS: macOS 15.3.1
 - Neovide Version 0.14.1
 - Neovim Version 0.10.4

**Additional context**
I installed neovide with homebrew : brew install --cask neovide

alloc33:
Same

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/cf063995-9caa-4838-8080-25a4f2ab7119" />

fredizzimo:
I think this should be fixed here
*  https://github.com/neovide/neovide/pull/3045

But it's untested, because I don't have a mac, so it would be great if you could test it. You can find the artifacts in the checks tabs, if you don't want to compile from source yourself.

alloc33:
> I think this should be fixed here
> 
>     * [fix!: simplify and fix the launching of neovim #3045](https://github.com/neovide/neovide/pull/3045)
> 
> 
> But it's untested, because I don't have a mac, so it would be great if you could test it. You can find the artifacts in the checks tabs, if you don't want to compile from source yourself.

I installed this branch from source. Neovide becomes unresponsive (but only using app bundle). Works fine if I run `neovide` from terminal.

Console logs:
```
default	02:41:20.438983+0700	lsd	com.neovide.neovide/Unknown Persona:5:5:1:1:Building bundle record for app
default	02:41:20.439205+0700	lsd	com.neovide.neovide/Unknown Persona:5:5:2:1:Built bundle record for app
default	02:41:20.554812+0700	lsd	com.neovide.neovide/Unknown Persona:5:4:2:1:_LSServerRegisterItemInfo result = 0
default	02:41:20.557039+0700	dmd	Received xpc stream event (distributed notification matching) with name: com.apple.LaunchServices.applicationRegistered user info: {
    CFBundleDisplayName = Neovide;
    bundleIDs =     (
[...]

Issue #3026, opened 2025-02-24
Neovide crash due to stdout from neovim binary (duplicated with old issue)


KhoaLee:
**Describe the bug**
Open the neovide app via the icon app on macOS instead of using CLI: neovide --fork. Error is in screenshot section

**To Reproduce**
1. Go to neovide.app
2. Click and open the neovide.app 

**Expected behavior**

**Screenshots**

![Image](https://github.com/user-attachments/assets/b59c38ec-5e51-4bc5-b601-f2ff874b7cae)

**Desktop (please complete the following information):**
 - OS: MacOS Sequoia 15.3.1
 - Neovide Version 0.14.1
 - Neovim Version 0.10.4

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

**Additional context**

fredizzimo:
You have some problem with you shell configuration, it prints "Reverting to default Node version" to stderr. 

Since Neovide uses stdio to communicate with Neovim, there can't be any extra output when the shell is launched as a non-interactive login shell, or the communication will break.

stormherz:
@fredizzimo It was actually working in 0.14.0. Should we really consider `stderr` always problematic even if `stdout` has provided proper output in expected format?

It's not always possible to just fix the shell. In my case, I'm running via WSL and using system-wide proxy on the host system and WSL will always spit out `wsl: A localhost proxy configuration was detected but not mirrored into WSL. blahblahblah...` in the shell. I can't really disable it on WSL side, but that prevents neovide 0.14.1 from booting up.

The simplest solution for me is to disable proxy, run neovide, then enable proxy, not a big deal, but takes time and there's looots of software that may litter shell output.

Judging from the command being invoked (I would guess you're verifying neovim version there) should we be bothered about the `stderr` output if `stdout` contains all the output we expect from neovim?

fredizzimo:
There was a bug in 0.14.0, that made it not always consider stderr output as errors, so that explain why it was working there.

You are right, we might be able to always ignore stderr, the communication does not rely on it. However, it has been assumed that stderr messages has been caused by failures to launch neovim, passing wrong arguments for example, so that's why it has been detected.

The reporting of that seem to be broken at the moment, so perhaps it's time to re-evaluate it a bit. Maybe by detecting actual crashes, or when Neovim exits too early.

stormherz:
[...]

Issue #2741, opened 2024-07-30
Neovide crashing on windows using wsl2 if brew eval is performed in zshenv


WizardStark:
**Describe the bug**
When trying to start Neovide on Windows it will crash if `eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"` is present in `~/.zshenv`.

**To Reproduce**
1. Install brew - `/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"`
2. Add `eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"` to `~/.zshenv`
3. Run `neovide --wsl` from Windows

**Expected behavior**
Neovide should open, and any binaries installed using brew should be executable. I understand Neovide does not start a login shell on startup, hence the `~/.zshenv`.

**Screenshots**
![image](https://github.com/user-attachments/assets/7cf973a7-350d-41da-a15c-4cca62977b8d)

**Desktop (please complete the following information):**
 - OS: Windows 11, 10.0.22631 with WSL2, 5.15.153.1-microsoft-standard-WSL2
 - Neovide Version 0.13.3
 - Neovim Version:
```
nvim -v
NVIM v0.10.0
Build type: RelWithDebInfo
LuaJIT 2.1.1713484068
```

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

```
TRACE [neovide] Neovide version: 0.13.3
DEBUG [neovide::settings::window_size] Loaded window settings: Maximized { grid_size: Some(278x76) }
DEBUG [neovide::bridge::command] Running which command: "wsl" "$SHELL" "-lc" "which nvim"
ERROR [neovide::error_handling] Neovide just crashed :(
This is the error that caused the crash. In case you don't know what to do with this, please feel free to report this on https://github.com/neovide/neovide/issues!

ERROR: Unexpected output from neovim binary:
	/usr/local/bin/nvim -v
stdout: NVIM v0.10.0
Build type: RelWithDebInfo
LuaJIT 2.1.1713484068
Run "nvim -V1 -v" for more info

stderr: 
	

Please check your WSL configuration.
wsl '$SHELL' -lc '{bin} -v'
```

**Additional context**
Tried to look at this myself, but with no error logs I really did not know where to start, the issue arose when I was unifying my dotfiles setup for mac, linux and wsl.
```
[...]

Issue #2024, opened 2023-09-13
Neovide takes a long time to launch on Windows 11


figurae:
**Describe the bug**
Neovide takes a long time to launch on Windows 11 when launched using the icon. I haven't measured it but it's around 2 seconds compared to less than 1 second it took before 0.11.2, so it's ~2-3 times slower.

I bisected a bit and the culprit seems to be 937decd850f2087a20e6488e42ffd1fafbec02e0. It looks like on Win11, this approach has to launch the console first to check nvim version (and I think it's even launched twice, as there are two windows that appear and disappear before Neovide itself launches), and on Win11 this takes a really long time.

Oh, and it also launches a bit longer than pre-0.11.2 when launched from the command line, but it's not as dramatic compared to launching from the icon.

**To Reproduce**
1. Launch Neovide 0.11.2 on Windows 11 by double clicking its icon
2. See that it has to launch two windows before launching Neovide and that it takes time

**Expected behavior**
I would like the previous startup time back, please. One of the main reasons I'm using modal editors is that they launch extremely quickly. If not by reverting to the previous method on Windows, then maybe by caching the information received using the new method or by providing an option to specify it manually so that Neovide doesn't have to check it every time?

**Desktop (please complete the following information):**
 - OS: Windows 11 10.0.22621
 - Neovide Version 0.11.2 and 937decd850f2087a20e6488e42ffd1fafbec02e0
 - Neovim Version 0.9.2

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
Surely that's not a real requirement, it's like 0.5 MB... attached instead.
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/12600866/neovide_rCURRENT.log)

MultisampledNight:
Hi, thanks for reporting!

I believe the relevant change comes from https://github.com/neovide/neovide/pull/1946, specifically the delay of the fork and window creation until the neovim instance has been constructed (and confirmed its versions matches). Could you (or someone who can reproduce it, I'm not on Windows) test before and after #1946? If it's not fast before it either, could you try a release where it worked fine and `git bisect` to find the relevant commit?

If #1946 is the cause, the reason for the move was to be able to detect Neovim errors before real launch and still print them if possible. A feasable approach for solving could be imo constructing the neovim instance on the neovim comms thread like before if one of both: 

1. A commandline option `--i-know-what-i-am-doing` or `--no-check-neovim-bin` as a manual override is passed, as you mentioned.
[...]

PR #3045, opened 2025-03-09
fix!: simplify and fix the launching of neovim

fredizzimo:
With https://github.com/KillTheMule/nvim-rs/pull/53, the complexity of command.rs can be almost completely eliminated. Both stdout and stderr are now reported normally, so there's no need to try to test running `nvim -v` . There's also no need to run `which`. That's now handled by `tokio::process::Command` or by the shell depending on if a shell is launched or not. 

A login shell is started for WSL and on macOS when running through the GUI like before. Otherwise `nvim` is started directly.

With https://github.com/KillTheMule/nvim-rs/pull/53, we are now tolerant to both stdout and stderr output. 
* fixes https://github.com/neovide/neovide/issues/2741
* fixes https://github.com/neovide/neovide/issues/3026

The macOS startup is now simplified (but untested yet).
* fixes https://github.com/neovide/neovide/issues/3030
* https://github.com/neovide/neovide/issues/3029
* https://github.com/neovide/neovide/issues/2960

WSL is now only started once, and therefore it's faster
* fixes https://github.com/neovide/neovide/issues/2024

- Yes, it re-opens this bug on platforms where the shell is not started. But I don't really see why we should support it, the user can trivially make a shell script that hides all complexity inside a single command.
* https://github.com/neovide/neovide/issues/2060

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -905,17 +905,6 @@ dependencies = [
 
 [[package]]
 name = "getrandom"
-version = "0.1.16"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8fc3cb4d91f53b50155bdcfd23f6a4c39ae1969c2ae85982b135750cccaf5fce"
-dependencies = [
- "cfg-if",
- "libc",
- "wasi 0.9.0+wasi-snapshot-preview1",
-]
-
-[[package]]
-name = "getrandom"
 version = "0.2.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c4567c8db10ae91089c99af84c68c38da3ec2f087c3f82960bcdbf3656b6f4d7"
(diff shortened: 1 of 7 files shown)


fredizzimo:
It might not work on macOS, the tests are failing. @falcucci it would be great if you can have a look, unless I can spot the issue.

falcucci (approved):
this seems to work seamlessly, merge when ready.

fredizzimo:
We are waiting for https://github.com/KillTheMule/nvim-rs/pull/53 to be approved, merged and released

fredizzimo:
A new nvim-rs version was released with the required changes, so I merged this.

harshasrisri:
I think this is the fix from 0.15 that breaks connecting `neovide` to remote `neovim` over `ssh` using `--neovim-bin` like this:
```
nohup neovide --neovim-bin="ssh remote_host -- /home/linuxbrew/.linuxbrew/bin/nvim" &
```

Connection to my remote_host happens over a proxy. I have tried using `neovide --server` to connect to the remote `neovim --headless --listen :6666` over a tunnel via `ssh -fNL . . .`.

But, sadly, it has not been very reliable compared to using `--neovim-bin`.

fredizzimo:
@harshasrisri, you should be able to make a shell script and pass the path to that as `--neovim-bin`
```bash
#!/usr/bin/env bash
ssh remote_host -- /home/linuxbrew/.linuxbrew/bin/nvim "$@"
```

fredizzimo:
You forgot to pass the parameters to nvim with `"$@"`, note that you need to escape it so that it's evaluated when the script is run

PR #3048, opened 2025-03-10
Enabled remapping of Command+Q on macOS

austinlongmn:
This feature enables the user to remap the `Command+Q` key on macOS.

Here is an example:

```lua
vim.keymap.set({"n", "i", "c", "v", "o", "t", "l"}, "<D-q>", "<CMD>wqall<CR>")
```

- Feature

- No


fredizzimo:
`ParallelCommand::Quit` currently does some special handling in `exit_handler.lua`. Could you move that to the lua folder and share the logic, so that we don't lose that functionality here?

fredizzimo (approved):
I think this looks good now. It's probably better to go through the `ParallelCommandQuit` like you did, so that we only have one code path for quitting, instead of calling the lua code directly.

fredizzimo:
@falcucci, can you test and merge this?

falcucci:
sure! i'll test tomorrow

PR #3051, opened 2025-03-12
linux: Support xdg activation

bbb651:
This allows the window to gain focus when it's created even with strict focus stealing prevention on wayland. On x11 it's also used for things like loading indicator between launching and the window appearing (I think we need to add `StartupNotify=true` to the desktop for this to work, I'm not very familiar with it so I didn't touch it here)
Tested on [niri](https://github.com/YaLTeR/niri) with:
```kdl
debug {
    strict-new-window-focus-policy
}
```
For more details see [this mutter blog post](https://blogs.gnome.org/shell-dev/2024/09/20/understanding-gnome-shells-focus-stealing-prevention/)

- Other

- No

### DIFF
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -21,7 +21,11 @@ use winit::{
 use winit::platform::macos::WindowAttributesExtMacOS;
 
 #[cfg(target_os = "linux")]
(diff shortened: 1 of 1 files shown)


fredizzimo:
Thank you!

PR #3054, opened 2025-03-14
feat: low latency cursor animation

fredizzimo:
The front of the cursor immediately moves to the destination, for a much snappier feel.

One thing that always bothered me is the additional latency the cursor animation causes, sometimes it's enough to make you go one step too far when travelling by word or `f`. But I have been too lazy to bother investigating and fixing it.

The issue was that all parts of the cursor were animating, so there was always some time until you got a visual indication of what the destination is. This is fixed by always snapping the leading edge of the cursor to the destination, only animating the trail

I considered making this an option, but I don't see any reason why anyone would prefer the old behaviour. If it feels too fast, then the animating length can be increased. But the default is plenty enough, at least for me, to give a good visual hint where the cursor is moving.

- No

### DIFF
--- a/src/renderer/cursor_renderer/mod.rs
+++ b/src/renderer/cursor_renderer/mod.rs
@@ -141,7 +141,7 @@ impl Corner {
         if (self.t - 1.0).abs() < f32::EPSILON {
             // We are at destination, move t out of 0-1 range to stop the animation
             self.t = 2.0;
-        } else {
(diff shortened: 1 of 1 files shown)


falcucci (approved):

falcucci:
lovely fix

xzbdmw:
This change makes me turn on cursor animation again ❤️

Issue #3059, opened 2025-03-18
Allow to disable the "low latency cursor animation"


jedrzejboczar:
**Is your feature request related to a problem? Please describe.**
I don't really like the instant animation introduced by https://github.com/neovide/neovide/pull/3054. While it might be good for short movements like `w`/`e`/`b`, it feels too sharp for me, especially for long movements (window to window) and I just had to switch back to older version of Neovide.

 **Describe the solution you'd like**
Allow to disable the new behaviour or to configure it to be instant only for short distance movements.

fredizzimo:
Have you tried increasing https://neovide.dev/configuration.html#animation-length?

fredizzimo:
Actually I just noticed, the number does not make much sense. The default is `0.06`, and you should just change it a tiny bit

jedrzejboczar:
Yes, I have value of `0.1` and changing doesn't fix my issue. If you increase this value, when the cursor moves from window to window it will jump instantly leaving a long trail. I'd say that it's rather that decreasing it helps because the instant jump of cursor start is less noticeable in contrast to cursor tail.

fredizzimo:
How are some settings like `neovide_cursor_animation_length=0.13`, `neovide_cursor_trail_size=0.1`. The animation length makes the end of the movement smoother, and the cursor trail size ensures that the trail doesn't become too overpowering.

There's also an undocumented `neovide_cursor_distance_length_adjust` setting that you can set to true. It increases the `neovide_cursor_animation_length` with the distance.

fredizzimo:
Also, I'm not sure which part of the effect bothers you. I don't notice any problems with long jumps myself. But I feel that the problem can be fixed by tuning, or slightly modifying the effect rather than adding lag back to long movements. BTW before the low latency mode, it always felt to me like long jumps were 2-3 characters short when I initially saw the cursor before it moved to the correct target.

jedrzejboczar:
Ok, next week I will try to tune the settings as you suggest and if I won't be able to find good values I'll try to make a video demonstrating what I mean.

fredizzimo:
Can you also test this https://github.com/neovide/neovide/pull/3061. With some of the suggested values from there.

jedrzejboczar:
Ok, I tested different configurations and your PR. I recorded some videos demonstrating the different behaviors:

[...]

PR #3061, opened 2025-03-22
feat!: more configurable cursor animation

fredizzimo:
https://github.com/neovide/neovide/pull/3054 added low latency cursor movement, but unfortunately the new behaviour doesn't suite everyone. This therefore adds more configuration possibilities. 

The `neovide_cursor_animation_length`, now controls the animation length exactly in milliseconds. And the trail size controls how much faster front of the cursor moves in relation to the back. With the default value of 1.0 the front jumps to the destination immediately, with minimum lag. For a smoother experience you can lower the value at the cost of more lag.

* fixes https://github.com/neovide/neovide/issues/3059

- Yes, the meaning of the values changes, and the `neovide_cursor_animation_length` quite drastically, the previous default was 0.06, which did not match the previous description of being specified in seconds.

We discussed this on discord, and I think it makes sense to do a big breaking change and change the defaults of all the cursor animations in the 0.15.0 release.

### DIFF
--- a/src/profiling/profiling_enabled.rs
+++ b/src/profiling/profiling_enabled.rs
@@ -317,8 +317,11 @@ macro_rules! tracy_named_frame {
 
 #[macro_export]
 macro_rules! tracy_plot {
-    ($name: expr, $dt: expr) => {
+    ($name: literal, $dt: expr) => {
         $crate::profiling::_tracy_plot($crate::profiling::cstr!($name), $dt)
+    };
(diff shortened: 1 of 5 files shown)


fredizzimo:
This will most likely change completely. It does not quite give us the full control that we want. Some other type of curve/configuration is needed. And perhaps also some control of the front of the cursor.

fredizzimo:
I merged this to start preparing for the 0.15.0 release and also to hopefully get some more testing before that.

PR #3067, opened 2025-03-25
Add background image continuation

Lartrax:
This pr relates to issue: #342 and pr: #2419.

This is a continuation on mysty0s pr.

Adds two new settings `vim.g.neovide_background_image` and `vim.g.neovide_background_transparency`.

- Feature

- No


src/renderer/mod.rs
```diff
@@ -307,9 +313,22 @@ impl Renderer {
...
+                    root_canvas,
+                    (opacity * 255.0) as u8,
+                    Background {
+                        color: default_background.with_a((255.0 * background_transparency) as u8),
+                        image: background_image,
+                    },
+                    &screen_rect,
+                    grid_scale,
+                )
+            })
```
fredizzimo, line 331: There's something wrong here, probably with these changes, but it could be something else. Without setting any transparency option, or background image, the border becomes transparent

image

fredizzimo, line 331: This is probably also a problem with this code, the normal opacity setting does not work properly (the right is Neovide main)
image

Without a background image it does not function at all
image

fredizzimo, line 331: The rendering stops working after resizing the window with `neovide_opacity` set and maximizing the window
image

website/docs/configuration.md
```diff
@@ -233,6 +233,32 @@ window to that value.
...
+```
+
+Lua:
+
+```lua
+vim.g.neovide_background_transparency = 0.8
+vim.g.neovide_background_image = "~/image.jpg"
+```
+
+Using a background image in conjunction with `g:neovide_opacity` can lead to some nice looking results. Especially when using a more subtle texture instead of a regular image.
```
fredizzimo, line 258: I think we need to document that the feature requires either `neovide_opacity` or `neovide_normal_opacity`, otherwise the image will be hidden behind the background

fredizzimo (changes requested):
Thanks for attempting to continue this!

Note: I didn't review very carefully yet, due to so many basics things being broken and I did not check the code at all.

OneOfOne:
are there any updates on this?

Lartrax:
Sorry for the long quiet. I don't see myself having time to continue this pr in the near future. I think there was some talk of supporting images more generally throughout Neovide, but I'm not up to date on how that's going. Maybe @fredizzimo has some input there.

@OneOfOne if you have the rust toolchain installed and don't resize your windows too often then you could compile the fork yourself and set up following the readme instuctions.

If anyone is willing to pick this up then I'm happy to transfer the fork ownership, if not then we should probably close this pr so others don't think this is being actively worked on. Then they are more likely to try making the solution themselves.

fredizzimo:
Yes, we are working on general image support, but that progress is also a little bit stalled at the moment, but I plan to continue working on it quite soon. 

It will most likely not cover the full use case of a background image though, since the plan is to only render images attached to neovim buffers, and there's no buffer covering the whole window, and especially not the border area outside of the Neovim area.

I will close this PR for now, it can be re-opened again if someone else picks it up.

Issue #3091, opened 2025-04-04
box drawing: half-circle characters not full-size with jetbrains mono


emmanueltouzery:
On my setup, the new box-drawing for half-circles are too small to cover the full vertical height:

![Image](https://github.com/user-attachments/assets/e29c32da-fbb5-4c1a-86e2-5d6644d6269d)

I found out that the issue is that the bounding box is not squarish:

```
[src/renderer/box_drawing/renderer.rs:328:9] &bounds = Box2 {
    min: Point2 {
        x: 2548.0,
        y: 0.0,
    },
    max: Point2 {
        x: 2555.84,
        y: 18.0,
    },
}
```

This line picks the min of the height and the width:
```rust
        let mut radius = (bounds.size().width).min(bounds.size().height / 2.0);
```

If I change it to pick the max, the half-circle covers the full height but is slightly clipped on the right.

Presumably an ellipse should be drawn instead of a circle, to fit in the bounds.

In addition, I think the:
```rust
        // Leave a small gap between the circles, and also allow them to move a bit to the side
        // depending on the pixel alignment of the cell.
        radius -= 1.0;
```
Is also making the circle not fit the height perfectly. Maybe the the -1.0 should be applied on the horizontal direction only, when it's drawn as an ellipsis.

**Desktop (please complete the following information):**
 - OS: Fedora 41

fredizzimo:
Are you sure that you are using the right circle from the powerline? https://github.com/ryanoasis/powerline-extra-symbols. The unicode u+25d7 is smaller.

I haven't seen any problems in our tests previously, but that doesn't mean that there aren't any problems https://github.com/neovide/neovide/blob/main/src/renderer/box_drawing/box_drawing_test.txt

fredizzimo:
Can you show the powerline section of that? It's a bit further down in the file

emmanueltouzery:
i guess this is that:

![Image](https://github.com/user-attachments/assets/02a27f5f-c44c-437c-b40c-8d19a6666f3e)

![Image](https://github.com/user-attachments/assets/32690ac3-47ec-41e3-a260-0f4f9d0b4e24)

emmanueltouzery:
I thought the info that I gave regarding the non-squarish bounding box and the min() was quite compelling.

fredizzimo:
[...]

PR #3092, opened 2025-04-05
fix: powerline half circle box drawing

fredizzimo:
It was previously assumed that the box drawing powerline half circles were perfect circles. But that's not the case, they should align with the rest of the powerline symbols, so in practice be a slightly oval shaped, depending on the font.

* fixes https://github.com/neovide/neovide/issues/3091

- No


emmanueltouzery:
works perfectly for me and fixes the issue :+1:

Issue #3070, opened 2025-03-26
Support for audio recording via external applications


helins:
**Is your feature request related to a problem? Please describe.**

It seems Neovide does not support directing input from the microphone to other apps like `sox`, at least not on MacOS.

E.g. Try `:!sox -t coreaudio default output.wav`, you will see no mic activity and the output file will be silent.

I wrote some fancy plugins for things like recording voice memos with `sox`, even fancy live speech-to-text powered by Whisper. It all works beautifully in Nvim TUI but sadly not in Neovide.

**Describe the solution you'd like**

Support for microphone input.

**Describe alternatives you've considered**

I thought it was a permission problem but I see in OS settings that Neovide is allowed to access the microphone (Settings > Privacy and Security > Microphone).

**Additional context**

If it is a configuration problem, I would greatly appreciate quick feedback.
I would like to demo those mentioned features next week using Neovide, if possible :)

fredizzimo:
@falcucci, we probably need to add entitlements to the plist https://developer.apple.com/documentation/bundleresources/entitlements

On the other hand, I'm not sure if we should do that, it sounds a bit strange if Neovide starts requesting microphone access. If we can get a permission dialog to only ask when some option is set, then it might be ok.

helins:
Actually, I am not sure why, but in my `Privacy and Security` settings, microphone was already allowed for Neovide. I do not remember ever allowing that manually. So isn't the "entitlement" already granted?

I understand your concern about Neovide, a text-editor, asking for microphone access, however:

- Permissions are typically requested on an as-needed basis, aren't they? For instance, accessing my `Downloads` directory with Telescope would initially request access in a modal
- In any case, it would be very sad to forbid things like speech to text in Neovide. In this AI age, that would sooner than later become a dealbreaker and Neovide is too awesome to be left behind. If Nvim TUI can do it, Neovide should be able to do it it.

helins:
After more research, there might be a confusion.
@falcucci, please correct me if I am wrong 🙏


1. I assume we are talking about the `NSMicrophoneUsageDescription` entitlement
    - <https://developer.apple.com/documentation/BundleResources/Information-Property-List/NSMicrophoneUsageDescription>
    - Which indeed does not currently appear in `Info.plist`
2. In my settings, Neovide actually does have permission to access the microphone
    - As shown in `Security & Privacy > Privacy > Microphone`
    - See screenshot below
3. How come Neovide even shows up in my settings if the entitlement is not requested?
[...]

PR #3095, opened 2025-04-06
feat: audio input support on macos (#3070)

helins:

Audio input is now possible on MacOS. As usual, the OS requires permission from the user on request, so nothing intrusive.

No


fredizzimo:
Thank you!

Based on your description she video I thick this is good.

PR #3109, opened 2025-04-16
fix(renderer): use saturating_sub for safer arithmetic in

davehorner:
rendered_window.rs
Neovide panics on Windows when the scrollback buffer has fewer lines than the window height. The subtraction

let max_delta = self.scrollback_lines.len() - self.grid_size.height as usize;
underflows when self.scrollback_lines.len() < self.grid_size.height as usize, causing:

Neovide panicked with the message 'attempt to subtract with overflow'. (File: src\renderer\rendered_window.rs; Line: 499, Column: 29)
This is a bug and we would love for it to be reported to https://github.com/neovide/neovide/issues
note: run with RUST_BACKTRACE=1 environment variable to display a backtrace

Steps to Reproduce
Launch Neovide on Windows with a fresh session.

Ensure the terminal buffer has fewer lines than the window height (e.g. start Neovide and do nothing).

Scroll up (e.g. with the mouse wheel or <C-y>).

Observe the panic.

- Fix

No

### DIFF
--- a/src/renderer/rendered_window.rs
+++ b/src/renderer/rendered_window.rs
@@ -496,7 +496,7 @@ impl RenderedWindow {
         if scroll_delta != 0 {
             let mut scroll_offset = self.scroll_animation.position;
 
-            let max_delta = self.scrollback_lines.len() - self.grid_size.height as usize;
+            let max_delta = self.scrollback_lines.len().saturating_sub(self.grid_size.height as usize);
             log::trace!(
                 "Scroll offset {scroll_offset}, delta {scroll_delta}, max_delta {max_delta}"
             );


fredizzimo:
Thank you. But I fear there's a bug somewhere else, so it would be good to figure out what's going on and fix the root cause instead.

Also, I can't reproduce the issue here, neither have I seen the issue reported. Maybe it's some plugin that triggers it? If you can post a logfile I can help finding the root cause, in case you don't find it yourself.

davehorner:
Ok.  I think this sort of panic should be protected.  but it does make sense that it would be something you haven't seen before.

If you would like to see how this occurs.  Please have a look at the code in the library interface I'm proposing
https://github.com/neovide/neovide/discussions/3108

 I will look into this further.  cheers.

Issue #1931, opened 2023-07-11
Preedit is not displayed when using IME on Neovide.


ZOI-dayo:
**Describe the bug**
Preedit is not displayed when using IME on Neovide.

https://github.com/neovide/neovide/assets/64856587/b812ff8c-ae13-446b-be8a-bad44219d259

**To Reproduce**
1. Open Neovide
2. Enter Japanese

**Expected behavior**
You should be able to see the string being converted even while you are typing using an IME.

↓ This is an example of expected behavior. (MacOS TextEdit)

https://github.com/neovide/neovide/assets/64856587/a5d2f097-6956-4a35-af32-da2dbb3177d5

**Screenshots**
There is a screen record in "Describe the bug" section.

**Desktop (please complete the following information):**
 - OS: macOS 13.3.1
 - Neovide Version : 372f29b79bebf2c2f40bf49bc2ad389ac9cf3369 (HEAD)
 - Neovim Version : NVIM v0.9.1

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide_rCURRENT.log](https://github.com/neovide/neovide/files/12014933/neovide_rCURRENT.log)

**Additional context**
I an using Google日本語入力(aka "mozc")
But the same bug occurs with the IME preinstalled in MacOS.

fredizzimo:
I have an experimental branch of that here https://github.com/neovide/neovide/compare/main...fredizzimo:neovide:fsundvik/ime, but note that it's a bit behind neovide main right now, so for example the IME fixes for X11 are not included, so it does not work on X11. The other platforms should be fine though.

But it's not completely ready, selections does not work, transparency does not work, and it also does not deal with line wraps in any way.

I will finish it up, only after https://github.com/neovide/neovide/pull/1870, since the transparency fixes will be much easier with that.

sevenc-nanashi:
Any updates on this issue? I'm facing to this problem too.

fredizzimo:
I'm not sure if it's working, but I rebased https://github.com/neovide/neovide/compare/main...fredizzimo:neovide:fsundvik/ime

It should work on all platforms, but it's till missing proper transparency support, selections, and some kind of wrapping or scroll when the window edge is reached.

If someone wants to continue working on it feel free to notify me and take it over. I'm not sure when I will return to it yet.

OliverChao:
[...]

PR #3110, opened 2025-04-18
feat: add Lua API of IME event handling

kanium3:
- Feature

- No

## About

This pull request introduces support for handling IME (Input Method Editor) preedit and commit events in Neovide, enabling better text input for languages requiring complex input methods. The changes include adding Lua scripts for IME handling, modifying the Rust backend to integrate with these scripts, and updating the keyboard manager to process IME-specific events.
Related to #1931 

## Sample

https://private-user-images.githubusercontent.com/67541916/444781118-15f24fbb-98d3-4865-b2b1-f6eacb42243d.mp4?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NTM1MjExMDcsIm5iZiI6MTc1MzUyMDgwNywicGF0aCI6Ii82NzU0MTkxNi80NDQ3ODExMTgtMTVmMjRmYmItOThkMy00ODY1LWIyYjEtZjZlYWNiNDIyNDNkLm1wND9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTA3MjYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUwNzI2VDA5MDY0N1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWZhYjAwMjhiM2M2YjU4ZGVhM2U5YWI1NDA0ZDIxZTdlNDdjNzg4ZDFhZjM3NTI2ODVhMTJmYzFkM2M2YmQ2MjQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.JuTd8C9uqsemVVSIFeHeejgWT8LPVSvSeFIdhhACz0s


fredizzimo:
Thank you, I will take a look at this a bit later (in a few days)

It's possible that it would make sense to just implement the base API in Neovide, and implement the rest in a separate plugin for easier development, and possible faster updates. But I'm not sure yet.

kanium3:
I don't know what you're unsure about, but there are plugins like equalsraf/neovim-gui-shim that separate GUI-related functions. I can't say that it works stably yet, so separating it as a plugin would probably be better in terms of development. Also, regarding handling IME with a Lua Plugin, in terms of performance, the focus is on the implementation using NeoVim's autocommands rather than on plugin loading.

Anyway, if you're going to merge only the base API, please let me know so I can make the change.

chyuaner:
Hi! I build and use your commit id: e58082d4cbb384874239dcdc7e61bd1e13004d01
In Manjaro KDE Plasma 6 Wayland + Fcitx5 chewing.

I try to insert mode with ime input at 3rd lines. 
But always insert at first line end character.

And I use direction keys move cursor.

Display Bug like this image:
Screenshot_20250421_052143

Is freeze status. Press Enter key is not response.

chyuaner:
I Record Screen. And found another bug!
If I have swp file will freeze. 
and output log:
```
ERROR [neovide::bridge::ui_commands] Input failed

Caused by:
    0: Error decoding response to request 'nvim_exec_autocmds'
    1: Error while reading from Reader
    2: EOF
ERROR [neovide::bridge::ui_commands] Input failed

Caused by:
    0: Error sending request 'nvim_exec_autocmds'
    1: Error writing to the Writer
    2: Broken pipe (os error 32)
ERROR [neovide::bridge::ui_commands] Input failed

Caused by:
    0: Error sending request 'nvim_input'
    1: Error writing to the Writer
    2: Broken pipe (os error 32)
ERROR [neovide::bridge::ui_commands] Input failed

Caused by:
    0: Error sending request 'nvim_input'
    1: Error writing to the Writer
    2: Broken pipe (os error 32)
ERROR [neovide::bridge::ui_commands] Input failed
```

https://github.com/user-attachments/assets/9067ed95-c732-4c4f-8679-1d8405549a01

------------

Back to yesterday's question.

> I would like to know what kind of behavior it has, so could you please send me the log data?
It doesn;t show any log. But I try to record my screen to reappear this bug.

I try to insert mode with ime input at 3rd lines.
But always insert at first line end character.

https://github.com/user-attachments/assets/2f7f5af1-2274-4a9f-a87f-fbd20f2f6442

```
處理 User Autocommands: "ImePreedit" 時發生錯誤:
Error executing lua callback: [string "<nvim>"]:63: Invalid 'end_col': out of range
stack traceback:
        [C]: in function 'nvim_buf_set text'
        [string "<nvim>"1:63: in function <[string "<nvim>"]:44>
Press ENTER or type command to continue
```

chyuaner:
> Regarding the swapfile issue, it looks like Neovim is either not responding at all, or crashed. It's possible that it's caused by something else and not these changes, @chyuaner, are you able to reproduce it without?

I change neovide version to Arch Linux Pacman Package: neovide-0.15.0-1.
And try reproduce the swapfile issue.

https://github.com/user-attachments/assets/94dffc7b-3aff-4264-9caa-dfdb5a44ba71

All hint text is show half one second and disappeared.
But key input is normal operation can use.

kanium3:
I will change my policy to implement a Lua API. I intend to separate the pre-edit preview and other functions into a separate plugin.

kanium3:
I tried to separate it into a Lua API and [a plugin that supports pre-edit](https://github.com/kanium3/neovide-ime.nvim). However, since the plugin only transferred what was written when opening pull_request, I would like to address the problem there.

fredizzimo:
What's the status of this, how well does it work currently? 

I think we could start thinking about getting this merged. But I think it probably needs some modifications, so that it works as before for IMEs that don't need pre-edit support.

kanium3:
Currently, the API part and the plugin are separated, and it is usable with the initial PR feel. 
Considering that the mechanism for inputting text into NeoVim's text buffer while supporting Preedit is handled by the plugin, I think the implementation of the part that provides the API is in the review stage.
And, For IMEs that do not require preedit support, an empty callback will be executed and the behavior should remain the same as before.

wxmup:
i am a fcitx5's user. i do not understand why this branch is not merged. i am so sad that i key chinese words in neovide.
图片

kanium3:
Thank you for your review! I've increased the amount of information that the command can hold and the number of arguments that can be passed to the handler.
I've also added a link to the FAQ explaining how to configure the Lua API.

fredizzimo:
It looks like you have the wrong line length configured. We should probably add an `.editorconfig` at some point to avoid that.

fredizzimo:
Thank you very much!

kanium3:
I'm grateful that you merged it. I'm impressed with the idea of ​​being able to customize the preedit display as a Lua API, rather than displaying it on the main body.

wxmup:
thansk you work! i am enjoy now.

sevenc-nanashi:
Thank you a lot, you're a lifesaver!!!!

fredizzimo:
@kanium3, I'm sorry but I think I will have to revert this until we resolve the issues with the freeze during messages

I'm afraid that we might need a Neovim patch to resolve it. I think that `exec_lua` is not allowed when a message is showing like  that, so the handler will never be called. But I have not debugged it completely.

Issue #2815, opened 2024-09-26
Click on folds cannot open, only close, need to "za" to re-open


barnacker:
**Describe the bug**
Works fine in nvim directly in the terminal but in neovide, can only close the folds.

**To Reproduce**
See desc and video.

**Expected behavior**
Open/Close folder with click

**Screenshots**
**On the left is classic nvim in terminal, on the right is Neovide.**

https://github.com/user-attachments/assets/66542036-938f-448d-ad90-84e873b146f6

**Desktop (please complete the following information):**
 - OS: NixOS 24.11.20240916.99dc878 (Vicuna) x86_64 
 - neovide 0.13.3
 - NNVIM v0.10.1

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide_rCURRENT.log](https://github.com/user-attachments/files/17150084/neovide_rCURRENT.log)

**Additional context**

fredizzimo:
Can you provide your configuration, or preferably a more minimal one? The default `zf` and `zo` works without any plugins installed, so some plugin is needed for this.

You can either provide a lazy based configuration, or one following the Neovim template https://github.com/neovim/neovim/blob/master/contrib/minimal.lua

I fear the problem might be caused by the suboptimal mouse handling in Neovim
* https://github.com/neovim/neovim/issues/27181
* https://github.com/neovim/neovim/issues/29135

In that case `neovide --no-multigrid` should "fix" the issue.

barnacker:
ok it works with that flag. no plugins required by the way I'm using native folding; I just changed the looks of them, but the click had the same issue before I even reskinned them with a plugin. I had this bug since day one with the barebone fold column basically.

Thank you so much!
Made my day :)

fredizzimo:
I'm re-opening this. It should work with multigrid, so that's something we or Neovim need to fix.

I wasn't aware of the `foldcolumn` setting, that's why I was not able to reproduce it. You need that for clickable folds.

fredizzimo:
Hm, for me the folds are clickable with `neovide -- --clean` (with multigrid enabled), and `set foldcolumn=auto`.

I'm not sure why they are not for you.

barnacker:
same results here, it's probably the mix with the diagnostic columns or something, multiple clickable elements

fredizzimo:
There's a fix on the way now https://github.com/neovide/neovide/pull/2919, but there might be some troubles geting things merged on the Neovim side.

Issue #2749, opened 2024-08-06
Mouse not working in statuscol


mikew:
**Describe the bug**
n the terminal and nvim-qt, I'm able to click `gitsigns` and see the diff. I'm also able to click the scrollview area to quickly jump to things. Neither works in neovide

**To Reproduce**
- Add https://github.com/lewis6991/gitsigns.nvim
- Add https://github.com/luukvbaal/statuscol.nvim
- Change a file in a git repo so you see gitsigns
- Try clicking on gitsigns

**Expected behavior**
With gitsigns, a preview of the diff should appear. With nvim-scrollview, it should scroll to the area.

**Screenshots**

terminal (konsole, but works in iTerm2 and elsewhere):

https://github.com/user-attachments/assets/39602ea3-1ef0-4b86-8f74-7b23c5b3a9ff

nvim-qt:

https://github.com/user-attachments/assets/d019d9a9-34e3-410d-ac43-01bbb503b04b

neovide:

https://github.com/user-attachments/assets/809e4948-6320-4d8a-be90-9b8dcf6992ff

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Neovide Version 0.13.3
 - Neovim Version v0.10.1

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

... a 30 second test generated a log that was ~5MB.

**Additional context**
None

fredizzimo:
This is a Neovim problem and unfortunately there's not much we can do until it's resolved there:
* https://github.com/neovim/neovim/issues/27181
* https://github.com/neovim/neovim/issues/29135
* https://github.com/neovim/neovim/issues/24801

What complicates the issue, is that it's a deliberate design choice in Neovim, so pull requests attempting to fix the issue will not necessarily be accepted until some of the maintainers there starts looking into the issue:
* https://github.com/neovim/neovim/pull/29316

mikew:
Dang, that's unfortunate. Thanks for the tip that it's due to `multigrid`, running with `--no-multigrid` makes the clicking work, and I wasn't a huge fan of the smooth scrolling (well, I am, it just works really awkward when there's network latency involved)

fredizzimo:
There's a fix on the way now https://github.com/neovide/neovide/pull/2919, but there might be some troubles geting things merged on the Neovim side.

Issue #2569, opened 2024-05-19
Inconsistent with terminal of vim.fn.screen* function


xzbdmw:
**Describe the bug**
vim.fn.screen* function only return statusline result

**To Reproduce**
repro.lua
```lua
vim.keymap.set("n", "gt", function()
    local col = vim.fn.screencol()
    local row = vim.fn.screenrow()
    local char = vim.fn.screenstring(row, col)
    -- __AUTO_GENERATED_PRINT_VAR_START__
    print([==[function#if#function char:]==], vim.inspect(char)) -- __AUTO_GENERATED_PRINT_VAR_END__
end)
```
move cursor to a char and type gt, the result is random string from statusline
**Expected behavior**
the result should be the same as cursor char
**Desktop (please complete the following information):**
macos 14.1
neovim 0.10 stable
neovide 13.0

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide_rCURRENT.log.zip](https://github.com/neovide/neovide/files/15370075/neovide_rCURRENT.log.zip)

**Additional context**

fredizzimo:
I think this might be related to this bug we reported to Neovim a while ago:
* https://github.com/neovim/neovim/issues/27181

Does it work properly if you jump using the keyboard?

xzbdmw:
Yes, mouse and keyboard is the same

fredizzimo:
I assume that you have tested this in the Neovim terminal app? It sounds strange that this is brokens, since Neovide does not manipulate any screen and cursor positions itself.

xzbdmw:
Yes, kitty print correct cursor char

yasen-pavlov:
Just wanted to chime in that I think I ran into this or a similar bug while I was trying to debug an issue I was having with the [statuscol](https://github.com/luukvbaal/statuscol.nvim) plugin. It uses `vim.fn.screenstring` to determine the sign in the folding column ([link to relevant code](https://github.com/luukvbaal/statuscol.nvim/blob/483b9a596dfd63d541db1aa51ee6ee9a1441c4cc/lua/statuscol.lua#L132)) when you click on it and it's working in the terminal, but I was getting some weird behaviour in neovide.

What I managed to figure out so far is that for some reason `vim.fn.screenstring` always returns an empty string for me in neovide. I was able to reproduce this on a clean nvim config by running the following lua code :
```
for row = 1, 3, 1 do
	for column = 1, 10, 1 do
[...]

Issue #2226, opened 2023-12-31
Mouse Scrollbar Unresponsive


mbway:
**Describe the bug**

Mouse events for [`nvim-scrollview`](https://github.com/dstein64/nvim-scrollview) do not work well in neovide.

In the terminal and neovim-qt, the scrollbar can be dragged with the mouse and is very responsive.

Up to neovide version 0.11.2, using the mouse to move the scrollbar does not update consistently.

Since neovide version 0.12.0 the scrollbar no longer responds to the mouse at all. This is probably related to https://github.com/neovide/neovide/issues/2222 but there were issues with the behaviour before as well.

**To Reproduce**

1. use this minimal `init.lua` (assuming lazy vim already installed)
```lua
local lazypath = vim.fn.stdpath("data") .. "/lazy/lazy.nvim"
vim.opt.rtp:prepend(lazypath)
vim.opt.number = true -- to show the view scrolling more easily
require("lazy").setup({
    {
        "dstein64/nvim-scrollview",
        config = function()
            require('scrollview').setup({})
        end
    },
})
```
2. run `neovide -- -u init.lua <large_file>`
3. attempt to use the mouse to move the scrollbar
4. run `nvim -u init.lua <large_file>`
5. attempt to use the mouse to move the scrollbar

**Expected behavior**
The scrollbar should respond to mouse input like in the terminal and other GUIs like neovim-qt

**Screenshots**

https://github.com/neovide/neovide/assets/4923501/f1c8dba5-4784-4b77-9326-86b8779bf6cd

**Desktop (please complete the following information):**
 - OS: Arch Linux, KDE Plasma 5, X11
 - Neovide Version 0.12.0
 - Neovim Version 0.9.4

fredizzimo:
It wonder why that plugin don’t use `<LeftDrag>`, which would work

https://neovim.io/doc/user/term.html#%3CLeftDrag%3E

But for now, I guess you need to set `mousemoveevent` to true.

Making it more responsive is tricky, because the library we are using `nvim-rs` does not support notifications, so we are forced to wait for a reponse from Neovim, before we can send the next event.  https://github.com/KillTheMule/nvim-rs/issues/7

mbway:
> But for now, I guess you need to set mousemoveevent to true

if you mean adding `vim.opt.mousemoveevent = true` to the init.lua, that seems to have no effect. 

[...]

PR #3111, opened 2025-04-18
fix: Send 0 for the mouse grid if Neovim supports it

fredizzimo:
* Split out from https://github.com/neovide/neovide/pull/2919

Fixes the following issues
* The mouse input does not work properly

This will close the following issues:
* Fixes https://github.com/neovide/neovide/issues/2569
* Fixes https://github.com/neovide/neovide/issues/2226
* Fixes https://github.com/neovide/neovide/issues/2815
* Fixes https://github.com/neovide/neovide/issues/2749

NOTE: Depends on and includes  https://github.com/neovide/neovide/pull/2919. The support detection also need to be changed before merging. We are waiting for https://github.com/neovim/neovim/pull/32535 on the Neovim side.

- No

### DIFF
--- a/src/bridge/api_info.rs
+++ b/src/bridge/api_info.rs
@@ -182,6 +182,16 @@ impl ApiInformation {
     pub fn has_event(&self, event_name: &str) -> bool {
         self.ui_events.iter().any(|event| event.name == event_name)
     }
+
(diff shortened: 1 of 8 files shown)


fredizzimo:
I managed to improve the detecting of supported Neovim nightly versions, so the code and documentation have changed a bit. Now it should be automatic in nearly all cases, but it's still possible to enable manually if it isn't.

Issue #2414, opened 2024-03-11
Hidden cursor stays hidden on macOS when neovide loses focus


jmaliksi:
**Describe the bug**
After turning on `neovide_hide_mouse_when_typing`, the mouse cursor sometimes remains hidden when moving focus between other desktop apps and neovide. 

**To Reproduce**
1. Focus on neovide window
2. alt tab to move focus to a different window
3. move mouse over unfocused neovide window
4. alt tab or click to bring focus back to neovide. Cursor disappears (bug)
5. doing a hjhjhjhj and moving the mouse unsticks the mouse hidden state and cursors appears again (workaround)

**Expected behavior**
Mouse cursor should appear as soon as any mouse event occurs

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOS 14.4
 - Neovide Version: 0.12.2—compiled from `main` this morning
 - Neovim Version: v0.9.4

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
(not actually sure how helpful the log will be and it also seems to be full of plugin noise. see additional context for more info, but also happy to provide the log if this might be more of a problem with my setup than the winit bug)

**Additional context**
It looks like the root cause is https://github.com/rust-windowing/winit/issues/2539 though [this comment](https://github.com/neovide/neovide/issues/1966#issuecomment-1686033380) mentioned that it had only been observed in macOS 13. Unfortuntely I have triggered it in macOS 14.4 so the bug seems to be alive and well.

I was able to make the mouse behave how I wanted by checking for window focus before hiding the cursor, and so was able to avoid `self.mouse_hidden` getting desynced from the "true" window cursor state
```
index fd0f586..9e62024 100644
--- a/src/window/mouse_manager.rs
+++ b/src/window/mouse_manager.rs
@@ -439,7 +439,7 @@ impl MouseManager {
                     renderer,
                     window,
                 );
-                if self.mouse_hidden {
+                if self.mouse_hidden && window.has_focus() {
                     window.set_cursor_visible(true);
                     self.mouse_hidden = false;
                 }
```

I'm super super new to rust and pretty much never do desktop application dev, so I don't know if this is a viable fix or not. Ideally I'd get the true cursor state from the window but I didn't see that in the winit API?

[...]

PR #3135, opened 2025-05-15
Check for window focus when showing or hiding cursor

jmaliksi:
This PR checks for window focus when showing/hiding the cursor to address the bug described in #2414.

## Testing
- successful local build and install on macOS 14.6.1
- verified cursor went hidden when typing in neovide window
- verified cursor stayed visible when opening iterm window via hotkey and tabbing back
- verified cursor stayed visible when tabbing to different window and tabbing back

- No


fredizzimo:
Thank you!

PR #3137, opened 2025-05-22
fix(Docs): update terminal paste suggested command

avegancafe:
If you use FZF, you probably already have bound <C-R> in terminal mode so this suggestion 100% works on everyone's terminal, unless they rebound <C-\\>C-N>

- ✅ Fix
- Feature
- Codestyle
- Refactor
- Documentation
- Other

- ✅ No


fredizzimo:
Thanks, but I think we should recommend bracketed paste, see this comment 
https://github.com/neovide/neovide/issues/1263#issuecomment-1972013043

But we need to test it on all platforms to ensure that it works first.

Issue #3106, opened 2025-04-12
Allow specifying underline offset


bodby:
**Is your feature request related to a problem? Please describe.**
I have my underlines in my terminal line up with the bottom of the cells so I can achieve the effect in the screenshot below. I can't do the same with Neovide.

**Describe the solution you'd like**
The ability to add or subtract an offset to underlines. Maybe under font settings?

**Describe alternatives you've considered**
Removing the background for my active tabs, or staying with foot (no cursor smearing, nooo).

**Additional context**
![image](https://github.com/user-attachments/assets/46d75bff-9e32-4acf-9065-c67b4760bd17)

fredizzimo:
There was already a pull request for it here https://github.com/neovide/neovide/pull/2647, but it got closed for some reason.

I think the bugs, that were found has been fixed now, so it could be resumed. @bravo1003, would you like to do that?

fredizzimo:
This was fixed by https://github.com/neovide/neovide/pull/3148

PR #3148, opened 2025-06-08
feat: Add underline offset config to adjust underline position

nomantis:
- Feature

- No

## About

This PR implements #3106.

This feature was implemented and reviewed in #2647, but the PR was closed without a resolution by the contributor. I have taken the feedback written there and reimplemented the feature.

I have also added a documentation entry to the Config File entry on the website.

## Sample

Originally this feature was made by the author to support the [lukas-reineke/indent-blankline.nvim](https://github.com/lukas-reineke/indent-blankline.nvim) neovim plugin, which has a visual annoyance if the underline does not connect with the vertical bar showing the scopes. In terminals such as kitty the underline position can already be adjusted.

https://github.com/user-attachments/assets/c8bc2e33-0015-4928-aca5-feb47d63ad72


website/docs/config-file.md
```diff
@@ -100,6 +101,9 @@ Settings `size`, `width`, `hinting` and `edging` can be found in
   - Font feature is a string with format `+<feature>`, `-<feature>` or `<feature>=<value>`,
     e.g. `+ss01`, `-calt`, `ss02=2`. `+<feature>` is a shorthand for `<feature>=1`,
     `-<feature>` is a shorthand for `<feature>=0`.
+- `underline_offset` is a float that defines the offset between the character baseline and the underline.
+  - This is decided automatically, either if the font contains the required metrics, or `-1.0` by default.
+  - Positive underline offset values will move the underline above the baseline, while negative values move it below.
```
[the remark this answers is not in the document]
nomantis: I was thinking of it as a coordinate system, where the baseline is the `x` axis. The default underline is `x = -1`, or 1 below the baseline.

Although I'm quite positive, that most people wouldn't want to move the underline above the baseline, I will invert the setting.

src/renderer/fonts/caching_shaper.rs
```diff
@@ -212,7 +212,9 @@ impl CachingShaper {
 
     pub fn underline_offset(&mut self) -> f32 {
         let metrics = self.metrics();
-        if metrics.underline_offset != 0. {
+        if self.options.underline_offset != 0. {
+            -self.options.underline_offset
```
fredizzimo, line 216: Sorry, for the delay reviewing this and for not spotting this the first time. But I think the options should contain an option, so that a setting of `0` offset is valid.

website/docs/config-file.md
```diff
@@ -100,6 +101,9 @@ Settings `size`, `width`, `hinting` and `edging` can be found in
   - Font feature is a string with format `+<feature>`, `-<feature>` or `<feature>=<value>`,
     e.g. `+ss01`, `-calt`, `ss02=2`. `+<feature>` is a shorthand for `<feature>=1`,
     `-<feature>` is a shorthand for `<feature>=0`.
+- `underline_offset` is a float that defines the offset between the character baseline and the underline.
+  - This is decided automatically, either if the font contains the required metrics, or `-1.0` by default.
```
fredizzimo, line 105: This should be changed to something like
"If not specified, the `underline_offset`, is determined automatically by the font metrics if present, or the stroke width.

src/renderer/fonts/caching_shaper.rs
nomantis: This was an oversight on my part, thank you for the feedback! I updated it, so it's now an Option, and if it has _any_ value, it takes priority over font metrics.

website/docs/config-file.md
nomantis: Updated it 🚀

fredizzimo:
Thank you!

I fixed a formatting issue with the website and merged

Issue #3150, opened 2025-06-17
statusline missing since a commit of neovim


tan-wei:
**Describe the bug**
I do a bisect with neovim-nightly, I find since this [commit](https://github.com/neovim/neovim/pull/34414/commits/2757a67a16f233a42be38253e14f0f5982bf5463), statusline could be missing. The maximize the window of neovide could reproduce this issue, but if the window changed into non-maximize, the statusline is shown correctly.

**To Reproduce**
1. Just open a simple Window

**Expected behavior**
Lualine is shown.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: archlinux
 - Neovide: 0.15.0
 - Neovim NVIM v0.12.0-dev-582+gd86d4bacc1

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide_rCURRENT.log](https://github.com/user-attachments/files/20776538/neovide_rCURRENT.log)

**Additional context**
Just before this commit, neovide works well, and I use git bitsect to confirm this.

tan-wei:
https://github.com/neovide/neovide/issues/2646 is not the case.

JeanNicolasdeLamballerie:
I have the same issue, on Windows 10. I just spent a good 20 minutes deciphering if this was my setup or neovide, but at this point I am nearly certain this is neovide related as it does not happen in terminal with virtually the same config.

In fact, I can even tell you where the bar goes (when it doesn't decide to draw random things in the middle of my screen) because I usually use a transparent bg : It is literally at the right position, but behind the background window.

![Image](https://github.com/user-attachments/assets/49593150-6bf5-463c-a1d6-9b3bfaee8685)

goes to :

![Image](https://github.com/user-attachments/assets/c3c6ff1a-4283-4ddb-a102-289a6942e445)

tan-wei:
Yes, terminal will not occur indeed. So it shows in the background?

JeanNicolasdeLamballerie:
Yes, I assume some visual wrapping is happening, where it's pushing the bottom of the window out when it's maximized. I have no idea why, though.

```lua
-- In your neovim lua configuration :

-- sets the neovide window itself to be transparent
vim.g.neovide_normal_opacity = 0.7

-- (...)
-- somewhere in the config for your colorscheme, according to their respective documentation... :
local gruvboxopts = {}  
local tokyonightopts = {}  
gruvboxopts.transparent_mode = true
  tokyonightopts.styles = {
[...]

PR #3165, opened 2025-07-13
fix: calculate grid_size everytime to avoid random blank lines appearing

cubewhy:

This pr makes Neovide calculate grid_size everytime when user run the program.

fix the problem described in
https://github.com/neovide/neovide/issues/3150

It removes the `grid_size` option in `neovide-settings.json` file

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -72,7 +72,7 @@ pub use channel_utils::*;
 #[cfg(target_os = "windows")]
 pub use windows_utils::*;
 
-use crate::settings::{load_last_window_settings, Config, PersistentWindowSettings, Settings};
+use crate::settings::{load_last_window_settings, Config, Settings};
 
 pub use profiling::startup_profiler;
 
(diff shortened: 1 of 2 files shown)


fredizzimo:
Thank you, and sorry for not reacting to the issues and this PR before. I took a break from Neovide and been using a custom Neovim version, so I didn't realize things were this broken. 

As for the fix, I will use that as a backup, if I don't find the actual problem soon as there must be a better way to fix the issue, unless it's something that needs to be fixed on the Neovim side.

cubewhy:
Thanks!

cubewhy:
> unless it's something that needs to be fixed on the Neovim side.

Neovim works fine with kitty, I think it has no chances to be a Neovim problem

By the way, after I increased the grid_size option, Neovide works fine. I think there are some problem getting the grid_size value. Neovide dumps a smaller value to .json file.

fredizzimo:
I merged my fix here https://github.com/neovide/neovide/pull/3168, so I think this can be closed. But I'm leaving it open until it's confirmed

cubewhy:
> I merged my fix here https://github.com/neovide/neovide/pull/3168, so I think this can be closed. But I'm leaving it open until it's confirmed

I'll confirm it later. If everything fine, I'll close this pr

cubewhy:
It works great! Thanks

PR #3193, opened 2025-08-22
Fix/mac window focus

jamylak:
- Fix

- No

## Previous Behaviour
- On Mac
- Open file with Neovide from a different desktop will **not** focus you on Neovide's window
- You will remain on your old desktop

https://github.com/user-attachments/assets/5654be91-b34b-495b-a332-ce8bef73e15d

## New Behaviour
- Your are moved to Neovide's active desktop

https://github.com/user-attachments/assets/4b32eb32-ef7c-47e9-aebe-92ebeb344f25

## Notes:
- I have absolutely no idea what I'm doing in Rust :D :P

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,6 +95,7 @@ winit = { version = "=0.30.9", features = ["serde"] }
 xdg = "2.5.2"
 notify-debouncer-full = "0.5.0"
 regex = "1.11.1"
+once_cell = "1.21.3"
 
 [dev-dependencies]
 scoped-env = "2.1.0"
--- a/src/window/window_wrapper.rs
+++ b/src/window/window_wrapper.rs
@@ -40,7 +40,7 @@ use {
 };
 
 #[cfg(target_os = "macos")]
-use super::macos::MacosWindowFeature;
+use super::macos::{MacosWindowFeature, WINDOW_FEATURE};
 
 const GRID_TOLERANCE: f32 = 1e-3;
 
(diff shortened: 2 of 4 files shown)


jamylak:
Just need to check one thing first, seems to have one issue

jamylak:
The issue should be fixed on my end but also i noticed CICD doesn't run due to an unrelated issue:

https://github.com/neovide/neovide/actions/runs/17156520753
`Invalid workflow file: .github/workflows/release.yml#L1(Line: 5, Col: 3): Unexpected value 'secrets'`

Cargo.toml
```diff
@@ -95,6 +95,7 @@ winit = { version = "=0.30.9", features = ["serde"] }
 xdg = "2.5.2"
 notify-debouncer-full = "0.5.0"
 regex = "1.11.1"
+once_cell = "1.21.3"
```
fredizzimo, line 98: `OnceCell` is in the standard
https://doc.rust-lang.org/std/cell/struct.OnceCell.html

src/window/window_wrapper.rs
```diff
@@ -509,10 +509,12 @@ impl WinitWindowWrapper {
...
         {
-            self.macos_feature = Some(MacosWindowFeature::from_winit_window(
-                window,
-                self.settings.clone(),
-            ));
+            let feature = MacosWindowFeature::from_winit_window(window, self.settings.clone());
+            WINDOW_FEATURE.with(|cell| {
+                cell.set(feature.clone())
+                    .expect("WINDOW_FEATURE already set");
+            });
```
fredizzimo, line 516: Can't you call `window_feature.ns_window.makeKeyAndOrderFront(None);` directly here, instead of using storing it in a thread local variable? Additionally it's already stored in `self.macos_feature`, so you should be able to read it from there.

fredizzimo (changes requested):

falcucci:
@jamylak thank you!

I have fixed the conflicts but need to go to the downtown, when I am back I improve the code and test it :)

falcucci:
@jamylak tested and works great, I have done some refactoring as well. thank you!

jamylak:
> @jamylak tested and works great, I have done some refactoring as well. thank you!

Thank you very much for updating this, will be good to be able to download the latest version now, sorry I didn't check in on this PR for a long time.

falcucci:
@jamylak it's available at our [nightly](https://github.com/neovide/neovide/releases/tag/nightly) releases :)

Issue #3198, opened 2025-08-29
Feature request: Capture mouse location even when outside of window to clamp later on


xNasuni:
**Is your feature request related to a problem? Please describe.**
when i use neovim and i want to copy part of the output, i always just hold click and move my mouse to the top left of the window, but i get frustrated that the mouse events only happen when inside the window, this means it always doesn't fully select everything i want to. it heavily slows down my workflow for copying output into notes, etc...

**Describe the solution you'd like**
the mouse events should somehow also be picked up even when the mouse isn't inside the window so that this would be fixed, and also adding an option to toggle this incase people like the original handling.

**Describe alternatives you've considered**
i tried to use fullscreen to mitigate this issue but it still occurs with the topbar leaving space of unselected characters. i haven't found any config option for this.

**Additional context**
visual demo:
![Image](https://github.com/user-attachments/assets/2c9e3ab2-994f-4bf7-bc1d-87854915b79b)

fredizzimo:
I don't think this can be supported, since if you click outside the window to start selecting, that click will be delivered to another window, or the desktop below.

But we can fix it to work with padding https://neovide.dev/configuration.html#padding

xNasuni:
i don't know whether you understood what i meant.

while starting to hold down click inside neovide, then dragging it outside the window without letting it go it still capturing mouse events, like i've seen in many other applications.

you said "if you click outside the window" which isn't what i was referring to so i'm not fully confident you understood

fredizzimo:
Ah yes I understood wrong, and was also only checking the last part of the video where the mouse enters the window.

Kethku:
https://github.com/neovide/neovide/pull/3203
Can you try this branch? I think it does what you want

Kethku:
Merged the branch as it works on two machines. Should be in the next release and or available if you build from main

PR #3203, opened 2025-09-03
Fix: drag events outside window

Kethku:
Allow mouse drags to continue even when the mouse has left the bounds of the window

- Fix

- No

Fixes https://github.com/neovide/neovide/issues/3198


fredizzimo (approved):
I tested this, and it appears to be fixing the issue without side effects. The code change also looks logical.

So, this can be merged once the `Cargo.lock` change is reverted. I updated all the dependencies yesterday, so it should not be needed

Issue #3066, opened 2025-03-24
Theme = "auto" doesn't work automatically.


0xDEAD-C0DE:
**Describe the bug**
Neovide doesn't switch themes automatically. I have set the `theme = "auto"` in the config.toml and always default to dark background. I a using LazyVim, but even having no init.lua, Neovide always has a background=dark. Is the `theme = auto` not implemented for Wayland?

**To Reproduce**
1. open nvim and the appropriate  theme is selected. Neovim switches to light/dark based on the OS dark/light theme. 

**Expected behavior**
launch Neovide and the dark or light theme should be selected
**Screenshots**

**Desktop (please complete the following information):**
 - OS: ArchLinux Cosmic epoch alpha 6
 - Neovide Version 14.1
 - Neovim Version 10.4

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[Neovide log](https://gist.github.com/0xDEAD-C0DE/d8a1a658cb0aa92a2fddd93668787c0a)
**Additional context**

fredizzimo:
Yes, winit does not support detection on Wayland, and apparently not going to support it either, see https://github.com/rust-windowing/winit/issues/1549. But we can detect it through the xdg portal instead with this library https://docs.rs/ashpd/latest/ashpd/

I think our theme support should be changed on all platforms. We should only need and option for detecting runtime changes. The detected defaults should always be passed when the UI connects, which can then be overridden in the `init.lua/vim`, just like the TUI does.

Amparose:
I seem to be having this issue on MacOS as well: always dark mode.

fredizzimo:
* This will be fixed here https://github.com/neovide/neovide/issues/3066

PR #3207, opened 2025-09-07
fix: Always auto-detect the background color, support setting the window theme

fredizzimo:
Neovide will now always autodetect the [`background`](https://neovim.io/doc/user/options.html#'background') option and set it at Neovim startup. It will also automatically change the `background` when your system theme changes, provided you haven't set `background` manually.

It has also been fixed, using https://github.com/bash/mundy to work on all systems, and there are no longer any race conditions at startup due to the way that it just sets the default.

Due to this, the old `theme` setting is no longer needed and has been re-purposed to set the window theme instead. `theme` is the most logical setting name, so I decided to keep it, but we can also use some other name, and deprecate the old `theme` setting completely.

Check out the documentation for more information.

* fixes https://github.com/neovide/neovide/issues/3066

- Yes, the theme setting now means something else


Kethku (approved):
imho repurposing the setting is kinda rough unless we can detect that the old value was used and send a message to the user to update. If we can do that then seems straightforward to me.

As a rule of thumb, people dont read patch notes. So changing the meaning of something is likely to cause problems

Kethku:
Actually its kinda a brutal outcome cause the meaning changes but the value names stay the same. Idk maybe since they are similar purposes its fine...

falcucci:
fixed conflicts, re-tested and works fine! thanks

Issue #3214, opened 2025-09-11
Neovide freezes on notify with `cmdheight=0`


Curve:
**Describe the bug**
Neovide freezes if `cmdheight` is set to `0` and `vim.notify` is called.

**To Reproduce**
1. Have the following `init.lua`:
```
vim.opt.cmdheight = 0
vim.notify("Test")
```
2. Start Neovide

**Expected behavior**
I expect to be able to press enter (or any key) do dismiss the notification, however, neovide is frozen.
_Everything works as expected in the terminal._

**Screenshots**

<img width="2543" height="1491" alt="Image" src="https://github.com/user-attachments/assets/8b256869-b1d5-4307-8738-43974b44d88d" />

**Desktop (please complete the following information):**
 - OS: Arch Linux (on Linux 6.16.5)
 - Neovide 0.15.1[^1]
 - Neovim: v0.11.4 (Build type: RelWithDebInfo, LuaJIT 2.1.1753364724)

[^1]: Interestingly, running `neovide --version` results in a segmentation fault:
        ```
        ❯ neovide --version
        neovide 0.15.1
        [1]    5070 segmentation fault (core dumped)  neovide --version
        ```

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
Log: [neovide_rCURRENT.log](https://github.com/user-attachments/files/22265376/neovide_rCURRENT.log)

**Additional context**
Seems similar to #2231

fredizzimo:
Hm. I can't reproduce here on Arch Linux

I tried Neoivde 0.15.0, 0.15.1 and the latest main, and Neovim 0.11.4 and nightly from a few days ago

Started with `neovide -- --clean -u init.lua`, but I also tested without `clean`

Is someone else able to reproduce?

Curve:
> Hm. I can't reproduce here on Arch Linux
> 
> I tried Neoivde 0.15.0, 0.15.1 and the latest main, and Neovim 0.11.4 and nightly from a few days ago
> 
> Started with `neovide -- --clean -u init.lua`, but I also tested without `clean`
> 
> Is someone else able to reproduce?

Hm... That's interesting I also tried main and 0.15.1 - Also tested with `neovide -- --clean` without any config file and then doing `:set cmdheight=0` and `:lua vim.notify("Test")` which also caused the same issue.

Do you reckon it might be some sort of system / dependency issue?

fredizzimo:
[...]

PR #3218, opened 2025-09-12
Revert "feat: add Lua API of IME event handling"

fredizzimo:
Reverts neovide/neovide#3110

Unfortunately, this breaks input when a message is showing see:
* https://github.com/neovide/neovide/issues/3214#issuecomment-3285109328

@kanium3, once we know how to fix the issues I will add this back.


kanium3:
I got it.

PR #3222, opened 2025-09-15
Add Go and Rust to Info.plist

Kreeblah:
- Feature

- No

This adds Go and Rust to the macOS Info.plist file so that Neovide can be set as the default editor for them.  I didn't see any particular ordering to the dicts in the file, so I just added these entries to the bottoms of each array.


fredizzimo:
Thank you!

PR #3242, opened 2025-10-02
fix: Use the Neovim grid clustering rules

fredizzimo:
This is mostly a refactoring, but it might fix some edge cases with double width and ambiwidth characters, where the Neovim and unicode segmentation views don't match.

Previously we did our own clustering using the Unicode segmentation crate. But Neovim already has its own clustering rules and single/double width character detection. These changes that information in use and removes the Unicode segmentation dependency. 

The word splitting is also done on the editor, rather than the rendering side, and word-based shaping cache is in use, instead of the style based one.

Some minor code cleanup has also been done.

This in itself does not fix many things, but it will allow us to adjust the size of the fallback fonts https://github.com/neovide/neovide/pull/3220 and fix some edge cases with the emoji rendering. It will also make it much easier to add support for more advanced text script systems and add bidi support.

- No

### DIFF
--- a/src/editor/window.rs
+++ b/src/editor/window.rs
@@ -213,8 +213,7 @@ impl Window {
 
         let line_fragment = LineFragment {
             text: text_range,
-            window_left: start as u64,
-            width: width as u64,
+            cells: start as u32..(start + width) as u32,
             style: style.clone(),
         };
 
--- a/src/editor/window.rs
+++ b/src/editor/window.rs
@@ -197,33 +197,30 @@ impl Window {
 
             width += 1;
 
-            if cluster.is_empty() && !current_word.cells.is_empty() {
-                current_word.cells.end += 1;
-            }
             let is_whitespace = cluster
                 .chars()
                 .next()
                 .is_some_and(|char| char.is_whitespace());
             if is_whitespace {
-                if !current_word.cells.is_empty() {
+                if !current_word.cluster_sizes.is_empty() {
                     // Finish the current word
                     words.push(current_word);
                     current_word = Word::default();
                 }
-            } else if current_word.cells.is_empty() {
+            } else if current_word.cluster_sizes.is_empty() {
                 // Properly initialize a new word
-                current_word.cells = width - 1..width;
-                current_word.text = text.len() as u32..text.len() as u32 + cluster.len() as u32;
+                current_word.cell = width - 1;
+                current_word.cluster_sizes.push(cluster.len() as u8);
+                current_word.text = text.len() as u32;
             } else {
-                current_word.cells.end += 1;
-                current_word.text.end += cluster.len() as u32;
+                current_word.cluster_sizes.push(cluster.len() as u8);
             }
 
             // Add the grid cell to the cells to render.
             text.push_str(cluster);
             text_range.end += cluster.len() as u32;
         }
-        if !current_word.cells.is_empty() {
+        if !current_word.cluster_sizes.is_empty() {
             words.push(current_word);
         }
 
--- a/src/editor/window.rs
+++ b/src/editor/window.rs
@@ -1,7 +1,6 @@
 use std::{collections::HashMap, sync::Arc};
 
 use log::warn;
-use unicode_segmentation::UnicodeSegmentation;
 
 use crate::{
     bridge::GridLineCell,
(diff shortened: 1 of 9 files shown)


src/editor/window.rs
```diff
@@ -195,23 +196,56 @@ impl Window {
...
 
             width += 1;
-            // The previous character is double width, so send this as its own draw command.
-            if character.is_empty() {
-                break;
+
+            // We can't deal with clusters that are longer than 255 bytes, so replace them with spaces.
+            // This should only happen if Neovim sends corrupted lines, or maybe in some pathological
+            // Unicode combining sequence.
+            let cluster = if cluster.len() > 255 { " " } else { cluster };
```
Kethku: is this a circumstance we should notify the user for? Or at least log?

src/editor/window.rs
```diff
@@ -352,3 +376,269 @@ impl Window {
...
+    fn make_window(grid: CharacterGrid) -> Window {
+        Window {
+            grid_id: 1,
+            grid,
+            window_type: WindowType::Editor,
+            anchor_info: None,
+            grid_position: (0.0, 0.0),
+        }
+    }
+
```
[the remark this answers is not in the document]
Kethku: I'm still trying to parse what this does, but can't it be a normal function? Using a macro here I think just makes it harder to understand and reason about.

src/renderer/grid_renderer.rs
```diff
@@ -217,39 +218,19 @@ impl GridRenderer {
...
         ) {
             return (text_drawn, true);
         } else if !text.is_empty() {
-            let mut paint = Paint::default();
-            paint.set_anti_alias(false);
-            paint.set_blend_mode(BlendMode::SrcOver);
             text_canvas.save();
 
             // We don't want to clip text in the x position, only the y so we add a buffer of 1
             // character on either side of the region so that we clip vertically but not horizontally.
```
Kethku: I recognize this comment isn't from this PR, but would it not be more correct to clip to the width of the window instead of doing just a single character of buffer zone? I think the logic might be simpler too cause we wouldn't have to reason about adding characters. Idk take it or leave it.

src/renderer/rendered_window.rs
```diff
@@ -13,21 +13,63 @@ use crate::{
...
+}
+
+#[derive(Debug, Default)]
+pub struct Word {
+    pub text: u32,
+    pub cell: u32,
+    pub cluster_sizes: Vec<u8>,
+}
+
+pub struct RenderedWord<'a> {
```
Kethku: Its a little weird for RenderedWord to not actually have anything Rendered in it. Would it make sense for this to be ClusteredWord instead?

src/editor/window.rs
fredizzimo: We have the raw grid information in the log, so it should be possible to detect the problem from that. And the user would report that some character is showing a space instead of the actual character. So, I don't think this is a problem.

Realistically this can only happen with [Zalgo](https://en.wikipedia.org/wiki/Zalgo_text) text, but we couldn't render that properly anyway when being restricted to one line.

src/renderer/grid_renderer.rs
```diff
@@ -262,33 +243,33 @@ impl GridRenderer {
             } else {
                 paint.set_color(style.foreground(&self.default_style.colors).to_color());
             }
-            paint.set_anti_alias(false);
-            // There's a lot of overhead for empty blobs in Skia, for some reason they never hit the
```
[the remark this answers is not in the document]
fredizzimo: Yes, we are only shaping actual words, so the trimming is done as part of the word splitting.

src/renderer/rendered_window.rs
fredizzimo: I will try to think of a better name. `Rendered` here refers to that this is the struct actually used when rendering and the `GridRenderer` constructs them. And it will end up being displayed on the screen after shaping and processed by Skia. But yes, it's true that it's not yet renderered.

Something like `ShaperWord` could work since it's the representation used by the shaper. But I don't like that too much either.

Kethku: Maybe RenderableWord?

fredizzimo: `RenderedWord` is now `Word` and `Word` was renamed to `WordData`. 

I also did some other refactoring related to this by exposing iterator functions instead of having to pass the line string around. See https://github.com/neovide/neovide/pull/3242/commits/2fa2bf3ba413b658a5ef435cdbe8d8553130de5d for the full refactoring

PR #3247, opened 2025-10-08
fix!: Remove default font and load "monospace" by default

fredizzimo:
The default embedded FiraCode Nerd font has several problems
1. It's outdated, and no longer compatible with many plugins
2. The old version (pre 3.0.0) overwrites some real glyphs most notably, which causes Neovide to render those glyphs wrong:
U+F900..U+FAFF	[CJK Compatibility Ideographs](https://en.wikipedia.org/wiki/CJK_Compatibility_Ideographs)
U+FB00..U+FB4F	[Alphabetic Presentation Forms](https://en.wikipedia.org/wiki/Alphabetic_Presentation_Forms)
U+FB50..U+FDFF	[Arabic Presentation Forms-A](https://en.wikipedia.org/wiki/Arabic_Presentation_Forms-A)
3. Only the normal variation is present, so bold and italic are missing
4. Falling back to a FiraCode Nerd font when the user have another font configured does not work as expected due to the symbols having different dimensions.

Instead of a supporting a constantly moving target and avoid the need to add the italic and bold variants, I removed the embedded font completely. The system default monospace font will be used instead. It probably doesn't support Nerd fonts, but Neovim does not need a Nerd font by default, and you also need to configure a font with support in terminal emulators. On some systems, Nerd Fonts symbols are still found as fallbacks, depending on how the system fallback fonts are configured.

I also removed some other completely unused fonts and improved the logging.

And finally, I fixed some problems when the primary font fails to load. This could previously cause Neovide to become really slow.

- The default font for users who don't configure anything might change
- The unused `embed_fonts` cargo feature has been removed, which might the packaging

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,6 @@ members = ["neovide-derive"]
 
 [features]
 default = []
-embed-fonts = []
 profiling = ["dep:tracy-client-sys"]
 gpu_profiling = ["profiling"]
 # Corresponds to https://github.com/nagisa/rust_tracy_client/blob/main/FEATURES.mkd


Kethku:
I sorta think updating is the more correct solution.

We added the font originally to make the default behavior more correct.

1. Is easily solved
2. Solved by 1
3. We can embed more
4. This is a problem with the default fallbacks too and not made worse I dont think.

Having default nerd fonts does help with the powerline/box drawing symbols. However if instead we default to using the rendered symbols with the new support (maybe if the configured fonts dont have them by default) then I feel much less strongly about the decision and the benefits of not embedding become more appealing

PR #3249, opened 2025-10-09
Fix Issue 3234: False information in documentation of config

DataKumpel:
Greetings from hacktober! Richard, Mo and I worked really REALLY hard to fix this issue in the docs.

- Fix
- Documentation

- No

### DIFF
--- a/website/docs/configuration.md
+++ b/website/docs/configuration.md
@@ -261,7 +261,7 @@ vim.g.neovide_title_background_color = string.format(
 vim.g.neovide_title_text_color = "pink"
 ```
 
-#### Window Blur (Currently macOS only)
+#### Window Blur
 
 VimScript:
 


fredizzimo:
Thanks, it does not support all the platforms though, so maybe it would be worth documenting the complete requirements.
* On windows we use this https://docs.rs/winit/latest/winit/platform/windows/trait.WindowExtWindows.html#tymethod.set_system_backdrop, so that means only Windows 11 is supported
* On the other platforms https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_blur. 
  * MacOS supported
  * Linux X11 unsupported
  * Linux Wayland supported on KDE Plasma, unsupported otherwise

Kethku:
Probably going to close this as we didn't get a response

PR #2872, opened 2024-11-13
refactor: multi-window support (WIP)

falcucci:
ℹ️ ⚠️ Breaking changes are expected. Doesn't work as expected yet. Do not use this as a daily build. Not ready for review.

## What kind of change does this PR introduce?
- Feature

## Did this PR introduce a breaking change? 
_A breaking change includes anything that breaks backwards compatibility either at compile or run time._
- The probability is high

PR #3252, opened 2025-10-15
refactor: Split the update loop into window specific loops and global application

### LINKED ISSUE #2868: `neovide --version` sporadically segfaults
**Describe the bug**

`neovide --version` command will sporadically panic with message 

```
Io error: Bad file descriptor (os error 9)
```

Note that this seems related to #2867, but here I'm running `neovide` in regular environment in which it usually works fine.

**To Reproduce**

1. `git clone https://github.com/neovide/neovide.git`
2. `cd neovide`
3. `git checkout -b 0.13.3 0.13.3`
4. `cargo build`
5. `export `RUST_BACKTRACE=1`
6. `./target/debug/neovide --version`

It sporadically coredumps with error `Io error: Bad file descriptor (os error 9)` and no backtrace.

Since I hadn't found reliable way to reproduce, you can try this:

```zsh
for i in {1..100}; do ./neovide --log --version; done 
```

Same thing happens with build downloaded from release `0.13.3`, [neovide-linux-x86_64.tar.gz](https://github.com/neovide/neovide/releases/download/0.13.3/neovide-linux-x86_64.tar.gz). I built it myself hoping I'd also get backtrace. No luck there.

**Expected behavior**

Shouldn't panic. 

**Screenshots**
N/A

- OS: `Ubuntu 24.04.1 LTS` `Linux 6.8.0-48-generic #48-Ubuntu SMP PREEMPT_DYNAMIC Fri Sep 27 14:04:52 UTC 2024 x86_64 x86_64 x86_64` `GNU/Linux`
- Neovide Version: `0.13.3`
- Neovim Version: `NVIM v0.10.3-dev-13+gb286ba419a Build type: RelWithDebInfo LuaJIT 2.1.1713484068`

No logs were emitted.

**Additional context**
N/A

fredizzimo:
This is a step towards multi-window support with separate renderers, vsync, and update rates. 

The main changes are
1. `WinitWindowWrapper`, which was a very confusing name has been renamed to `NeovimWindow` to indicate that it deals with a single Neovim instance
2. `NeovideApplication` has been added. Its main purpose is to store the global application state and ensures that everything is freed before the event loop has finished. It also is the entry point for all events, and further delegates them to a possible active update loop. The plan is to support multiple active update loops in the future when actual multi-window support is added. Finally, it gathers the information from the update loop to schedule how long to wait for the next event.
3. The `UpdateLoop` is no longer unique, and there can be multiple different ones. Right, now only one is active at a time, but it can update either the `NeovimWindow` or the `ErrorWindow`. So, the code can be re-used to run both.
4. A `Window` trait has been added. It's used to make it possible for the same update loop code to run different types of windows.
5. The Neovim/Tokio runtime is now stored and managed by the application, to make sure that it's stopped before it's too late. The `NeovimWindow` launches its neovim instance using that runtime. Currently, only a single one, but with some small changes, it will be possible to launch multiple instances. The idea is that the events sent back will get a window id, so that they are routed back to the correct Window. And instead of the global `send_ui` function the window will get a dedicated channel to send events to the instance.
6. The `RunningTracker` has been removed, and the `NeovideApplication` now stores it's exit code, communicated through events like almost everything else.
7. The lifetime of the clipboard is now managed by the application, so that it's destroyed before the event loop. This fixes https://github.com/neovide/neovide/issues/2868  
8. The Vsync is now owned and encapsulated inside the `SkiaRenderer`, with a few trait function to access the functionality needed
9. The error handling that might need a window has been moved to `NeovideApplication`, communicated through events. Startup errors and output due to the command line processing, is handled before the application is run 

NOTE: This is just the first step towards the goal, but it contains some important fixes, without breaking any existing functionality.

## Differences from #2872

The approach here is quite different from https://github.com/neovide/neovide/pull/2872, which keeps a single update loop and a single `WinitWindowWrapper` and instead introduces a new concept called routes. This means that almost all functions in there had to be changed to deal with finding the correct window. While the approach here will be restricted to a handful of small functions in the `NeovideApplication`.

The other big difference is the plan here is to make it possible to update and animate all windows smoothly at the same time, while #2872 mostly only animates a single focused window. 

#2872 also makes some small changes to the update loop which makes it less efficient, for example we currently try to animate each time the event loop goes back to idle, to avoid having to do too much work when it's actually time to render, and to be able to do something useful instead of just waiting. But #2872 does all its work when the loop is woken due to an event.

#2872, also changes the way the windows are created, they are now created with the wrong size to be later resized. This is a quite big thing, since delaying the window creation in https://github.com/neovide/neovide/pull/2562 fixed a lot of issues. 

Finally, I'm avoiding the needless ownership changes of #2872, where a lot of things, like the Window and Skia renderer now are reference counted instead of having a single clear owner. In particular shared ownership of the window can be dangerous, since there's a very defined order for dropping it, see https://github.com/neovide/neovide/blob/d26376444c971f305717817bbd79a9959f86a0fb/src/renderer/opengl.rs#L225-L245

All these issues, and the fact that I wanted to fix #2868, motivated me to make this. And it does not fundamentally invalidate the approach in #2872, a lot of the other changes there are applicable on top of the changes here.

- No

### DIFF
--- a/src/error_handling.rs
+++ b/src/error_handling.rs
@@ -2,9 +2,6 @@ use anyhow::{Error, Result};
 use itertools::Itertools;
 use log::error;
 
-#[cfg(target_os = "windows")]
-use crate::windows_attach_to_console;
-
 use crate::bridge::{send_ui, ParallelCommand};
 
 fn show_error(explanation: &str) -> ! {
--- a/src/renderer/d3d.rs
+++ b/src/renderer/d3d.rs
@@ -89,7 +89,7 @@ pub struct D3DSkiaRenderer {
     gr_context: DirectContext,
     swap_chain: IDXGISwapChain3,
     swap_chain_desc: DXGI_SWAP_CHAIN_DESC1,
-    swap_chain_waitable: HANDLE,
+    _swap_chain_waitable: HANDLE,
     pub command_queue: ID3D12CommandQueue,
     buffers: Vec<ID3D12Resource>,
     surfaces: Vec<Surface>,
@@ -265,7 +265,7 @@ impl D3DSkiaRenderer {
             command_queue,
             swap_chain,
             swap_chain_desc,
-            swap_chain_waitable,
+            _swap_chain_waitable: swap_chain_waitable,
             gr_context,
             _backend_context: backend_context,
             buffers: Vec::new(),
@@ -454,17 +454,15 @@ impl SkiaRenderer for D3DSkiaRenderer {
     }
 
     fn refresh_interval(&self) -> f32 {
-        self.vsync.get_refresh_rate(self.window(), &self.settings)
+        self.vsync.get_refresh_rate(&self.window, &self.settings)
     }
 
     fn request_redraw(&mut self) -> bool {
-        let window = self.window.as_ref().unwrap();
-        self.vsync.request_redraw(window)
+        self.vsync.request_redraw(&self.window)
     }
 
     fn update_vsync(&mut self) {
-        let window = self.window.as_ref().unwrap();
-        self.vsync.update(window);
+        self.vsync.update(&self.window);
     }
 
     fn wait_for_vsync(&mut self) {
--- a/src/renderer/metal.rs
+++ b/src/renderer/metal.rs
@@ -17,12 +17,12 @@ use skia_safe::{
     },
     Canvas, ColorSpace, ColorType, PixelGeometry, Surface, SurfaceProps, SurfacePropsFlags,
 };
-use winit::{event_loop::EventLoopProxy, window::Window};
+use winit::window::Window;
 
 use crate::{
     profiling::tracy_gpu_zone,
     renderer::{RendererSettings, SkiaRenderer, VSync},
-    window::{macos::get_ns_window, UserEvent},
+    window::macos::get_ns_window,
 };
 
 use super::Settings;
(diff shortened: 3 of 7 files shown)


fredizzimo:
I forgot to mention, that one further improvement that should be done later is to decouple the Neovim instance from the actual window, so that you can switch between instances within the same window. And even further down the road, to be able to have multiple instances per window to render splits with different instances, but that probably requires quite a bit of further refactoring.

Edit: Multiple instances, might not need that much, just multiple `Renderer`s for one `SkiaRenderer` and some changes to the mouse and keyboard handling to the active instance.

Edit2: That seems to indicate that what we currently call `Renderer` is actually the full state of one Neovim instance. A global renaming pass of our modules might be in order. But some of that state is currently also in `NeovimWindow / WinitWindowWrapper`, so it needs to be moved.

Some further things to note, is that part the state, like the font cache should be more global. Skia probably also has some way to partially share the caches, so it needs to be split into a global state and states for each surface/window.

falcucci (changes requested):
good improvements in general

suggestion: the multi-window feature made at https://github.com/neovide/neovide/pull/2872 is already a working state solution that needs to be *refined*, I would put your approach on top of https://github.com/neovide/neovide/pull/2872 or even redo that work using this approach instead but making sure everything keeps working.

next steps would be probably stacking both PRs. they are way too big.

falcucci:
closing since it's no longer needed and stalled, thank you

Issue #2929, opened 2025-01-05
macOS: Dock icon not changeable anymore since 0.14


chrisgrieser:
**Describe the bug**
Prior to 0.14, you could easily change the dock icon by setting a custom one via `cmd+i` on the app and pasting a custom one, or by using an icon-replacer like replacion. However, since 0.14, that is not the case anymore. The custom icon is only used when Neovide is not running; as soon as Neovide is running, the original icon is still used.

**To Reproduce**
1. Go to the `Applications` folder
2. Select `Neovide.app`
3. `cmd+i`, select the icon
4. Paste a custom `.icns` file

**Expected behavior**
The custom icon always being in use.

**Screenshots**
when neovide is running:
<img alt="Showcase" width=40% src="https://github.com/user-attachments/assets/3a1a7de9-ee4a-4a68-83e1-c3f7475562e3">

when neovide is not running:
<img alt="Showcase" width=40% src="https://github.com/user-attachments/assets/d18de9a4-e19a-4249-a2e0-0b72aade7c06">

**Desktop (please complete the following information):**
 - OS: macOS 15.1 (arm)
 - Neovide Version: 0.14
 - Neovim Version: NVIM v0.10.3

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
[neovide_rCURRENT.log](https://github.com/user-attachments/files/18310678/neovide_rCURRENT.log)

**Additional Context**
macOS apps usually use a rounded-square-shaped icon, and most apps comply to that. For UI consistency, it is probably also worth a thought to use a properly shaped, nice icon in the first place.

falcucci:
hey @chrisgrieser nice, thanks for reporting.

A minor note here is that this isn't really a bug. Neovide have never offered this as a feature *but* I like the idea of it.

Basically this was changed to make sure to load and set the app icon for any context, for example when you spawn a neovide instance from our released binary (not the .dmg or bundle one) we had a default console process icon, which is bad.

I am interested in set a better icon for macOS tho, or even give the option officially to open this possibility about changing or choose a pre configured one for the end user.

chrisgrieser:
> A minor note here is that this isn't really a bug. Neovide have never offered this as a feature

This feels like a technicality, tbh. *Every*  macOS app allows you to customize its icon. Thus, Neovide is now deviating from widely accepted OS-standard which I do consider bug-ish at the very least.

> I am interested in set a better icon for macOS tho

[...]

PR #3272, opened 2025-10-29
feat: handling a custom icon path if provided

falcucci:
This should address some user [complaints](https://github.com/neovide/neovide/issues/2929) about previous support of custom icons when we tried to fix the console default process icon in general.

It's better to free users to customize it using their favorite neovide custom icon if they wish providing an option.

### DIFF
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -141,6 +141,10 @@ pub struct CmdLineSettings {
     )]
     pub x11_wm_class_instance: String,
 
+    /// The custom icon to use for the app.
+    #[arg(long)]
+    pub icon: Option<String>,
+
     #[command(flatten)]
     pub geometry: GeometryArgs,
 
--- a/src/platform/macos/mod.rs
+++ b/src/platform/macos/mod.rs
@@ -30,7 +30,7 @@ use crate::{cmd_line::CmdLineSettings, error_msg, frame::Frame};
 
 use crate::window::{WindowSettings, WindowSettingsChanged};
 
-static NEOVIDE_ICON_PATH: &[u8] =
+static DEFAULT_NEOVIDE_ICON_BYTES: &[u8] =
     include_bytes!("../../../extra/osx/Neovide.app/Contents/Resources/Neovide.icns");
 
 define_class!(
@@ -75,18 +75,25 @@ pub fn get_ns_window(window: &Window) -> Retained<NSWindow> {
     }
 }
 
-fn load_neovide_icon() -> Option<Retained<NSImage>> {
+fn load_icon_from_custom_path(icon_path: &str) -> Option<Retained<NSImage>> {
+    let path = NSString::from_str(icon_path);
+    NSImage::initWithContentsOfFile(NSImage::alloc(), &path)
+}
+
+fn load_icon_from_default_bytes() -> Option<Retained<NSImage>> {
     unsafe {
         let data = NSData::dataWithBytes_length(
-            NEOVIDE_ICON_PATH.as_ptr() as *mut c_void,
-            NEOVIDE_ICON_PATH.len(),
+            DEFAULT_NEOVIDE_ICON_BYTES.as_ptr() as *mut c_void,
+            DEFAULT_NEOVIDE_ICON_BYTES.len(),
         );
-
-        let icon_image: Option<Retained<NSImage>> =
-            NSImage::initWithData(NSImage::alloc(), data.as_ref());
-
-        icon_image
+        NSImage::initWithData(NSImage::alloc(), data.as_ref())
     }
+}
+
+fn load_neovide_icon(custom_icon_path: Option<&String>) -> Option<Retained<NSImage>> {
+    custom_icon_path
+        .and_then(|path| load_icon_from_custom_path(path))
+        .or_else(load_icon_from_default_bytes)
 }
 
 #[derive(Debug)]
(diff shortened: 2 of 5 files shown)


falcucci:
suggestion: I would like to merge this before moving to the new platform specific code approach.

Kethku:
Couple things come to mind.

1. This is missing docs for the new commandline argument. We should add that as a part of this PR.
2. We should double check that this works on other platforms as the icon should load there as well. I think it might just work but I didnt dig too deeply, so just wanna double check.
3. I wonder if this should be a change in the config file instead of or maybe in addition to via the cmdline options. I figure its the sort of thing youd want to set and forget rather than passing every time.

fredizzimo:
> 1- We should double check that this works on other platforms as the icon should load there as well. I think it might just work but I didnt dig too deeply, so just wanna double check.

Yes, it can and should be done on the other platforms, where the icon is loaded here  https://github.com/neovide/neovide/blob/cb3b3dfadb34c2c22f62786decdeb7c3c61d1e15/src/window/mod.rs#L140

Unfortunately macOs has to be remain a special case until this is implemented https://github.com/rust-windowing/winit/pull/4381, but it probably makes sense to move the code to the same place already at this stage.

> 2. I wonder if this should be a change in the config file instead of or maybe in addition to via the cmdline options. I figure its the sort of thing youd want to set and forget rather than passing every time.
The icon is loaded in a different place on them

Yes, I think so. I would even go as far and support setting it from within Neovim itself with lua as the highest priority setting. That way you could have project specific icons when you change the working directory for example, or whatever logic you want.

falcucci:
1. will be addressed thanks 
2. I guess it works since we aren't doing anything different but I'll double check
3. That's tricky because macOS supports hot reload icons if we consider it from the config, but I am not sure about other platforms. I would just add a cmd option for now and then add it in the configs in case we are sure about compatibilities

fredizzimo:
> 3\. That's tricky because macOS supports hot reload icons if we consider it from the config, but I am not sure about other platforms. I would just add a cmd option for now and then add it in the configs in case we are sure about compatibilities

The config file and command line can be considered the same. Both are loaded very early at the startup.

falcucci:
> 3. That's tricky because macOS supports hot reload icons if we consider it from the config, but I am not sure about other platforms. I would just add a cmd option for now and then add it in the configs in case we are sure about compatibilities

> The config file and command line can be considered the same. Both are loaded very early at the startup.

Shouldn't be considered the same. Frame for example isn't hot reloaded in case we want to support hot reload in the future. Configs like that usually tends to work as a reactive thing.

fredizzimo:
Frame should probably be part of the config file too. And not the whole file is hot-reloaded, for example the `neovim-bin` can't be hot reloaded, but it's part of the config.

I think the frame setting could even be changed at runtime, for example this can, but I have not checked all the macOS specific settings https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_decorations

When we have multi-window support most settings can be hot-reloaded and applies to following window or neovim process that is started.

falcucci:
seems like we have some points. anyway I think the item 3 can be addressed in another incremental pull request regardless which focus in the config setup support. Then we centralize this discussion there.

fredizzimo:
I think it should be part of the same PR. I would like to see that all PRs have the associated documentation and settings included whenever possible. 

We have made many mistakes with settings in the past, and therefore I think that we should start to think about them from all angles as part of each feature, and not something that's done as an afterthought.

falcucci:
Not necessarily, that's not the pull request who will fix the config settings neither is the intention of it. I'll push into another pull request since they can be splitted up and I prefer small changes.

fredizzimo:
And I prefer to review complete features

It's impossible to review anything if everyone splits everything into separate pull requests for functionality, documentation and configuration.

falcucci:
in that case we agree to disagree

falcucci:
It seems like you don't read my messages https://github.com/neovide/neovide/pull/3272#issuecomment-3464631791 https://github.com/neovide/neovide/pull/3272#issuecomment-3465858411 https://github.com/neovide/neovide/pull/3272#issuecomment-3465858411

from there on there is nothing I can do.

fredizzimo:
I have read them but you haven't' given me one valid reason for not including a three line change in this PR, that both I and @Kethku have requested?

I'm not asking you to fix an existing config option, I'm asking you to add the option for the new feature you are adding

falcucci:
> I have read them but you haven't' given me one valid reason for not including a three line change in this PR, that both I and @Kethku have requested?
I'm not asking you to fix an existing config option, I'm asking you to add the option for the new feature you are adding

I mean I literally said I was going to fix items 1 and 2 in this PR

falcucci:
I already gave you, I want a specific channel to discuss the config support change. I dont really care if it's 2 lines of code, but I'll need to also test on all platforms the new config and I would like to unblock this feature at least in the meantime for the users.

falcucci:
@Kethku the changes has been made, please review again while I test the startup on other platforms (if possible could you please open the app on windows?). I'll open a new pull request for the config so we centralize this specific config topic there.

Kethku:
looks good code wise. ill test on windows shortly

Kethku:
Unfortunately it turns out I dont have time to test this myself today. I may come back to it in a day or two unless someone else can test it. Imho that shouldn't block a merge tho as on inspection it looks alright so we can probably optimistically merge

falcucci:
tested on windows and macOS. all good here.

PR #3274, opened 2025-10-30
feat: include a new icon field in the configuration file

falcucci:
Upon discussion at https://github.com/neovide/neovide/pull/3272 some ideas has been made like creating and prioritizing within neovim itself an option in lua, so let's bring the config option first and discuss here to support or not other ways to load icons if needed.

### DIFF
--- a/website/docs/config-file.md
+++ b/website/docs/config-file.md
@@ -33,6 +33,7 @@ backtraces_path = "/path/to/neovide_backtraces.log" # see below for the default 
 fork = false
 frame = "full"
 idle = true
+icon = "/full/path/to/neovide.ico" # Example path. Default icon is bundled. Use .icns on macOS.
 maximized = false
 mouse-cursor-icon = "arrow"
 neovim-bin = "/usr/bin/nvim" # in reality found dynamically on $PATH if unset
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -142,7 +142,7 @@ pub struct CmdLineSettings {
     pub x11_wm_class_instance: String,
 
     /// The custom icon to use for the app.
-    #[arg(long)]
+    #[arg(long, env = "NEOVIDE_ICON")]
     pub icon: Option<String>,
 
     #[command(flatten)]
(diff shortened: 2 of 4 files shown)


fredizzimo:
This looks good, but should be merged with https://github.com/neovide/neovide/pull/3272, so that I can review and test the former properly.

falcucci:
you should take this pr to test its purpose, but this branch has everything what https://github.com/neovide/neovide/pull/3272 has so nothing is blocking you to test everything and address your comment in the right place.

Kethku:
@fredizzimo you're taking a stand here for what seems to me is of little consequence. Sure its nice to have prs merged together, but we need to consider contributions at least somewhat optimistically. Especially for contributors that have demonstrated extended interest in the project. If falcucci feels strongly about doing the work this way, its our responsibility as people considering the contribution to at least consider that for him it might be easier like this.

This is not the hill to die on and the conflict makes future contributions and progress *much* harder for little upside.

Kethku:
That said, its clear to me that you two have animosity toward each other. I'm done weighing in as its not my responsibility to mediate. It hasn't really worked in the past.

falcucci:
@Kethku btw it worked well on Windows and macOS here. let me know if you have any request

website/docs/config-file.md
```diff
@@ -33,6 +33,7 @@ backtraces_path = "/path/to/neovide_backtraces.log" # see below for the default
 fork = false
 frame = "full"
 idle = true
+icon = "/full/path/to/neovide.ico" # for macOS the icon should be in `.icns` format.
```
Kethku: This is not a default value. The value should have the value that would be used by default give the line above the code block

Kethku: no this is about the docs not the functionality. maybe take another look?

falcucci: thank you, misunderstood. check again please.

Kethku (approved):

kutsan:
First of all, thanks for your work. This is already merged but it would be great if the file path supported tilde expansion. For example, if I place my icon at `~/.config/neovide/icon.icns`, I currently have to write:

```
icon = "/Users/username/.config/neovide/icon.icns
```

I use my config on multiple machines and the usernames can vary between them. It would be nice if the tilde for the home directory worked directly:

```
icon = "~/.config/neovide/icon.icns"
```

falcucci:
@kutsan that's indeed more convenient, i'll take a look soon

falcucci:
@kutsan should be working at https://github.com/neovide/neovide/pull/3295

PR #3275, opened 2025-10-31
feat: add :NeovideConfig command to open the config file

falcucci:
Currently, editing the configuration file requires users to know and **find** or **create** the specific file path for their operating system

- $XDG_CONFIG_HOME/neovide/config.toml
- ~/Library/Application Support/neovide/config.toml
- etc. 

This can be inconvenient and requires looking up documentation, especially for new users.

Just for practical and convenience reasons adding `:NeovideConfig` command, that directly opens the `config.toml` file for editing within the running Neovide instance would be really cool to have.

https://github.com/neoclide/coc.nvim has inspired this feature.

### DIFF
--- a/website/docs/commands.md
+++ b/website/docs/commands.md
@@ -18,3 +18,10 @@ window containing Neovide to the front and activate it. This
 is useful for tools like neovim_remote which can manipulate
 neovim remotely or if long running tasks would like to
 activate the Neovide window after finishing.
+
+## Open Config File (Unreleased yet)
+
+Running the `NeovideConfig` command will open your Neovide
+configuration file for editing. This provides a simple and
+discoverable way to access your settings without needing to
+know the platform-specific path to the file.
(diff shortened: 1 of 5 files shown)


Kethku:
I really like this idea but it needs to catch when neovide is run from a remote server because the config file might not be accessible. Ideally we should send a message to the user in this case

website/docs/commands.md
```diff
@@ -18,3 +18,10 @@ window containing Neovide to the front and activate it. This
 is useful for tools like neovim_remote which can manipulate
 neovim remotely or if long running tasks would like to
 activate the Neovide window after finishing.
+
+## Open Config File
```
Kethku: This should probably indicate that this isn't released

falcucci: done

falcucci:
@Kethku done, I have also double checked if the `NEOVIDE_CONFIG` var handles correctly its value.

Kethku (approved):

PR #3279, opened 2025-11-05
feat: add animated progress bar with nvim_echo event support

falcucci:
introduces an animated progress bar to provide visual feedback for long-running background tasks by hooking into the native vim.api.nvim_echo fn for progress events, this feature works out-of-the-box.

inspired from https://github.com/neovim/neovim/pull/34846 https://github.com/neovim/neovim/pull/35973

https://github.com/user-attachments/assets/2ae0b06d-71d5-4141-8aaa-ea2fb75d042b

real use-case show-off.

https://github.com/user-attachments/assets/bf3099fa-31bb-498a-9ec4-58a8dadef772

the following new settings can customize the appearance and behavior of the progress bar. 
```
vim.g.neovide_progress_bar_enabled = true
vim.g.neovide_progress_bar_height = 5.0
vim.g.neovide_progress_bar_animation_speed = 150.0
vim.g.neovide_progress_bar_hide_delay = 1.0

```

### DIFF
--- a/lua/init.lua
+++ b/lua/init.lua
@@ -166,4 +166,25 @@ vim.api.nvim_create_user_command("NeovideConfig", function()
     end
 end, {})
 
+local function progress_bar(data)
+    -- Wrap inside pcall to avoid errors if Neovide disconnects.
+    pcall(rpcnotify, "neovide.progress_bar", data)
+end
+
+pcall(vim.api.nvim_create_autocmd, 'Progress', {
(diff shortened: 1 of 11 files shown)


falcucci:
this merge is aligned with @Kethku which will double check pre-releases afterwards anyway before our releases in general. I am merging this after some testing. in case we encounter something wrong at the neovide nightly releases a patch will be sent.

PR #3280, opened 2025-11-07
ci: automate nightly cross-platform nightly-release builds

falcucci:
- add a github actions workflow to build and publish nightly pre-release builds for windows, macOS and linux
- automate cross-platform builds, packaging, and upload of release assets for each os
- ensure nightly releases are overwritten and published as drafts before final release
- include steps for installing build toolchains and platform-specific dependencies
- support workflow scheduling via cron and manual dispatch

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,300 @@
+name: Nightly Release
+
+on:
+  schedule:
+    - cron: "0 5 * * *"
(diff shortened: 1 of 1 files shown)


falcucci:
this merge is aligned with @Kethku which will double check pre-releases afterwards anyway before our releases in general. I am merging this after some testing at my fork. in case we encounter something wrong for the nightly releases a patch will be sent.

PR #3283, opened 2025-11-08
ci: fix and limit mac deployment steps to main branch only

falcucci:
restrict mac code signing and app store connect key steps to run only on the main branch.


falcucci:
this merge is aligned with @Kethku which will double check pre-releases afterwards anyway before our releases in general. I am merging this after some testing at my fork. in case we encounter something wrong for the nightly releases a patch will be sent.

PR #3286, opened 2025-11-11
ci: automate nightly artifact creation for windows and macos binaries

falcucci:
Just to keep it as a standard we should create the artifact binaries for each platform

(we should release the neovide cli since the bundle isn't equivalent)


falcucci:

PR #3287, opened 2025-11-14
docs: update discord badge to show live member count

falcucci:
badgen.net is broken, no status or updates in the world wide web.


falcucci:

Issue #3289, opened 2025-11-14
Nightly freezes on open (Neovide-x86_64-apple-darwin.dmg)


9mm:
This isnt a lot of detail right now because i have like 50 tabs open and im right in the middle of something. Pulled latest to re-compile, noticed that the .app no longer works. when i open it, neovide opens with all my plugins, then immediately freezes where i have to force quit it

i downloaded fresh from nightly build and same effect

I can provide more or try ot figure out the last commit it occurs on later, but for now just making an issue for it

falcucci:
@9mm thank you, good to see people using our nightly releases as testers, that's what we need.

i'll double check here as well, please share any more info you find out.

akio-ito:
Probably related to: Forward stdin if it's piped on Linux and MacOS (https://github.com/neovide/neovide/pull/3206).
Works for me if I revert this commit.

falcucci:
@akio-ito thank you very much

I can confirm, i am preparing a fix

9mm:
@akio-ito your full name is 1 letter off from my gf's mom's name! Akiko Ito

falcucci:
@9mm @akio-ito feel free to test the artifact generated at the [latest](https://github.com/neovide/neovide/actions/runs/19374922722?pr=3290) pipeline.

here it works just fine now.

9mm:
awesome1 it works!

akio-ito:
It works!
Arigatou gozaimasu.

PR #3290, opened 2025-11-14
fix: detect real stdin descriptors to stabilize embedded neovim ingestion

falcucci:
embedded launch path now forwards `stdin` only when the descriptor truly represents redirected input.

Instead of using `stdin.is_terminal()` we now uses the `rustix::fs::fstat` to classify stdin by file type and hands it to neovim only for regular files, FIFOs or sockets.

Finder/Dock launches therefore retain their character-device stdin and no longer hang, while `ls | neovide -` correctly propagates the pipe, since macOS bundled apps have stdin connected to a char e.g `/dev/null` so the check misidentified them as pipes previously.

closes https://github.com/neovide/neovide/issues/3289

it fix the bug introduced from https://github.com/neovide/neovide/pull/3206 at nightly builds

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -175,7 +175,7 @@ csscolorparser = "0.7.0"
 shlex = "1.3.0"
 
 [target.'cfg(not(target_os = "windows"))'.dependencies]
-rustix = "1.0.8"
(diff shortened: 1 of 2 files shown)


falcucci:

PR #3292, opened 2025-11-16
chore: remove old workspace configuration files

falcucci:
out-dated stuff


falcucci:

PR #3293, opened 2025-11-16
fix(mouse): Cursor is invisible on unfocused window

Geobert:
If the cursor is hidden and we Atl-Tab, it will stay invisible. On focus loss, this PR restore visibility (https://github.com/neovide/neovide/issues/2414#issuecomment-3538787842 )

- Fix

- No

### DIFF
--- a/src/window/mouse_manager.rs
+++ b/src/window/mouse_manager.rs
@@ -464,6 +464,13 @@ impl MouseManager {
                     }
                 }
             }
+            WindowEvent::Focused(false) => {
+                let window_settings = self.settings.get::<WindowSettings>();
+                if window_settings.hide_mouse_when_typing && self.mouse_hidden {
+                    window.set_cursor_visible(true);
+                    self.mouse_hidden = false;
+                }
+            }
             _ => {}
         }
     }


falcucci:
this one unfortunately didnt work on macOS for me, winit short circuits probably duplicate visibility requests and AppKit doesn't refresh cursor rects while the window is unfocused. That means a show request issued while the window is unfocused gets ignored and the cursor stays hidden.

allow me to propose a more consistent hotfix so we track a `cursor_resync_needed` flag and once focus returns, force winit to rebuild the cursor state by toggling visibility. ideally winit would handle this internally, but this workaround keeps neovide cursor hidden feat usable until that happens, but I can create a pull request targeting winit itself later to fix [this](https://github.com/rust-windowing/winit/issues/1295)

falcucci:
@Geobert could you please check it this [pull request](https://github.com/neovide/neovide/pull/3294) also fix it for you?

it has been working great for me now. let me know

falcucci:
as confirmed, closing this in favor of https://github.com/neovide/neovide/pull/3294

your contribution has been included there, thank you!

Issue #3284, opened 2025-11-10
CTRL-] defunct on Mac with German keyboard layout


quadruple-output:
**Describe the bug**
I use Neovide on Mac with a German keyboard layout. The key combination `CTRL-]` (i.e. holding down the `CTRL` key while tapping the key combination that should produce the closing square bracket) does not produce the desired effect. This effectively means that navigation within help texts is not straight-forward.

When I enter `CTRL-v` `CTRL-]` in INSERT mode, it prints `^^` instead of the expected `^]`. This is only the case in Neovide, but not when I open nvim in a terminal.

On a German keyboard, the character `]` is triggered by the key combination `Option 6`. The key combination `Shift+Option+6` results in `^` (as a "dead" key). 

**To Reproduce**
1. Open Neovide and enter INSERT mode
2. Press `CTRL-v`, followed by `CTRL-]`
3. The resulting text is `^^` instead of `^]`

**Expected behavior**
It should be `^]`.

**Screenshots**

**Desktop (please complete the following information):**
 - macOS 26.0.1
 - Neovide Version 0.15.2
 - Neovim Version 0.11.4

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

\[Edit: I originally pasted irrelevant key events, here. Updating with the relevant sequence (starting in INSERT mode):\]

```text
TRACE [neovide::window::keyboard_manager] KeyEvent {
    physical_key: Code(
        ControlRight,
    ),
    logical_key: Named(
        Control,
    ),
    text: None,
    location: Right,
    state: Pressed,
    repeat: false,
    platform_specific: KeyEventExtra {
        text_with_all_modifiers: None,
        key_without_modifiers: Named(
            Control,
        ),
    },
}
TRACE [neovide::window::keyboard_manager] Modifiers { state: ModifiersState(CONTROL), pressed_mods: ModifiersKeys(RCONTROL) }
TRACE [neovide::window::keyboard_manager] KeyEvent {
    physical_key: Code(
        KeyV,
    ),
    logical_key: Character(
        "v",
    ),
    text: Some(
        "v",
    ),
    location: Standard,
    state: Pressed,
    repeat: false,
    platform_specific: KeyEventExtra {
        text_with_all_modifiers: Some(
            "\u{16}",
        ),
        key_without_modifiers: Character(
            "v",
        ),
    },
}
TRACE [neovide::window::keyboard_manager] Key pressed <C-v> ModifiersState(CONTROL)
[...]

PR #3298, opened 2025-11-17
fix: key text normalization with macos support

falcucci:
we previously relied on text/logical_key when formatting normal keys, but macos clears those when control is held, leaving only the physical key resulting `CTRL+]` on a [german layout](https://github.com/neovide/neovide/issues/3284) to have a  `<C-6>` interpretation.

winit `KeyEvent` already exposes the true control character via `text_with_all_modifiers()` so we can recover the intended printable symbol **when option isn't meta** ofc.

https://github.com/user-attachments/assets/10760319-e603-4bc8-9515-341a088c5f04

closes https://github.com/neovide/neovide/issues/3284

### DIFF
--- a/src/window/keyboard_manager.rs
+++ b/src/window/keyboard_manager.rs
@@ -1,11 +1,11 @@
-use std::sync::Arc;
+use std::{borrow::Cow, sync::Arc};
 
 use crate::{
     bridge::{send_ui, SerialCommand},
     settings::Settings,
 };
 
-#[allow(unused_imports)]
+#[cfg(target_os = "macos")]
 use winit::platform::modifier_supplement::KeyEventExtModifierSupplement;
 use winit::{
     event::{ElementState, Ime, KeyEvent, Modifiers, WindowEvent},
@@ -18,6 +18,14 @@ use {
 };
 
 use crate::profiling::tracy_named_frame;
+
+/// See https://en.wikipedia.org/wiki/ASCII#Control_characters
+#[cfg(target_os = "macos")]
+mod ascii {
+    pub const CTRL_OFFSET: u32 = 0x40;
+    pub const CTRL_RANGE_START: u32 = 0x00;
+    pub const CTRL_RANGE_END: u32 = 0x1f;
+}
 
 fn is_ascii_alphabetic_char(text: &str) -> bool {
     text.len() == 1 && text.chars().next().unwrap().is_ascii_alphabetic()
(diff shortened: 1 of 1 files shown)


falcucci:
@quadruple-output has tested and [confirmed](https://github.com/neovide/neovide/issues/3284#issuecomment-3547492510) that patch works

Kethku:
I'm very dubious about this change. https://github.com/rust-windowing/winit/blob/6a46610632fa01cb8ce7726597af59ed0048e29a/winit-appkit/src/event.rs#L69 for instance theres a lot of handling in this file for this kind of behavior and just shifting the ascii code just feels like it might be fragile. Are we sure that this handles all cases? I'm not super paged in on this logic but my spidey senses are tingling

falcucci:
@Kethku I actually totally get the concern given. but the change is less dangerous than it looks like I guess.

Just to make it explict, starting from what we have changed, we **only** touch macOS and **only** when `ctrl + option` is pressed **and** if `option` isn't *Meta*.

walked through the trace made by the author issue to confirm we only hit this path in the broken scenario, so we are just aligning with how terminal already does to neovim itself: deliver the control byte and neovim resolves it to `<C-...>`

If you have any particular case you'd like me to test let me know or even create a macOS only setting for nightly releases, otherwise I think this is the minimal fix that unblocks ctrl+option navigation without changing the rest of the keyboard stack.

I am also running this fix since yesterday to see if I get any edge cases but everything works perfectly.

falcucci:
@Kethku you're right. this should be considered a workaround for now until winit threads that information into logical_key or adds an official control printable representation somehow. 

winit macOS pipeline [intentionally](https://github.com/falcucci/winit/blob/781d8ae212cfa1b8cb8297bdd3dd7cb1dd456cbc/src/platform_impl/macos/event.rs?plain=1#L129-L138) strips Control/Command from logical_key. thats why Ctrl+Option+6 ends up as 6 even though macOS told us the actual control byte.


In fact, alacrity does have the same problem, but has workarounds with key-bindings for windows and linux I guess but macOS still remains. https://github.com/alacritty/alacritty/issues/8308 https://github.com/alacritty/alacritty/issues/8274

fredizzimo:
You should report this issue to winit, and preferably fix it there if you can. 

We already have two similar bug reports for windows
* https://github.com/rust-windowing/winit/issues/4052
* https://github.com/rust-windowing/winit/issues/3012

I'm pretty sure that this hack will cause other issues

falcucci:
@fredizzimo this is all about macOS right now, but the issue will be addressed upstream, we can always create a toggle setting for nightly releases and see if we get edge cases and unblock @quadruple-output which I would be also glad to hear if you get any other issues. at this point I would prefer to unblock users instead of not allowing them to use it.

> I'm pretty sure that this hack will cause other issues

can you please report the cases you are sure?

@Kethku let me know what you think

fredizzimo:
I'm just saying that there are similar unfixed issues on Windows as well, the keyboard support in winit is still not bug free.

Shift+CMD used to be buggy on macOS too https://github.com/rust-windowing/winit/issues/3078

falcucci:
@fredizzimo got it, yea that's true. i'll report this with more details upstream

falcucci:
@quadruple-output to avoid unknown unknowns, I can propose to keep this branch up-to-date for you or anyone with the same problem as a workaround. merging this knowing that alacrity is also broken and that the issue is upstream is a bit risky, I would love to unblock this to nightly releases but fixing all cases where we dont have fully mapped at our end to fix in the client can become an anecdote.

quadruple-output:
I understand the raised concerns, and I would favor stability (which requires code maintainability) over possibly unstable fixes.
@falcucci Thank you for your offer of keeping this branch up-to-date. However, for me personally, this would not be required. I am currently only aware of a single key mapping (`<C-]>`) that does not work for me, and 
I have a work-around (using `g[`). If you start maintaining separate branches, I'll be worried about your long-term mental health 😉.

There is one more thing I would like to mention: I use [wezterm](https://github.com/wezterm/wezterm) which is also written in Rust and does not seem to have this issue. It might be interesting to learn how they do it.

Kethku:
notably i believe wezterm does not use winit at all but favors platform specific apis. could be wrong about that i cant look myself rn

Kethku:
i was right https://github.com/wezterm/wezterm/blob/main/window/Cargo.toml wezterm rolls their own windowing layer. Such a thing might be viable at some point but is a pretty huge maintenance burden.

PR #3308, opened 2025-11-27
refactor: improve performance for cursor animation

cubewhy:
I notice the function `animate` calls every time when render, using Itertools sometimes causes performance issues on large codebase (like linux)
Improve the sorting logic without actually resorting them
Some part of the code is AI generated, but it works fine to me.

Refactor

No

### DIFF
--- a/src/bridge/clipboard.rs
+++ b/src/bridge/clipboard.rs
@@ -2,11 +2,14 @@ use std::error::Error;
 
 use rmpv::Value;
 
-use crate::clipboard;
+use crate::clipboard::Clipboard;
 
-pub fn get_clipboard_contents(register: &Value) -> Result<Value, Box<dyn Error + Send + Sync>> {
+pub fn get_clipboard_contents(
+    clipboard: &mut Clipboard,
+    register: &Value,
+) -> Result<Value, Box<dyn Error + Send + Sync>> {
     let register = register.as_str().unwrap_or("+");
-    let clipboard_raw = clipboard::get_contents(register)?.replace('\r', "");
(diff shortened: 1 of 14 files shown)


falcucci:
It is indeed slightly faster but double-loop loses those micro-optimizations within `sorted_by` which is optimized by std lib routines and increases branch mis-predict chances, so depending on cache behavior it can show slightly higher retired instructions per frame even if overall frame time falls so I am not sure.



```rust
for (i, corner) in self.corners.iter_mut().enumerate() {
    let mut rank = 0;
    for (j, other_val) in alignments.iter().enumerate() {
        let my_val = alignments[i];
        // The comparison `j < i` ensures stability for equal values
        if *other_val < my_val || (*other_val == my_val && j < i) {
            rank += 1;
        }
    }

    corner.jump(
        &settings,
        center_destination,
        cursor_dimensions.into(),
        rank,
    )
}
```

Maybe a better win like lower frame time with same or less CPU before trading the more declarative code loops would be ideal? Maybe cache locality could explain the bump of the new loop that keeps the alignments array hot, so the zone finishes faster but higher cache residency also means the CPU spends more cycles doing useful work and fewer idling.

CleanShot 2025-12-02 at 20 22 51@2x


----

CleanShot 2025-12-02 at 20 33 25@2x

---


I'll continue later to keep an eye on it.

Kethku:
Yeah I'm inclined to not take this change. It doesn't make the code appreciably faster and to my eyes slightly obscures the purpose of this logic. Performance improvements should be driven by a profile trace rather than rules of thumb.

Issue #325, opened 2020-07-13
Can't build using "cargo build --release" windows 10.


sohaibjkhan:
Kinda new to all this but here's the error: 

```
error: failed to run custom build command for `sdl2-sys v0.34.0`

Caused by:
  process didn't exit successfully: `C:\Program Files\neovide\target\release\build\sdl2-sys-4f674920fe02e694\build-script-build` (exit code: 101)
--- stdout
running: "cmake" "C:\\Program Files\\neovide\\target\\release\\build\\sdl2-sys-c3057c18a50b6eb5\\out\\SDL2-2.0.10" "-G" "Visual Studio 16 2019" "-Thost=x64" "-Ax64" "-DSDL_SHARED=OFF" "-DSDL_STATIC=ON" "-DCMAKE_INSTALL_PREFIX=C:\\Program Files\\neovide\\target\\release\\build\\sdl2-sys-c3057c18a50b6eb5\\out" "-DCMAKE_C_FLAGS= -nologo -MD -Brepro" "-DCMAKE_C_FLAGS_RELEASE= -nologo -MD -Brepro" "-DCMAKE_CXX_FLAGS= -nologo -MD -Brepro" "-DCMAKE_CXX_FLAGS_RELEASE= -nologo -MD -Brepro" "-DCMAKE_BUILD_TYPE=release"
-- Selecting Windows SDK version 10.0.18362.0 to target Windows 10.0.18363.
-- Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE)
--
-- SDL2 was configured with the following options:
--
-- Platform: Windows-10.0.18363
-- 64-bit:   TRUE
-- Compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/BuildTools/VC/Tools/MSVC/14.26.28801/bin/Hostx64/x64/cl.exe
--
-- Subsystems:
--   Atomic:    ON
--   Audio:     ON
--   Video:     ON
--   Render:    ON
--   Events:    ON
--   Joystick:  ON
--   Haptic:    ON
--   Power:     ON
--   Threads:   ON
--   Timers:    ON
--   File:      ON
--   Loadso:    ON
--   CPUinfo:   ON
--   Filesystem:        ON
--   Dlopen:    ON
--   Sensor:    ON
--
-- Options:
--   3DNOW                  (Wanted: ON): OFF
--   ALSA                   (Wanted: OFF): OFF
[...]

PR #3312, opened 2025-11-30
fix: ensure teardown via context to avoid clipboard outliving event loop

falcucci:
just cherry-picking only the isolated clipboard lifetime fix from https://github.com/neovide/neovide/pull/3252

Instead of a whole refactor, the clipboard is now owned centrally by the app: we build a single `Arc<Mutex<Clipboard>>` from the event loop, forward a weak handle through the runtime, error handling, RPC layers, etc and each consumer locks that shared handle when it needs clipboard access.

This removes the old global OnceLock and ensures the clipboard drops before the event loop shuts down, fixing the crash we hit when winit begins to tears things down.

fixes https://github.com/neovide/neovide/issues/2868
fixes https://github.com/neovide/neovide/issues/3311

### DIFF
--- a/src/bridge/clipboard.rs
+++ b/src/bridge/clipboard.rs
@@ -2,11 +2,14 @@ use std::error::Error;
 
 use rmpv::Value;
 
-use crate::clipboard;
+use crate::clipboard::Clipboard;
 
-pub fn get_clipboard_contents(register: &Value) -> Result<Value, Box<dyn Error + Send + Sync>> {
+pub fn get_clipboard_contents(
+    clipboard: &mut Clipboard,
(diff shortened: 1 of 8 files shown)


fredizzimo:
This misses the point of the fix. You have to guarantee that the clipboard is deleted before the internal wayland socket, which is owned by the winit event loop.

Additionally  https://github.com/neovide/neovide/pull/325, takes care of the ownership of the renderers, which has similar requirements

falcucci:
@fredizzimo yea that's true.

I prepared a fix already after my push since I was focused on lifetime management only and forgot the actual issue.

PR #3313, opened 2025-12-01
fix: ensure clean up during teardown and on drop

falcucci:
An easier and consistent fix for the Wayland edge case issue for the clipboard crash.

After https://github.com/neovide/neovide/pull/3312 the issue persisted on some edge cases and the clipboard worker kept outliving the event loop, which kept libwayland aborting because we dropped the connection before the worker stopped.

the ownership in the UpdateLoop now holds both the runtime and the clipboard Arc. It has a fn teardown() helper for the Drop impl that first drains the clipboard ensuring the clear up while the event loop is still alive. That ordering seems to be important to prevent the Wayland crash.

As a side note: I finally could reproduce this in another laptop using wayland but mainly, thanks @ewhac for reporting and [testing](https://github.com/neovide/neovide/issues/3311#issuecomment-3597758854) it constantly without hesitate for even a second. that was tricky but very appreciated, sir. lovely.

### DIFF
--- a/src/error_handling.rs
+++ b/src/error_handling.rs
@@ -1,7 +1,7 @@
 use std::{
     io::{stdout, IsTerminal},
     process::ExitCode,
-    sync::Arc,
+    sync::{Arc, Mutex},
 };
 
 use anyhow::{Error, Result};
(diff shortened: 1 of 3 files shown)


falcucci:
and thanks @paulobressan for renting for some minutes his machine to reproduce the fix.

Issue #3299, opened 2025-11-18
Performance issue on 5K screen


trongrg:
**Describe the bug**
So I just got a new 5K monitor, and Neovide, which is my main code editor, start to have a lot of UI lags to the point that I cannot use it anymore

**To Reproduce**
1. Open neovide
2. Drag it to the 5K monitor
3. Maximize it
4. Do anything with the editor: open a file, open picker, scroll
5. See that there is a massive lag on every rendering. Sometimes the picker window get re-rendered before disappearing

**Expected behavior**
No lags

**Screenshots**
[neovide-screen-recordings.zip](https://github.com/user-attachments/files/23604849/neovide-screen-recordings.zip)

**Desktop (please complete the following information):**
 - OS: macOS 26.1
 - Neovide: 0.15.2
 - Neovim Version: 0.11.5

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**

[neovide-log.zip](https://github.com/user-attachments/files/23604810/neovide-log.zip)

**Additional context**
Moving neovide window from the 5K monitor to the 4K monitor and the issue stops

diegodario88:
I'm experiencing micro-stuttering when scrolling, and it only happens on my 4K display. When I scroll just a small amount, I can see the dedicated GPU usage spike up to 75%. However, when I lower the resolution to 1920x1080 (Full HD), the scrolling becomes smooth, and I see very few spikes; if they do occur, they peak at a maximum of 50%.

# Version
- Neovide: 0.15.2
- Neovim: 0.11.5

## Hardware Information:
- **Hardware Model:**                              Dell Inc. Precision 3591
- **Memory:**                                      32.0 GiB
- **Processor:**                                   Intel® Core™ Ultra 7 165H × 22
- **Graphics:**                                    Intel® Graphics (MTL)
- **Graphics 1:**                                  NVIDIA RTX 1000 Ada Generation Laptop GPU

## Software Information:
- **OS Name:**                                     Fedora Linux 43 (Workstation Edition)
- **GNOME Version:**                               49
- **Windowing System:**                            Wayland
- **Kernel Version:**                              Linux 6.17.11-300.fc43.x86_64

# Lua Config
```lua
vim.g.neovide_floating_shadow = false
vim.g.neovide_hide_mouse_when_typing = true
vim.g.neovide_scroll_animation_far_lines = 1
```

# Yaml Config
```yaml
frame = "none"
maximized = true
[...]

PR #3324, opened 2025-12-21
fix(renderer): prevent font cache thrashing on high-DPI displays

diegodario88:
The previous implementation triggered `cleanup_font_cache` aggressively whenever the render loop was waiting for VSync or idle. On high-resolution displays (like 4K), this caused excessive cache thrashing, forcing the GPU to re-upload font textures every frame and significantly increasing GPU usage.

Closes: #3299

- Fix

- No

### DIFF
--- a/src/renderer/fonts/caching_shaper.rs
+++ b/src/renderer/fonts/caching_shaper.rs
@@ -365,6 +365,11 @@ impl CachingShaper {
     }
 
     pub fn cleanup_font_cache(&self) {
+        // Only purge if we are truly about to exhaust the cache.
+        // See: https://github.com/neovide/neovide/issues/3299
+        // On high-DPI displays the old unconditional purge invalidated
+        // glyphs every frame, which forced the GPU to keep
+        // re-uploading textures and tanked performance.
         let limit = font_cache_limit();
         let used = font_cache_used();
 
--- a/src/window/update_loop.rs
+++ b/src/window/update_loop.rs
@@ -303,6 +303,9 @@ impl UpdateLoop {
             self.animate();
             self.schedule_render(skipped_frame);
         } else {
+            // Cache purging should only happen once we become idle; doing it while throttling for
+            // vsync caused Skia to evict glyphs mid-animation and re-upload them every frame.
+            // See https://github.com/neovide/neovide/pull/3324
             if self.num_consecutive_rendered > 0 {
                 self.window_wrapper
                     .renderer


fredizzimo:
Hm, there's something strange going on here.

I guess there could be a macOS specific issue. But on the other platforms the font cache usage never gets even close to filling up, even when rendering chinese with a lot of characters on an ultrawide. But it's possible that macOS uses a lot more for some reason. Can you provide some [tracy logs](https://neovide.dev/troubleshooting.html#performance-profiling) (the compress very well), it should show the usage?

Secondly, it should never purge the cache when waiting for vsync during an animation, it's only done when Neovide goes back to idle, and there's probably actual time to spare. But I don't own a mac, so I have never properly verified that the macOS render loop works as it should, so there could be bugs. But the tracy logs should confirm it. It's done on purpose there, to ensure that there's always plenty of space available during the actual animation and avoid stuttering. According to my measurements (but I don't have a mac, so it was done on other platforms), a single screen never consume even close to half of the cache, so purgin to that size seems like a good idea. And Skia never purges anything that is still in use, so it's always safe. So I suggest that code is left, or some alternative way of ensuring that there's space available, without unbounded growth.

With  the current caching limit, it should be set so that half the cache is a reasonable maximum working limit. You should never be able to fill that with just standard ascii characters, or a realistic screen full of different characters. The other half should be just an extra safety, for when you are scrolling around in a document with many different unicode characters, and all the characters in the whole document don't fit. So you need different sets depending on which part of the document you look at. You will take a slight performance hit, but not worse than scrolling to that part of the document the first time. As mentioned pathological cases where the screen don't fit half of the cache is still OK, since skia should see that all is in use and don't purge it.

diegodario88:
Thanks for the feedback! Just to clarify, I'm not on macOS; I’m actually running Linux Wayland. I suspect the high-DPI scaling on my display might be pushing the glyph sizes higher than usual, which could be why I'm hitting those limits sooner than expected.

Regarding the purge logic, I'll check the Tracy logs to see if it’s being triggered more often than it should during what I perceive as idle. I'll get those logs over to you shortly so we can figure out if it's a bug or if we just need a different approach for high-res screens. Cheers!

diegodario88:
Attached are the Tracy traces for comparison (captured on Linux Wayland with high-DPI scaling):

- [neovide-prob.zip](https://github.com/user-attachments/files/24281400/neovide-prob.zip) — Current main: Demonstrates the cache thrashing/stuttering.
- [neovide-fix.zip](https://github.com/user-attachments/files/24281426/neovide-fix.zip) — With the proposal: Shows the improved behavior and stable cache usage.

Hopefully, this helps clarify the behavior on high-resolution displays

fredizzimo:
I can see the improvement, but I don't immediately see what's wrong. I will look a bit closer later.

The font cache usage never seem to go over 500 kb, if Skia reports correctly, and there's no indication that it shrinks in the graphs either as far as I can see. It looks like all CPU timings are OK, so it's probably doing something on the GPU, probably triggering a stall. 

I think one thing that could be tried, is to manually check if say 90 % is filled, and only then trigger the cleanup. That should prevent it from being constantly called, while still making sure that there's normally plenty of space.

I also have a quite high resolution monitor 3840 × 1600, and I'm running at 144hz but I have not spotted similar problems on that.  It's possible that the problem is GPU specific. 

Edit: It does seem to call `cleanup_font_cache` when it wakes up because of some other event than the signal to render. So the condition is wrong, it should only do that when it's really idle. I think that can be fixed by moving the call to the else block here, so that it is called when the animation is finished. Note I'm not 100% sure it's the right place, since I don't remember the code flow exactly and only took a quick look. Alternatively I think you can check for `ShouldRender::Wait` after `prepare_and_wait` has been called, since wait should only be set when it's idle.
https://github.com/neovide/neovide/blob/40f6b24c64de48f053dead24feb653b9d017304b/src/window/update_loop.rs#L306-L318

diegodario88:
Updated the PR and collected fresh logs. Comparison between `neovide-fix-v2` and the original `trace_dump_prob`:

1.  Elimination of Font Cache Thrashing:
    *   Before: `purge_font_cache` was called 5,168 times.
    *   After: `purge_font_cache` was called 0 times.
    *   This confirms that the aggressive purging has been completely stopped in this session, which is expected since the cache (8MB) likely didn't reach the 90% threshold.
2.  Stability of Font Loading:
    *   load_font calls remained constant (15 calls in both traces), indicating that the fix didn't negatively impact the ability to load fonts when needed.
3.  Improved Caching Efficiency:
    *   shape_cached calls increased significantly (from ~13k to ~57k), likely due to a longer or more active session, but the average time per call decreased from 3,384ns to 2,247ns.
    *   This suggests that looking up glyphs in the cache is now faster or more stable because the cache isn't being constantly invalidated.


[neovide-fix-v2.zip](https://github.com/user-attachments/files/24299637/neovide-fix-v2.zip)

falcucci:
thanks for the pull request @diegodario88 I'll analyze and take a look at this one soon.

falcucci (commented):
It is always great preventing needless GPU uploads.

I recommend adding some comments between the vsync throttling and cache purges for future maintenance, before I approve.

Also we can update the description of the pr where it says

> Increases the FONT_CACHE_SIZE from 8MB to 32MB to better accommodate the larger glyph sizes required for high-DPI scaling.

as far as I see it we dont really need it anymore.

and thank you for taking the ownership to fix the stuttering at such edge cases for you and others.

falcucci:
I had to generate the traces on macOS just to double check, thank you!

PR #3325, opened 2025-12-21
classic division by zero panic.

falcucci:
```
Neovide panicked with the message 'attempt to calculate the remainder with a divisor of zero'. (File: src/utils/ring_buffer.rs; Line: 89, Column: 38)
```

During the :restart impl we just found this bug at the draw line whenever the window hasn’t received a position update yet or its backing buffers are empty. 

This prevents stale batches that arrive after the renderer reset from indexing a zero-length ring buffer, stopping the [rem_euclid(0)](https://doc.rust-lang.org/stable/std/primitive.i32.html#method.rem_euclid) panic while we blank the UI for :restart.

In short: just ignore invalid draw operations briefly so we trade a harmless no-op for stability.

### DIFF
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -23,7 +23,6 @@ use std::{
 };
 
 use itertools::Itertools;
-use log::error;
 use progress_bar::{ProgressBar, ProgressBarSettings};
 use skia_safe::Canvas;
 
@@ -461,10 +460,10 @@ impl Renderer {
                         }
                         _ => {
                             let settings = self.settings.get::<CmdLineSettings>();
-                            // Ignore the errors when not using multigrid, since Neovim wrongly sends some of these
+                            // Neovim can emit draw commands before a window is positioned when using multigrid.
                             if !settings.no_multi_grid {
-                                error!(
-                                    "WindowDrawCommand: {command:?} sent for uninitialized grid {grid_id}"
+                                log::warn!(
+                                    "Ignoring {command:?} sent for uninitialized grid {grid_id}"
                                 );
                             }
                         }
--- a/src/renderer/rendered_window.rs
+++ b/src/renderer/rendered_window.rs
@@ -395,6 +395,15 @@ impl RenderedWindow {
             WindowDrawCommand::DrawLine { row, line } => {
                 tracy_zone!("draw_line_cmd", 0);
 
+                if self.actual_lines.is_empty() || !self.valid {
+                    log::warn!(
+                        "Ignoring DrawLine for grid {} row {} because the window is not ready yet",
+                        self.id,
+                        row
+                    );
+                    return;
+                }
+
                 let line = RenderedLine {
                     line,
                     background_picture: None,


Kethku:
is this an order problem? like if we queue up these draws and render them when the window becomes valid would that be more correct?

falcucci:
hmmm I think no, this change would actually be better because we align better with neovim behavior, we just ignore unusable work until the UI has enough information, then we let the fresh batch fill the window.

as far as I am understanding it's not like an ordering bug we can fix in the renderer, it's more like when Neovim sends the DrawLine before its Position, we could maybe also stash every pre-position draw command in a side queue, but we would still have to guess whether the window will be the same size once it's positioned so I am not really sure if worth it.

The redraw protocol seems to guarantee that Position defines the grid, then neovim immediately resends every line for that grid. If we stash pre-position draws, we still can't apply them until the grid exists anyway and once it exists the data is already re-sent.

fredizzimo:
I don't think Neovim ever sends drawline before the position updates, unless the have introduced a new regression. `:restart` should not affect that either, since the Neovim process is restarted. It's more likely that you keep some buffers on the Neovide side that should have been cleared.

fredizzimo:
Also silently ignoring errors without even a log message is a dangerous path to take that quickly makes it almost impossible to find bugs that possibly occur further down the line (for example ignoring drawline will likely cause screen corruption). You did it once before here https://github.com/neovide/neovide/pull/3109 (hint, it masks a real bug in the code)

falcucci:
This is a preparetion to avoid the next :restart pull request to crash that I got during the impl.

what bit us here is that we clear its RenderedWindow buffers as soon as :restart is requested as a defensive method at my local branch. but there can still be old draw batches waiting in our queue. After the reset those batches drain and try to write into the now-empty ring buffer, which is why the guard was added. it isn't Neovim reordering events problem, it's Neovide processing stale work after we've thrown the window state away.

This is a simple approach. ignoring those leftovers fixes the crash and the guard is the last line of defense in case any slip through.

> Also silently ignoring errors without even a log message is a dangerous path to take that quickly makes it almost impossible to find bugs that possibly occur further down the line (for example ignoring drawline will likely cause screen corruption). You did it once before here https://github.com/neovide/neovide/pull/3109 (hint, it masks a real bug in the code)

We didn't make it fully silent - we downgraded the log from error! (which spammed every restart) to trace!, so it's still recorded when we want to troubleshoot. honestly keeping it at error! was misleading because it implied Neovim was misbehaving; switching to trace! acknowledges that the bug is just our own cleanup window. *If we ever see a real ordering problem, we can re-elevate the log*, but until then the trace-level message plus the guard keeps the logs usable without hiding anything from debugging.

And this is what I wrote in the pull request you just mentioned

> I think that defensive programming doesn't hurt as well, just the fact that this can panic is worth to merge IMO. saturating_sub is highly recommended for safe op.

> Root cause must still be investigated tho

fredizzimo:
> but there can still be old draw batches waiting in our queue
Working around that by ignoring errors, instead of properly emptying the queues sounds like a nightmare to me, but it's your choice.

>  honestly keeping it at error! was misleading because it implied Neovim was

But Neovim is misbehaving if that happens

This is the critical error that I'm talking about
```rust
    if self.actual_lines.is_empty() || !self.valid {
        return;
}
```

And that will cause screen corruption, unless Neovim sends another update later

falcucci:
That guard only trips during the window after we've cleared RenderedWindow on :restart to make sure we have the correct state and before Neovim has reattached and resent the full state.

In that interval the only draw commands we can receive are leftovers from the pre-reset session and those are guaranteed to be overwritten because the freshly spawned Neovim sends a complete Position + DrawLine sequence as soon as it reattaches.

So we're not discarding legitimate updates mid-session. As soon as Neovim redraws, which it always does right after restart and everything renders normally, no corruption.

fredizzimo:
I still think it's a nightmare of architecture to mix commands from different sessions, ignore errors and hope for the best.  Especially since you have already done multi-window support that *should* completely isolate the instance, or at least be able to wait until an instance has completely quit. My refactoring certainly does, since different queues are used.

But if you think this is the way to go, then so be it.

Kethku:
This is unrelated to this pr and off topic. If you wanna talk architecture please either do it in dms, in the discord server, or in relevant issues.

PR #2101, opened 2023-11-04
docs(faq): Clean up cmd-c/cmd-v keymap examples

lkhphuc:
## What kind of change does this PR introduce?
- Documentation

## Did this PR introduce a breaking change? 
- No

Clean up the mapping for copy and paste, also add the terminal mode mapping.

PR #3337, opened 2026-01-21
docs: fix and simplify clipboard bindings example

mwinters0:
The existing example did not work in insert mode in Linux, and the nvim api robustly handles paste in all modes.

- Documentation

- No

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,7 +72,7 @@ jobs:
         uses: cargo-bins/cargo-binstall@main
 
       - name: Install cargo-nextest
-        run: cargo binstall -y cargo-nextest
+        run: cargo binstall -y --locked cargo-nextest
 
       - uses: Swatinem/rust-cache@v2
         with:
--- a/src/main.rs
+++ b/src/main.rs
@@ -68,7 +68,7 @@ use renderer::{
 };
 use running_tracker::RunningTracker;
 use window::{
-    create_event_loop, determine_window_size, UpdateLoop, UserEvent, WindowSettings, WindowSize,
+    create_event_loop, determine_window_size, Application, UserEvent, WindowSettings, WindowSize,
 };
 
 pub use channel_utils::*;
@@ -122,7 +122,7 @@ fn main() -> ExitCode {
     ) {
         Err(err) => handle_startup_errors(err, event_loop, settings.clone(), clipboard),
         Ok((window_size, initial_config, runtime)) => {
-            let mut update_loop = UpdateLoop::new(
+            let mut application = Application::new(
                 window_size,
                 initial_config,
                 event_loop.create_proxy(),
@@ -131,7 +131,7 @@ fn main() -> ExitCode {
                 clipboard,
             );
 
-            let result = event_loop.run_app(&mut update_loop);
+            let result = event_loop.run_app(&mut application);
 
             match result {
                 Ok(_) => running_tracker.exit_code(),
@@ -286,6 +286,7 @@ fn setup(
         settings,
         colorscheme_stream,
     )?;
+
     Ok((window_size, config, runtime))
 }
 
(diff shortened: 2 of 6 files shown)


falcucci:
I am going to double check those soon, there is also https://github.com/neovide/neovide/pull/2101 which is pretty similar, maybe we can mashup things to have a common value after validation.

falcucci:
@mwinters0 going through the comments like [this one](https://github.com/neovide/neovide/issues/1263#issuecomment-1972013043), I have ended up that this would be the most consistent alternative, can you please double check if everything works for you?

```lua
if vim.g.neovide then
  local function paste()
    vim.api.nvim_paste(vim.fn.getreg("+"), true, -1)
  end

  vim.keymap.set({ "n", "v" }, "<D-c>", '"+y', { silent = true })
  vim.keymap.set({ "n", "v", "i", "c" }, "<D-v>", paste, { silent = true })
  vim.keymap.set("t", "<D-v>", [[<C-\><C-n>"+P]], { silent = true })
end
```

mwinters0:
@falcucci

- Investigating a bit further, `nvim_paste()` works in mode `t` as well.
- The nvim docs suggest mapping to function rather than key sequence to avoid a second round of keymap interpretation, which allows the command to execute faster and works regardless of weird keymaps.  I converted Copy and Save to this.
- Added `desc` and `silent`.

I should add:
- I'll be at FOSDEM if you'd like to discuss :)

falcucci:
@mwinters0 thanks for updating but indentation is not really good. plus I think it's safer to use the one I have [suggested](https://github.com/neovide/neovide/pull/3337#issuecomment-3800259252), since I have double checked it works in terminal buffers. so I want to avoid people having a specific terminal mapping. can you please double check if the suggested works for you? If so, I can update your pull request and commit some changes.

mwinters0:
- I don't understand.  In what way is your suggested approach safer?  It requires default keymaps, which are not guaranteed, whereas the lua functions work regardless of other keymaps.  I have tested that this approach works in terminal buffers.
- What suggestions would you have for changing the indentation?  I have no opinions about code style.

Edit:
- I've made the functions one line.

falcucci:
@mwinters0 cool, it seems [this](https://github.com/neovide/neovide/issues/1263#issuecomment-1972013043) approach worked better for people in general, I guess we can endeup recommending this

```lua
if vim.g.neovide then
  local function save() vim.cmd.write() end
  local function copy() vim.api.nvim_cmd({ cmd = "yank", reg = "+" }, {}) end
  local function paste() vim.api.nvim_paste(vim.fn.getreg("+"), true, -1) end

  vim.keymap.set({ "n", "i", "v" }, "<D-s>", save, { desc = "Save" })
  vim.keymap.set("v", "<D-c>", copy, { silent = true, desc = "Copy" })
  vim.keymap.set({ "n", "i", "v", "c", "t" }, "<D-v>", paste, { silent = true, desc = "Paste" })
end
```

with a slightly change for the nvim_paste passing true so neovim treats the clipboard content as a real paste, preserving line endings and bracketed‑paste behavior. That avoids the "paste only after another key" delay in the command‑line and makes plugins that depend on proper paste handling for example fzf‑lua, to work properly, as far as I can see.

falcucci:
I have also added https://github.com/lkhphuc as a co-author since he had a very [similar](https://github.com/neovide/neovide/pull/2101) open pull request.

lkhphuc:
thank you @falcucci . 
just curious I switched my old keymaps to this new ones and for some reason the copy keymap only copy the last line of my visual selection.
Wondering if it's just me?

falcucci:
@lkhphuc yep, let me know if this one fix for you so I can update the docs.

```lua
if vim.g.neovide then
  local function save() vim.cmd.write() end
  local function copy() vim.cmd([[normal! "+y]]) end
  local function paste() vim.api.nvim_paste(vim.fn.getreg("+"), true, -1) end

  vim.keymap.set({ "n", "i", "v" }, "<D-s>", save, { desc = "Save" })
  vim.keymap.set("v", "<D-c>", copy, { silent = true, desc = "Copy" })
  vim.keymap.set({ "n", "i", "v", "c", "t" }, "<D-v>", paste, { silent = true, desc = "Paste" })
end
```

lkhphuc:
yep it's working now. thanks.

PR #3361, opened 2026-02-12
feat: support  `title-hidden` option in full frame for macOS

TymekV:
Clarified `--title-hidden` option's behavior, according to the following code snippet:

https://github.com/neovide/neovide/blob/7a3f56830caf671ad884caff4990741336282a1e/src/window/mod.rs#L235-L248

- Documentation

- No

### DIFF
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -234,7 +234,7 @@ pub fn create_window(
 
     #[cfg(target_os = "macos")]
     let mut window_attributes = match frame_decoration {
-        Frame::Full => window_attributes,
+        Frame::Full => window_attributes.with_title_hidden(title_hidden),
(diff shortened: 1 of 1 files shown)


falcucci:
thank you @TymekV 

It seems like there is a need for users who want to keep the frame = full but using tittle-hidden.

I see no reason to not just add this customization for them?

```rust
Frame::Full => window_attributes.with_title_hidden(title_hidden),
```

then you could remove your updated docs

TymekV:
Yss, that would actually be a better solution. I thought that there were some macOS limitations preventing this or something similar.

falcucci:
@TymekV feel free to update it, then I guess we are good to merge

TymekV:
Done

falcucci:
thank you @TymekV

PR #3363, opened 2026-02-13
fix: properly detach from the terminal when forking

### LINKED ISSUE #3362: Does not detach from the terminal properly with `--fork`
**Describe the bug**
Neovide doesn't detach from the terminal properly when used with `--fork`, so if it's invoked from a process managing the terminal and that process exits (taking the terminal with it), Neovide is killed.

**To Reproduce**
1. Run `env -u TMUX tmux new 'neovide --fork; sleep 3'`
2. Notice that Neovide starts.
3. Notice that Neovide then dies after about three seconds.

**Expected behavior**
I expected Neovide to detach fully from the terminal, as Neovim-Gtk and Neovim-QT do.

**Screenshots**

 - OS: macOS 26.3 and Debian unstable
 - Neovide Version 0.15.2 (macOS) and `4165be85db4f180756c8efd877400db2ea9e8937` (Debian)
 - Neovim Version 0.11.6 (Debian)

Attached below, since it's too large for the issue body.

[neovide_rCURRENT.log](https://github.com/user-attachments/files/25275421/neovide_rCURRENT.log)

**Additional context**
This is a contrived example with tmux, but I'm using software to spawn Neovide on a different system and my software gives the shell invoking Neovide a PTY.  When the first Neovide process exits, so does the shell, and then Neovide dies.  tmux is just a much easier way to reproduce this problem.

I _believe_ (but have not verified) that this problem occurs because when the terminal goes away, the kernel sends `SIGHUP` to all of its children in the same process group.  The double-fork technique (e.g., from `daemon` in the `fork` crate v0.4.0 or newer) may fix this.

bk2204:
If Neovide is run in forking mode with a terminal and the shell from which it is invoked exits, Neovide is killed.  This can be easily seen with `env -u TMUX tmux new 'neovide --fork; sleep 3'`.

To avoid that, we must detach from the terminal using the double-fork technique, which calls `setsid` and closes the file descriptors after the first fork so that we detach from the terminal.  Then, when the caller exits, we will not receive a SIGHUP and will continue running.

This is implemented in the `fork` crate, which contains an analogue of the `daemon` function found on Linux and the BSDs, so use that here to detach properly.

Fixes #3362

- Fix

- No

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -911,6 +911,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "aa9a19cbb55df58761df49b23516a86d432839add4af60fc256da840f66ed35b"
 
 [[package]]
+name = "fork"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "30268f1eefccc9d72f43692e8b89e659aeb52e84016c3b32b6e7e9f1c8f38f94"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "fsevent-sys"
 version = "4.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


falcucci:
really appreciated contribution @bk2204 thank you

Issue #3136, opened 2025-05-16
Subpixel antialias isn't working on Windows


danlkv:
**Describe the bug**
Subpixel antialias isn't working on Windows with neovide 15.0.1

**Expected behavior**
Colored edges like in #283

**Screenshots**
Neovide:
![Image](https://github.com/user-attachments/assets/a57afa7a-1ca4-4b76-95e1-2a18bd42e6e7)

Expected:
![Image](https://github.com/user-attachments/assets/19551050-4e3e-4d94-ab15-f3ebfee8069f)

**Desktop:**
 - OS: Windows 10
 - Neovide Version 0.15.0 (also tried with 0.14.1 and 0.13.2)
 - Neovim Version 0.11.1
 - Font configuration:

```
vim.opt.guifont = "Victor Mono:h10:W500:#e-subpixelantialias:#h-full" 
```

danlkv:
https://github.com/mono/SkiaSharp/issues/2308#issuecomment-1328363832 - this issue is new in rust-skia 2.88, but it's strange that I saw the same problem with older neovide

aqilc:
Encountering something similar on my end.

NoCodeNoBlunder:
Same here

danlkv:
FYI I did get to work this on my local build, but it's not in a shape appropriate for an MR. A proper implementation would need to support the settings correctly. But for anyone curious, here's the patch:


<details>
<summary>Subpixel antialias support patch</summary>

```diff
diff --git a/src/main.rs b/src/main.rs
index 0977ab6..881fedf 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -104,6 +104,7 @@ fn main() -> ExitCode {

     let running_tracker = RunningTracker::new();
     let settings = Arc::new(Settings::new());
+    println!("Neovide version: {}", env!("CARGO_PKG_VERSION"));

     match setup(
         event_loop.create_proxy(),
@@ -256,6 +257,7 @@ fn setup(
     };

     let mut runtime = NeovimRuntime::new()?;
+    println!("Settings: {settings:?}");
     runtime.launch(proxy, grid_size, running_tracker, settings)?;
     Ok((window_size, config, runtime))
 }
diff --git a/src/renderer/d3d.rs b/src/renderer/d3d.rs
index 9cc85a8..e619ca3 100644
--- a/src/renderer/d3d.rs
+++ b/src/renderer/d3d.rs
[...]

PR #3367, opened 2026-02-16
Add g:neovide_pixel_geometry setting

khyperia:
Fixes https://github.com/neovide/neovide/issues/3136

In testing this PR, it appears that once you set `vim.g.neovide_pixel_geometry`, you need to resize the window (or trigger a redraw or something, I suppose) for the content to update to rerender with the new setting. Also, it appears that RGBV/BGRV do nothing on Windows at least (setting it to it makes text greyscale-aliased for me), but that's probably some Skia issue, and might be different on different platforms - the vertical setting is at least there for the future, in case Skia changes.

Not implemented in this PR is fetching the value from the OS. I know that Wayland exposes this value in its protocol, and surely Windows has some API to fetch the current cleartype pixel geometry setting, same for Mac, but I have no idea how that value might be plumbed through whatever graphics API that we can fetch it from. At least this PR allows the setting to be "hardcoded" by users - fetching it from the OS can be done in a later PR, IMO.

In my silly little opinion, this setting should default to Unknown, but in the issue, @fredizzimo said:

> I think we could set `RGBH` as the default, since the majority of the monitors have that layout.

which, valid, I suppose! :sparkles:

### DIFF
--- a/src/renderer/d3d.rs
+++ b/src/renderer/d3d.rs
@@ -7,7 +7,7 @@ use skia_safe::{
         BackendRenderTarget, DirectContext, FlushInfo, Protected, SurfaceOrigin, SyncCpu,
     },
     surface::BackendSurfaceAccess,
-    Canvas, ColorSpace, ColorType, PixelGeometry, Surface, SurfaceProps, SurfacePropsFlags,
+    Canvas, ColorSpace, ColorType, Surface, SurfaceProps, SurfacePropsFlags,
 };
 use windows::core::{Interface, Result, PCWSTR};
 use windows::Win32::Graphics::DirectComposition::{
@@ -369,7 +369,7 @@ impl D3DSkiaRenderer {
 
             let surface_props = SurfaceProps::new_with_text_properties(
                 SurfacePropsFlags::default(),
-                PixelGeometry::default(),
+                render_settings.pixel_geometry.into(),
                 render_settings.text_contrast,
                 render_settings.text_gamma,
             );
--- a/src/renderer/fonts/font_options.rs
+++ b/src/renderer/fonts/font_options.rs
@@ -2,13 +2,14 @@ use std::{collections::HashMap, fmt, iter, num::ParseFloatError, sync::Arc};
 
 use itertools::Itertools;
 use log::warn;
+use rmpv::Value;
 use serde::Deserialize;
 use skia_safe::{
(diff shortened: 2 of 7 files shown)


khyperia:
> fetching it from the OS can be done in a later PR, IMO

I believe the most appropriate place for this to be implemented is `winit`, and `winit` does not expose this information. I traced the place it's provided on wayland, and that's where it stops - `winit-wayland` doesn't expose the `subpixel` field provided by `smithay_client_toolkit::output`: https://smithay.github.io/client-toolkit/smithay_client_toolkit/output/struct.OutputInfo.html#structfield.subpixel

falcucci:
> In testing this PR, it appears that once you set vim.g.neovide_pixel_geometry, you need to resize the window (or trigger a redraw or something, I suppose) for the content to update to rerender with the new setting.

to fix it just add the PixelGeometry variant matching [here](https://github.com/neovide/neovide/blob/main/src/window/window_wrapper.rs?plain=1#L362) and it should trigger and update it automatically :)

khyperia:
aha, thank you!

falcucci:
about we defaulting to RGBH,

skia seems to [intentionally](https://api.skia.org/SkSurfaceProps_8h.html#ab546a3fcdf190266bb1e5b4be38db98f) defaults it to Unknown, so rust skia does forward its behavior and I guess they do this to eliminate the margin error of guessing even if small, which is safer indeed. I guess we should keep it Unknown. and btw here you were very precise: 

> Not implemented in this PR is fetching the value from the OS. I know that Wayland exposes this value in its protocol, and surely Windows has some API to fetch the current cleartype pixel geometry setting, same for Mac, but I have no idea how that value might be plumbed through whatever graphics API that we can fetch it from.

thank you for going forward on this. it could probably work on Wayland but macOS for example doesn't have the subpixel exposed officially on its protocol as far as I could check, there is some internal/private options tho but that enforces your point.

and as you said this can be done in another issue/pull request targeting specific platforms.

khyperia:
> I guess we should keep it Unknown

Right! Changed it to default to Unknown.

> thank you for going forward on this

thank _you_ for helping make such a great project!

khyperia:
Ah whoops forgot to update the docs too with the switch to default Unknown, did so now.

falcucci:
thank you very much @khyperia very appreciated contribution!

PR #3376, opened 2026-02-25
improv: on macos use ppid heuristic for gui launch

falcucci:
we now stop using `$TERM` to decide when to wrap neovim in a login shell, so we check and treat ppid as a gui launch signal.

https://en.wikipedia.org/wiki/Launchd#Components

### DIFF
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -116,6 +116,16 @@ fn std_command_from_spec(spec: CommandSpec) -> StdCommand {
     result
 }
 
+#[cfg(target_os = "macos")]
+fn launched_from_desktop() -> bool {
+    // On macOS, apps launched from Finder or `open` are spawned by launchd = PPID 1.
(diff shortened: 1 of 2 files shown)


src/bridge/command.rs
```diff
@@ -116,18 +116,27 @@ fn std_command_from_spec(spec: CommandSpec) -> StdCommand {
...
     args: &[String],
     _cmdline_settings: &CmdLineSettings,
 ) -> CommandSpec {
-    use std::env;
     use uzers::os::unix::UserExt;
-    if env::var_os("TERM").is_some() {
-        // If $TERM is set, we assume user is running from a terminal, and we shouldn't
-        // re-initialize the environment. See https://github.com/neovide/neovide/issues/2584
+    if !launched_from_desktop() {
+        // If we're not launched from the desktop, assume a terminal launch and avoid
```
falcucci: using the 'desktop' terminology just for general understanding

PR #3388, opened 2026-03-07
feat: add --no-embed-arg command line argument

khyperia:
I'm a silly goose. I want this to work:

```sh
neovide --neovim-bin ssh -- my-server nvim
```

but it doesn't:

- `-p` is automatically **prepended** to the arguments to neovim-bin (able to be disabled with `--no-tabs`)
- `--embed` is automatically **prepended** with no way to disable it

resulting in the neovim-bin commandline being `ssh --embed -p my-server nvim`, which obviously fails - it needs to be `ssh my-server nvim --embed -p`.

there's two ways of fixing this:

- make `--embed` be **appended** rather than **prepended**
- add a `--no-embed-arg` to disable the automatic addition of the `--embed` argument

this PR implements the second option. (I can open a PR with the first option if you'd prefer that instead!). So, this works, with this PR:

```sh
target/debug/neovide --no-embed-arg --no-tabs --neovim-bin ssh -- my-server nvim --embed
```

### DIFF
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -35,9 +35,11 @@ pub fn create_nvim_command(settings: &Settings) -> TokioCommand {
     create_tokio_nvim_command(&cmdline_settings, true)
 }
 
-pub fn create_restart_nvim_command(details: &RestartDetails) -> TokioCommand {
+pub fn create_restart_nvim_command(settings: &Settings, details: &RestartDetails) -> TokioCommand {
     let mut cmd = TokioCommand::new(&details.progpath);
-    cmd.arg("--embed");
+    if !settings.get::<CmdLineSettings>().no_embed_arg {
+        cmd.arg("--embed");
+    }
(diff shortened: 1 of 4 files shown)


falcucci:
the problem with `--no-embed-arg` is that we add an exception that breakes :restart on local nvim spawns once you remove it, which would require you to also treat it there but it seems a bit odd to me in comparison with the first option.

- make --embed be appended rather than prepended

this approach seems more natural to me since we already use --embed as default and would keep doing it. appending it by default in this case without breaking :restart for example and avoiding an extra cmd option + solving your use case :)

```
target/debug/neovide --no-tabs --neovim-bin ssh -- my-server nvim --embed
```

khyperia:
> this approach seems more natural to me

made https://github.com/neovide/neovide/pull/3393 for that approach!

falcucci:
closing in favor of https://github.com/neovide/neovide/pull/3393

PR #3393, opened 2026-03-10
move --embed to be last argument

khyperia:
See https://github.com/neovide/neovide/pull/3388 for context, this is a different way of solving the same issue (if this PR gets merged, that one can get closed)

This makes `neovide --no-tabs --neovim-bin ssh -- my-server nvim` function correctly!

This _might_ be a breaking change for some people's setups, which is why I originally went with the other option, but it might be enough of a [spacebar heating](https://xkcd.com/1172/) moment that it should be fine, unless someone comes back reporting an issue the other way around :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2186,18 +2186,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "objc2-ui-kit"
-version = "0.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


falcucci:
I think it's better append it without deduplicating :)

i'll send a commit with some helpers and you let me know if everything works for you as the same

khyperia:
> let me know if everything works for you as the same

yeah, just checked, my use case works great!

falcucci:
@khyperia thank you!

PR #3407, opened 2026-03-16
hotreload: support window geometry as one setting

falcucci:
size and grid are not independent runtime options. together with `maximized` they define one mutually exclusive geometry choice, so we don't model them as separate hot reload events.

if reload emitted a size/grid independently, then a transition like size -> grid would depend on event order and could briefly drive an invalid state. in the end, we make geometry hot reload work, but we do it in a way that preserves the actual model.

One geometry setting, one parser, one reload event, one place to apply it.

todo: we need to improve how we declare mutually exclusive config options in the config file.

### DIFF
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -215,6 +215,28 @@ impl MouseCursorIcon {
     }
 }
 
+impl GeometryArgs {
+    pub fn from_config(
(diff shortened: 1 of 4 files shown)


falcucci:
tested also on windows.

PR #3422, opened 2026-03-24
fix: split route cwd from caller cwd on handoffs

falcucci:
the mistake here was treating `--chdir` and a relative file arg as if they described the same directory from its caller.

they don't.

For a handoff like

```
cd ~/neovide
neovide --reuse-instance --new-window src/main.rs --chdir ~
```

the file comes from the caller shell cwd, so it should still mean `~/neovide/src/main.rs` but the new neovim route itself should start with cwd `~`

now we take both meanings explicitly through the handoff path:
- cwd: the per-route startup cwd
- caller_cwd: the base directory for resolving relative file drops

### DIFF
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -1,5 +1,7 @@
 #[cfg(target_os = "windows")]
 use std::os::windows::process::CommandExt;
+use std::path::Path;
+use std::path::PathBuf;
 use std::process::Command as StdCommand;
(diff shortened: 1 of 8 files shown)


varp:
@falcucci Thank you for the great job! Could I ask you to run the Nightly action manually with the commit, please?

varp:
Cron came and did the job

Issue #3427, opened 2026-03-26
Command line windows retain a line of text after being cleared by 'Enter'


pcw3:
**Describe the bug**
I enter a command which produces a large output such as `:ls` or `:version`. After I press the 'Enter' key the first line of the previous command' output is still visible. I don't see this issue when using Neovim in a terminal.

**To Reproduce**
Steps to reproduce the behaviour:
1. Enter `:version'`
2. Press `'Enter'` key
3. See unwanted text

**Expected behaviour**
The command line is clear of any text

**Desktop (please complete the following information):**
 - OS: Kubuntu 24.04
 - Neovide Version: v0.12.0-dev-2707+gc9e961994b
 - Neovim Version: nightly-119+g20cf79e

**Please run `neovide --log` and paste the contents of the `.log` file created in the current directory here:**
No recent log found

falcucci:
@pcw3 I could reproduce and that's fixed at #3430 

let me know if you have any issue still on this,

thank you

pcw3:
Confirmed fixed in Neovide 0.16.0, thank you.

PR #3430, opened 2026-03-28
fix: render the correct message window geometry

falcucci:
note: neovim [reports](https://github.com/neovim/neovim/blob/f2d0b06ecbb50541db748cd7974a24816b609dd3/src/nvim/message.c?plain=1#L205-L254) message grids with the full default-grid height even when only the bottom few rows are actually on screen.

we were already compensating for that in one place when clamping the target position, but then we turned around and used the full backing-grid size for pixel_region()

that means the renderer still treated the message window as if all rows were visible, which is wrong.

fixes https://github.com/neovide/neovide/issues/3427


falcucci:
@pcw3 would be great if you could test this and double check if it works as expected for you as well.

PR #3439, opened 2026-03-31
fix: handle nvim-initiated too-large resizes

CGamesPlay:
If you do `set lines=40` followed by `set lines=999`, Neovide detects that it couldn't make the window 999 lines tall and updates itself to the correct value. But if you issue another `set lines=999`, the window size doesn't change, so Neovide doesn't do the check. This change just causes Neovide to check the real window size after it attempts to resize, in case the attempt to resize the window failed for some reason (like it was already the maximum allowable size).

I don't think this is a breaking change. Should be a safe improvement.


falcucci (approved):
good catch, thank you!

PR #3440, opened 2026-03-31
handoff: Add neovim-bin and neovim arguments

CGamesPlay:
Love the multi-window support! I was thrilled to see that land and ported all of my remote ssh commands to using it. Unfortunately, the version in 0.16.0 doesn't allow the windows to use different values for `--neovim-bin` and friends.

This change just adds --neovim-bin, neovim arguments, and files to open to the handoff struct. As a result, I can `neovide --neovim-bin=ssh -- myhost cd directory "&&" exec nvim` and it works with `--reuse-instance --new-window`.

I also put `files_to_open` into the handoff. This is perhaps a bit more contentious, but fixes a minor bug: "new window" would re-open the same files as the initial invocation, which I don't think is generally the desired behavior.

This change will break users who use `neovide --neovim-bin && neovide --reuse-instance --new-window`, because they now need to include the `--neovim-bin` in the subsequent invocations.

I've tested that this works on macOS with `--reuse-instance --new-window`. I think it should not cause problems in other configurations, but would appreciate pointers if I am missing something.

### DIFF
--- a/src/bridge/command.rs
+++ b/src/bridge/command.rs
@@ -23,6 +23,8 @@ const FORKED_FROM_TTY_ENV_VAR: &str = "NEOVIDE_FORKED_FROM_TTY";
 pub struct OpenArgs {
     pub files_to_open: Vec<String>,
     pub tabs: bool,
+    pub neovim_bin: Option<String>,
+    pub neovim_args: Option<Vec<String>>,
 }
 
 /// Mode is how neovide should populate argv when launching an embedded nvim instance.
@@ -102,12 +104,26 @@ fn build_nvim_command_parts(
     embed: bool,
     mode: OpenMode,
 ) -> (String, Vec<String>) {
-    let bin = cmdline_settings.neovim_bin.clone().unwrap_or_else(|| "nvim".to_owned());
-    let mut args = cmdline_settings.neovim_args.clone();
+    // For OpenMode::Args (handoff/new-window routes), use the route's own
+    // bin/args. None means "not specified" = default nvim, NOT the global
+    // CmdLineSettings value (which belongs to the original invocation).
+    // For Startup/None modes, use global CmdLineSettings as before.
+    let (bin, mut args) = match &mode {
+        OpenMode::Args(open_args) => {
+            let bin = open_args.neovim_bin.clone().unwrap_or_else(|| "nvim".to_owned());
+            let args = open_args.neovim_args.clone().unwrap_or_default();
+            (bin, args)
+        }
+        _ => {
+            let bin = cmdline_settings.neovim_bin.clone().unwrap_or_else(|| "nvim".to_owned());
+            let args = cmdline_settings.neovim_args.clone();
+            (bin, args)
+        }
+    };
+
     if embed {
         append_embed_arg(&mut args);
     }
-
     args.extend(build_open_args(cmdline_settings, mode));
 
     (bin, args)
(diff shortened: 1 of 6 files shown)


falcucci:
hey thank you!

I'll review a bit later when I have more time. one thing to say is that the `files_to_open` should be fixed now by https://github.com/neovide/neovide/pull/3441 since we had to impl the new `:restart` protocol (it changed right before the neovim 0.12 release) so it makes more sense to me to fix it there.

the remaining `neovim_bin` and `args` makes sense to add in the handoff payload to solve your problem, so you will just need to adapt a bit your changes after solving the conflicts.

CGamesPlay:
Rebased! I think there's probably a useful refactor to split a CmdLineSettings into a (new struct, OpenArgs), which could potentially remove some surface area for errors, but it's not necessary to fix the issue I'm having.

falcucci:
@CGamesPlay thanks!

> This change will break users who use neovide --neovim-bin && neovide --reuse-instance --new-window, because they now need to include the --neovim-bin in the subsequent invocations.

still speaking about the scope here before the review, note that this should not be a breaking change neither affect the user current experience to interact through the cli to create new windows, requiring `--neovim-bin` for every subsequent invocations will sound off for a normal workflow.

the intention of `--reuse-instance --new-window` is to open a new window as the existing one by default but for different projects, buffers, chdirs. users may also have neovim-bin on their config as a safeguard too, so we should preserve it.

what you could do here is to change the dynamics a bit, behave more as an override **when** needed, more like explicit requests wins for those cases like yours, would be preferable. that way we go to the most common (respecting the globals user oriented config) to the less common instead.

CGamesPlay:
I’m not opposed to having a sensible default that is user configurable… but it seems strange that the first instance is the one that decides what binary/arguments will be used by default for the lifetime of a neovide process. The first instance just isn’t really special in my mind, it just happens to be the one the user started with. Does it perhaps make more sense to put these default values into the neovide config file?

As a quick counterpoint to having it always inherit the first invocation’s args: what if I want to launch a new instance that has no neovim args? Then I need to use some new flag that clears existing arguments.

falcucci:
not strange at all to me. as I said, imo most common use cases comes first but also give all the flexibility for the exceptions or less common, including new flags if needed. users (including myself) launches neovide based on their user oriented config from the Dock or the cli and from there starts working as desired.

> Does it perhaps make more sense to put these default values into the neovide config file?

anything we can add into the config file is good to have, some options are currently missing but all of the reasonable ones would be ideal to be there, where people can default their common values. If they need a new custom or with different args just request an override somehow.

those aren't written in stones, we could also change and include a breaking change in the future if really needed, but now I think is unnecessary, solving your problem without breaking the default workflow is more important atm.

CGamesPlay:
So, my proposal would be:

- add a `neovim_command: Vec<String>` field to the toml, it specifies the default value of neovim-bin and arguments
- when `--neovim-bin` is given on the command line, override the config value just for that invocation
- when additional (non-filename) arguments are given, append them to the config value for just that invocation

This way, users who currently pass a custom `--neovim-bin` and rely on `--reuse-instance --new-window` reusing the same server will be able to preserve their workflow by moving the `--neovim-bin` into their config file, and all users will be able to use custom neovim-bins on a per-invocation basis.

Does this sound acceptable? If so I'll make that change. If not, I would really appreciate understanding what the use case is where the proposal breaks down, because I don't understand it (Activating Neovide from the Dock never passes `--neovim-bin` and so isn't broken by this change).

falcucci:
> This change will break users who use neovide --neovim-bin && neovide --reuse-instance --new-window, because they now need to include the --neovim-bin in the subsequent invocations.

my comment was based on this info from you, since i was going to test and review later, but if there is no breaking change, we do not need to change the approach, that was the only concern.

let me check later when I am at home to test it properly and see if it breaks anything, the new config values could live in another pull request tbd.

falcucci:
@CGamesPlay sorry the delay on this one, have been busy. the changes looks really good!

I have tested and used a bit myself. I just noticed that the route cwd wasn't kept when duplicating a window in pair with its neovim bin/args. that would make more sense now since we are keeping some ctx. Also made some very minor [changes](https://github.com/neovide/neovide/pull/3440/changes/8244daf0018998709bda1754e915d2a13d6781fd) for readability and maintenance.

thank you!

looshch:
thanks a ton @CGamesPlay, i believe this fixes [this](https://github.com/neovide/neovide/issues/3466) issue i opened recently

PR #3445, opened 2026-04-01
fix: reintroduce error window for launch failures

CGamesPlay:
Multi-window support changed from using the error window to using `.expect`. This means that if nvim fails to start in a new window, the entire Neovide process aborts, ending all of your nvim sessions. This change catches these errors and sends them to the main loop as `NeovimLaunchError`, and extends the main loop to be able to handle arbitrarily many of these error windows without blocking.

Tested by:
1. `open -a Neovide --args --wsl`
2. `open -na Neovide --args --reuse-instance --new-window --neovim-bin=false`

Before this would abort Neovide, now it shows the error window again.

### DIFF
--- a/src/window/application.rs
+++ b/src/window/application.rs
@@ -22,7 +22,7 @@ use winit::{
 
 use super::{
     CmdLineSettings, EventPayload, EventTarget, RouteId, WindowSettings, WindowSize,
-    WinitWindowWrapper, save_window_size,
+    WinitWindowWrapper, error_window, save_window_size,
 };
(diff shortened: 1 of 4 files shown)


falcucci:
thank you! the changes are good,

the only issue I found for now is that the error window doesn't have a control flow for the eventloop scheduler which at this state wakes up sporadically and as consequence consumes high CPU.

I made a commit over here with also a couple of minor changes.

falcucci (approved):

PR #3450, opened 2026-04-02
feature: use local open for remote neovim

CGamesPlay:
This adds a hook into `vim.ui.open` when Neovide is running remotely to enable opening URLs using the normal way (`gx`).

I opted to make the call a bit selective: if the caller specifies `neovide_no_remote_open` or `cmd`, then it falls through to the default (nvim-local) implementation. The logic is: giving a `cmd` means the caller has opinions about how the call should be processed, and that opinion was formed on the remote system. Also, not doing so is basically remote code execution through Neovide.

Shouldn't break anyone's workflow, and can be disabled via a documented init.lua variable if necessary.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1314,6 +1314,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "is-docker"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "928bae27f42bc99b60d9ac7334e3a21d10ad8f1835a4e12ec3ec0464765ed1b3"
+dependencies = [
+ "once_cell",
+]
+
+[[package]]
+name = "is-wsl"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "173609498df190136aa7dea1a91db051746d339e18476eed5ca40521f02d7aa5"
+dependencies = [
+ "is-docker",
+ "once_cell",
+]
+
+[[package]]
 name = "is_terminal_polyfill"
 version = "1.70.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1629,6 +1648,7 @@ dependencies = [
  "objc2-foundation 0.3.2",
  "objc2-metal 0.3.2",
  "objc2-quartz-core 0.3.2",
+ "open",
  "parking_lot",
  "rand 0.9.2",
  "raw-window-handle",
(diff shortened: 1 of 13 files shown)


CGamesPlay:
Not sure what this test failure is about but it looks highly unrelated to my change (403 in external service).

falcucci:
Is this not achievable by a personal custom remote opener?

I am looking at it and I'm not fully comfortable enabling this through neovide.

even with opt.cmd filtered out, this still lets a remote neovim instance trigger local OS handlers
which is quite dangeours. I think it's wrong assume the users are aware of it (even if they should)

also im not sure even if we should have this opt-in at all, at least understand the nuances of the [open](https://crates.io/crates/open) crate on
windows, linux and macOS when we *supposely* execute those arguments locally from a remote machine.
but that would require also a clever validation on our side since it is a trust boundary change, so I would personally prefer the user to take full control over this instead of delegating it to us.

CGamesPlay:
Hey @falcucci thanks for taking the time to review! Would you be open to accepting if it was allowlisted to only support `http://` and `https://`?

> Is this not achievable by a personal custom remote opener?

Within the scope of nvim, I don't think so. Unless Neovide has a feature that enables this that I missed? This might be an interesting idea (like a remote<->local lua plugin system), but right now the only use I have now is opening URLs from source comments in my browser.

>  this still lets a remote neovim instance trigger local OS handlers
which is quite dangeours. I think it's wrong assume the users are aware of it (even if they should)

This is valid. So, browsers handle this by a prompt for any external URL (like `spotify://`), but they always open `http://` and `https://` URLs without any prompting. Would you be open to allowlisting this to only support `http://` and `https://`? These two URL schemes are always "safe" (if annoying) to open automatically, and a browser-style prompt could be implemented later if it is ever needed.

> also im not sure even if we should have this opt-in at all, at least understand the nuances of the [open](https://crates.io/crates/open) crate on
windows, linux and macOS when we supposely execute those arguments locally from a remote machine.

Hmm, checking the source, the `open` crate uses the `std::process::Command` abstraction, so any encoding problem would be a vulnerability in Rust stdlib. Pragmatically, I think among the 31 million total downloads for the open crate, a majority of them would be shocked to find out that malformed URLs result in arbitrary code execution. [append] But, if this is a real concern, we could also validate that it's a well-formed HTTP URL, using e.g. [url](https://crates.io/crates/url).

80avin:
If I may, I want to highlight the way VSCode does it.
In VSCode, even if we click on links - even if popular ones, it gives a prompt to open it.

I feel this is a much more sensible implementation than avoiding the feature or adding filters over URLs. As such, the RPC handler can prompt the user and then proceed to open.

Here's VSCode's implementation
image

CGamesPlay:
What I've done now is added a new setting to the Neovide toml config, `remote.allowed-url-patterns`. This will define a list of URL patterns that are permitted. When a URL is not permitted, the user gets an error: `neovide.open failed: URL rejected by allowlist: URL`. This setting is hot-reloadable.

The default configuration is empty, meaning no URLs are allowed. My recommended setting is `["http://*", "https://*"]`, which will allow web URLs and block everything else. This setting only has an effect when Neovide is in "remote mode", which is when `--wsl` or `--server` is passed today. It still adds the `vim.g.neovide_no_remote_open` flag to bypass the Neovide open handler entirely (which means that `gx` won't work unless you've got a custom setup for it).

@80avin I took some time to look into this idea. The key problem is that no native dialog functionality exists today. We could do it in lua, like Neovim's confirm quit message, but that defeats the purpose because it runs in the remote nvim instance. Anything else has to be coded from scratch, because our GUI dependencies don't offer it, either.

PR #3477, opened 2026-04-14
fix: keypad * mapping (use NumpadMultiply instead of NumpadStar)

### LINKED ISSUE #3383: kMultiply (* on numerical keyboard) not handled correctly
All numeric keyboard keys correctly send <kX> codes (like kPlus, kDivide, etc.), except "\*", which sends "\*" instead of `<kMultiply>` (see `:help keypad-multiply`). This makes it impossible to map it separately from "\*" on main keyboard (shift-8 on us keyboard).

I found `KeyCode::NumpadStar => Some("kMultiply"),` line in src/windows/keyboard_manager.rs, but it doesn't seem to work, please check.

`imap <kMultiply> TEST` - doesn't have any effect
`imap * TEST` - writes "TEST" for both shift-8 and "\*" on numerical keyboard

Tested on both current release (0.15.2) and nightly build (2026.03.04) Linux x86 binary (Ubuntu 24.04).

After searching maybe just changing NumpadStar to NumpadMultiply may work.

pvxx:
Simple one-line fix, no breaking change.

Neovide currently maps KeyCode::NumpadStar to "kMultiply".

According to winit docs:

- NumpadMultiply = keypad "*"
- NumpadStar = telephony "*" key

This causes keypad "\*" to be treated as regular "\*" in Neovim.

Fix: use NumpadMultiply instead.

Resolves #3383.

### DIFF
--- a/src/window/keyboard_manager.rs
+++ b/src/window/keyboard_manager.rs
@@ -127,7 +127,7 @@ impl KeyboardManager {
         let is_numlock_key = key_event.text.is_some();
         match physical_key_code {
             KeyCode::NumpadDivide => Some("kDivide"),
-            KeyCode::NumpadStar => Some("kMultiply"),
+            KeyCode::NumpadMultiply => Some("kMultiply"),
             KeyCode::NumpadSubtract => Some("kMinus"),
             KeyCode::NumpadAdd => Some("kPlus"),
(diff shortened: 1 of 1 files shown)


falcucci:
@pvxx thank you!

PR #3481, opened 2026-04-15
feat: settings for menu shortcuts on macOS

vapourismo:
Love Neovide. Cheers for making the Neovim experience so nice! :)

## Context

I use `cmd+{h,j,k,l,n}` in my workflows. Since Neovide 0.16, some menu item shortcuts like "Window > New Window" prevent these key combinations from being forwarded to Neovim.

As I am on a Mac, I would normally remap the key shorts using macOS' "App Shortcuts" feature, which lets you change the shortcuts. Unfortunately, this doesn't seem to work for Neovide on my machine - I'm not sure why. 

So I find myself in a pickle. I want to keep using Neovide, but I also don't want to maintain a monkey patched version going forward. I decided to be pragmatic and contribute this PR, which adds settings for all the key shortcuts used for the macOS menu bar.

- Feature

- No

## How does it work?

When adding the menus on macOS, the settings type is now passed in as well. Each menu adder method can extract the key combination that has been configured. 

I have reused the `KeyCombo` type for key parsing. Please tell me if you want me to move this into a more generic place, given I left it in the tab navigation module for now.

## Problems with named keys

I am not very familiar with how `setKeyEquivalent` works for named keys. I did a bit of research, and it seems this requires ad-hoc mappings. I left this out for now to keep the PR more focused. I would tackle this in a follow-up if there is interest.

### DIFF
--- a/src/cmd_line.rs
+++ b/src/cmd_line.rs
@@ -118,6 +118,69 @@ pub struct CmdLineSettings {
     #[arg(long = "no-system-native-tabs", action = ArgAction::SetTrue, value_parser = FalseyValueParser::new())]
     _no_system_native_tabs: bool,
 
+    /// Set the Window > New Window shortcut
+    #[cfg(target_os = "macos")]
+    #[arg(
+        long = "system-new-window-hotkey",
+        env = "NEOVIDE_SYSTEM_NEW_WINDOW_HOTKEY",
+        default_value = "cmd+n"
+    )]
+    pub system_new_window_hotkey: String,
+
+    /// Set the Neovide > Hide shortcut
+    #[cfg(target_os = "macos")]
+    #[arg(
+        long = "system-hide-hotkey",
+        env = "NEOVIDE_SYSTEM_HIDE_HOTKEY",
+        default_value = "cmd+h"
+    )]
+    pub system_hide_hotkey: String,
+
+    /// Set the Neovide > Hide Others shortcut
+    #[cfg(target_os = "macos")]
+    #[arg(
+        long = "system-hide-others-hotkey",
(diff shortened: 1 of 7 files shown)


falcucci:
@vapourismo thank you, the only problem found was the config parser not accepting false as a value for those new entries, so I have added a deserializer to accept that. but after all it looks good to me, still validating some stuff.

```
Error while parsing config file /Users/falcucci/.config/neovide/config.toml:
TOML parse error at line 22, column 28
   |
22 | system-new-window-hotkey = false
   |                            ^^^^^
invalid type: boolean `false`, expected a string

Continuing with default config.
Error while parsing config file /Users/falcucci/.config/neovide/config.toml:
TOML parse error at line 22, column 28
   |
22 | system-new-window-hotkey = false
   |                            ^^^^^
invalid type: boolean `false`, expected a string

Continuing with default config.
ERROR [neovide::window::window_wrapper] Failed to launch neovim runtime: Could not locate or start neovim process

Caused by:
    No such file or directory (os error 2)
```

vapourismo:
> @vapourismo thank you, the only problem found was the config parser not accepting false as a value for those new entries, so I have added a deserializer to accept that. but after all it looks good to me, still validating some stuff.

Cheers! I noticed this as well, but kind of ignored it because the `is_disabled_keyword` is unchanged.

Perhaps we can amend the config docs to leave out the reference to `false` for now, then address the parsing issue separately?

falcucci:
@vapourismo that's fine, I guess you dont need any changes from here in these changes. I just need to fix the 2024 rust edition warnings and it would be ready to go.

falcucci (approved):
thank you!

PR #3491, opened 2026-04-21
Resolve relative paths earlier

CGamesPlay:
This is a minor cleanup PR. It removes `caller_cwd` from the handoff message. It is only used to resolve filenames before giving them to nvim, but there's no reason to do that; we can just resolve the paths in the calling process.

- Codestyle

- No

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -81,7 +81,7 @@ use crate::{
 };
 
 #[cfg(target_os = "macos")]
-use crate::utils::resolved_cwd;
+use crate::utils::{resolve_relative_path, resolved_cwd};
 
 pub use profiling::startup_profiler;
 
@@ -316,11 +316,16 @@ fn maybe_handoff(settings: &Settings) -> HandoffOutcome {
     let CmdLineSettings { files_to_open, tabs, new_window, neovim_bin, neovim_args, .. } =
         cmdline_settings;
 
+    let cwd = std::env::current_dir().ok();
+    let files_to_open = files_to_open
+        .into_iter()
+        .map(|path| resolve_relative_path(&path, cwd.as_deref()))
+        .collect();
+
     let request = ipc::handoff::HandoffRequest {
         version: BUILD_VERSION.to_owned(),
         files_to_open,
         cwd: resolved_cwd(cmd_line::argv_chdir().as_deref()),
-        caller_cwd: resolved_cwd(None),
         tabs,
         new_window,
         neovim_bin,
--- a/src/ipc/handoff.rs
+++ b/src/ipc/handoff.rs
@@ -27,8 +27,8 @@ const LISTENER_POLL_INTERVAL: Duration = Duration::from_millis(100);
 pub struct HandoffRequest {
     pub version: String,
     pub files_to_open: Vec<String>,
+    /// Directory to start nvim in, accounting for --chdir
     pub cwd: Option<String>,
-    pub caller_cwd: Option<String>,
     pub tabs: bool,
     pub new_window: bool,
     #[serde(default)]
@@ -44,7 +44,6 @@ impl HandoffRequest {
             version: BUILD_VERSION.to_owned(),
             files_to_open: Vec::new(),
             cwd: None,
-            caller_cwd: None,
             tabs: true,
             new_window: false,
             neovim_bin: None,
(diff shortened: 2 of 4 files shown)


src/main.rs
```diff
@@ -316,11 +316,17 @@ fn maybe_handoff(settings: &Settings) -> HandoffOutcome {
     let CmdLineSettings { files_to_open, tabs, new_window, neovim_bin, neovim_args, .. } =
         cmdline_settings;
 
+    let cwd = match std::env::current_dir() {
+        Ok(cwd) => cwd,
+        Err(err) => return HandoffOutcome::Error(format!("cwd is invalid: {err}")),
+    };
+    let files_to_open =
+        files_to_open.into_iter().map(|path| resolve_relative_path(&path, Some(&cwd))).collect();
+
```
falcucci, line 324: please, dont let an unavailable caller cwd become a startup failure.

```suggestion
let cwd = std::env::current_dir().ok();
let files_to_open = files_to_open
    .into_iter()
    .map(|path| resolve_relative_path(&path, cwd.as_deref()))
    .collect();

```

falcucci (changes requested):

src/main.rs
CGamesPlay: Done. Now an invalid caller PWD preserves the existing behavior of opening the wrong files in nvim.

CGamesPlay:
I'm not sure the UX of opening the wrong files is better, but it is what you asked for, so I've restored the behavior.

CGamesPlay:
Need to mark this as a draft. I just realized that it doesn't properly handle the `--chdir` flag. In my mind we should basically treat `--chdir DIR` as an immediate `cd DIR` before anything else in the startup path. Notably, relative paths passed as arguments should be resolved relative to the `--chdir` value and *not* the original CWD.

The original code doesn't do this either, I think, but I'll fix this bug before marking this PR ready.

falcucci:
> I'm not sure the UX of opening the wrong files is better, but it is what you asked for, so I've restored the behavior.

the code changes was basically treating an unavailable cwd as a fatal startup error, that's wrong. nvim already gives a warning on such cases and we shouldn't deviate from that.

CGamesPlay:
> the code changes was basically treating an unavailable cwd as a fatal startup error, that's wrong. nvim already gives a warning on such cases and we shouldn't deviate from that.

Maybe the right thing to do is somewhere in the middle then. This path is for handoff: talking to a different nvim process in a potentially different CWD. If the handoff is creating a new window, then what you're saying is right (modulo the `--chdir` flag which I will add support for). But if it isn't, then it will treat relative paths as relative to the directory that nvim happens to be in (opening the wrong files).

falcucci:
hmm no, I dont agree about changing semantics between modes and I would avoid big semantic changes here. the file path should mean what the caller meant, if it exists or not. and *never* being fatal about the path being inexistent.

to clean things up, that's how I got the fatal error previously (already addressed and fixed in review)

```
cd whatever
rm whatever
neovide --reuse-instance --new-window
```

CGamesPlay:
Yes, 100% agreed, "the file path should mean what the caller meant". The problem is *in master right now* that a relative file path is interpreted as local to the existing nvim instance's CWD. aka, the wrong path. This PR doesn't change that behavior, it preserves the "open the wrong file" behavior that already exists.

Another problem, again in master right now: `neovide --reuse-instance --chdir ~/Downloads README.md` should open `~/Downloads/README.md`, but it actually opens `README.md` in the existing nvim's CWD. This second one is what I would like to fix before unmarking this as a draft.

falcucci:
> Another problem, again in master right now: neovide --reuse-instance --chdir ~/Downloads README.md should open ~/Downloads/README.md, but it actually opens README.md in the existing nvim's CWD. This second one is what I would like to fix before unmarking this as a draft.

no, it behaves exactly as it should on main that's why I made an explicity caller cwd, do not change this specific semantics. README.md should still mean what the caller meant and --chdir should affect the route cwd. they are different.

```
cd /whatever
neovide --reuse-instance --new-window --chdir ~/Downloads README.md
```

this should open `/whatever/README.md` and `--chdir` should only affect the route cwd in this case.

the same for

```
cd ~/neovide
neovide --reuse-instance --new-window src/main.rs --chdir ~/
```

should open `~/neovide/src/main.rs` and affect the route cwd `~/` which is already the correct behavior on main.

CGamesPlay:
I'm going to abandon this PR. It was supposed to be a minor cleanup I encountered as I was looking through other things, but it clearly isn't. Not worth your time to review.

I would appreciate if you did offer some feedback on #3450, as that's something I would really like to see committed. And if there are any other issues or PRs that you think would benefit from some attention, maybe I can take a look at those as well.

PR #3504, opened 2026-05-10
fix: surface startup errors before first grid - 0.12+ only

### LINKED ISSUE #3499: Neovide fails to launch
**Describe the bug**

Neovide hangs before showing its first window if Neovim emits an error-style message during startup.

The minimal trigger appears to be `nvim_echo(..., { err = true })` before the first drawable frame/window exists. This also affects `vim.notify(..., vim.log.levels.ERROR)` on Neovim 0.12.1, because the default implementation uses `nvim_echo(..., { err = true })`.

When this happens:

- No Neovide window appears.
- `neovide_rCURRENT.log` stops at `INFO [neovide::bridge] Neovim process attached`.
- There are no `renderer::rendered_window` log entries.
- Neovide uses high CPU indefinitely.
- The child `nvim --embed` process is idle/waiting.

**To Reproduce**

1. Create a one-line init file that emits an error-style startup message:

```sh
printf '%s\n' 'vim.api.nvim_echo({ { "startup error before first draw" } }, false, { err = true })' > /tmp/neovide-repro.lua
```

2. Launch Neovide with that init file:

```sh
neovide --log --no-fork -- -u /tmp/neovide-repro.lua
```

3. Observe that no window appears and Neovide stays running with high CPU.

Equivalent repro:

```sh
printf '%s\n' 'vim.notify("startup error before first draw", vim.log.levels.ERROR)' > /tmp/neovide-repro.lua
neovide --log --no-fork -- -u /tmp/neovide-repro.lua
```

Control case that renders successfully:

```sh
printf '%s\n' 'vim.notify("startup warning before first draw", vim.log.levels.WARN)' > /tmp/neovide-control.lua
neovide --log --no-fork -- -u /tmp/neovide-control.lua
```

**Expected behavior**

Neovide should create/show a window and display or otherwise surface the startup error message.

At minimum, Neovide should not spin CPU indefinitely while waiting for initial window creation.

**Screenshots**

No screenshot is available because no Neovide window appears.

- OS: macOS 15.7.4 (24G517), arm64
- Neovide Version: 0.16.1
- Neovim Version: NVIM v0.12.1

I also reproduced related plugin-triggered cases on locally built Neovide `0.16.2` and current `main`.

The full log is short in the failing case and stops at attach. Relevant tail:

```text
TRACE [neovide::channel_utils] neovim_handler OptionSet { gui_option: AmbiWidth("single") }
TRACE [neovide::channel_utils] neovim_handler OptionSet { gui_option: ArabicShape(true) }
TRACE [neovide::channel_utils] neovim_handler OptionSet { gui_option: Emoji(true) }
TRACE [neovide::editor] Option set AmbiWidth("single")
TRACE [neovide::editor] Option set ArabicShape(true)
TRACE [neovide::editor] Option set Emoji(true)
INFO [neovide::bridge] Neovim process attached
```

There are no `renderer::rendered_window` entries after this in the failing run.

**Additional context**

Neovim 0.12.1 implements default `vim.notify(..., vim.log.levels.ERROR)` via `nvim_echo(..., { err = true })`:

```lua
vim.api.nvim_echo(chunks, true, { err = level == vim.log.levels.ERROR })
```

So this can happen from plugins/config that emit startup errors before Neovide creates its first window.

This was originally hit via plugins/config that emitted startup errors. One concrete real-world trigger was `hinell/lsp-timeout.nvim`, which emitted:

```text
LspStart or LspStop commands are NOT found.Make sure lsp-config.nvim is installed
```

Suppressing the startup `vim.notify(..., ERROR)` / `nvim_echo(..., { err = true })` made Neovide render normally.

Debugging notes:

- Sampling showed CPU is in Neovide, not Neovim.
- Neovide process: ~75-95% CPU.
- Child `nvim --embed` process: ~0% CPU.

Sampling the child `nvim` process showed it waiting in:

```text
wait_return -> vgetc -> input_get -> inbuf_poll -> uv__io_poll -> kevent
```

With a debug build of Neovide, I instrumented `Application::schedule_next_event`. During the hang, it repeatedly reported:

```text
routes=0 route_cores=1 ui_state=Initing render_states=0 pending_window=false create_allowed=true errors=0 deadline=None
```

It reached over 100k `schedule_next_event` calls in roughly 2 seconds.

This suggests two related issues:

1. Neovide can wait forever for the first drawable frame/window-create condition if Neovim emits an early error message and enters `wait_return`.
2. While in that no-window/no-render-state state, Neovide busy-loops because `deadline=None` maps to `ControlFlow::WaitUntil(now)`.

Relevant code path appears to be:

```rust
fn next_control_flow_for(
    deadline: Option<Instant>,
    has_error_windows: bool,
    now: Instant,
) -> ControlFlow {
    match deadline {
        Some(deadline) => ControlFlow::WaitUntil(deadline),
        None if has_error_windows => ControlFlow::Wait,
        None => ControlFlow::WaitUntil(now),
    }
}
```

Changing the final branch to `ControlFlow::Wait` stops the CPU spin, but the first window still does not appear. So that likely fixes only the spin, not the startup deadlock.

falcucci:
fixes https://github.com/neovide/neovide/issues/3499

neovim can hit wait-return during startup before the first grid update. at that point we have no message area to show the error and no useful window to accept the Enter that would let startup continue.

now temporarily we request external messages during startup, capture msg_show events until the first rendered batch, restore the built-in message and cmdline UI, then replay the captured messages through nvim_echo without duplicating message history.

_we keep this path gated to nvim **0.12+**, since 0.10/0.11 externalize cmdline prompts in ways we do not render during startup_, so we preserve cmdheight across the temporary external-message mode where the startup capture does not silently rewrite the user's message area.

### DIFF
--- a/src/bridge/events.rs
+++ b/src/bridge/events.rs
@@ -71,7 +71,7 @@ pub struct GridLineCell {
 
 pub type StyledContent = Vec<(u64, String)>;
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum MessageKind {
     Unknown,
     Confirm,
@@ -106,6 +106,33 @@ impl MessageKind {
             _ => MessageKind::Unknown,
         }
     }
+
+    pub fn is_error(self) -> bool {
+        matches!(
+            self,
+            MessageKind::Error
+                | MessageKind::EchoError
+                | MessageKind::LuaError
+                | MessageKind::RpcError
+        )
+    }
+
+    pub fn highlight_group(self) -> Option<&'static str> {
+        match self {
+            MessageKind::Warning => Some("WarningMsg"),
+            MessageKind::Error
+            | MessageKind::EchoError
+            | MessageKind::LuaError
+            | MessageKind::RpcError => Some("ErrorMsg"),
+            _ => None,
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct StartupMessage {
+    pub kind: MessageKind,
+    pub content: String,
 }
 
 #[allow(unused)]
@@ -417,6 +444,7 @@ pub enum RedrawEvent {
         kind: MessageKind,
         content: StyledContent,
         replace_last: bool,
+        append: bool,
     },
     MessageClear,
     #[allow(unused)]
@@ -436,6 +464,8 @@ pub enum RedrawEvent {
         entries: Vec<(MessageKind, StyledContent)>,
     },
     Suspend,
+    NeovimSessionStarted,
+    StartupMessageUiRestored,
     NeovideSetRedraw(bool),
     NeovideIntroBannerAllowed(bool),
 }
@@ -476,7 +506,7 @@ fn extract_values_with_optional<const REQ: usize, const OPT: usize>(
         for (index, value) in values.into_iter().enumerate() {
             if index < REQ {
                 required_values[index] = value;
-            } else {
+            } else if index - REQ < OPT {
                 optional_values[index - REQ] = Some(value);
             }
         }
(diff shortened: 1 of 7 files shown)


griwes:
For the record, this breaks noice.nvim's initialization; even when loaded at VeryLazy with lazy.nvim, it still sees an UI client that has ext_messages and ext_cmdline enabled, and refuses to initialize. A polling loop that delays its setup until there's no ext_cmdline / ext_messages UI clients works, but it feels like an awful hack (though definitely preferable to monkey patching noice's init). Any chance we could make the behavior of this PR configurable?

alex35mil:
+1 to the noice.nvim issue. I had to do some unpretty thing to delay noice loading.

falcucci:
thanks for the report.

I could replicate this for noice and with #3517 you would be able to disable within your config file or by using 

	neovide --no-startup-message-capture
		
if you prefer.

```toml
startup-message-capture = false
```

PR #3513, opened 2026-06-10
add --force-register-clipboard

khyperia:
Related: https://github.com/neovide/neovide/pull/3393

The above PR enabled the workflow of `neovide --no-tabs --neovim-bin ssh -- SERVER nvim`. However, the heuristics for whether neovide registered its own custom clipboard provider made neovide think this is a local neovim session. It's not! So no clipboard provider would be registered, and I couldn't copy/paste.

There are many different ways of solving this "issue". This PR is trying out adding a new `--force-register-clipboard` commandline option, that skips the "is this neovim instance local" heuristics, and just uses the value provided instead. Works great for my use case, but, let me know if you'd prefer a different approach and I'm happy to explore alternatives!

### DIFF
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -200,10 +200,10 @@ async fn create_neovim_session(
 
     let cmdline_settings = settings.get::<CmdLineSettings>();
 
-    let remote = cmdline_settings.wsl || cmdline_settings.server.is_some();
     // This is too verbose to keep enabled all the time
     // log::info!("Api information {:#?}", api_information);
-    setup_neovide_specific_state(&session.neovim, remote, &api_information, &settings).await?;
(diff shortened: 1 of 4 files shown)


CGamesPlay:
I'm not the maintainer here, just another contributor who uses Neovide primarily with remote nvim. I'm trying to get #3450 merged, and I think it's worth considering how `--force-remote-clipboard` might evolve in the future. My initial reaction is that perhaps it could be renamed to `--remote`, which would control all of the current and future "remote bridge" functionality that needs to exist. This is basically what `--wsl` does on non-Windows platforms today, so implementation-wise we might even be able to simplify `--wsl` to imply this flag.

khyperia:
ooh yeah makes sense! I would also want that behavior for `gx`. I'm happy to change this PR to be `--remote` or `--force-remote`, but I'll wait for a maintainer to chime in on what is preferred.

PR #3522, opened 2026-06-25
feat: add TERM_PROGRAM and TERM_PROGRAM_VERSION environment variables

vegerot:


vegerot:
cc @falcucci

falcucci:
will check soon

falcucci:
@vegerot empty pull request descriptions are not welcome. elaborate precisely why you need this patch, in case you need to setup your config you have [g:neovide](https://neovide.dev/configuration.html#initvim-and-initlua-helpers) as you can see in the docs.

PR #3535, opened 2026-07-25
ci: fix AppImage updates for nightly builds

mareklangiewicz:
Nightly AppImages were built with the same update info as stable (`latest`) and the `.zsync` file wasn't uploaded to the nightly release, so AppImageUpdate couldn't follow the nightly channel.

Point LDAI_UPDATE_INFORMATION at the `nightly` tag and upload `neovide.AppImage.zsync` the same way we do for regular releases.


falcucci (approved):

falcucci:
thank you!

PR #3544, opened 2026-08-10
fix: render COLRv1 fonts on Linux

### LINKED ISSUE #3493: No support for COLRv1 emoji format
**Describe the bug**
I know that neovide used to render emojis correctly. I had to add this to make it work:
```
vim.opt.guifont = "JetBrainsMono Nerd Font,Noto Color Emoji:h9.8"
```

However now it doesn't render them for me anymore. There's a blank space where the emoji would be, but it's not drawn. The emojis are drawn in neovim in the terminal (kitty). They also don't render if I comment my vim.opt.guifont option.

**To Reproduce**
1. Create a new file
2. Try to paste this emoji: 🤦
3. It doesn't appear

 - OS: fedora linux
 - Neovide Version 0.16.2 nightly b5363a36b7e4995e3b87f5c170f87596ec5739f4
 - Neovim Version 0.12.1

attached a gzip.

[neovide_rCURRENT.log.gz](https://github.com/user-attachments/files/26980331/neovide_rCURRENT.log.gz)

fzlzjerry:
Closes #3493.

## Summary

- build Skia with its embedded FreeType on Linux so COLRv1 color glyph painting is compiled in
- add a renderer regression test using the official Fontations COLRv1 fixture
- document the additional Clang and Ninja dependencies for Linux source builds

## Root cause

Font lookup and shaping are working for the affected Noto Color Emoji release: swash and Skia resolve the same glyph ID. The glyph disappears later, when the cached Linux Skia library tries to paint it. That library does not contain Skia's FreeType COLRv1 paint path because the path is gated by the FreeType headers used at compile time.

Enabling `skia-safe`'s `embed-freetype` feature builds Skia against its bundled current FreeType. The same Noto font then renders, and the regression fixture changes from a monochrome outline fallback to a correctly colored COLRv1 glyph.

## Validation

- `cargo test --offline` (188 passed)
- `cargo clippy --all-targets -- -D warnings`
- `cargo fmt --all -- --check`
- `cargo build --release --locked`
- verified the regression test fails without `embed-freetype` and passes with it
- reproduced and verified the fix with the current Noto Color Emoji COLRv1 font

## Build trade-off

rust-skia 0.97.2 does not currently publish a cached Linux binary for this `embed-freetype` feature combination, so a first Linux source build compiles Skia locally. The locked release build completed successfully in 8 minutes in my environment. The stripped release binary increased from 36,127,656 to 36,940,160 bytes (+812,504 bytes, about 2.25%). The installation docs now include the required Clang and Ninja packages.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -881,6 +881,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77ce24cb58228fbb8aa041425bb1050850ac19177686ea6e0f41a70416f56fdb"
 
 [[package]]
+name = "font-test-data"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7d99914ae8c4aee9e814f54be404e15182a9a5556eb2d37f4189e6cff44d5a63"
+dependencies = [
+ "font-types 0.12.2",
+]
+
+[[package]]
 name = "font-types"
 version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -888,6 +897,12 @@ checksum = "39a654f404bbcbd48ea58c617c2993ee91d1cb63727a37bf2323a4edeed1b8c5"
 dependencies = [
  "bytemuck",
 ]
+
+[[package]]
+name = "font-types"
+version = "0.12.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0a7299a780854a6d391be2ae1c8521c9368471b559dbfd6a8dbd9f407eaff100"
 
 [[package]]
 name = "foreign-types"
@@ -1605,6 +1620,7 @@ dependencies = [
  "derive-new",
  "dirs",
  "flexi_logger",
+ "font-test-data",
  "fork",
  "futures 0.3.31",
  "gl",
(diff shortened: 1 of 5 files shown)


falcucci:
thank you, I have just double checked and even the latest rust-skia 0.99.0 doesnt include the embed-freetype cache key in the skia [binaries](https://github.com/rust-skia/skia-binaries/releases) repo.

I would only merge this if we are able to send a patch to support the key `embed-freetype` in rust-skia upstream first to use the cached version.

would you be keen to do it?

fzlzjerry:
Yes, I can take that upstream. I’ll trace how rust-skia keys and publishes binary-cache feature combinations, add support for `embed-freetype`, and link the upstream PR here before updating this branch.

fzlzjerry:
Upstream patch is open: rust-skia/rust-skia#1317. It adds the exact Linux `embed-freetype,gl,textlayout` release combination, producing the `ftembed-gl-jpegd-jpege-pdf-textlayout` cache key for x86_64 and aarch64 GNU/Linux. Android is excluded because it already embeds FreeType and would otherwise duplicate existing artifacts. Once that lands and rust-skia publishes the corresponding release assets, I’ll update this branch to the cached version.

falcucci:
I have bumped skia-safe to the current latest version so when they release your fix you can just bump it again in this patch without complications.

emmanueltouzery:
the new skia-safe release 0.153.2 now includes the changes, however it also has presumably other changes, that neovide maybe would have to be ported to.

falcucci:
@fzlzjerry thank you, really appreciated!

